PR #589, opened 2014-09-27
usable objects module - contribution

Henddher:
usable objects modules: use and @mkusable commands allow to create puzzle-like objects that can be "materialized" by combining other objects.


Griatch:
Since there were no further communication on this and it seems this contribution is still in need of further development, I'm closing this pull request now. You are welcome to make a new request when ready.

PR #609, opened 2014-11-03
(WIP) Collab

Kelketek:
Opening this pull request as WIP to get feedback and more eyes on it. It is not ready for proper merge, but a few commits might be worth cherry-picking early, such as the Typeclass-aware QuerySets, and the ability to check different lock types from one lock string. Just let me know if you're going to do that so I can handle any rebasing/commit fiddling accordingly on my end.

### DIFF
--- a/evennia/contrib/collab/collab_settings.py
+++ b/evennia/contrib/collab/collab_settings.py
@@ -1,0 +1,74 @@
+# To enable collaborative building in your game, add this line to your settings.py file:
+#
+# from evennia.contrib.collab.collab_settings import *
+#
+# ...After the line:
+#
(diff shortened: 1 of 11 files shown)


ahmedcharles:
Wouldn't it be better to have this PR against devel instead of master?

Kelketek:
@ahmedcharles This PR predates the devel branch.

Kelketek:
@Griatch This has been rebased. If you could look over it again at your next opportunity, that would be great. I think I addressed most of the notes you had previously, but it's been a good long time since then.

Kelketek:
Closing in favor of another PR which shall be internal instead of across forks for better visibility.

PR #637, opened 2015-01-14
Devel: Setup.py and configuration

Kelketek:
This PR is a work in progress. It will provide a setup.py and the configuration within it to allow Evennia to be installed as a system package.


Kelketek:
@Griatch This should be ready for merge now. You will be able to install this package using pip install. During development, it will be most helpful to use the -e flag with pip, which will softlink the installation so that the package is in the path, but editing the files will be immediately available instead of requiring a reinstall.

I have found that the web stuff is broken, but I haven't had a chance to track down the causes yet. If you merge this PR, I'll look into adding fixes for the web stuff in a new PR.

Owlsoup:
Guess 'exceptions.ImportError: No module named urls'? got to do with this 'broken web stuff'. @Griatch told me that the devel webclient wasn't functional yesterday. But I do however, find it strange that I don't get this traceback when starting the server - I get it when the server is running, and with no apparent reason (without trying to connect with the webclient!).

evennia/VERSION.txt
```diff
@@ -0,0 +1 @@
+0.9
```
Griatch: I guess we need to start discussing just what versioning system scheme to use here.

Kelketek:
@Owlsoup The web stuff is handled by Django, and Django's URL/view/template machinery is not initialized until the first web request comes in. The only thing that's validated ahead of time is the Django models. :)

Owlsoup:
Ok. Good knowledge! ^^

Issue #645, opened 2015-01-18
Create more insightful testing structure


Kelketek:
The current command tests that we run enable us to test that commands have run and check their output to ensure that the displayed text from the commands are correct. However it doesn't tell us much about the function of the commands themselves-- whether they've set the properties we expect, etc.

To fix this, we will probably want to make use of [Mocks](http://www.voidspace.org.uk/python/mock/) and a more robust set of test base classes that have convenience functions for examining the ins and outs of command effects.

Kelketek:
Another idea: Add Travis for Continuing Integration tests.

Griatch:
Travis was added. What is the goal of this issue - that is when will it be considered fulfilled to an extent that it can be closed?

Kelketek:
@Griatch I need to sit down and examine the tests and see if there's anything else in particular worth doing before calling this one closed. I'll try to post a more specific scope, if any, sometime on Sunday.

Griatch:
Since this concerns structure (rather than the amounts of tests, like #120), maybe the removal of un-finished test modules in `evennia/tests/` could fit this Issue. 

I was also briefly considering moving all those spurious `test.py` from their various app-folders and into the main `tests/` folder ... It would make it a lot easier to locate all tests in one place (assuming the models got good names), but not keeping the tests in respective app might completely ruin our chances of running specific tests from the command line ...?

Kelketek:
@Griatch Tests should stay in their module. App code should stay together, and, yes, it would make running specific tests more difficult.

Griatch:
We already have problems with the Commands in this regards (since commands is not an app). Do we really need the central `tests/` folder at all if we want to keep the app structure intact?

Kelketek:
@Griatch My feeling is no. We could make commands an app just by dropping an empty models.py into it and adding it to INSTALLED_APPS, though.

Griatch:
I'm pretty sure modern Django don't require the `models.py` anymore actually. Worth looking at if only for the testing reason.

gtaylor:
@Griatch that came in with AppConfig in Django 1.7, IIRC. As long as you don't care about backwards compatibility, it'd be safe to no longer have models.py.

Griatch:
@gtaylor 

Right. We require 1.7 for the migrations anyway.

Kelketek:
[...]

PR #646, opened 2015-01-25
(WIP) Test improvements, and small things

Kelketek:
Some of the commits I was going to roll into the collab PR (but which are generally more useful) are included in this one instead. This includes the ability to check different lock types from one lock string, and the ability to clear tags by category.

However the meat and real concern of this PR is the testing system refactor it includes. This worked fine when played atop the current master but has broken in devel. I'm not certain why. The changes in the testing infrastructure here aren't quite enough to satisfy what I had in mind for #645, but they do take care of some issues-- breaking down each command into its own test, making tests transactional, and better handling of the settings for testing.

The current issues post-rebase are:

1. I'm not certain the paths are being loaded right for the testing environment. It's difficult to load some typeclasses and imports may require overqualification.
2. Old objects are removed from the DB per the transaction cases (I think) but Evennia's caching structure does not reset all the way, causing ghost objects to interfere with subsequent test runs.

I believe these are the primary causes of tests failing. To test, create a new game, migrate it, then run `evennia test`.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -276,6 +276,45 @@ class CmdCpAttr(ObjManipCommand):
     locks = "cmd:perm(cpattr) or perm(Builders)"
     help_category = "Building"
 
+    def check_from_attr(self, obj, attr, clear=False):
+        """
+        Hook for overriding on subclassed commands. Checks to make sure a
+        caller can copy the attr from the object in question. If not, return a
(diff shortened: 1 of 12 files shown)


Kelketek:
@Griatch This is probably good enough to merge now. The only test failure left is the one about batch commands, and I'm not sure how you want to handle that.

PR #658, opened 2015-02-09
Development Branch Preview PR

Kelketek:
Running PR for the development branch to look over its current status and for easy viewing of the Travis builds at a glance.


Griatch:
I think the pieces remaining before merging devel are currently as follows:

- Resolve remaining merge requests against devel (e.g. #659)
- Discuss with @Kelketek what needs to be done to close #645
- Add a more lenient mechanism when introducing syntax errors in typeclassed modules. It's currently raising a traceback (which can kill the entire cmdset loading too) rather than loading a temporary base class and logging the error as before. 
- Merge my local rewrite of the tutorial_world (this is done, but it does not currently build correctly)
- Fix Windows installation to work using `pip install -e .`
- I also need to take a pass over the web features to make sure they work okay. 

Anything else? 
I think other issues can be applied towards master once devel has merged.

ahmedcharles:
Does migration work again?

Griatch:
@ahmedcharles Good point, need to test this too. Added to the list.

PR #661, opened 2015-02-13
Remove redundant check in perm().

ahmedcharles:


Kelketek:
Are you sure this check is redundant? The check should result in a 0 if either one of the items is 0, but the second item if they're both nonzero.

Griatch:
No, it looks okay, probably a remnant from a refactor.

PR #672, opened 2015-02-22
Contrib fixes

Kelketek:
Makes the line editor and menu system work again. Also forces evennia to initialize when running tests.


Kelketek:
@Griatch Would like your feedback on if this is the best way to handle initializing evennia in test cases. This allowed the initialization to run for me when testing collab individually. Presumably other apps may have the same problem.

Griatch:
Yes, this is the correct way to initiialize the `evennia` central API (it is required since otherwise lots of the imports load before the django have initialized, it's a sort of catch-22)

PR #674, opened 2015-02-23
New test runner

Kelketek:
I've swapped out the old deprecated test runner with a new one that follows Django's current best practices, and started using Mock in a few places where it makes sense. I also removed the test stubs and moved the remaining real tests to their appropriate folders.

After doing so, the new test runner detected new tests which appear to have been forgotten. This revealed several broken modules. I also did some spot repair of modules that were broken in a similar fashion after some searching around. Once this was completed, it was clear the `idmapper` tests and `ampoule` tests were having trouble. I started to work on the `ampoule` tests, and got them to load, at last. Once loaded, I saw it was having trouble with imports and fixed that.

However, once the tests were finally able to run, they really wrecked the place. They all fail. Part of this may be because of changes that are now active in the server and part of it may be due to some things the current test environment doesn't set up. The tests appear so broken, however, they cause the test count to report incorrectly at the end of the suite, perhaps some sort of threading issue is at work there. I'm not really sure where to begin on fixing these tests as I don't understand the code they are testing.


Griatch:
Merged with a rebase; removed the procpool for now (into a separate repo [here](https://github.com/evennia/procpool)) and refactored the idmapper as well as got its unittests to work.

PR #675, opened 2015-02-25
TYPO IN COMMENTS of CMD @copy

tajmone:
TYPO IN COMMENTS: file "/src/commands/default/building.py"

one exact copt of the original
    ....copT = copY...
copy of the ....


Kelketek:
Thanks for this!

PR #678, opened 2015-02-27
Multisession fixes

### LINKED ISSUE #673: Bug: [Devel]: MULTISESSION_MODEs behaving strangely
#### Steps to reproduce the issue:

1. Start devel with MULTISESSION_MODE 1, 2 or 3
2. Try various combinations of puppeting to the same Character from different login sessions
3. The session "view" is not properly translated, with one of the sessions still receiving the data also after it has been de-puppeted. 

#### What I expect to see and what I actually see (tracebacks, error messages etc):

The MULTISESSION_MODEs should work as advertised

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve:

Devel branch.

Kelketek:
Fixes multisession modes to work as advertised.

@Griatch , please take note of a few behavioral changes this introduces. The primary one is that, by default, characters get their messages mirrored between sessions, but players do not, unless a message is intended to go to all sessions a player has. I'm not sure if that was the default before, because it's been eons since I used the multisession modes. But, for instance, one doesn't get 'command not found' error messages in every session if an OOC session makes an error.

Additionally, this adds an alias to `@quit`, `quit`, so that the same command works when unlogged in or logged in.

Resolves #673

### DIFF
--- a/evennia/commands/cmdhandler.py
+++ b/evennia/commands/cmdhandler.py
@@ -100,7 +100,7 @@ def _msg_err(receiver, string):
         string (str): string with a {traceback} format marker inside it.
 
     """
-    receiver.msg(string.format(traceback=format_exc()))
+    receiver.msg(string.format(traceback=format_exc(), _nomulti=True))
 
 
 # custom Exceptions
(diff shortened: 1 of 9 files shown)


Griatch:
Merged with rebase.

PR #685, opened 2015-03-03
i18n Italian : ITALIAN Localization

tajmone:
Evennia localizated to Italian: 1st draft. 100% translated. A few doubts were marked as "fuzzy" and commented (in En + It ) in .po file. The .mo file was django-compiled.


Griatch:
Merged with rebase.

Issue #625, opened 2014-12-10
mxp_send in EvTable


volundmush:
This is NOT working.

```python
bbtable.add_row(mxp_send(board.num,"+bbread %s" % board.num), acc, mxp_send(board,"+bbread %s" % board.num), board.last_post(), board.posts.all().count())

def mxp_send(text="", command=""):
    if text:
        return "{lc%s{lt%s{le" % (command, text)
    else:
        return "{lc%s{lt%s{le" % (command, command)
```

I have confirmed that the problem is EvTable. it's stripping out the MXP somewhere - it looks like it's involved with ANSIString but I'm not sure how exactly.

If the problem is 'splitting text across multiple lines' then I would easily accept having only the first line's text be clickable. every single wrapped character doesn't need it. The tables I use that implement this would generally not wrap anyways.

On that note, is there any chance of getting the MXP features for sending weblinks or SEND MENUS working?

Griatch:
MXP works in tables, but there is currently a lingering issue with the length of the cell content being one character off compared to the borders. So not closing this yet.

Griatch:
Closing this and split the remainder of this issue into a Feature request, see #704.

PR #688, opened 2015-03-07
Pending merge: ANSIString+mxp issue

Griatch:
@Kelketek 

As discussed I made a separate branch with a unittest for the ANSIString + mxp issue. This goes towards resolving the main point of #625.

### DIFF
--- a/evennia/utils/ansi.py
+++ b/evennia/utils/ansi.py
@@ -339,7 +339,7 @@ class ANSIParser(object):
     ansi_map = dict(mux_ansi_map + ext_ansi_map)
 
     # prepare matching ansi codes overall
-    ansi_re = r"\033\[[0-9;]+m"
+    ansi_re = r"(\033\[[0-9;]+m|\{lc(.*?)\{lt|\{le)"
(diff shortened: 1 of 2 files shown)


Kelketek:
@Griatch Rebased this against master and tweaked the test code layout a little as well. Please look at this and let me know if you see any issues with the proposed solution.

Kelketek:
Per discussion in IRC, having ANSIString be MXP aware is not the correct approach.

Griatch:
Not sure closing this was the conclusion of the IRC chat. I though the conclusion was that the length of ANSIString should be MXP aware although it should not explicitly map the ansi tag internally?

With this closed we are otherwise nowhere nearer to resolving #625.

Kelketek:
@Griatch ANSIString's length should not be MXP aware. Instead, a function should be placed in `utils` named m_len() that checks if an object is of class base_string, and if it is, uses a regex substitution to determine the length of the string without MXP, and if it isn't, just uses len(). This is because if we made ANSIString's len aware of MXP, it would make its length reporting inconsistent with its indexes, and this could have odd consequences.

PR #697, opened 2015-03-10
Added m_len and made ANSIString ignore MXP.

Kelketek:
As discussed earlier, this change makes ANSIString essentially ignore MXP, and creates the m_len function.

It also tweaks parse_ansi slightly to not add state changes to the parser object during the function, since the object has a lot of shared references via ANSIString.

### DIFF
--- a/evennia/utils/ansi.py
+++ b/evennia/utils/ansi.py
@@ -82,7 +82,7 @@ class ANSIParser(object):
         """
         return self.ansi_map.get(ansimatch.group(), "")
 
-    def sub_xterm256(self, rgbmatch):
+    def sub_xterm256(self, rgbmatch, convert=False):
         """
         This is a replacer method called by `re.sub` with the matched
         tag. It must return the correct ansi sequence.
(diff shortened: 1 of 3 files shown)


Griatch:
I tried to get this to work with EvTable and found that no matter what I did the Cell comes out as too small (empty, in fact) if using `m_len`. I eventually tracked this down to the `ANSI_PARSER.strip_mxp()` removing the entire MXP-decorated string in some circumstances. Now, after a lot of print statements I found that `ANSIStrings`' `_char_indexes` being empty after adding other strings to the ANSIString (which Evtable does, for padding). This is how to reproduce (you don't need MXP at all in the string): 

```python
from evennia.utils.ansi import ANSIString, ANSI_PARSER

a = ANSISTring("Test")
print a._char_indexes
>>> [0, 1, 2, 3]
print ANSI_PARSER.strip_mxp(a)
>>> ANSIString(u'Test', decoded=True)

b = ANSIString(" ") + a 
print b._char_indexes
>>> [0]
print ANSI_PARSER.strip_mxp(b)
>>> ANSIString(u' ', decoded=True)
```

It's clear when adding the empty string, but you could add text strings too and get very strange results, with strip_mxp curiously enough replacing the original string with the newly added one ... not sure what is going on there.

Kelketek:
@Griatch I'm rather surprised that bug had been around before being discovered. I've added a test for it so it shall not recur. It was a simple one-line bug-- a copy-paste error where I had not changed the name of the variable I was pulling from. This should work. :)

This branch has been rebased against the current master.

Griatch:
Merged into master. There is a remaining issue with mxp width but this seems to be an issue in evtable.

PR #706, opened 2015-03-14
Typeclass tweaks

Kelketek:
A couple of typeclass tweaks to be merged ahead of time for collab, but without having any collab-specific code.

### DIFF
--- a/evennia/typeclasses/models.py
+++ b/evennia/typeclasses/models.py
@@ -128,8 +128,8 @@ class TypeclassBase(SharedMemoryModelBase):
 
 class DbHolder(object):
     "Holder for allowing property access of attributes"
-    def __init__(self, obj, name):
-        _SA(self, name, obj.attributes)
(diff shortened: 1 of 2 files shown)


evennia/typeclasses/models.py
```diff
@@ -126,6 +126,31 @@ class Meta:
         return new_class
 
 
+class DbHolder(object):
```
Griatch: Should probably break out NdbHolder too, just for the symmetry of it then.

Kelketek: @Griatch Alright. I'll get that to you here in a bit, then.

Kelketek: @Griatch Done. NDBHolder was almost verbatim the same as DBHolder, so I consolidated them into DBHolder with an argument.

Griatch:
Merged with rebase.

PR #711, opened 2015-03-18
(WIP) Collab

Kelketek:
This Pull request is the feature branch for comparison of Collab. Collab is an in-progress feature of Evennia that will allow public building. The default Evennia building commands assume builders are trusted. Collab assumes they are untrusted, and that anyone may build. Instructions for adding Collab to one's game are present in `collab_settings.py`.

Collab comes with its own permissions checking model that piggybacks on the existing locks and tags systems to create a concept of object ownership. Objects can have both a 'true owner', the player that created them, and a 'display owner'-- the character object that created them. This allows a layer of hiding between characters and players so that alts are not immediately revealed, a feature often important in games like these.

Collab provides a set of base collaborative typeclasses, as well as a 'collab mixin' that performs the meat of what these typeclasses do. The Collab typeclasses expose an interface of attribute handlers that are scoped to permission. For example, .usr is a version of 'db' that has a few indicative lock functions you can check against in the collab settings module. For an example of how this works, and what conventions should be presented to the user for setting attributes via these handlers, see the `@set` command.

Collab comes with its own implementation of the build commands. These are the meat of the PR at current, and a few are left to be implemented.

This PR **MUST** be highly tested. Several tests are already written, and tests need to be written for every feature. Full coverage, or as close as possible, must be maintained, as this feature is **designed to take abuse**.

**Workflow notes:** This PR is frequently rebased against master, and often commits are squashed. If anyone other than @Kelketek contributes a PR against this one, their commits may be reorganized in rebase slightly. At the moment, the current five commits keep getting squashed into by new code, but in order to give credit, outside commits won't be squashed into his existing commits (without permission).

This feature is a long work in progress, and notes are welcome on architecture, bugs, documentation, or anything that may help it mature and be ready for inclusion in the master branch.

### DIFF
--- a/evennia/contrib/collab/commands.py
+++ b/evennia/contrib/collab/commands.py
@@ -1,0 +1,894 @@
+import evennia
+
+from django.conf import settings
+from evennia.contrib.collab.perms import (
+    quota_check, collab_check, PermissionsError, 
+    is_owner, set_owner, attr_check, get_owner, prefix_check, get_handler)
+from evennia.commands.cmdhandler import get_and_merge_cmdsets
+from evennia.commands.default.muxcommand import MuxCommand
+from evennia.commands.default.building import (
+    CmdSetObjAlias, CmdListCmdSets, CmdName, CmdLock, CmdSetAttribute, CmdCpAttr, CmdMvAttr, CmdCopy
+)
+from evennia.utils import utils, lazy_property
+from evennia.utils.ansi import ANSIString as A
+from evennia.utils.eveditor import EvEditor
+from evennia.utils.utils import inherits_from
+
+
+# Reminder of some of the things in the default Build commands. Removing ones
+# that don't matter or we've already implemented as we go.
+# __all__ = (
+#           "CmdTunnel",
+#           "CmdTypeclass", "CmdWipe",
+#           "CmdExamine", "CmdFind", "CmdTeleport",
+#           "CmdScript", "CmdTag", "CmdSpawn")
+
+def pre_collab(command, syntax_err, perm_err, locks=None, call_super=True, extra_check=None):
+    """
+    Used for commands that affect an object. Can optionally call the super if
+    this is the only difference that needs to be handled.
+
+    If extra_check is provided, will run that as a final check on the object before
+    the super is called (if it is to be called)
+    """
+    if not command.lhs:
+        command.msg("{r%s{n" % syntax_err)
+        return False
+    obj = command.caller.search(command.lhs)
+    if not obj:
+        return False
+    if not collab_check(command.caller, obj, locks=locks):
+        command.msg("{r%s{n" % perm_err)
+        return False
+    if extra_check:
+        if not extra_check(obj):
+            return False
+    if call_super:
+        super(command.__class__, command).func()
+    return True
+
+
+class BaseCreationCommand(MuxCommand):
+    """
+    Creates objects of different typeclasses as specified in the settings.
+    """
+    locks = "cmd:perm(create) or perm(builders)"
+    help_category = "Building"
+    not_permitted = []
+
+    def pre_load(self):
+        """
+        Used for typeclass-frontloaded commands like @dig.
+        """
+
+    def get_location(self):
+        """
+        Used to determine what object will be placed in the object's
+        destination field upon creation.
+        """
+        if not self.caller.location:
+            if hasattr(self.caller, 'player'):
+                return self.caller
+            else:
+                raise ValueError(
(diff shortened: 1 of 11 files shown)


evennia/contrib/collab/commands.py
```diff
@@ -0,0 +1,817 @@
...
+        try:
+            self.perm_check(key, destination, location)
+        except PermissionsError as err:
+            self.msg(str(err))
+
+        return evennia.create_object(key=key, aliases=aliases, location=location,
+                                     destination=destination, typeclass=typeclass)
+
+
+class CmdCreate(BaseCreationCommand):
```
Griatch: If wanting to stick to the layout used by the default cmdsets, then the docstring should look like this: 

```python
Create an object. 

Usage:
  @create object name;object alias 1;alias2

Eventual further details here.
```

evennia/contrib/collab/commands.py
```diff
@@ -0,0 +1,817 @@
...
+
+class CmdDesc(MuxCommand):
+    """
+    Describe an object. To give an object a one-line description, use:
+
+    @desc object=description
+
+    To enter a line editor for a longer description, use:
+
+    @desc/long object
```
Griatch: Is `@desc/long` a good switch name? I see what the purpose is, but would it not be more logical to call this something like `@desc/edit` or `@desc/editor`, that is, describing what it does rather than the type of content you are creating.

evennia/contrib/collab/commands.py
```diff
@@ -0,0 +1,817 @@
...
+    def get_attr(self, obj, attr):
+        """
+        Hook for overriding on subclassed commands. Do any preprocessing
+        required and get the attribute from the object.
+        """
+        attr_name, handler = prefix_check(obj, attr)
+        return handler.get(attr_name)
+
+
+class CmdExamine(MuxCommand):
```
Griatch: Interesting. Re-implementing the examine command like this is a good exercise but may be hard to consistently maintain. I wonder if one could maybe make the main command more pluggable. The ability to selectively show various sections (as shown here) could also be more generalized in core I reckon.

Kelketek: @Griatch `@examine` is perhaps one of the oldest commands in our codebase. It's probably due for an overhaul. I could look into making a PR to re-implement it in a more modular way, and then updating this one to take advantage.

It stands to reason that people will want to override the core `@examine` command just as they may want to override any other, so making it easier to override would be a good idea, I think.

evennia/contrib/collab/commands.py
Kelketek: @Griatch I could go for either. `@desc/edit` is fine with me.

Perhaps a more interesting question while we're here, though, is concerning the line editor. Use of collab, by including this, causes the line editor to be a dependency. There's nothing necessarily wrong with this, but I wonder if the Line editor should become part of core instead of a contrib, and then be overridable by setting, and then this feature backported to the existing `@desc` command.

evennia/contrib/collab/commands.py
Griatch: Breaking up `@examine` into more easily readable and extensible chunks sounds like useful thing.

Griatch:
@Kelketek 
So what is the progress on this, which aspects are left to do, apart from unittest suites? 
.
Griatch

evennia/contrib/collab/commands.py
Griatch: @Kelketek  For reference, the line editor is now called EvEditor and is located in evennia.utils.eveditor.

Kelketek:
@Griatch Sorry it's taken me so long to answer here. A handful of building commands still need to be made and a few decisions on how to make scripts accessible need to also be made. There is also the open question of how to have some sort of functionality similar to 'parent rooms' on other servers. Parent rooms make their commands available to rooms that are contained within them. Presumably, there should be some elegant solution with commandsets that doesn't end up ruining the abstraction they give you.

Griatch:
@Kelketek 

As for "parent rooms", I'm not convinced this is something needed unless one wants to try to mimic a full prototype kind of inheritance. If desired, it shouldn't be too hard to have a typeclass use super() to get the same cmdsets added by its parent.

Griatch:
@Kelketek 

There has been no code activity on this PR for more than a year now. Do you expect to ever proceed with this? I propose we close this PR and pick it up again when you have time to continue with it, what do you say?

Kelketek:
@Griatch I'm alright with that. It'll be a while yet. Business is going well but still bootstrapping. So long as I'm coding directly for work, coding won't be a displacement activity for me.

PR #715, opened 2015-03-18
Typo in docstring : \evennia\server\initial_setup.py

tajmone:
Just a typo...


Kelketek:
@tajmone Please look at the diff on this PR. There are some issues here. It looks like you have some noise and distress signals from struggling with git. :)

tajmone:
Kelketek: I don't know.... Since I've used GitHub for Windows GUI everything went wrong.
I've been using only Git via shell now, and locally everything is fine. I just can't work out how to fix the mess that GitHub for Windows made here on GitHub. Geeeee ... one would think that GitHub would produce some neat software ... then I read reviews about it and realized I'm not the only one experimenting nightmare.

Anyhow, this was the mistake I was trying to point out:

PostgreSQL-psycopg2 compatAbility fix:
PostgreSQL-psycopg2 compatIbility fix:

I'm just getting too frustrated. I'll destroy the fork and start over again...

NEVER USE GITHUB FOR WINDOWS

PR #717, opened 2015-03-19
fixed a small typo in docstring

tajmone:
Now, I really hope that I've managed to fix all the problems with GitHub. it seems clean and now only the single file I've updated is showing up!!!


Griatch:
Merged with rebase. To avoid future issues - make sure to delete your version of this work-branch and make a new branch when you make further fixes. This is because master is, due to the rebase, now out-of-sync with your branch.

tajmone:
;-) Got it ... (better late then ever)

PR #723, opened 2015-03-21
Typos correction in `contrib` folder

tajmone:
Some typos in docstrings, comments and user-output strings.

Please note: in example_batch_code.py (10-11):
```python
# of created objects etc. You can however also run a batch-code python file
# directly using Python (and can also be de).
```
`can also be de`?? something missing here! Couldn't correct that because I have no clue what it's missing.

Also, at 31-35:
```python
# #INSERT filename - this includes another code batch file. The named file will
#            be loaded and run at this point. Note that code from the inserted
#            file will NOT share #HEADERs with the importing file, but will
#            only use the headers in the importing file. Make sure to not
#            create a cyclic import here!
```
Is not too clear: is it (1) the imported file doesn't share its headers with the importing file; or (2) the imported file doesn't have access to the importing file's headers; or (3) the imported file doesn't share its headers with the importing file but will make use of the importing file's header.
This sentence could be made simpler and clearer (in my opinion).

### DIFF
--- a/evennia/contrib/email_login.py
+++ b/evennia/contrib/email_login.py
@@ -167,7 +167,7 @@ class CmdUnconnectedCreate(MuxCommand):
             session.msg(string)
             return
         if not re.findall('^[\w. @+-]+$', playername) or not (0 < len(playername) <= 30):
-            session.msg("\n\r Playername can max be 30 characters or fewer. Letters, spaces, dig\
+            session.msg("\n\r Playername can be max 30 characters, or less. Letters, spaces, dig\
 its and @/./+/-/_ only.") # this echoes the restrictions made by django's auth module.
             return
         if not email or not password:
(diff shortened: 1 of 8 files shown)


Griatch:
Thanks, clarified docstring in batchcode and merged this with a rebase. :)

tajmone:
I don't understand why the last file `menusystem.py` didn't pass the test. I've only changed comment lines in it, and the file was in synch. Strange. Travis CI gives some error number that I can't make much sense of. No big deal, just curious as to what went wrong...

Griatch:
That was not your fault, but a change in master (that I did) that caused the test to fail a little earler. It was fixed before merger. The Travis system just checks the whole merger and determines if the result will pass, it doesn't know from where the error comes.

PR #727, opened 2015-03-26
Collab backports

Kelketek:
Pulls the line editor into the main area, allows it to be overridden, adds long desc option, and also adds an exit to the test scenarios.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -515,24 +515,66 @@ class CmdDesc(MuxCommand):
     """
     describe an object
 
-    Usage:
+    Usage 1:
       @desc [<obj> =] >description>
 
-    Setts the "desc" attribute on an
+    Sets the "desc" attribute on an
     object. If an object is not given,
     describe the current room.
+
+    Usage 2:
+      @desc/edit [obj]
+
+    Opens up a line editor to set the "desc"
+    attribute on an object for advanced editing.
     """
     key = "@desc"
     aliases = "@describe"
     locks = "cmd:perm(desc) or perm(Builders)"
     help_category = "Building"
 
+    def edit_handler(self):
+        if self.rhs:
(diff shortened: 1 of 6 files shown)


evennia/commands/default/building.py
```diff
@@ -515,12 +515,18 @@ class CmdDesc(MuxCommand):
...
+    Usage 1:
       @desc [<obj> =] >description>
 
-    Setts the "desc" attribute on an
+    Sets the "desc" attribute on an
     object. If an object is not given,
     describe the current room.
+
+    Usage 2:
+      @desc/long [obj]
```
Griatch: This seems to refer to "/long" still, whereas the switch is /edit in the code below.

evennia/commands/default/building.py
```diff
@@ -531,22 +537,60 @@ def func(self):
...
-        else:
+            obj.db.desc = self.rhs
+            desc_set = True
+        elif 'edit' not in self.switches:
             obj = caller.location
-            desc = self.args
-        # storing the description
-        obj.db.desc = desc
-        caller.msg("The description was set on %s." % obj.key)
+            obj.db.desc = self.args
```
Griatch: This seems like a rather hard-to-read way to do it. Why not just an if ... else statement for the editor or non-editor cases?

evennia/utils/utils.py
```diff
@@ -1257,3 +1257,10 @@ def m_len(target):
     if inherits_from(target, basestring):
         return len(ANSI_PARSER.strip_mxp(target))
     return len(target)
+
+
+def get_line_editor():
```
Griatch: Nice - customizable line editor location. :)

Kelketek: Seemed to me we'd better have it if commands were expected to depend on it. :)

PR #734, opened 2015-04-01
Add a worldloader.

luyijun:
It can load game world from CSV files. Please read README.md.

Here is the brief introduction:
Copy folder ```worldloader``` to your game folder, then copy folder ```worldloader/example_tutorial_world/worlddata``` to your game folder.

Add these at the end of the ```/server/conf/settings.py```:
```
from worlddata import world_settings
INSTALLED_APPS = INSTALLED_APPS + (world_settings.WORLD_DATA_APP,)
```

Run ```evennia migrate```.
  
Add these to ```/commands/default_cmdsets.py```'s beginning:
```
import worldloader.command as loadercmd
```

Add these to ```CharacterCmdSet```'s ```at_cmdset_creation``` in ```/commands/default_cmdsets.py```:

```
        self.add(loadercmd.CmdImportCsv())
        self.add(loadercmd.CmdBatchBuilder())
        self.add(loadercmd.CmdSetDataInfo())
```

There are three new commands:

1. ```@importcsv``` It can load data from CSV files to game db. CSV files' path is defined in ```/worlddata/world_settings.py```'s ```CSV_DATA_FOLDER```.

2. ```@datainfo``` It can set an object's data info in game directly.

3. ```@batchbuild``` It can setup the whole game world with data in CSV files.

Run Evennia and login.

As a builder in game, you should move yourself to Limbo and input:
```
@datainfo #2=limbo
```
(This adds a unique key to Limbo.)

Then input:
```
@batchbld
```

If everything is OK, the tutorial should be build.


Griatch:
Thanks for the contribution! Being able to read and create a world from a csv file is a cool idea. A csv file in itself is not very readable though. Is this intended to go together with some sort of external editor?

I/we have some questions on the actual implementation of this. If I understand correctly you are essentially replacing the batch-code/cmd builder syntax with a CSV one. Cool. But why is there a need for several new models and an entirely new app to do this?

evennia/contrib/worldloader/LICENSE.txt
```diff
@@ -0,0 +1,28 @@
+BSC license
```
Griatch: Should this actually read "BSD"?

evennia/contrib/worldloader/README.md
```diff
@@ -0,0 +1,126 @@
...
+    |
+    --- ...
+    |
+    --- worlddata
+```
+
+Add these at the end of the ```/server/conf/settings.py```:
+```
+from worlddata import world_settings
+INSTALLED_APPS = INSTALLED_APPS + (world_settings.WORLD_DATA_APP,)
```
Griatch: Why is there a need for a new django app to read a few CSV files?

evennia/contrib/worldloader/README.md
```diff
@@ -0,0 +1,126 @@
+# Intro
+It is a tool for Evennia(http://github.com/evennia/evennia) which can load game world from CSV files.
+
+
+## Installation
+You need to install Evennia and create your game first. Evennia's wiki is here: http://github.com/evennia/evennia/wiki.
+
+Copy the folder ```/example_tutorial_world/worlddata``` to your game folder. The directory should look like
```
Griatch: I would think the example data could remain in the evennia/contrib folder? Why does it need to be copied to the game folder? Would it not be easier for the user to simply supply the path to the CSV file to load, wherever it is?

evennia/contrib/worldloader/README.md
```diff
@@ -0,0 +1,126 @@
...
+key,name,alias,typeclass,desc,lock,attributes,location,destination
+```
+
+These values will set to object's relative property. Data in ```attributes``` field should be in form of Python's dict. All other fields except above ones will be set to object's attributes too.
+
+Objects in ```world_rooms.csv```, ```world_exits.csv```, ```world_objects.csv``` are unique. Every record should has one and only one object in the world.
+
+Objects in ```personal_objects.csv``` are not unique, a record can has zero or multiple objects.
+
+The ```world_details.csv``` is used to set object's detail only.
```
Griatch: This is cool, but seems to enforce the use of the ExtendedRoom contrib. I don't see this explicitly set or mentioned anywhere.

evennia/contrib/worldloader/command.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        # this can be removed in your child class, it's just
+        # printing the ingoing variables as a demo.
+        super(MuxCommand, self).func()
+
+
+#------------------------------------------------------------
+# import csv tables
+#------------------------------------------------------------
+class CmdImportCsv(MuxCommand):
```
Griatch: For a contrib, the commands should probably be broken out into their own module without the default commands being in there. Makes the module easier to read - the commands are to be imported into the actual default anyway.

evennia/contrib/worldloader/command.py
```diff
@@ -0,0 +1,281 @@
...
+    """
+    Usage:
+      @importcsv
```
Griatch: This seems to be an incomplete usage string. Should it be `@importcsv <modelname>` perhaps?

evennia/contrib/worldloader/command.py
```diff
@@ -0,0 +1,281 @@
...
+        models = self.args
+        if models:
+            models = [arg.strip() for arg in models.split(',')]
+        else:
+            models = world_settings.WORLD_DATA
+
+        # import models one by one
+        for model_name in models:
+            # can only import models in world_settings.WORLD_DATA
+            if not model_name in world_settings.WORLD_DATA:
```
Griatch: Why is this limited to models in `world_settings.WORLD_DATA`? As a safety precaution?

evennia/contrib/worldloader/loader.py
```diff
@@ -0,0 +1,491 @@
...
+    obj.attributes.add("model", model, category=DATA_INFO_CATEGORY, strattr=True)
+    obj.attributes.add("key", key, category=DATA_INFO_CATEGORY, strattr=True)
+    
+    if (not app) or (not model) or (not key):
+        return True
+
+    return load_data(obj)
+
+
+def load_data(obj):
```
Griatch: I must admit I don't understand why this "data_info" is needed. You are basically describing the model's natural_key (I think), which you can get directly from any model at any time. Storing it as three separate Attributes seems unnecessary.

evennia/contrib/worldloader/command.py
```diff
@@ -0,0 +1,281 @@
...
+        
+        if not self.rhs:
+            if self.args == self.lhs:
+                # no "="
+                obj_name = self.args
+                obj = caller.search(obj_name, location=caller.location)
+                if not obj:
+                    caller.msg("Sorry, can not find %s." % obj_name)
+                elif obj.db.info_db and obj.db.info_key:
+                    caller.msg("%s's datainfo is %s" % (obj_name, obj.db.info_db + "." + obj.db.info_key))
```
Griatch: This info_db Attribute seems (by reading ahead) appears related to the natural key of the model, which is unique to each app+model. From there on finding an object is a matter of locating its dbid. I might be confused here, but why is there a need for setting this thing explicitly?

evennia/contrib/worldloader/command.py
```diff
@@ -0,0 +1,281 @@
...
+      
+    Build the whole game world with data in CSV files.
+    """
+    key = "@batchbuilder"
+    aliases = ["@batchbld"]
+    locks = "perm(Builders)"
+    help_cateogory = "Builders"
+    arg_regex = r"\s.*?|$"
+
+    def func(self):
```
Griatch: This seems like a very limited command - you can only ever use it once until someone changes the settings. Would it not be more reasonable for this command to take a CSV file as input and build using it?

evennia/contrib/worldloader/example_empty_world/worlddata/models.py
```diff
@@ -0,0 +1,121 @@
+from django.db import models
+
+
+#------------------------------------------------------------
+#
+# store all rooms
+#
+#------------------------------------------------------------
+class world_rooms(models.Model):
```
Griatch: I'm sorry if I'm nitpicking here, but why is this intermediary step necessary - that is, why is a separate database table needed holding only the text and not the actual objects? If uniqueness is (for some reason) required, why not just check this at creation-time? I see no documented use for these extra models, but maybe I'm not just seeing it.

evennia/contrib/worldloader/example_tutorial_world/worlddata/csv/world_details.csv
```diff
@@ -0,0 +1,54 @@
+key,name,location,desc
```
Griatch: I presume this is to be used with some sort of external editor? CSV text is cool for transfering data etc but it's after all not easy to read nor write by humans:  especially since the order of elements in each row is important and missing any one comma will mess up all that comes after it ... If so, this should probably be made more explicitly clear.

evennia/contrib/worldloader/example_tutorial_world/worlddata/models.py
```diff
@@ -0,0 +1,125 @@
...
+    key = models.CharField(max_length=255, primary_key=True)
+    name = models.CharField(max_length=255)
+    alias = models.CharField(max_length=255, blank=True)
+    typeclass = models.CharField(max_length=255)
+    desc = models.TextField(blank=True)
+    location = models.CharField(max_length=255, blank=True)
+    home = models.CharField(max_length=255, blank=True)
+    lock = models.CharField(max_length=255, blank=True)
+    attributes = models.TextField(blank=True)
+    tutorial_info = models.TextField(blank=True)
```
Griatch: This database field seems very tutorial-world specific to belong in a full app. Or is `tutorial_info` intended as a general feature for learning a game?

evennia/contrib/worldloader/loader.py
```diff
@@ -0,0 +1,491 @@
...
+        if not field.name in known_fields:
+            attributes[field.name] = info.serializable_value(field.name)
+
+    set_obj_attributes(obj, attributes)
+
+    return True
+
+
+
+def set_obj_typeclass(obj, typeclass):
```
Griatch: You're gonna need to explain this one to me, I'm not quite following. You are here loading data from a custom model and apply it onto the object tied to that data store ... why is this data not applied directly at your `create_object` step?

luyijun:
It has two main purposes.
1. Game designers who know nothing about programing can use these CSV files to design the game. They can create rooms and objects freely.

2. It is easy to modify rooms and objects. If the game world is established and you want to change some objects' description, you should find them and type in new descriptions one by one in game. It is a hard work. And it is hard to input multiline descriptions. With CSV files, you can modify these files and type only one command in game to apply these changes.

You can use Excel in Windows or Numbers in Mac or any text editor you like to edit CSV files and export them. When using Excel, you must convert the output CSV files to UTF8.

I put worlddata in user's game folder because the data models may be different between different games and it can be modified by game designers.

evennia/contrib/worldloader/LICENSE.txt
luyijun: Yes, it is a typo.

evennia/contrib/worldloader/README.md
luyijun: If the game world is large, there may be a series of tables and a lot of CSV files. So I think it's better to load them into db and use a new app to hold db models.

evennia/contrib/worldloader/README.md
luyijun: If a player want to create his own game world, he will edit these CSV files and even edit db models. So it's better to put these files in the game folder.

evennia/contrib/worldloader/README.md
luyijun: The ```world_details.csv``` is used for ExtendedRoom only, but other files are for common use. If a user does not use ExtendedRoom, he can leave it alone.

evennia/contrib/worldloader/command.py
luyijun: Yes, the ```<modelname>``` is lost.

evennia/contrib/worldloader/command.py
luyijun: There are rooms, exits, common objects and maybe other types, each CSV file must match a db model. So I limited the models.

evennia/contrib/worldloader/loader.py
luyijun: The ```model``` is useful when removing an object. For example, Room01 was in table ALL_ROOMS before, but it is not in this table now. From this room's ```model``` attribute, the program know it is a room and it should not exist, so it can be removed. If there is no ```model``` attribute, the program could not know whether it should be removed.

The ```app``` is not quite unnecessary.

evennia/contrib/worldloader/README.md
luyijun: CSV files are just one source. You can load data from other sources to create the game world too.

evennia/contrib/worldloader/command.py
luyijun: With ```app``` and ```model```, the program can find the key quickly. But not all object in game has these values, such as Limbo. If I want to add an exit to it, I need to add a key to Limbo manually. Then the program can find it and add an exit to it.

evennia/contrib/worldloader/command.py
luyijun: Game world data relate to each other. If only load one data file, there may be errors. For example, I removed a room and move objects of this room to another room. If I only load the room data, these objects may be moved to wrong positions. So it is better to reload all data.

evennia/contrib/worldloader/example_empty_world/worlddata/models.py
luyijun: These tables store game world data. Currently CSV file is the only data source, but in the future, other files could be data source too, or even modify data online. All these data sources will be imported into these tables, and use these tables to create the game world.

evennia/contrib/worldloader/example_tutorial_world/worlddata/csv/world_details.csv
luyijun: Excel on Windows and Numbers on Mac can handle these CSV files. When using Excel, the outputted CSV files must convert to UTF8.

evennia/contrib/worldloader/example_tutorial_world/worlddata/models.py
luyijun: Data in example_tutorial_world are used for tutorial world only. If someone want to create his own game world, he can modify these models. The models in example_empty_world haven't ```tutorial_info```.

evennia/contrib/worldloader/loader.py
luyijun: This function can change an object's data after creation. If an object has already been created, its data still can be modified. For example, some players have equipped swords, and you want to change swords' description. If removing old swords and creating new swords, some data may lose.

evennia/contrib/worldloader/example_empty_world/worlddata/models.py
luyijun: I'll add documents to them.

Griatch:
@luyijun 

Having looked over this I must admit I'm not convinced about your arguments for the design of this. Basically, what you are doing is replicating the database structure in CSV files. One line in your CSV file is practically one line in a database table. The main difference is that you handle some things, like Attributes inline. 

Let's study the two main use-cases you present, namely the ease of creating new objects by modifying the CSV file and the ease of changing the existing objects by changing those same CSV files. 

1. For creating new objects, all you need to do is to read the CSV file and for each line you need to create a new object from the parsed data in that file. For this no extra database model is needed, all you need is a parser linked to the correct create method. 
2. For modifying existing objects, I see your point for a database model, but also in this case you are already choosing to represent all your data with your CSV files - without them your modification is moot anyway. This means that there is pretty much a 1:1 relation between your CSV data file and the database. When the CSV file changes what you need is just the *diff* between your previous CSV file and your current version to apply. No extra complexity is needed; you could even simply save the last version of your CSV file (for doing the diff) and still take up less storage space with a lot less complexity than keeping a custom database model that the end-users must even themselves  modify down the line. 

Overall, your design requires hard-coded loading of fixed locations in the game folder. I don't agree that this is necessary in order to maintain inter-connectivity; if CSV files have an inter-dependence, this should be defined in the CSV file (such as in the header, telling the parser to also load whichever CSV file it depends on). The in-game command then only requires you to supply the top-level CSV file that you want to load. This way your code can be applied freely as needed without needing massive modifications to the game folder setup. 

As a side note, I should also point out that all django databases can natively import and export JSON, XML or YAML files via the `evennia dumpdata` and `evennia loaddata` commands. It might be worth exploring this as an alternative implementation (but I'm admittedly not sure how well it works with updates). 
.
Griatch

tajmone:
I followed with interest all these question and answers. Which brings to mind an issue of higher order: When a proposed contrib is not eligible to be included in the main Evennia download, where should it go? I mean ... it would still be nice to have some kind of official (or semi-official) place in which such contribs might end up.

This contrib--while I agree with Griatch's arguments and see the points that make it more of a "special case" contrib than an ordinary one (too many global changes and restrictions)--is indeed interesting and somehow promising in its potential.

I get the impression that at this beta stage Evennia would benefit from sample codes, tutorials, and all kind of snippets that might help starters in finding support and/or inspiration. Forks could be a dispersive way to collect snippets, and not a good solution for small scripts.

Shouldn't it be worthy considering some web-space to gather snippets, tutorials, examples, and even custom variations on these, that did not make it to the contrib folder? This could allow to keep Evennia's contrib folder for practical and not-too-esoteric code, yet not loose trace/contact with other contribs that might be of interest (in themselves or as a base for development).

In my past experience with software which relied on users' extensions, those who "centralized" them from the start were much easier to follow up then those which evolved in a dispersive manner, with many different sites offering bits and pieces of extensions (very often the same one but out of synch when it came to updates).

Griatch:
@tajmone 

First, I should point out that this thread represents a *discussion* about the push request. This happens for all possible pull requests to a bigger or smaller extent. The thing with adding things to `contrib/` is that adding it to the core Evennia repo also means that *we* accept to maintain it along with the rest of Evennia. It will not be merged as it stands but that may change as the discussion progresses. My comments also only represents *my* opinions. If @luyijun doesn't agree with my feedback, it's perfectly fine and I would be happy to link to his repository for his worldloader to make it available to users. 

Our [links section](https://github.com/evennia/evennia/wiki/Links#third-party-evennia-links) acts as the grouping page for any related projects or plugins hosted outside Evennia's core. People can (and do) also announce eventual own projects on the mailing list or here in the issue tracker. Github itself acts as the best repository of adjoining code we can wish for - not only does it conveniently allow for the creator to keep updating their work, it easily allows people to track those updates and even revert back to older versions as necessary. 

The future of the `contrib/` folder is a little under debate. As said, on one hand it adds external functionality and helps extend the bare-bones base, but on the other hand it represents code that core devs needs to maintain, which is not always easy (it's not easy even for code I've written myself actually, since this code is not included by default - it needs unittests which have not yet been written). It might be that `contrib/` should be lifted out into a separate repository or some other solution, but nothing is decided yet.

Griatch:
@luyijun 

From the looks of [muddery](https://github.com/muddery/muddery) (cool name!) it looks like you are taking the worldloader in another direction. Am I to understand from the lack of response that you are not going to modify/discuss this contribution any further and that this pull request should be closed?

I put up a link to muddery from our [link page](https://github.com/muddery/muddery) at any rate, hope the projects can see continued collaboration. :) 
.
Griatch

luyijun:
Yes, I  created a new project named Muddery. It still under construction.

One of Evennia's principals is trying to impose as few game-specific things on users as possible. But a lot of game designers do not like programming, they like to concentrate on planning the game world. So I think it's better to use Evennia as core and provide useful game models in my project. It can make building game worlds easier.

Griatch:
@luyijun 

Sure, fair enough. Good luck with muddery and keep in touch!
.
Griatch

PR #759, opened 2015-06-12
@lock typo fix and added note about apostrophes in help

butterdroid:
Added a missing apostrophe to the end of lockstring form line.

Added a note about the use of apostrophes when passing the lockstring to the command.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -1654,7 +1654,7 @@ class CmdLock(ObjManipCommand):
     object.
 
     Lockstring is on the form
-       'access_type:[NOT] func1(args)[ AND|OR][ NOT] func2(args) ...]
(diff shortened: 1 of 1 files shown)


butterdroid:
Just realized issue is a bit more complicated.
Passing multiple locks in lockstring causes the first lock to break ex: " 'get:all() " will cause the object to become un-gettable.
Gonna modify my gentle note to a DO NOT DO THIS. But it might be nice for the command to not take apostrophes or handle them somehow.

Griatch:
Thanks for the help, but I am closing this request un-merged since the warning should not be needed anymore; revision 4d69e8078775e78cdf3950c385a05656cfe917da changes the `@lock` command to strips extraneous quotes before passing it on to the lockhandler.

Issue #524, opened 2014-06-26
Feature request: Add dynamic name constructor


Kelketek:
#301 brought this up some time ago, and at the time it seemed a bit excessive but as more use cases have come up a more solid case for it seems to be bubbling up.

There should be a way to get 'display names' of objects as seen by users. This function should be handed who is looking at the object so that its appearance can change based on context. For instance, when staffers walk into a room, it would be handy if the names included the DBREFs, like sword(#4356), and many games might want to enhance and colorize certain typeclasses. It might also be the case that conditions of the room change the names of objects. A dark room might have 'A thick book' where more light might say 'UNIX in a Nutshell'.

Likewise, sdesc systems could make good use of this function, grabbing a nick for something if desired instead of using the actual key.

This would require doing a few things. The first would be to make sure that certain commands used for normal interaction will display the display name, and the second would be to update search utilities so that contents in the user's area of influence (their inventory and the current room) are pulled in for matching against. This wouldn't work across the database, of course.

Griatch:
The tools relevant for this was implemented in merge #763. Closing.

PR #763, opened 2015-06-20
Dynamic name handling

Kelketek:
Allows for users to override name handling, and improves handling of disambiguation functions.


Griatch:
Referencing #524 which can be closed with this.

PR #771, opened 2015-07-25
Login fail

volundmush:
So earlier I requested there be a hook called in case a login fails - correct user, but wrong password.

Why is this important? Well, put simply, it's good to know if there's someone up to no good. This could alert the admin or players of suspicious login attempts, which is very important for security. There is no good reason to not have this.

Penn even supports it via the SOCKET EVENT 'SOCKET`LOGINFAIL'.

So here, I made it myself. It was simple. :)


Griatch:
Merged with rebase.

PR #775, opened 2015-07-26
Create cleanup

volundmush:
Okay, this one is a bit ambitious. I'm sure you're either going to be very happy or very furious at the proposed changes. C/P'd from the commit notes:

- Implemented a new Character Slots feature. Player and Character typeclasses now has methods/properties suitable for overloading. @charcreate was changed to use it.

- Decided that having @charcreate and evennia/evennia/commands/default/unloggedin's UnconnectedCreate do a lot of the work that create.py's functions were already doing was kind of silly. Centralized as much as possible to create.py.

- Added create_character() to create.py. It is basically a wrapper for create_object.

- The inits called by create.py (init_new_player, init_new_character) are now solely responsible for setting default descs, adding to public channel, etc. Now the relevant hooks can be overloaded.

- Altered how CmdUnconnectedConnect treats banned connections. It uses the same session method as CmdUnconnectedQuit to disconnect and send the message. Why? I figure a command depending explicitly on another command existing as asking for trouble in the future.

- Added a 'first puppet' hook to the default character typeclass. Player typeclass has an equivalent, so why not?

- Slightly changed how the Guest typeclasses worked. Added an option to define the default Guest CHARACTER typeclass. (By default it's the same as base character.)


Griatch:
Thanks for the code! I don't have time to look at it today (vacations and all) but will try to do so soon.
Just remember to split new features into separate branches and commits, I'll most likely rebase and cherry-pick any accepted changes, using branches will make it easier for your master to not go out of sync with a rebased upstream.

Griatch:
While plenty of good ideas here, these should have been at least six separate commits. It'll take me some time to go through this. 

 - The more finely grained slots for the character seems like a good idea: :) 
 - The reasons `@charcreate` et al specify so many things is that the create methods are specifically meant to be able to be modified by their calls. Setting things up at the command level was originally meant as being instructive and to invite people to modify things there. But I agree that since the default commands don't change the defaults anyway, it's unnecessary to be that verbose.  
 - Not sure about the create_character and your Character is just a typeclass of object, it is not a fixed thing in the engine in the way the object is. Hard-coding the character creation is breaking a design principle of Evennia. Same goes with other examples of character being hard-coded, like in setting up channels by calling from utils.utils; this modification should happen by modifying the typeclass, not by hardcoding calls in the core. 
 - Good point about using the session method to quit rather than calling the quit command, makes sense.
 - first_puppet hook for the character typeclass is a good idea but may be mis-named since it's the Character which is the puppet. A more consistent name might be at_first_puppeting or something like that. Will need to ponder that. 

I can't cherry pick from this since it's all in one commit, so I'll need to copy things out and test things separately. I'll get back to you when I'm back in full gear.

volundmush:
I'll probably need instruction on how to better do commits for pull requests. The problem here is that I just started coding changes and some changes depended on other changes and I started headscratching at how to present it...

RE: init, create_character.
Yeah, probably. I was trying to base off of what was already there. To me, characters are a very SPECIAL kind of Object that'll be doing a lot of very specific things that other Objects don't need to worry about and I don't see how you're going to be getting around that issue. Although I'm not really sure why utils is called in the first place for what I copied it from, since all it does is... trigger some methods on the created object EITHER WAY. Why is that even IN utils? Seems like the create function could just trigger a method on the typeclass....

Griatch:
The reason for the utils function is because these hooks are called from other places too and was thus put into utils to homogenize the API, something not possible if hidden away inside the `create_*` function. But I agree it looks a bit strange, it might be clearer to accept code duplication in this case, or maybe put something more on the typeclass so as to be able to homogenize it in another way.

Griatch:
I'm sorry Volund - I've really tried but I cannot merge this as one block and I really don't want to wade through the code line by line to copy only the pertinent things over. Could you make separate commits for the points you make above and then make a new merge request? Best is probably to use a new branch since the one you are using is pretty messy. 

Closing this for now, sorry for the inconvenience. :/

PR #789, opened 2015-08-22
Edge case bug w/ Multisession_mode!= 0 on Evennia initial_setup.py

Quae:
Fix for NoneType exceptions when initial_setup.py is run where multisession_mode is not set to 0. To replicate:
1) Set up a new mygame
2) In the mygame folder conf/settings.py, set MULTISESSION_MODE = 3
3) run evennia start
3.5) It'll ask you to make a super user
4) Exceptions, Mister Anderson!

Note the same error for line 108, 111 (of original code), respectively.

error 1
error 2

### DIFF
--- a/evennia/commands/default/player.py
+++ b/evennia/commands/default/player.py
@@ -83,10 +83,12 @@ class CmdOOCLook(MuxPlayerCommand):
         sessid = self.sessid
         # get all our characters and sessions
         characters = player.db._playable_characters
-        if None in characters:
-            # clean up list if character object was deleted in between
-            characters = [character for character in characters if character]
-            player.db._playable_characters = characters
+
+        if characters is not None:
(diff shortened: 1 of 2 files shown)


Griatch:
I'll add the fix, but I cannot reproduce this error at any `MULTISESSION_MODE`. Did you make any previous modifications to the database file or other changes to the game dir before starting?

Griatch:
The fix was merged in with a rebase (don't use your master branch to commit changes, use a temporary branch you can throw away - that way you can avoid clashes between your master and Evennia's master when we rebase commits). 

I would still need to know what was the cause of this though, since I cannot reproduce it.

Quae:
Thanks for the tip! I'll reach out regarding any other changes that may have precipitated the bug.

PR #801, opened 2015-09-07
Assorted PEP8 cleanups in evennia_launcher.py

gtaylor:
This is not a 100% thorough pass, but I knocked off many of the easier violations.


Griatch:
Merged!

PR #804, opened 2015-09-07
Make evennia -i block until process is closed. Fixes #803.

### LINKED ISSUE #803: Bug: evennia -i no longer starts interactively
#### Steps to reproduce the issue:

1. evennia -i start server

#### What I expect to see and what I actually see (tracebacks, error messages etc):

It looks like the process is getting foregrounded or daemonized (not sure which, haven't had time to look into). This causes some process supervisors and Docker to immediately exit.

The -i flag should keep the process foregrounded. I suspect this may have been broken when logic was added to start both processes if none are specified (you can't stay interactive to the portal and the server processes at the same time in one tty).

On a related note, perhaps we should fail loudly if -i is passed to ``evennia start all`` or ``evennia start``, since we can't actually start both processes interactively.

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve:

Currently on 538228af0411d862d558b3413decd6aa762f49be
Ubuntu 15.04 with Python 2.7
My Docker container is based off of the official python:2.7 image, which is Debian-backed (I think).

To see what I've got on the docker-compose side (and see the symptom on display): https://github.com/gtaylor/evennia-docker

gtaylor:
SIGINT handling will need some more work, but this will at least restore interactive mode to being interactive (which is critical for production deployments under a process supervisor like upstart, systemd, supervisor, etc).

### DIFF
--- a/evennia/server/evennia_launcher.py
+++ b/evennia/server/evennia_launcher.py
@@ -1033,7 +1033,9 @@ def server_operation(mode, service, interactive, profiler):
             GAMEDIR, TWISTED_BINARY, SERVER_LOGFILE,
             PORTAL_LOGFILE, HTTP_LOGFILE])
         # start the server
-        Popen(cmdstr, env=getenv())
+        process = Popen(cmdstr, env=getenv())
+        if interactive:
+            process.wait()
 
     elif mode == 'reload':
         # restarting services


Griatch:
Merged with rebase together with keyboard interrupt handling.

PR #807, opened 2015-09-27
Playersupdate

volundmush:
Just adding in idle_time and connection_time to the DefaultPlayer typeclass. Hope I'm doing this right.


Griatch:
Merged with a rebase!.
.
Griatch

PR #824, opened 2015-10-16
Improve help command.

ahmedcharles:


ahmedcharles:
My motivation for this was actually because the menusystem adds an h alias, which means that he is interpreted as help, which is just weird.

In general, I'd rather have commands that have crisp names, because they increase my confidence in users understanding what is going to happen when they type something and in me understanding what will happen when users type something. (Note, I think the default commands should be held to a higher standard given that they need to work in as many different muds as possible.)

PR #826, opened 2015-10-17
Add add_lines option to EvMenu.

ahmedcharles:
This allows removing the lines before and after the node texts.
Also only strip newlines, not spaces or tabs from the node text.

### DIFF
--- a/evennia/utils/evmenu.py
+++ b/evennia/utils/evmenu.py
@@ -272,7 +272,7 @@ class EvMenu(object):
     """
     def __init__(self, caller, menudata, startnode="start",
                  cmdset_mergetype="Replace", cmdset_priority=1,
-                 allow_quit=True, cmd_on_quit="look"):
+                 allow_quit=True, cmd_on_quit="look", add_lines=True):
         """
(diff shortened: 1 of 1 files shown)


Griatch:
This is a good idea. May I suggest this be more flexible by adding something like `text_prefix`, `text_postfix` instead? That way, doing `text_prefix="\n\r"` etc would accomplish the same thing while allowing the user the ability to customize as they see it. I think for it to be the most flexible, I guess there should be a `text_separator_char` as well, to mark what the separator between the node text and the options ("default is `"_"`).

ahmedcharles:
I'd much rather have a util which formats an arbitrary string with _'s before and after, which could be used outside EvMenu and just have this display the text as is, potentially with newlines between the text and the list of options.

Griatch:
@ahmedcharles 

Hm, How about a `node_formatter` keyword then, which allows you to supply it a callback with whatever formatting you want. It should have the callsign `callback(nodetext)` and return the text in whatever formatting you want, including the separator. And if it's given, the default formatting/line managing is turned off with no added line breaks or spaces.

PR #828, opened 2015-10-17
Refactor EvMenu so it doesn't use 'look' anymore.

ahmedcharles:
This should be completely backwards compatible.

It's split into multiple commits to make reviewing easier.


Griatch:
Merged with rebase.

PR #830, opened 2015-10-17
Fix idle timeouts.

ahmedcharles:
The fix for rate limiting removed the code which updated cmd_last,
which is used to determine whether someone is idle.


Griatch:
Merged with rebase.

PR #831, opened 2015-10-18
Update menu_login to use EvMenu.

ahmedcharles:
This relies on the EvMenu updates, so those should be merged first.

### DIFF
--- a/evennia/contrib/menu_login.py
+++ b/evennia/contrib/menu_login.py
@@ -1,14 +1,14 @@
 """
-Menu-driven login system
+EvMenu-driven login system
 
-Contribution - Griatch 2011
+Contribution - Ahmed Charles 2015
 
+This is heavily influenced by the previous menu login by Griatch.
 
 This is an alternative login system for Evennia, using the
-contrib.menusystem module. As opposed to the default system it doesn't
-use emails for authentication and also don't auto-creates a Character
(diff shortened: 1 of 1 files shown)


Griatch:
This version of menu_login seems to have severely diminished functionality compared to the old non-EvMenu version, and as such I can't accept it in this form. 

The old menu_login did validation at every step and asked to to correct your input before continuing. It also allowed you to return back to the root menu to start over if you changed your mind by giving an empty input (which will never be a valid playername/password anyway).

This one will allow any input, including empty space and then checks the total input only at the end - which leads to abruptly dropping connection completely if any of the previous outputs were wrong - this is very frustrating to the end user and now how menu login systems work normally.

ahmedcharles:
My main goal here was simplifying the flow of the code.

I'll make some changes and test them and submit an updated PR.

ahmedcharles:
I'm refactoring the rest of the login code so this will hopefully get easier.

Griatch:
Any progress on this refactoring?

ahmedcharles:
I've been busy, but I'll take a look this weekend, hopefully.

Griatch:
This still seems like a very useful update, hope it's still in the works.

Griatch:
This was instead implemented from scratch by vlegoff in  #961, so closing this.

PR #832, opened 2015-10-18
Move to python3 style except.

ahmedcharles:
We already use this syntax in 3 places, so these should be fairly safe.


Griatch:
Merged with rebase.

PR #834, opened 2015-10-18
Move to python3 style exec.

ahmedcharles:
Most of the places that exec is used, uses exec() style, already.


Griatch:
Merged with rebase.

PR #835, opened 2015-10-18
Move to python3 style raise.

ahmedcharles:
Most of the places that use raise, use this still, already.


Griatch:
Merged with rebase.

PR #836, opened 2015-10-18
Remove commented out print statements.

ahmedcharles:


Griatch:
Many of these are very useful debug statements for quick checks. But I guess it does make the code cleaner to remove them when not used.

ahmedcharles:
I'm trying to consolidate all of the logging so it uses utils.logger, if possible.

Unfortunately, there doesn't seem to be a way to setup logging so that it only logs some messages some of the time based on log level. I'll look at the twisted docs and see how easy it would be to set those up and then we could add a log_debug that is off by default but can be turned on for development.

But if that existed, I'd still want it to be more structured, such that we wouldn't end up with debug logs all over, with lots of inconsistencies.

Griatch:
print debugging is almost exclusively for my personal devel use so it's a good point to remove them from view. Merged with rebase.

PR #838, opened 2015-10-19
Printing traces to the user is usually bad.

ahmedcharles:
This could, in theory, expose more information about how the server
works than would be advised, benefiting potential attackers.


Griatch:
Hm; complete hiding of error is also pretty annoying to the user. But it's a valid point.

ahmedcharles:
Well, we could suggest in the message that developers look in the log, since all of the tracebacks should be there.

Griatch:
Nah, it makes sense to hide it. Merged with rebase.

PR #842, opened 2015-10-19
Move to python3 style division.

ahmedcharles:
Note: / and // have the same behavior in python2.

### DIFF
--- a/evennia/commands/cmdparser.py
+++ b/evennia/commands/cmdparser.py
@@ -5,6 +5,7 @@ replacing cmdparser function. The replacement parser must accept the
 same inputs as the default one.
 
 """
+from __future__ import division
 
 import re
 from django.conf import settings
--- a/evennia/commands/default/system.py
+++ b/evennia/commands/default/system.py
@@ -3,6 +3,7 @@
 System commands
 
 """
+from __future__ import division
 
 import traceback
 import os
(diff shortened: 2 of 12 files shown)


Griatch:
@ahmedcharles 

No they don't: 
```python
print 1 / 2.0
<<< 0.5
print 1 // 2.0
<<< 0.0 
```

Python does automatic type conversion if one of the divisors is a float. // turns off this functionality. What 3.0 changes is to change integer divisors to always return floats (contrary to Py2). 

So these changes may be ok, but if // is used explicitly it was to enforce a particular form of divisor behavior which you leave up to the divisor types if you use /.

ahmedcharles:
Ah, this is more nuanced than I thought. I'll review them more carefully.

Russell-Jones:
Have a look at [past.utils.old_div](http://python-future.org/compatible_idioms.html?highlight=old_div#division)

ahmedcharles:
I think I actually got it correct this time. Mind talking a look?

Griatch:
I don't know exactly how prettytable is supposed to work since that's a third-party lib we include, but it looks ok at a glance. Will do some more tests though.

ahmedcharles:
I could remove the change from prettytable, though that wouldn't mean it would work correctly under python3 anyways. What about the other changes?

Griatch:
I've ran this through a few tests and prettytable seems to display correctly as well, along with the websocket system used by the webclient. So merging it in. :)

PR #854, opened 2015-10-24
Fix bug in examine.

ahmedcharles:
Examine sorts the original stack used by the object that it examines,
which is bad.


Griatch:
Merged with rebase

PR #860, opened 2015-11-01
at_post_puppet should be called after cache_lock_bypass.

ahmedcharles:
at_post_puppet does a look, which results in a perms check, which is wrong
for the superuser because the lock bypass is not up to date.


Griatch:
Merged with rebase.

PR #862, opened 2015-11-01
Use comprehension rather than map.

ahmedcharles:


ahmedcharles:
This is for python 3 compatibility.

Griatch:
Merged with rebase.

PR #865, opened 2015-11-01
Use logger rather than print for these messages.

ahmedcharles:


Griatch:
Good point. This is very old code. Merged with rebase.

PR #867, opened 2015-11-01
Move normal login code into it's own function.

ahmedcharles:


evennia/commands/default/unloggedin.py
```diff
@@ -118,6 +117,45 @@ def guest_player(session):
         return True, new_player
 
 
+def normal_player(session, name, password):
```
Griatch: A more instructive name would be `create_normal_player`. This also needs docstring.

ahmedcharles:
Updated.

Griatch:
Merged with rebase.

PR #868, opened 2015-11-01
Import python3 compatible object.

ahmedcharles:


Griatch:
What is the difference? And what happens when you mix this "py3-object" with the normal one used in py2 (like every end user is bound to do)?

ahmedcharles:
Unfortunately, the docs aren't super clear. I assume it's supposed to work, though I'm not exactly sure how users objects would interact with evennia's objects in an incompatible way?

1. If they call a function on an evennia object from their own object, then it should probably work? If that doesn't work, then this could never be a transitional approach.
2. If they inherit from evennia's objects, they will inherit from this new object as well, so it should also work.

Which other interactions would there be?

Griatch:
I guess it's less mixing and more just how critical the new-style is. We are importing user's modules and run them as part of the serve r- it's a big deal for our functionality; does people using old-style objects in an imported module suddenly make the running server *non*-py3 compliant on the fly?

ahmedcharles:
I would hope not. If that was the case, using this library would be pointless, right?

Though, I admit, it's not obvious that it is supposed to work, since the documentation is lax.

Griatch:
@ahmedcharles Maybe, not all python programs dynamically load third-party modules the way Evennia does either though. But ah well, one problem at a time I guess.

ahmedcharles:
https://github.com/PythonCharmers/python-future/blob/master/src/future/types/newobject.py

That's the implementation. The only potential issue I see is __str__ returning unicode, potentially. Granted, I'm not sure how great evennia's unicode story is currently, I just know that it's hard in general.

Griatch:
We store everything in unicode behind the database. We are not using explicit unicode in other cases. `__str__` returning unicode may be an issue, not sure about that.

ahmedcharles:
Well, there is always testing, though I'm not sure what the right approach would be.

Griatch:
I don't see any particular errors with this. Merged with rebase.

Russell-Jones:
See also https://github.com/PythonCharmers/python-future/issues/176 "future.builtin.bytes calls __str__() not __bytes__()"
My preference would be to use b"" or bytes() (2 and 3) when we mean bytes and "" or str() (2 and 3 with future.builtins / from __future__ import unicode_literals) when we mean text, but I think we likely need the ability to use def __bytes__(self): in 2 for that to work (I'd like to use that in ANSIString).

PR #869, opened 2015-11-01
Import python3 compatible cmp.

ahmedcharles:
I'm not exactly sure what the difference is between python3's cmp and python2's cmp, so I'm just putting this out there for discussion, though I'm sure the code works. :)


Griatch:
Merged with rebase.

Russell-Jones:
Ideally, use of cmp would be replaced http://python3porting.com/preparing.html#keycmp-section

PR #870, opened 2015-11-01
Don't use long, since it doesn't exist in python3.

ahmedcharles:


Griatch:
Luckily long is never actually used in Evennia besides those checks. Merged with rebase.

PR #872, opened 2015-11-02
Have every class inherit from object.

ahmedcharles:


Griatch:
Using Meta this way is a Django style (I don't think I've ever seen Meta declared like this in any tutorial) and I'm not sure it's actually *needed* to have it explicitly object in py3 - everything's a new-style class anyway? Anyway, seems to work like this. Merged with rebase.

PR #873, opened 2015-11-02
Replace filter with if/comprehension.

ahmedcharles:


Griatch:
`filter` is a pretty rare use case in Evennia, we've used list comprehensions for the most part, good catch. Merged with rebase.

PR #878, opened 2015-11-09
Text2html fix

feend78:
This PR addresses an as-yet unreported bug, as far as I can tell.

I was working on putting together an Ainneve-specific description for the Limbo room, incorporating web links to certain resources, when I found some issues with the regex used to convert URLs to links in the text2html parser. The behavior I saw was that any URL ending in s, p, a, or n was having its link cut off early, and for URLs followed only by whitespace and a newline, the link was wrapping including the first characters on the following line. Also, URLs beginning with `https` were not recognized. 

**Changes:**
* Modified the regex to recognize URLs beginning with `https`
* Fixed the regex boundary problems and converted it to compiled.
* Moved the compiled URL regex out to the module level, similar to the others used within.
* Updated function `convert_urls` to use the new compiled regex.
* Wrote unit tests for the modified function. Then, _really_ fixed the boundary errors in the regex.

### DIFF
--- a/evennia/utils/tests.py
+++ b/evennia/utils/tests.py
@@ -230,3 +230,82 @@ class TestMLen(TestCase):
 
     def test_dict(self):
         self.assertEqual(utils.m_len({'hello': True, 'Goodbye': False}), 2)
+
+
+from .text2html import TextToHTMLparser
+
+class TestTextToHTMLparser(TestCase):
+    def setUp(self):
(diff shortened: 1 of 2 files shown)


Griatch:
Cool catch, especially the unit tests are very appreciated and useful for things like this. Your patch works fine as far as I can tell thanks for the contribution!

Merged (with a rebase, so closing this manually)
.
Griatch

PR #879, opened 2015-11-09
Use list* from future.utils.

ahmedcharles:
dict.keys() -> list(dict)
dict.values() -> listvalues(dict)
dict.tems() -> listitems(dict)

### DIFF
--- a/evennia/commands/cmdset.py
+++ b/evennia/commands/cmdset.py
@@ -26,11 +26,11 @@ Set theory.
     to affect the low-priority cmdset.  Ex: A1,A3 + B1,B2,B4,B5 = B2,B4,B5
 
 """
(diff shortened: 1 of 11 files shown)


ahmedcharles:
This process was mostly manual rather than automated, in order to pick between the list and view versions of the functions. This patch should be safe in that it only creates lists rather than generators, which may use more memory, but should always be safe to do.

Griatch:
As far as I can tell, this patch creates lists only for cases where the list itself is stored, so it should use the same memory as before, no?

ahmedcharles:
Yes. I meant the comment more in general, rather than specific to this commit. The next commit, which will use viewitems/viewkeys/viewvalues, will have that tradeoff, because in many cases it will be going from producing a list to producing a generator.

PR #884, opened 2015-11-15
Move logger to core.

ahmedcharles:


Griatch:
@ahmedcharles 

Motivation?

ahmedcharles:
A subset of the motivation is in the README.md file. The entire motivation is something like this:

Currently, it's hard to reason about some modules in evennia due to not knowing what the dependencies of that module are. The utils module is particularly 'bad' in this sense, because it has things like logger that are supposed to have as few dependencies as possible while also having things like EvMenu and EvTable, which depend on almost all of the rest of evennia.

Longer term, I'd like to have Evennia be split in a set of modules with known 'layering'. Layering being the term for having groups of functions/classes/etc split in layers, where any given layer can only depend on the layer below it. Like, building a multi-layer cake or something. :)

Griatch:
@ahmedcharles 

Inter-depencies is of less convern than the API presented to the user. This sounds like you have ideas for a bigger (very big) structural changes down the line. Could you maybe elaborate some more on how you envision this "layered" module structure to look in the end?

ahmedcharles:
I'm not sure I understand the distinction between user API and non-user API, since all of the code I've changed in Evennia so far has formed some part of the user API. Is there some other code that I'm not seeing that's not part of the user API that I should know about? Perhaps the code in the server directory related to the portal isn't? (My assumption being that most of Evennia is user API.)

My ideas so far are somewhat small and involve taking things with few dependencies and separating them so it's easy document that and keep them so. When there are no more 'obvious' changes, I'm mostly interested in documenting the dependency relationships between things so that they are more of a conscious design decision rather than purely organic in how they happen.

So, ultimately, it should be lots of incremental improvements which make things clearer over time. And hopefully, in a way that doesn't break too many users.

And obviously, if this gets accepted, my next step is moving all internal uses of logger to using this.

Griatch:
@ahmedcharles 

It's all API. Technically some stuff is less likely to be useful to an Evennia developer than others but there is no established "off-limits" parts.

 I was more trying to understand the reasoning of re-structuring *anything* by order of inter-dependencies. Inter-dependencies (i.e. that logger sits in `utils` despite having less dependencies than other modules) seems utterly irrelevant to the API - in the sense of the user going to utils.logger and using the resources there - what does it matter to the user if the module they use have few or many dependencies with other modules? 

You write 

> "Currently, it's hard to reason about some modules in evennia due to not knowing what the dependencies of that module are." 

I guess I'm looking for more elaboration on what you find hard or what you want to express but cannot.

ahmedcharles:
Oh, the higher level goal is to avoid circular dependencies, in general. It's not so much that users care if the module has lots of dependencies, but they do care if that module has circular dependencies that results in their code not working.

An example of such a circular dependency is the way execute_cmd("look") was used in various places and now it isn't. Granted, that's not a circular import dependency, but the same could happen with those.

Ideally, one can strip away the 'upper' layers of evennia and use only the 'lower' layers and have it still work. Unfortunately, right now, it's hard to prove which layers are above or below which other layers.

One of my goals (when starting on writing a game) was to be able to replace/remove every command in evennia, so it would be possible to have a game that does nothing. That's possible now, but wasn't before. As you add more commands back, so login is possible, I'm not sure which other things break.

Similarly, what happens if I try to do this with other things. What things break?

And in the end, I hope this makes things slightly more testable.

Griatch:
Hm. I'm fine with cleaning up the code structure as needed, and it's excellent that you are willing to look over code this way! We have indeed done such refactors in the past in order to clean out too-organic growth. I'm not sure I'm comfortable with breaking out new folders like "core" without fully grasping what the "layers" you refer to are, though. Evennia's inter-dependencies is a pretty complex affair and it sounds like a refactor on the form you seem to envision would lead to more reshuffling down the line.  

It sounds like it would be a better solution (for now) to explore this refactor in a separate fork and merge it into master if/when it is at a point where it is more complete and the advantages of the restructuring is more clear. That way, one could potentially toy with refactoring without intermediary care for end users. 

I'd be fine with keeping this PR open to follow the progress of this if you are ok with this solution.

ahmedcharles:
Sure, let's see where it goes.

Griatch:
It has been four months. If there is no activity on this, maybe the PR should be closed for now?

ahmedcharles:
I'll reopen at some point if/when I get back to this. Thanks.

Griatch:
Ok, closing for the time being.

Issue #912, opened 2016-01-09
channel aliases not displayed in @channels command output


orkim:
When joining a channel you can add an alias with `addcom <alias> = <channel>`.

Even after being joined to a channel you can repeat this command to add multiple aliases if you like.

However, once an alias is added, the output from the `@channels` command, in the "my alias" column shows the channel name, and not the defined alias(es).

With multiple aliases the channel name will be repeated for each one added.

If you do know the alias (since you can't view them) you can still remove them with "delcom <alias>" just fine.

I would expect to be able to view the aliases listed with the @channel command.

This is a fresh install with no custom code.

orkim:
fixed with PR #915

PR #915, opened 2016-01-11
fix alias display on @channels/comlist

orkim:
I believe this is a fix for issue #912 which addresses the output of the `@channels` and `comlist` commands.  Please review and let me know if I did something wrong.


Griatch:
Looks good! Merged.

Issue #911, opened 2016-01-09
default "nick" command fails to list all defined nicks


orkim:
On a fresh install I was playing around with the "nick" command.  After reading the help file I entered the following commands (lines starting with "###" are the commands issued):

    ### nick/list
    No nicks defined.
    ### nick hi = say Hello there.
    
    Nick set: 'hi' -> 'say Hello there.'.
    ### nick/player user = long user name
    
    Nick set: 'user' -> 'long user name'.
    ### nick/object rock = a big boulder
    
    Nick set: 'rock' -> 'a big boulder'.
    ### nick/list
    Defined Nicks:
    +-----------+----------+------------------+
    | NickType  | Nickname | Translates-to    |
    +-----------+----------+------------------+
    | inputline | hi       | say Hello there. |
    +-----------+----------+------------------+

As you can see, the "nick/player" and "nick/object" do indeed show as set, but are not listed with "nick/list" as one would expect.

It makes no difference if the object/player actually exists or not.

orkim:
fixed with PR #916

PR #916, opened 2016-01-11
display nicks from all categories

orkim:
This should make the `nick/list` command display the defined nicks from each category.  Please let me know if there are any issues with implementation.  This addresses issue #911 completely.


Griatch:
Seems to work fine, thanks for the PR! Merged.

PR #922, opened 2016-02-08
Internationalization for comms

kaihuang201:


Griatch:
@kaihuang201 

Thanks for the PR! The changes to the commhandler looks good and correct. Note however that you are including an already merged commit  in this PR, namely 94023d9. I can fix this on my end, but in the future it's easiest if you make a new branch off master for new features as you add them.

Griatch:
[Merged with rebase](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) (which means your branch will be out of sync with master, so create a new branch for further developments).

PR #929, opened 2016-03-13
Add .at_rename() hooks to Objects and Players.

volundmush:
Hooks! We can always use a lot of these. These would be nice to have because it ties in with some of my Athanor code and I'm sure other people might find it useful. No reason not to have these!

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -1126,9 +1126,11 @@ class CmdName(ObjManipCommand):
                     if not obj.access(caller, "edit"):
                         caller.mgs("You don't have right to edit this player %s." % obj)
                         return
+                    oldname = str(obj.username)
                     obj.username = newname
                     obj.save()
                     caller.msg("Player's name changed to '%s'." % newname)
+                    obj.at_rename(oldname, newname)
                     return
             # object search, also with *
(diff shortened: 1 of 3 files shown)


Griatch:
I agree an `at_rename` hook can be useful. Also good docstrings and code overall!

 However, I don't think it's all that useful to call a default on-typeclass hook (which is a "core" thing that game devs should not need to change) from a default command (which is meant to be overridden). If we do it this way, we need to make some sort of pseudo-API docs for default commands that say 

> "Oh, and if you ever make your own `@name` command, make sure to always call `at_rename`".

This is not very clean. Also, what happens when you, in a command body (somewhere else than in `@name`) does `target.key = "The ENRAGED orc captain"` or something - the dev should usually expect that to trigger `at_rename` as well. 

In short, this call should not be up to each dev to remember to do. Core should always do the call regardless of what game commands are used. 

So I think this should change from calling this hook from `@name` to instead be called whenever the `key` property on the typeclass changes. That way, *any* change to `key` will trigger the hook. Evennia's metaclasses wrap the `db_key` field to `key` in the idmapper, but if you were to make an actual `key` get/set/delete override in typeclasses/models.TypedObject (for example using `@property`), Evennia will use that rather than its own wrapper.

PR #930, opened 2016-03-13
Character conn idle

volundmush:
Tried to do this earlier but didn't quite understand my git-fu at the time.

Adding a few handy methods to DefaultCharacter. Seeing no reason that they shouldn't have these! If people want to use different logic they can just overload 'em.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -1126,9 +1126,11 @@ class CmdName(ObjManipCommand):
                     if not obj.access(caller, "edit"):
                         caller.mgs("You don't have right to edit this player %s." % obj)
                         return
+                    oldname = str(obj.username)
                     obj.username = newname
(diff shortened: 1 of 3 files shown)


Griatch:
The idle_time and connection_time looks useful, thanks! 

From a git-fu standpoint, it looks like you have based this branch off your at_rename branch, which mixes up the two PRs. To completely separate your features you need to base each branch off your "clean" master. No worries now though, I'll just git cherry-pick the relevant commit. :)
.
Griatch

Griatch:
Testing this code in master branch leads to a traceback: 

```
idle = [session.cmd_last_visible for session in self.sessions]
<<< TypeError: 'ObjectSessionHandler' object is not iterable
```

`self.sessions` is a handler, you need to use its `get` method to use it.

PR #934, opened 2016-03-19
msg-changes to evennia/commands/default/player.py

BlauFeuer:
## Propose CmdOption session option msg be sent to current session only.

Since the `@option command affects sessions separately, sending the output of the command to other sessions causes confusion with messages like **Encoding was changed from 'utf-8' to 'latin-1'.** or **Screen reader mode was turned on.** being sent to a session that was not receiving the option changes.

This change only affects the `@option command when in multi-session mode.

### DIFF
--- a/evennia/commands/default/player.py
+++ b/evennia/commands/default/player.py
@@ -406,11 +406,11 @@ class CmdOption(MuxPlayerCommand):
             sencodings = settings.ENCODINGS
             string += " Custom: %s\n Server: %s" % (pencoding, ", ".join(sencodings))
             string += "\n{wScreen Reader mode:{n %s" % self.session.screenreader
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks for the PR! 

> I merged with cherry-pick to avoid the unnecessary merge-commit in your version history. Since I saw you made your changes in your master branch, this will likely lead to a conflict when you pull down the upstream changes (easiest way to fix this is to revert your master branch so it's in sync and then apply the change on that, google is your friend). In the future you'll probably have an easier time if you make new branches for any changes/PRs you do and always keep your clone of master clean compared to upstream (see also [here](https://github.com/evennia/evennia/wiki/Version%20Control).

Anyway, thanks again. Closing.

PR #936, opened 2016-03-26
Added support for SSL IRC connections.

cptmashek:
Adding the ability to connect to IRC with SSL.

The IRC server I use is SSL-only.

Introduces the dependency of `pyopenssl`.

### DIFF
--- a/evennia/commands/default/comms.py
+++ b/evennia/commands/default/comms.py
@@ -755,7 +755,8 @@ class CmdIRC2Chan(MuxCommand):
     link an evennia channel to an external IRC channel
 
     Usage:
-      @irc2chan[/switches] <evennia_channel> = <ircnetwork> <[+]port> <#irchannel> <botname>
+      @irc2chan[/switches] <evennia_channel> = <ircnetwork> <port> <#irchannel> <botname>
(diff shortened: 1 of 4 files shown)


Griatch:
Cool, thanks for the PR! Some comments: 

Firstly, `pyopenssl` should not be added as a requirement. The way we've done with existing SSL ports is to treat it as an optional. In our existing SSL-telnet implementation we simply check for the library and warn that it's needed to continue with SSL support. This could be added in the same way for the `@irc2chan` command, whenever SSL is desired. Something like: 

```python
try:
    import OpenSSL
except ImportError:
    self.caller.msg("To use SSL, the PyOpenSSL library must be installed.")
    return
```

As for the command syntax, `+port` looks nice, but it's not consistent with any other use of SSL elsewhere (SSL is a separate port for Evennia and does not use + to denote itself). Default commands tend to use the `/switches` for these kind of extra features, so maybe an `/ssl` switch would be more in line with other commands?

cptmashek:
Oh right, I saw that in the code but it escaped me to do the same thing. Thank you! :D 

I will change the `+port` to `/ssl`, since it does make a lot of sense.

PR #944, opened 2016-04-04
Introducing the Evennia Game Directory service contrib.

gtaylor:

This PR adds a preliminary-but-functioning [Evennia Game Directory](http://evennia-game-directory.appspot.com/) service contrib. 

The idea was to have an opt-in contrib that automatically sends details on your game to a central listing. This will increase discoverability and also give us some interesting metrics for seeing what Evennia is doing out in the wild.

As of now, the set of things that we are sending is very limited. See [_form_and_send_request()](https://github.com/gtaylor/evennia/blob/gamedir_client/evennia/contrib/gamedir_client/client.py#L61). We don't show all of these in the directory, and I'd expect there to be a bunch of other things that we could add.

To see how to set this up and configure it, refer to the [README](https://github.com/gtaylor/evennia/blob/gamedir_client/evennia/contrib/gamedir_client/README.md). Since this is an early MVP, we don't have very granular configuration. Future development will be dictated by how many actually end up using this thing.

The matching server for this client is located [here](https://github.com/gtaylor/evennia-game-directory). We are currently acting incredibly naively in that we're accepting any and all properly formed requests as truth. If this ends up seeing use, I can pitch in some notion of ownership and listing management. This was a quick weekend project to get something out the door to the users.

### DIFF
--- a/evennia/contrib/gamedir_client/README.md
+++ b/evennia/contrib/gamedir_client/README.md
@@ -1,0 +1,139 @@
+# Evennia Game Directory Client
+
+Greg Taylor 2016
+
+This contrib features a client for the [Evennia Game Directory]
+(http://evennia-game-directory.appspot.com/), a listing of games built on 
+Evennia. By listing your game on the directory, you make it easy for other 
+people in the community to discover your creation.
+
+*Note: Since this is still an early experiment, there is no notion of 
+ownership for a game listing. As a consequence, we rely on the good behavior 
+of our users in the early goings. If the directory is a success, we'll work 
+on remedying this.*
+
+## Listing your Game
+
+To list your game, you'll need to enable the Evennia Game Directory client. 
+Start by `cd`'ing to your game directory. From there, open up 
+`server/conf/server_services_plugins.py`. It might look something like this 
+if you don't have any other optional add-ons enabled:
+
+    """
+    Server plugin services
+    
+    This plugin module can define user-created services for the Server to
+    start.
+    
+    This module must handle all imports and setups required to start a
+    twisted service (see examples in evennia.server.server). It must also
+    contain a function start_plugin_services(application). Evennia will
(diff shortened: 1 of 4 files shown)


Griatch:
This looks awesome! I can't examine it in detail right now, but the idea is great. If not already in place, one thing that sounds reasonable is to relate this to the setup of the [MSSP protocol](https://github.com/evennia/evennia/blob/wclient/evennia/server/portal/mssp.py), which is the existing telnet solution that MUD-listing sites use. All of those settings may not be relevant but it's probably as close to a "standard" as the MU*-world has.

gtaylor:
I had thought about at least matching some of the MSSP key names, but ended up diverging a bit since this is very much focused on Evennia (with no goal to be general-purpose). Many of these variables would not be sent in the future once listings became managed by the directory itself (more on this below).

What I'd expect as far as potential paths forward:

* Throw this to the users. Make it clear that this is experimental and may change in breaking ways (not as in breaking their games, but the sending may break).
* See if we get any kind of adoption at all (we might not).
* Figure out what kind of info we'd like to show on the directory as we go.
* Figure out if there is any other info we'd like to record for your benefit but not show (OS + Python versions, etc).
* Eventually, if this got serious enough we'd need to implement listing ownership and security. This would prevent listing over-writing and allow us to stop sending a bunch of what we are sending right now with every request.

This is one of those low-effort tinker projects I dig. I'd be glad to tinker with this some and see how we do. I'm running it on App Engine right now because we'll probably fit in the free tier, thus "guaranteeing" a pretty stable hosting situation for this. Nobody likes stuff randomly vanishing when the bill stops getting paid.

Griatch:
@gtaylor 

Cool, as it's a contrib that does not affect any other systems, I see no reason not to merge it and see how it goes. :)

Griatch:
Merged with rebase.

Issue #924, opened 2016-02-15
Task: Devel "wclient" branch: Things to do (and help with)


Griatch:
The wclient branch deals with the webclient and web infrastructure update. It is described in detail [here on the mailing list](https://groups.google.com/forum/#!msg/evennia/xWQu_YVm14k/0wdXvvr5GwAJ). 

## Fun / easy things to help with

- [x] Go over the web/webclient/ HTML, JS and CSS files to make the default (simplistic) webclient more cross-browser compliant and following best practices. This could be multiple steps, but an experienced web developer should be able to help out a lot here. 
  - Look over the JS/HTML/CSS files to see if they follow "best pracices" (whatever that may be). Especially when it comes to extendability and performance. Our goal is to offer a good, performant baseline for people to read and extend.
 - I kept the dependencies to jquery only and tried to not use JS things that are only available in the latest browsers. But I did this just by the seat of my pants - there are for example no compatibility exceptions anywhere or any real feeling for what the minimal browser version is. Our base client (especially the evennia.js library) should ideally offer as good backwards compatibility as possible (we can after all fall back to AJAX for browsers not supporting websockets).
  - After seeing what can be done about compatibility, we should be able to state just what the minimum browser versions are, we can't do this at the moment.
The client, simple as it is, will currently not rescale well for different screen sizes and shapes - the inputline should stably stay at the bottom and the scrolling text always end up in the right area regardless of client or platform. I'm sure this is a CSS issue I'm not clear on.
  - The goal with future versions of the web client will be to allow for extending the GUI component with multiple windows people can send messages/data to (via added gui hooks). This is not critical for this first iteration of the client, but if anything could be done to ease future expansions already at this point, it should be kept in mind.
- [x] At the protocol level, remap GMCP standard handshakes (Core.Hello and Core.Supports.Set) into Evennia inputfuncs. 
- [x] At the protocol level, remap [MSDP standard commands](http://tintin.sourceforge.net/msdp/) (SEND, LIST etc) into Evennia inputfuncs.

## Fun but tricky/involved things to help with

- [x] Add/update server/inputfuncs.py with many more useful inputfuncs, especially for answering default GMCP/MSDP commands (like requesting all commands available, monitor an attribute etc).
- [x] Update evennia/scripts/monitorhandler.py to remove all OOBHandler references, make it more useful as a resource for inputfuncs to use. 

## Maybe less fun things to help with

- [x] SSH protocol converted to new Protocol Infrastructure
[...]

PR #948, opened 2016-04-10
WebClient fixes and enhancements

NatePri:
This work is in reference to [issue 924](https://github.com/evennia/evennia/issues/924).

- Fixed issue with input area positioning incorrectly on window resize
- Added "use strict" to the scripts, fixed some incorrectly scoped variables as a result.
- Made some CSS changes to improve the input area look & feel
- Changed onKeyDown handler to only do history navigation when the cursor is at the end of the line. (May be less necessary with the following change.)
- Added ability to save the currently editing text in the history stack (so that stray up/down arrows won't kill what you've been typing).
- Added input area resize-on-type functionality.

Make the web client easier to deal with, both from the user and developer perspective.

The AMD loader changes I mentioned on the mailing list will be in a separate pull request.

### DIFF
--- a/evennia/web/webclient/static/webclient/css/webclient.css
+++ b/evennia/web/webclient/static/webclient/css/webclient.css
@@ -124,7 +124,7 @@ div {margin:0px;}
 #inputfield, #inputsend, #inputsizer {
     display: block;
     box-sizing: border-box;
-    min-height: 50px;
+    height: 50px;
     background: #000;
     color: #fff;
     padding: 0 .45em;
--- a/evennia/web/webclient/static/webclient/css/webclient.css
+++ b/evennia/web/webclient/static/webclient/css/webclient.css
@@ -97,6 +97,8 @@ div {margin:0px;}
     position: absolute;
     overflow: auto;
     padding: 1em;
+    -webkit-box-sizing: border-box;
+    -moz-box-sizing: border-box;
     box-sizing: border-box;
     top: 0;
     left: 0;
@@ -123,6 +125,8 @@ div {margin:0px;}
 /* Input field */
 #inputfield, #inputsend, #inputsizer {
     display: block;
+    -webkit-box-sizing: border-box;
+    -moz-box-sizing: border-box;
     box-sizing: border-box;
     height: 50px;
     background: #000;
@@ -132,20 +136,20 @@ div {margin:0px;}
     font-family: 'DejaVu Sans Mono', Consolas, Inconsolata, 'Lucida Console', monospace;
 }
 
-#inputsizer {
-    margin-left: -9999px;
-    line-height: normal;
-}
-
-#inputfield {
+#inputfield, #inputsizer {
     float: left;
     width: 95%;
     border: 0;
     resize: none;
+    line-height: normal;
 }
 
 #inputfield:focus {
     outline: 0;
+}
+
+#inputsizer {
+    margin-left: -9999px;
 }
 
 /* Input 'send' button */
--- a/evennia/web/webclient/static/webclient/css/webclient.css
+++ b/evennia/web/webclient/static/webclient/css/webclient.css
@@ -8,7 +8,7 @@
 --- */
 
 /* Overall element look */
-html, body { height: 100% }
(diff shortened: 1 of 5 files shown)


Griatch:
This looks great! Just the kind of input and fixes needed from someone who knows more about web development. I'll look this over and test it tonight, but on a first glance it looks very good. :+1:

evennia/web/webclient/static/webclient/js/evennia.js
```diff
@@ -50,6 +50,7 @@ An "emitter" object must have a function
 */
 
 (function() {
+    "use strict"
```
Griatch: I was not aware of "use strict", it makes sense to use it. Out of curiosity, my understanding is that "use strict" can be used at the file level rather than the function level. Is there some reason not to put it on the file level?

evennia/web/webclient/static/webclient/css/webclient.css
```diff
@@ -94,31 +94,54 @@ div {margin:0px;}
...
-    padding: 0 .8em;
-    overflow: auto
+    width: 100%;
+    padding: 0;
 }
 
 /* Input field */
+#inputfield, #inputsend, #inputsizer {
+    display: block;
+    box-sizing: border-box;
```
Griatch: I know I looked at box-sizing for the original code but got the impression it was a relatively new css feature. Then again, it all comes down to where to put the lowest browser denominator, so I guess this is okay. I did get the impression that one needs vendor prefixes to handle medium-old browsers though, such as mentioned here: https://css-tricks.com/box-sizing/#article-header-id-7

evennia/web/webclient/static/webclient/js/webclient_gui.js
```diff
@@ -76,36 +96,91 @@ function doSendText() {
...
+        var scrollh = inputfield.prop("scrollHeight");
+        var clienth = inputfield.prop("clientHeight");
+        var newh = 0;
+        var curr_text_len = inputfield.val().length;
+        
+        if (scrollh > clienth && scrollh <= max_height) {
+            // Need to make it bigger
+            newh = scrollh;
+        } 
+        else if (curr_text_len < prev_text_len) {
```
Griatch: This looks like an interesting solution! Out of curiosity, is using an offscreen element really the only way to measure the size of a text? Shouldn't this somehow be dependent on inputsizer having the same css properties as the field we are trying to measure as well (it doesn't look like it does)? Not saying it's wrong, just confused about how it works. :)

evennia/web/webclient/static/webclient/css/webclient.css
NatePri: Ah yes, I'll add those. I'm used to the CSS preprocessor I use at work, which adds them automatically. But yeah, the vast majority of browsers can handle it these days, and even IE8 had support for it. There's a scary-looking warning on it for MDN, but in practice I see it used quite a bit.

evennia/web/webclient/static/webclient/js/evennia.js
NatePri: The reason for not doing it at the file level is mostly to do with using script minifiers; if you ever use a minifier that concatenates all the script into a single script, then you're forced to run all of that script in strict mode. This would be fine for us currently, but if you ever take a dependency on a script that doesn't run in strict mode you wouldn't be able to minify it together with the rest of your scripts.

In any case, it's a best practice to enclose JS in a self-executing function to avoid leaking globals.

evennia/web/webclient/static/webclient/js/webclient_gui.js
NatePri: That I'm aware of, yeah. The only alternate solution I can think of would be to figure out the metrics for whatever font is being used and then use the character count, but this would get hairy and wouldn't account for line-height. I'll make the CSS a bit more consistent; at the moment with the current set of CSS rules, it works fine as written (the extra rules don't affect the dimensions of the sizer or textarea differently), but it'd be too easy to slip a rule in later to throw things off as it stands, so I'll fix that.

Issue #949, opened 2016-04-11
Unable to use "l" as an option in a default EvMenu


shollen:
EvMenu redisplays a node when the single letter 'l' is supplied as input to a default EvMenu node. 

1) To reproduce the "bug," set up a default node, like...
```
def node0(caller):
    caller.msg("\nWhat is your first name?")
    options = ({"key": "_default",
                "goto": "node1"})
    return 'test', options
```
2) Start your menu with the @py command: @py from evennia.utils.evmenu import EvMenu; EvMenu(self, "path to the module containing node0", startnode="node0")

Instead of moving on to node1 with 'l' as the raw_string, the code redisplays 'test' and leaves execution in node0.

I'm using the latest version of Evennia on OSX 10.11.3.

Griatch:
I would think the string "look" causes the same effect. Also "q" and "quit" will exit the menu with this configuration. This is expected behavior with the current design: the idea being that one should always be able to re-show the node or quit the menu regardless of the state one is in. 

The automatic `q/uit` command can be turned off when initializing the evmenu, but `l/ook` currently cannot. Maybe the best way to handle this is to allow the user to turn off the auto-look in the same way. That means it's up to the menu creator to handle how nodes are re-shown.

shollen:
Yes. You’re exactly right. I replied just as you did.

I’’ll take a look at implementing an auto-look toggle.

Thanks.

-Shan

Griatch:
Will be closed when wclient branch merges.

PR #959, opened 2016-04-17
Adds allow_look option to EvMenu. 

shollen:
Solves the issue of the default EvMenu 'look' command interfering with custom code for permutations of ‘look’ in a default node.

I added an allow_look option similar to the allow_quit option, but without the option of a replacement function. allow_look defaults to True.

I implemented this change when I found that I couldn't use the single letter 'l' as input  in a _default node asking for a character's dexterity orientation, as the code uses 'left'.startswith and 'l' is an alias for the default 'look' command.

### DIFF
--- a/evennia/utils/evmenu.py
+++ b/evennia/utils/evmenu.py
@@ -312,7 +312,7 @@ class EvMenu(object):
     """
     def __init__(self, caller, menudata, startnode="start",
                  cmdset_mergetype="Replace", cmdset_priority=1,
-                 allow_quit=True, cmd_on_quit="look",
+                 allow_quit=True, cmd_on_quit="look", allow_look=True,
                  nodetext_formatter=None, options_formatter=None,
                  node_formatter=None):
         """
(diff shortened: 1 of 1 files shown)


Griatch:
@shollen 

Thanks for the PR! Terribly sorry but I interpreted your latest mailing list post as if you weren't  going to move forward with this - so I added the functionality in the wclient branch a few hours ago (you can see the update in issue #949). I should have commented in the issue as well though, to make sure you saw it. Your PR looks fine and uses correct branching and everything so now I feel bad. :-(

I did add few more changes though (which will not show until wclient branch merges). Notably I changed the name of `allow_quit` to `auto_quit` (since the effect of it is to automatically add a quit command). That also made more sense when adding additional `auto_look` and `auto_help` keywords (`auto_help` will turn off the automatic use of "h" and "help" which I suppose could have been a similar problem as for "l" and "look" ). 

...So I'm closing this for now (you can still comment if you want), hope it doesn't discourage you from future PRs! 
.
Griatch

shollen:
I’m not just new to Python and Evennia. I’ve never used Git before either, so it took me a bit to get my local stuff configured correctly and set up to do things the right way. I love Python. I love Evennia. And I want to help the codebase grow however I can. But game design and development is in my blood, so my focus is finding/building the best tools I can to get the job done. I learned a ton about Git by getting the EvMenu stuff pushed. The issue has been fixed. Doesn’t matter to me whose code it is.  :-)  So don’t feel bad! I learned!

Issue #951, opened 2016-04-11
Devel: Making the webclient aware of closing server


Griatch:
#### Brief summary of issue 

When the server shuts down or is killed, the devel-branch webclient is currently unaware of it - one can still use the interface (although you don't get any returns, obviously). The effect is quite confusing for the player. 

#### Expected result of feature

The client should somehow notify the user that the server has gone away, either by a popup or by some sort of text. Either way one should not be able to continue using the client when the server is no more able to retrieve what it sends. 

I wonder if the client could somehow regularly ping the server to see it is still there. If it is easier I suppose we could also send an explicit command from the server telling the client that it's going away (this will not work if the server is killed in an uncontrolled way).

Griatch:
pinging @natepri,  maybe you have some ideas/input on this?

NatePri:
Looking at what happens in the debugger when I stop evennia, it fires connection_close and connection_error events from the emitter, but they seem to be silently ignored by the GUI. It seems like we should be able to just hook into these to show some sort of error message, without requiring any special messaging from the server.

Ideally, you'd want there to be a way to try to re-establish the connection as well. I'm not sure what sort of behavior you'd want for reconnection; should it try to reconnect automatically (probably with a limited number of retries to avoid inadvertently DoSing the server whenever there's a hiccup), or wait for user input to try to do the reconnect? Either way, I'd think it shouldn't be hard to do.

Griatch:
Hah, you are right about connection_close/connection_error. I forgot about those, they are indeed the way to go for server info to the server. In an older version, there was a window alert-box for this, not sure if that is the way to go, or if one should display it some neater way. 

The AjaxCometConnection should ideally work in the same way as the websocket one. I suppose the close/error messages should work with the ajax connection too, but if it was possible to have it identify an erroneous connection it may be useful (I don't know how hard that would be though). 

Good point about allowing the client to reconnect. I think maybe a combination of the two options may be best for the user: Try automatically a few times (telling the user what is going on), and then pop up some sort of button to let the player choose when to try again.

Griatch:
I implemented callbacks for the gui to report when the server goes away. I don't try to re-connect though, probably better to let this be an explicit action. This seems to work just fine for both websocket and AJAX/COMET. 

[...]

PR #960, opened 2016-04-17
Improve webclient connection error handling

NatePri:
- Refactored WebSocket client to not return websocket object directly.
- Unified Connection API between WebsocketConnection and AjaxCometConnection.
- Changed AjaxCometConnection to emit close events in a way that mirrors WebsocketConnection.
- Added ability to reconnect client without a page refresh, incl. confirmation dialog.

Abstracts away some of the differences between Websocket and Ajax connections, makes the webclient UI behavior more consistent with native telnet clients for dropped connections.

- Issue resolved: Devel: Making the webclient aware of closing server #951


evennia/web/webclient/static/webclient/js/evennia.js
```diff
@@ -196,62 +214,87 @@ An "emitter" object must have a function
     var WebsocketConnection = function () {
         log("Trying websocket ...");
         var open = false;
```
Griatch: I used 'open' like you use 'ever_open', so I don't think 'open' needs to be there anymore?

Griatch: Never mind, looking at it from my phone I didn't notice it's used in isOpen.

Griatch:
Thanks for this, it looks really good. I think I will add a catch for the window-close and reload-window events too, to avoid users accidentally disconnecting from the server (very easy to do if hitting the F5 key).

PR #961, opened 2016-04-18
Add a new contrib to support menu_login with EvMenu

vlegoff:


evennia/contrib/evmenu_login.py
```diff
@@ -0,0 +1,338 @@
...
+    when the player first connects. It sets up the menu before handing off
+    to the menu's own look command.
+    """
+    key = syscmdkeys.CMD_LOGINSTART
+    locks = "cmd:all()"
+    arg_regex = r"^$"
+
+    def func(self):
+        "Execute the menu"
+        menu = EvMenu(self.caller, {
```
Griatch: This way of specifying the menu tree works, but requires you to keep it updated if you should ever add any new nodes. if you just give the `menu` keyword as `"evennia.contrib.evmenu_login"`, then every global function whose name does not start with "_" in this module will be interpreted as a node in the tree.

vlegoff: True.  Since I had to define the functions and the MenuTree in the same file, however, I didn't know how to point to the same module.  It sounds like a good altrnative if defined in a separate module, however.

Griatch: It's designed to work in the same module, this is how the test menu in the evmenu.py module itself works. :)

evennia/contrib/evmenu_login.py
```diff
@@ -0,0 +1,332 @@
...
+            # to handle tracebacks ourselves at this point. If we don't, we
+            # won't see any errors at all.
+            caller.msg(dedent("""
+                |rAn error occurred.|n  Please e-mail an admin if
+                the problem persists.
+            """.strip("\n")))
+            logger.log_trace()
+        else:
+            text = ""
+            caller.msg("Welcome, you're new account has been created!")
```
Griatch: This should be "your".

Griatch:
I've tested this now. Seems to work fine. 

It's pretty abrupt though - you don't even get a notification about which game you are connecting to. It seems this implementation lacks the ability to use the CONNECTION_SCREEN setting, which is commonly used for a greeting "graphic".

vlegoff:
I've added the connection screen this time.  It's displayed even in case of errors (or when typing B to go back to the start node), so perhaps it needs modification, but I think that's okay.
I also restructured the nodes to adopt more conventional nodes.  The error messages have also been updated to reflect this change.

PR #971, opened 2016-04-30
Removed 'self' from _get_mysql_db_version()

cptmashek:
There was an error when checking for mysql version, so I removed `self` from `_get_mysql_db_vesion(self)`, as it is not in an object.

File modified: `evennia/utils/dbserialize.py`


Griatch:
Thanks. :)

PR #1007, opened 2016-07-12
Bug: when calling create_script on a script that had kwargs set as de…

biscuitWizard:

…fault/globals on the class, create_script would override them even when no kwargs were specified in the command. Code now checks to see if there's a create_script kwargs override and if that doesn't exist, it tries to pull the default from the typeclass global itself.

### DIFF
--- a/evennia/utils/create.py
+++ b/evennia/utils/create.py
@@ -180,13 +180,13 @@ def create_script(typeclass=None, key=None, obj=None, player=None, locks=None,
 
     # validate input
     kwarg = {}
-    if key: kwarg["db_key"] = key
+    kwarg["db_key"] = key if key else typeclass.key if typeclass.key else None
     if player: kwarg["db_player"] = dbid_to_obj(player, _ScriptDB)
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks for the contribution. While the idea of this is good, I need need to look into the repercussions of this change before merging it. It does after all manage to kill every unittest involving script creation (as seen in the travis build above).

Griatch:
I can't merge this PR for the following reasons: 

1. The solution crashes every use of `create_script` (including unittests). I don't know how it could have worked for you since at this point the database fields are not initialized with their values. What you get with typeclass.key is a property object, not its value.
2.  I can't reproduce the bug it's supposed to fix. I tested by making a Script class with some of the variables defined (like `interval` etc) on the class level. I then called `create_script` with some or no arguments (sometimes replacing, sometimes unset so default is used). The results was that everything I didn't put explicitly in the call to `create_script` was automatically taken from the class variables set on the class. This is the expected behavior (and how models work in general). 

So I'm closing this. If you can give some example code for how to reproduce the bug you mention (where `create_script` overrides class variables despite not being given any kwargs), please open a new issue report for it.

PR #1019, opened 2016-08-23
Contrib Proposal: ASCII Map Builder

CloudKeeper1:
A contrib for the @mapbuilder command and corresponding build_map() function.

This is a command capable of taking a reference to a basic 2D ASCII map stored in the Evennia directory and  generating the rooms and exits  necessary to populate that world. The characters of the map are iterated over and compared to a list of trigger characters. When a match is found it triggers the corresponding instructions. These triggers and instructions are intended to be changed on a case by case basis.

You then call the command in-game using the path to the module and the name of the variable holding the map string. The path you provide is relative to the evennia or your mygame folder.
    

`@mapbuilder <path.to.module.VARNAME>`

The ability to easily and intuitively create a large playing area based on easily understandable input (a map) for basic users and prototyping.

### DIFF
--- a/evennia/contrib/mapbuilder.py
+++ b/evennia/contrib/mapbuilder.py
@@ -1,207 +1,312 @@
 # -*- coding: utf-8 -*-
 
 """
-
 Evennia World Builder
 
 Contribution - Cloud_Keeper 2016
 
-This is a command capable of taking a reference to a basic 2D ASCII
-map stored in the Evennia directory and  generating the rooms and exits 
-necessary to populate that world. The characters of the map are iterated
-over and compared to a list of trigger characters. When a match is found
-it triggers the corresponding instructions. Use by importing and including 
-the command in your default_cmdsets module. For example:
+Build a map from a 2D ASCII map.
+
+This is a command which takes two inputs:
+
+≈≈≈≈≈
+≈♣n♣≈   MAP_LEGEND = {("♣", "♠"): build_forest,
+≈∩▲∩≈                 ("∩", "n"): build_mountains,
+≈♠n♠≈                 ("▲"): build_temple}
+≈≈≈≈≈
+
+A string of ASCII characters representing a map and a dictionary of functions
+containing build instructions. The characters of the map are iterated over and
+compared to a list of trigger characters. When a match is found the
+corresponding function is executed generating the rooms, exits and objects as
+defined by the users build instructions. If a character is not a match to
+a provided trigger character (including spaces) it is simply skipped and the
+process continues.
+
+For instance, the above map represents a temple (▲) amongst mountains (n,∩)
+in a forest (♣,♠) on an island surrounded by water (≈). Each character on the
+first line is iterated over but as there is no match with our MAP_LEGEND it
+is skipped. On the second line it finds "♣" which is a match and so the
+`build_forest` function is called. Next the `build_mountains` function is
+called and so on until the map is completed. Building instructions are passed
+the following arguments:
+    x         - The rooms position on the maps x axis
+    y         - The rooms position on the maps y axis
+    caller    - The player calling the command
+    iteration - The current iterations number (0, 1 or 2)
+    room_dict - A dictionary containing room references returned by build
+                functions where tuple coordinates are the keys (x, y).
+                ie room_dict[(2, 2)] will return the temple room above.
+
+Building functions should return the room they create. By default these rooms
+are used to create exits between valid adjacent rooms to the north, south,
+east and west directions. This behaviour can turned off with the use of switch
+arguments. In addition to turning off automatic exit generation the switches
+allow the map to be iterated over a number of times. This is important for
+something like custom exit building. Exits require a reference to both the
+exits location and the exits destination. During the first iteration it is
+possible that an exit is created pointing towards a destination that
+has not yet been created resulting in error. By iterating over the map twice
+the rooms can be created on the first iteration and room reliant code can be
+be used on the second iteration. The iteration number and a dictionary of
+references to rooms previously created is passed to the build commands.
+
+Use by importing and including the command in your default_cmdsets module.
+For example:
 
     # mygame/commands/default_cmdsets.py
-    
+
     from evennia.contrib import mapbuilder
-    
+
     ...
-    
+
     self.add(mapbuilder.CmdMapBuilder())
 
-You then call the command in-game using the path to the module and the
-name of the variable holding the map string. The path you provide is
-relative to the evennia or your mygame folder.
+You then call the command in-game using the path to the MAP and MAP_LEGEND vars
+The path you provide is relative to the evennia or mygame folder.
 
-    @mapbuilder <path.to.module.VARNAME>
+Usage:
+    @mapbuilder[/switch] <path.to.file.MAPNAME> <path.to.file.MAP_LEGEND>
 
-For example to generate from the sample map in this module:
+Switches:
+  one - execute build instructions once without automatic exit creation.
+  two - execute build instructions twice without automatic exit creation.
 
-    @mapbuilder evennia.contrib.mapbuilder.EXAMPLE_MAP
+Example:
+    @mapbuilder world.gamemap.MAP world.maplegend.MAP_LEGEND
+    @mapbuilder evennia.contrib.mapbuilder.EXAMPLE2_MAP EXAMPLE2_LEGEND
+            (Legend path defaults to map path)
 
-Whilst this map is contained here for convenience, it is suggested
-that your map be stored in a separate stand alone module in the
-mygame/world folder accessed with @mapbuilder world.gamemap.MAP
-
-The rooms generated for each square of the map are generated using 
-instructions for each room type. These instructions are intended 
-to be changed and adapted for your purposes. In writing these 
-instructions you have access to the full API just like Batchcode.
+Below are two examples showcasing the use of automatic exit generation and
+custom exit generation. Whilst located, and can be used, from this module for
+convenience The below example code should be in mymap.py in mygame/world.
 
 """
-# ---------- #
 
-# This code should be in mymap.py in mygame/world.
+from django.conf import settings
+from evennia.utils import utils
+
+# ---------- EXAMPLE 1 ---------- #
+# @mapbuilder evennia.contrib.mapbuilder.EXAMPLE1_MAP EXAMPLE1_LEGEND
 
 # -*- coding: utf-8 -*-
 
-EXAMPLE_MAP = """\
-≈≈≈≈≈≈≈≈≈
-≈≈≈≈≈≈≈≈≈
-≈≈♣♠♣♠♣≈≈
-≈≈♠n∩n♠≈≈
-≈≈♣∩▲∩♣≈≈
-≈≈♠n≈n♠≈≈
-≈≈♣♠≈♠♣≈≈
-≈≈≈≈≈≈≈≈≈
-≈≈≈≈≈≈≈≈≈
-"""
-
-# ---------- #
-
-#Add the necessary imports for your instructions here.
+# Add the necessary imports for your instructions here.
 from evennia import create_object
 from typeclasses import rooms, exits
 from evennia.utils import utils
 from random import randint
 import random
 
-def build_map(caller, raw_map):
-    """
-    This is the part of the code designed to be changed and expanded. It
-    contains the instructions that are called when a match is made between
-    the characters in your map and the characters that trigger the building
-    instructions.
+# A map with a temple (▲) amongst mountains (n,∩) in a forest (♣,♠) on an
+# island surrounded by water (≈). By giving no instructions for the water
+# characters we effectively skip it and create no rooms for those squares.
+EXAMPLE1_MAP = """\
+≈≈≈≈≈
+≈♣n♣≈
+≈∩▲∩≈
+≈♠n♠≈
+≈≈≈≈≈
+"""
+
+
+def build_forest(x, y, **kwargs):
+    """A basic example of build instructions. Make sure to include **kwargs
+    in the arguments and return an instance of the room for exit generation."""
+
+    # Create a room and provide a basic description.
+    room = create_object(rooms.Room, key="forest" + str(x) + str(y))
+    room.db.desc = "Basic forest room."
+
+    # Send a message to the player
+    kwargs["caller"].msg("Forest Room Created.")
+
+    # This is generally mandatory.
+    return room
+
+
+def build_mountains(x, y, **kwargs):
+    """A room that is a little more advanced"""
+
+    # Create the room.
+    room = create_object(rooms.Room, key="mountains" + str(x) + str(y))
     
-    """
+    # Generate a description by randomly selecting an entry from a list.
+    room_desc = ["Mountains as far as the eye can see",
+                 "Your path is surrounded by sheer cliffs",
+                 "Haven't you seen that rock before?"]
+    room.db.desc = random.choice(room_desc)
     
-    #Create a tuple containing the trigger characters
-    forest = ("♣", "♠")
-    #Create a function that contains the instructions.
-    def create_forest(x, y):
-        #This has just basic instructions, building and naming the room.
-        room = create_object(rooms.Room, key="forest" + str(x) + str(y))
-        room.db.desc = "Basic forest room."
-        
-        #Always include this at the end. Sets up for advanced functions.
-        caller.msg(room.key + " " + room.dbref)
-        room_list.append([room, x, y])
- 
- 
-    mountains = ("∩", "n")
-    def create_mountains(x, y):
-        #We'll do something fancier in this one.
-        room = create_object(rooms.Room, key="mountains" + str(x) + str(y))
-        
-        #We'll select a description at random from a list.
-        room_desc = [
-        "Mountains as far as the eye can see",
-        "Your path is surrounded by sheer cliffs",
-        "Haven't you seen that rock before?"]
-        room.db.desc = random.choice(room_desc)
-        
-        #Let's populate the room with a random amount of rocks.
-        for i in xrange(randint(0,3)):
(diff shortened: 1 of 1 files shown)


evennia/contrib/mapbuilder.py
```diff
@@ -0,0 +1,256 @@
...
+        #Display map retrieved.
+        caller.msg("Creating Map...") 
+        caller.msg(map)
+        
+        #Pass map to the bulid function.
+        build_map(caller, map)
+        caller.msg("Map Created.")                              
+
+#Helper function for readability.
+def prepare_map(map):
```
Griatch: Maybe call this `map_to_list` to make it clearer what it does? Also, as a helper function used in `build_map` and `CmdMapBuilder`, it should be moved up to be just above `build_map` in the source file. If it's not to be used outside this module but is only meant as a helper, it's also praxis to start its name with an underscore.

evennia/contrib/mapbuilder.py
```diff
@@ -0,0 +1,256 @@
...
+# ---------- #
+
+#Add the necessary imports for your instructions here.
+from evennia import create_object
+from typeclasses import rooms, exits
+from evennia.utils import utils
+from random import randint
+import random
+
+def build_map(caller, raw_map):
```
Griatch: How is `build_map` intended to be overloaded? Should one copy&paste the entire file into mygame or is there some pluggable way? It's not documented.

evennia/contrib/mapbuilder.py
```diff
@@ -0,0 +1,256 @@
...
+EXAMPLE_MAP = """\
+≈≈≈≈≈≈≈≈≈
+≈≈≈≈≈≈≈≈≈
+≈≈♣♠♣♠♣≈≈
+≈≈♠n∩n♠≈≈
+≈≈♣∩▲∩♣≈≈
+≈≈♠n≈n♠≈≈
+≈≈♣♠≈♠♣≈≈
+≈≈≈≈≈≈≈≈≈
+≈≈≈≈≈≈≈≈≈
```
Griatch: There seems to be no instructions in `build_map` for how to build `≈` or am I just missing it? If it's using the default, this might be good to mention.

evennia/contrib/mapbuilder.py
```diff
@@ -0,0 +1,256 @@
...
+        room_list.append([room, x, y])
+ 
+ 
+    mountains = ("∩", "n")
+    def create_mountains(x, y):
+        #We'll do something fancier in this one.
+        room = create_object(rooms.Room, key="mountains" + str(x) + str(y))
+        
+        #We'll select a description at random from a list.
+        room_desc = [
```
Griatch: Since this function appears to actually be meant to supply hard-coded data, maybe the instructions for building each of these terrain types could be moved outside of the mapbuilder code and into something equivalent to how evmenu nodes work? 
Here's a suggested alternative: Make the call `@mapbuilder <mapfile> [legendfile]`, where legendfile defaults to the same file as the map file. The legend file defines a "legend dict" that can either have a fixed value or one you give as part of the path. The legend dict maps map symbols to functions in that same module. Here's an example:

```python
# file map_legend.py 

# map might optionally be up here^

MAP_LEGEND = \
    {"▲": build_temple, 
     "∩": build_mountain,
     "n": build_mountain}

def build_mountain(x, y):
    # build mountain
    return mountain

def build_temple(x, y):
    # build temple
    return temple
```

This would abstract the build mechanism from your map-builder function completely - for every position on the grid, the system queries `MAP_LEGEND` and calls the linked function ...

dwesely:
I really like the idea of parsing a map file to make it easier to create lots of rooms at once. It looks like @CloudKeeper1 would like to be able to take a [Dwarf Fortress map](http://dwarffortresswiki.org/index.php/DF2014:Map_legend) and bulk convert it right into an Evennia world. That would be very cool. :-)

An additional use case I have in mind is this:

- The designer starts with an idea/sketch of a single scene like a D&D [One Page Dungeon](http://www.dungeoncontest.com/)
- The designer makes a text file in a similar format to the example below (using plain old text editor or something more full-featured like [JavE](http://jave.de/))
- The designer runs the parser, which outputs a batch script to create the scene as described
- The designer looks through the batch script, tweaks it as necessary for their needs
- The designer can run (or share) their batch script as needed

Parsing a map straight into the game seems like it would be a bit error-prone, so this would keep the parsing step offline. Also, the parser wouldn't need to change when the default values change for each scene/setting.  The user can have access to some defaults through the map input format, and the scene/setting details can stay separate from the parser function. The output of the parser would be the batch script to create the scene, with some placeholders to fill in blanks. The user would correct the output for the scene they are creating, not necessarily change the parser to match the setting.

I hope that all makes sense, I'm very new to Evennia, and fairly new to Python. Does this seem like it would be too much of a headache for the user to go through the extra step, or do you guys think this would actually make it easier to do world-building a scene at a time?

```
# Identify build parameters up front
#
# Draw a map using any of the following characters: .,|-+/\:'"
# Identify room numbers using a number in square brackets in the room
# Identify exits with an [X] between the rooms being connected
# Identify up/down exists with [U:#] or [D:#], where # is the room target
#
# Below the map, describe the room and objects

default_exit_name = 'door'
wrap_descriptions_at_column = 50
justify_descriptions = False

    .-----------------------.-------------------------.
    |                       |             |           |
    |               [U:5]   |            [X]   [3]    |
    |                       |             |           |
    |        [1]           [X]     [2]    |           |
    |                       |             +-----------|
    |               [D:6]   |             |  [4]      |
    |                       |             |           |
    |                       |            [X]          |
    |                       |             |           |
    '---------[X]-----------"-------------------------'

    .-------------------.         .-------------------.
    |                   |         |                   |
    |                   |         |                   |
    |       [6]         |         |       [5]         |
    |           [U:1]   |         |           [D:1]   |
    |                   |         |                   |
    '-------------------'         '-------------------'

# r#:Room Title
# d#:Description
# o#.n:Object title for nth object in room #


r1:Living Room
d1:You see a common country living room with standard furniture.
o1.1:Couch
d1.1:This couch is old and threadbare.
o1.2:Cabinet

r6:Basement
d6:This smelly basement is small, but it smells.

r2:The Dining room
d2:Blah blah blah, etc etc.......
```

Griatch:
@dwesely 

It sounds like an interesting way to do it. While it's @CloudKeeper1 's call, I would think what you suggest is something different from what is intended with this particular PR though.

This PR is meant as a wilderness map with one room per map character; what you propose requires the parsing of large connected swathes of ascii map positions to form the area of each room with its inline markings. It's possible to do (see EvForms for parsing swathes of characters in a different context if you are interested), but it's a quite different beast from the code of this PR. 

There is also no concept of creating a batchfile dynamically anywhere in Evennia -- all batchfiles are (at least so far) intended to be created outside the game and then loaded. I suppose, if one parsed it off a map anyway, one could have the parser create the file like a batchcmd- or batchcode file rather than create it outright. 

But agreed, it sounds like an interesting way to handle building. Maybe something for a PR of your own?

BlauFeuer:
If @CloudKeeper1 keeps it abstract enough, taking some of @Griatch 's suggestions, then @dwesely might be able to extend or reuse more of the `@mapbuilder` code.

Personally, I like the idea of having an external program (even if it is just a text editor) make large-scale map building easier.  The in-game commands are still useful for small-scale building, but submitting files to build is a good idea, and I have seen it used elsewhere.

dwesely:
I think I'll take a crack at an offline tool for parsing a structured map file. I've been meaning to work on ascii-art edge detection for another project anyway, and this would be a good step in that direction, because I can define the structure of the input.

CloudKeeper1:
I'm having a look at this today. Moving the build instruction functions outside the module means I lose the ability to easily keep track of the objects built inside them (For example, keeping track of rooms which I now do with the room_list). What are your thoughts on this solution: each build function can optionally return a list of objs which are appended to a reference dictionary. The dictionary is passed to each function to allow references to the objects created in other functions. The dictionary would look something like : 
reference_dict = {(0,1): [room], (0,2): [room, npc]....etc } 

dwesely: I like your idea, although at that stage I'd be more inclined to do a GUI that outputs batchcode.

Griatch:
@CloudKeeper1 

That seems unnecessarily complex. There is no reason for a build function to know about other rooms the way the map is built here. I can see three possible use cases for doing so (correct me if you have something other in mind). 

1. You aim to support something like "To the west you see the beginning of a forest". But that information the build function can get from its own coordinate (which it has) and the full map (which can be easily loaded at the top of the build module for those wanting it). 
2. You aim to support something like "To the west you see a large flag pole", where the flag pole is an object in that room. In this case you *would* need to have a reference to the western room (you can then do `room.contents` to find the flag pole). But sending a list of references the way you suggest won't do - that list is incomplete as the rooms are being built. If you want this kind of ability you must implement it as a second pass over the previously created map.
3. You want to support irregular exits. In this case you need to change the map to support exit symbols, then do a first pass to build rooms and a second to open the exits between them.

As this map system is originally designed, I don't see why the build-mechanic would need to hold a reference to the room at each coordinate at all. It has all it needs from the map already. All it would need (in principle) is a True/False return to make sure that room was built without errors and that it could move on to the next. Only if you want to implement a second pass do you do you need to store the growing list of rooms and coordinates.

Coordinates can be stored on the room as tags btw - you could then easily find a room with the only combination of two tags that match a single x,y position (or get all rooms in a collumn etc).

CloudKeeper1:
I raise this issue precisely in preparation of second passes and irregular exits. It seems to be the first functionality enquired about (or at least, the first thing Blau enquired about) and can reasonably be added without disrupting the original intended usage. I agree that it could become complex, which is why I asked if anyone could think of a better way.

I hear what you say about coordinate tags, but when I originally wrote the code I steered away from hard coding their use due to the possibility that if you ran the map twice it would turn to absolute spaghetti. Keeping track of the rooms internally seemed the cleanest way before the complete abstraction.

Griatch:
@CloudKeeper1 

If your goal is to expand this with a second pass, still don't see why you'd need to make it very complex just because you are calling an external module to get the build-functions. Here's an idea for how it could work.

First the look of the build functions (imported from outside, as discussed): 

```python
def build_forest(ix,iy, mapping=None):
    # ...
    return new_room
```

Returning the just created room should be all you need - everything in this room can be found from this object later. 

During the first pass you loop over the map and call the matching build functions as desired. Some of those calls may be to functions that will not do anything yet, because they are looking for a valid `mapping` object which does not yet exist. If so they'll return None. 

The rooms/Nones are collected into a 2D grid `mapping=[[room00,  ..., roomN0], ... [room0N, ... roomNN]] `. The ix,iy coordinate on the map "graphic" should match the indices of this mapping. As said, not all positions may actually contain a room (and never will) but is intended to show an exit connection, for example. 

Now for the second pass, you loop through the map again, but now you also send the full map from the first pass: `buildfunc(ix, iy, mapping=mapping)`. It's up to each build function to determine if it should do anything on the second pass or not. This they do by simply identifiying that `mapping` is not `None`. On this pass all build functions returns None. Instead they operate directly on the mapping: maybe they are inserting/updating  a room in the mapping or connecting it to its neighbours. Since `mapping` is a mutable changes made to it in each method will change the mapping itself and after the pass it will have been changed to the final version. 

I don't know if this is where you want to take this, but it's an idea for how it could work.

CloudKeeper1:
Thanks for the feedback!

My prototype is doing something similar but with a dictionary with coordinate tuple's as the key rather than a list of lists. So that the room is accessed via `mapping[(1,1)]` as opposed to `mapping[1][1]` and is flexible enough for multiple rooms in a single location. I had thought that might have been simpler to grasp but given your suggested method mirrors how the map itself is treated in code it may be that it's more consistent to use the list of lists method.

I was passing an additional value for passes: `def build_forest(ix,iy, mapping=None, pass=2):` because I basically just wrapped the execution code into a `for pass in xrange(len(passes)):` where the amount of passes is determined by the switch used. Code in the building functions is then handled exactly how you outlined with `if mapping is None` except with `if pass == 2`.

I'll polish what I have over the weekend.

CloudKeeper1:
Updated.
The code has been changed following peer review.
*Both the map and build instructions have been abstracted out of the module.
*Functionality has been expanded to include multiple passes allowing custom exit creation.
*Documentation expanded and a further example provided.

evennia/contrib/mapbuilder.py
```diff
@@ -0,0 +1,458 @@
...
+    if kwargs["iteration"] > 0:
+        return
+
+    room = create_object(rooms.Room, key="forest" + str(x) + str(y))
+    room.db.desc = "Basic forest room."
+
+    return room
+
+
+def build_verticle_exit(x, y, **kwargs):
```
Griatch: Typo: "verticle" -> "vertical".

Griatch:
I have tested this now, seems to work fine for the single pass (I have some minor suggestions and things I noticed only). The second-pass of the example doesn't seem to work though: It creates the rooms and the exits but when I teleport to one of the newly created forest rooms they lack exits.

evennia/contrib/mapbuilder.py
```diff
@@ -0,0 +1,458 @@
...
+EXAMPLE1_MAP = """\
+≈≈≈≈≈
+≈♣n♣≈
+≈∩▲∩≈
+≈♠n♠≈
+≈≈≈≈≈
+"""
+
+
+def build_forest(x, y, **kwargs):
```
Griatch: I don't think this function will ever be used, since there is a same-named function below that will replace it.

evennia/contrib/mapbuilder.py
```diff
@@ -0,0 +1,458 @@
...
+  |   |
+≈ ♣ ♣ ♣ ≈
+  | | |
+≈ ♣-♣-♣ ≈
+
+≈ ≈ ≈ ≈ ≈
+"""
+
+
+def build_forest(x, y, **kwargs):
```
Griatch: This will replace the `build_forest` function defined above.

evennia/contrib/mapbuilder.py
```diff
@@ -0,0 +1,458 @@
...
+
+        # OBTAIN MAP FROM MODULE
+        
+        # Breaks down path_to_map into [PATH, VARIABLE]
+        path_to_map = args[0]
+        path_to_map = path_to_map.rsplit('.', 1)
+        
+        try:
+            # Retrieves map variable from module or raises error.
+            game_map = utils.variable_from_module(path_to_map[0],
```
Griatch: Entering the wrong name here yields an `AttributeError` which is not caught by the error handling: 
```python
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/commands/cmdhandler.py", line 464, in _run_command
    ret = yield cmd.func()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/mapbuilder.py", line 388, in func
    build_map(caller, game_map, legend, iterations, build_exits)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/mapbuilder.py", line 405, in build_map
    game_map = _map_to_list(game_map)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/mapbuilder.py", line 280, in _map_to_list
    list_map = game_map.split('\n')
AttributeError: 'NoneType' object has no attribute 'split'
```

evennia/contrib/mapbuilder.py
```diff
@@ -0,0 +1,458 @@
...
+        
+        # Breaks down path_to_legend into [PATH, VARIABLE]
+        path_to_legend = args[1]
+        path_to_legend = path_to_legend.rsplit('.', 1)
+        
+        # If no path given default to path_to_map's path
+        if len(path_to_legend) == 1:
+            path_to_legend.insert(0, path_to_map[0])
+        
+        try:
```
Griatch: Wrong name here yields `AttributeError` as above.

evennia/contrib/mapbuilder.py
```diff
@@ -0,0 +1,458 @@
...
+
+    Usage:
+        @mapbuilder[/switch] <path.to.file.MAPNAME> <path.to.file.MAP_LEGEND>
+
+    Switches:
+      one - execute build instructions once without automatic exit creation
+      two - execute build instructions twice without automatic exit creation
+
+    Example:
+        @mapbuilder evennia.contrib.mapbuilder.EXAMPLE_MAP MAP_LEGEND
```
Griatch: This example is wrong; it should be `EXAMPLE1_MAP` etc.

evennia/contrib/mapbuilder.py
```diff
@@ -0,0 +1,458 @@
...
+                 "Haven't you seen that rock before?"]
+    room.db.desc = random.choice(room_desc)
+    
+    # Create a random number of objects to populate the room.
+    for i in xrange(randint(0, 3)):
+        rock = create_object(key="Rock", location=room)
+        rock.db.desc = "An ordinary rock."
+
+    # Send a message to the player
+    kwargs["caller"].msg("Mountain Room Created.")
```
Griatch: One thing I noticed with this was how to get to the area just created. I know this is just an example and that people can change the input as they want, but since all these rooms are named the same, it might set a good precedent to include the #dbref of the newly created room in this return text, so one can easily jump to it and check things out afterwards.

CloudKeeper1:
Addressed the above.

1. Second pass not working, room has no exits when using example 2:
The suggested commands for the example code didn't include the relevant switch. Wasn't an issue with the code, it just only iterated once.
2. First `build_forest` function not being reached due to duplicate function name:
Changed name of example code build functions to unique names. Funnily enough Pycharm didn't pick that up either.
3. Attribute errors not caught:
Added appropriate return statements.
4. Example on line 297 wrong:
Changed examples.
5.Suggestion - pass back Dbref:
Dbref is now returned on creation. This was in the original code I just overlooked it.

Griatch:
Looks good! Thanks for your work on this, it's now merged into the contrib folder! 
Closing.

PR #1026, opened 2016-08-28
Fixed the bug that prevented restarting a script more than once after unpausing. Resolves #1023.

### LINKED ISSUE #1023: Scripts only restart once when unpaused, become inactive afterwards

I have tested a simple script that sends out a message to the owner whenever they restart. I manually pause, then unpause it, and find that it only restarts a single time after being unpaused; afterwards, the script's is_active variable is set to False, as though it were still paused, and the script does not send its message again, suggesting that it is indeed paused. Clearly, it should run continuously after being unpaused, but it doesn't.

1. Create the following script in a file:
    ```
from evennia import DefaultScript
class Script1(DefaultScript):
    def at_script_creation(self):
        self.key = "Script1"
        self.desc = "Test script to check how restarting works."
        self.interval = 5  # seconds
        self.repeats = 0  # repeat only a certain number of times
        self.start_delay = False # wait self.interval until first call
        #self.persistent = True

    def at_start(self):
        self.obj.msg("Script has started.")

    def at_repeat(self):
        self.obj.msg("Script is being repeated.")
        self.restart(interval = 5, repeats = 0, start_delay = True)
    ```
2. From inside Evennia, Run "@script self = <whatever the script's Python path is>" to load it on the character.

3. Assuming there are no other scripts on the character, run "@py self.scripts.all()[0].pause()"

4. Then run "@py self.scripts.all()[0].unpause()" 

"@py self.scripts.all()[0].is_active" will return False. The script will not be sending any further messages after the first repeat, indicating that it is inactive.

Evennia version 0.6.0

andrei-pmbcn:

### DIFF
--- a/evennia/scripts/scripts.py
+++ b/evennia/scripts/scripts.py
@@ -395,6 +395,11 @@ class DefaultScript(ScriptBase):
             # if this script was paused manually (by a direct call of pause),
             # it cannot be automatically unpaused (e.g. by a @reload)
             raise RuntimeError
+
+        # Ensure that the script is fully unpaused, so that future calls
+        # to unpause do not raise a RuntimeError
+        self.db._manual_pause = False
+
         if self.db._paused_time:
             # only unpause if previously paused
             self.is_active = True


Griatch:
Looks good, thanks! Merged with rebase. You can delete your local working branch now.

PR #1039, opened 2016-09-05
Added missing access() method to Msg for its lockshandler.

TehomCD:
I just noticed that the access() method for Msg was missing. I actually use locks for msg objects here or there, so I went ahead and just copied the version from TempMsg, figuring that their access() methods should be identical.


Griatch:
Merged with a rebase. Thanks!

PR #1040, opened 2016-09-05
Actions

andrei-pmbcn:

Added the action system contrib.

Griatch invited me.

Please let me know what you think and give it a try. I'll make any changes to the code and the documentation that are necessary.


evennia/contrib/actions/README.md
```diff
@@ -0,0 +1,372 @@
...
+
+And then subclass the appropriate [typeclasses](https://github.com/evennia/evennia/wiki/Typeclasses "typeclasses").
+Each of the ActionCharacter and ActionRoom objects has an attribute called "actions" that stores all methods and properties related to the action system. When the object is initialized, its actions.setup() method runs, loading the default parameters for the object. Likewise, running setup() from the evennia.contrib.actions modulecalls the actions.setup() methods for all objects that are subclassed from either ActionCharacter or ActionRoom.
+
+In any subclass you make for the ActionCharacter and ActionRoom classes, you must be sure to fill in several members of the "actions" attribute for the object being created, as well as call [super](https://docs.python.org/2/library/functions.html#super "the super function"), within at_object_creation. 
+
+For subclasses of __ActionCharacter__, you should do:
+
+> at_object_creation(self):
+>     self.actions.movespeed = <your function here>
```
Griatch: I presume `self.actions` is some sort of handler rather than an Attribute as suggested here.

Without looking at the code, won't placing the super() call reset those custom-set actions to their default values?

evennia/contrib/actions/README.md
```diff
@@ -0,0 +1,372 @@
...
+> at_object_creation(self)
+>     self.actions.view = <your function here>
+>    *[...]*
+>     super(ActionRoom, self).at_object_creation()
+
+The view function must have two arguments. The first of these is the object that is being viewed and the second is the character viewing it. It will return either a formatted string or the value False, the former when the object being viewed can be seen by the viewer, and the latter when the object being viewed is invisible to the viewer. The string can contain the name of the object being viewed, its description or anything else the coder desires. The view function is used automatically inside the ActionRoom's "actions.display" method and should never need to be caled directly.
+
+Finally, when you subclass from __ActionExit__, you should do:
+> at_object_creation(self)
+>     self.distance = <your float here>
```
Griatch: This does not look like a persistent property. It's either an Attribute (and should be written `self.db.distance` or this won't survive a server reload.

andrei-pmbcn: self.distance is in fact a property in the barebones python sense, i.e. it has the @property decorator. It points correctly to self.db.distance.

evennia/contrib/actions/README.md
andrei-pmbcn: I was using the term "attribute" to mean a barebones python attribute, the way y is an attribute of x.y . I will call it a "handler" instead to avoid confusion.

The super() call will call self.actions.setup() without the override flag, meaning that if the custom-set properties have already been set, they will not be set again. It's not an issue.

evennia/contrib/actions/README.md
```diff
@@ -0,0 +1,372 @@
...
+>    *[...]*
+>     super(ActionExit, self).at_object_creation()
+
+The distance value of an exit represents how far one must travel to move into the room from the exit, and to move from the room to the exit. It is used in measuring travel times, together with the character's movement speed.
+
+The action system comes with two sets of __[commands](https://github.com/evennia/evennia/wiki/Commands "commands")__, one for characters and the other for staff. To load them, simply add them to one of your [CmdSets](https://github.com/evennia/evennia/wiki/Command%20Sets "command sets"), such as the ones in your game's commands/default_cmdsets.py file:
```
Griatch: It might be a good idea to make a short section higher up in this document that gives a short listing of all the various components you need to add to the game to try this contrib out. Something like

## Installation in brief
- Import and inherit your Character, Exit and Room classes from contrib.actions.ActionCharacter, ActionExit and ActionRoom 
- Customize these Attributes as needed (see below for details): ...
- Import and add the following cmdsets to your default cmdset: ...
- ... and so on 

Experienced Evennia users might get quite far with just that list and can dig into the details later as needed.

evennia/contrib/actions/README.md
Griatch: I see. This should probably be clarified since it's can be (and is, as you notice) confusing compared to other Evennia APIs. 

Standard Evennia tries to be clear with what is persistent or not - it has four cases: 

1. .db for persistent Attributes, 
2. .ndb for non-persistent Attributes.
3. Handlers, which are called as `obj.handlername.add()/remove()` etc (never on their own). 
4. Model-centric database fields which are few and non-changing, like `self.key` and `self.location`.
5. All other properties are non-persistent. 

By introducing a new type of persistent property you are introducing a sixth case. If done for a good reason this is not wrong per se, but it's different from Evennia standard API and should be pointed out.

evennia/contrib/actions/README.md
```diff
@@ -0,0 +1,372 @@
...
+
+ActionRoom objects have the method "actions.display(viewed_object, message, target=None, data="", default=False)", which will show everyone in the room a processed version of the message string in the arguments list according to the room's view function (if you have set it) or in a default way if no view function exists. The target should be an object, while the data should be a string or an object. If default is set to True, and the viewed object is invisible to a given viewer, the name of the viewed object will be replaced (with "Someone" if the object is a character or "Something" otherwise) in the version of the message given to that viewer. If default is setto False, and the viewed object is invisible to the viewer, the message will not be shown to the viewer at all. Thus, the display method can be used to implement some features of stealth.
```
Griatch: Markdown formatting point a function example like this should be put in `backtick` markers rather than quotes.

evennia/contrib/actions/README.md
```diff
@@ -0,0 +1,372 @@
...
+* Do not have a target, despite has_target being set (the error string is "NoTarget ")
+* Have a target that is not inside their room (the error string is "TargetNotInRoom ")
+
+The strings of all errors encountered are concatenated and passed to the room's actions.handle_invalid method, which then notifies the MudInfo channel about the errors and attempts to clean up the action. If you know what you're doing, you can overload actions.handle_invalid to process invalid actions in a different way.
+
+
+## Actions
+
+To create your own actions, you will want to import the Action class as shown above, then subclass it. The \_\_init\_\_ function of your subclass will have a call to [super](https://docs.python.org/2/library/functions.html#super) that will include some or all of the following keyword arguments:
+
```
Griatch: Wrap your python code in three backticks, followed by the string `python` to get correct syntax highlighting in your examples. See here: https://help.github.com/articles/creating-and-highlighting-code-blocks/

evennia/contrib/actions/README.md
```diff
@@ -0,0 +1,372 @@
...
+To create an action and initialize it, simply create an object of your subclass of Action. Say you have a class MyAction that inherits from Action; then simply do MyAction(owner, room, target, etc). As long as the original at_creation method of the Action class is called (whether directly or through super), your action will be properly processed so that it will show up either in the list of ongoing actions in the room or in the owner's actions queue.
+
+
+## Commands
+
+There are two command sets in the evennia.contrib.actions module, ActionCmdSet and ActionDebugCmdSet. The commands in ActionCmdSet are meant for all characters and include the following:
+
+#### alarm
```
Griatch: Maybe I'm not familiar with the original intent of this system, but "alarm" seems like a strange term for going into turn-based mode. As a new user I would personally have more expected the command for switching to this mode to be named maybe "turns", "pause", "turnbased" or something to that effect. Is there a particular motivation for chosing the term "alarm"?

evennia/contrib/actions/README.md
```diff
@@ -0,0 +1,372 @@
...
+
+To have a functional action, you must supply your subclass of Action with some or all of the at_creation, at_attempt, at_completion, at_failure and at_cancel  methods. These will overload the methods in the Action class. at_creation, at_completion, at_failure and at_cancel should always call [super](https://docs.python.org/2/library/functions.html#super), unless you intend to re-implement some of the base functionality of the action system itself. at_attempt simply returns True by default, meaning that the action always succeeds, and thus at_attempt can be safely overriden.
+
+To create an action and initialize it, simply create an object of your subclass of Action. Say you have a class MyAction that inherits from Action; then simply do MyAction(owner, room, target, etc). As long as the original at_creation method of the Action class is called (whether directly or through super), your action will be properly processed so that it will show up either in the list of ongoing actions in the room or in the owner's actions queue.
+
+
+## Commands
```
Griatch: I don't think you need to list the call signature/help entry of all commands here - this is available in their code and help entries. Besides, if you change the code, this text may easily go out of date. A simple list briefly mentioning what each command does, eventually with some more explanation of unintuitive behavior is probably enough for this section.

evennia/contrib/actions/README.md
```diff
@@ -0,0 +1,372 @@
...
+* In turn-based mode, PCs get 60 seconds to complete their turn, while NPCs get 5. There is no way to customize this at present.
+
+All of these issues can be resolved, and will be resolved if people who are planning to use the system contact the maintainer, Andrei Pambuccian, about it. My github name is andrei-pmbcn.
+
+Other limitations that are unlikely to be resolved include the following:
+
+* There needs to be at most one exit pointing to a given room B in any room A if B also has an exit to A. If two exits in A point to the same room B, which has an exit to A, then when moving into A after having left B, the distance used to calculate the movement action's duration will be the distance value assigned to the first of the exits from A to B, whichever of the two exits that happens to be. Confused? Just remember that when two rooms have more than one exit to each other, and the distances of these exits are different, the action system will pick one of these distances, which may be the wrong one.
```
Griatch: In order to be reproducible, I guess it could automatically go for the shortest of the multiple possible exit routes?

evennia/contrib/actions/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+from evennia.contrib.actions.actions import Action
+from evennia.contrib.actions.handlers import CharacterActionHandler, RoomActionHandler
+from evennia.contrib.actions.scripts import ActionSystemScript
+from evennia.contrib.actions.typeclasses import ActionCharacter, ActionRoom, ActionExit
+from evennia.contrib.actions.commands import (CmdAlarm, CmdActSettings, CmdActions,
+    CmdStop, CmdDone, CmdQueue)
+from evennia.contrib.actions.debug import ActionDebugCmdSet
+from evennia.contrib.actions.setup import setup
+
+class ActionCmdSet(CmdSet):
```
Griatch: The base `ActionCmdSet` seems to belong and be easier to find in the `command.py` module then imported here rather than defined here. There is (in principle, at least) no need to import the individual commands here since people can just import the cmdset and add it to their default cmdset in one go (but you may want to have the commands available for people to pick-and-choose from I suppose).

evennia/contrib/actions/actions.py
```diff
@@ -0,0 +1,516 @@
+"""
```
Griatch: A brief mention of the contents and purpose of this module is needed here.

evennia/contrib/actions/README.md
andrei-pmbcn: "Alarm" seemed fitting because in nearly every case where someone will enter turn-based mode, they will do so because a conflict of some sort is starting. It happens when people are about to start fighting, or a potential enemy has just arrived into the room, or a fire is breaking out, or someone is poisoned and slowly dying away, or someone has just spotted a thief. Under these circumstances, what they would want to type is "alarm", because something alarming, a crisis situation, was happening. The term "turnbased" would take too long to type in a crisis situation, "turns off" wouldn't be correct because you don't disable turn-based mode simply by lowering your alarm (everyone else's alarm has to also be off in order for that to happen); likewise for "pause", which would also be anticlimatic. By contrast, "alarm" is short and climatic.

andrei-pmbcn:
@Griatch, I have addressed all issues you raised apart from the use of the name "alarm", which is still in discussion. Let me know if there is anything more to change.

BlauFeuer:
Would "**alert**" be a better term for the notice given when an intended action is not completed?

andrei-pmbcn:
"alarm" is unrelated to whether or not any actions are being carried out in the room. I have considered using "alert" as well and wouldn't mind picking that instead, if Griatch prefers it.

Griatch:
I think "turns" is likely the most obvious name since it doesn't gice any value judgement as to when it's used. 
The turn-based system may well be invoked in a non-combat situation depending on the game, one alternative might for example be to help control emote spamming of a large RP scene with many players (allowing one emote per turn, for example).

I've not had time to go through the code or run this yet - that may take some time. Others are of course welcome to help go through the contrib as well.

andrei-pmbcn:
Griatch, if you want a neutral term for the command, the best one would be "tb". It's short, meaning it can be used quickly in crisis situations, and "tb on" / "tb off" is a lot more appropriate than "turns on" / "turns off", since the latter alludes incorrectly to the terms "to turn on" / "to turn off". "tb" is also short for turn-based, and suggests that the ~character~ is entering or leaving turn-based mode, rather than that turns are being assigned in the room (which may or not happen regardless of whether the command is issued, since the command only affects the character who issues it and does not necessarily switch the room to RT or TB mode).

Griatch:
@andrei-pmbcn 

I started testing but ran into some issues out of the gate. I changed my default Character, Room and Exit to inherit from their Action- equivalents. I also add the two cmdsets, I then start with a fresh database and default setup. 

Trying to run most of the @@-type debug actions lead to tracebacks however: 

```
@@actfast/@@actslow/@@actlarm/@@actrarm and turnbased
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/commands/cmdhandler.py", line 464, in _run_command
    ret = yield cmd.func()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/debug.py", line 72, in func
    DebugActionFast(self.caller, self.caller.location)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/debug.py", line 88, in __init__
    duration=duration,
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/actions.py", line 170, in __init__
    self.at_creation(action)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/actions.py", line 179, in at_creation
    if not action['owner'].actions.active and action['cancellable']:
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/handlers.py", line 251, in active
    return self.owner.db.actions['active']
TypeError: 'NoneType' object has no attribute '__getitem__'
```

```
@@actsetup
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/commands/cmdhandler.py", line 464, in _run_command
    ret = yield cmd.func()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/debug.py", line 260, in func
    actions.setup(override=False)
NameError: global name 'actions' is not defined
```

evennia/contrib/actions/README.md
```diff
@@ -0,0 +1,313 @@
+# The action system
```
Griatch: Trying to read this file in an actual code editor made it clear the lines here are immensely long, making it very hard to read. My editor is 150 characters wide and it's nowhere wide enough. Please shorten the line length to something like 80-100 characters to make it be readable by most normal editors (see existing contribs for examples).

Griatch:
Deleting a previous comment since I might have found where I went wrong - so likely those errors were user error on my part. On a fully reset database I run into this error on first startup however (I can't start the server): 

```python
 File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/twisted/application/app.py", line 450, in getApplication
    application = service.loadApplication(filename, style, passphrase)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/twisted/application/service.py", line 411, in loadApplication
    passphrase)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/twisted/persisted/sob.py", line 224, in loadValueFromFile
    eval(codeObj, d, d)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/server.py", line 481, in <module>
    EVENNIA = Evennia(application)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/server.py", line 165, in __init__
    self.run_initial_setup()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/server.py", line 257, in run_initial_setup
    initial_setup.handle_setup(0)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/initial_setup.py", line 200, in handle_setup
    setup_func()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/initial_setup.py", line 91, in create_objects
    nohome=True)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/utils/create.py", line 123, in create_object
    new_object.save()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/utils/idmapper/models.py", line 377, in save
    super(SharedMemoryModel, self).save(*args, **kwargs)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/db/models/base.py", line 708, in save
    force_update=force_update, update_fields=update_fields)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/db/models/base.py", line 745, in save_base
    update_fields=update_fields, raw=raw, using=using)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/dispatch/dispatcher.py", line 192, in send
    response = receiver(signal=self, sender=sender, **named)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/typeclasses/models.py", line 77, in post_save
    instance.at_first_save()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/objects/objects.py", line 879, in at_first_save
    self.at_object_creation()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/typeclasses.py", line 38, in at_object_creation
    self.actions.setup()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/handlers.py", line 56, in setup
    self.movetype = None
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/handlers.py", line 202, in movetype
    if owner.actions.movebps and owner.actions.movetype:
exceptions.NameError: global name 'owner' is not defined
```

evennia/contrib/actions/debug.py
```diff
@@ -0,0 +1,286 @@
...
+            reach='contact',
+            data="diabolically",
+            cancellable=True,
+            invokes_tb=True,
+            non_turn=False,
+            duration=duration,
+            )
+
+
+#class CmdDebugActAnyArm(Command):
```
Griatch: Should this be removed?

evennia/contrib/actions/handlers.py
```diff
@@ -0,0 +1,752 @@
...
+from evennia import DefaultCharacter 
+from evennia.utils import logger
+from evennia.contrib.actions.actions import Turn
+from evennia.contrib.actions.scripts import ActionSystemScript
+from evennia.contrib.actions.utils import (process_queue, format_action_desc,
+    handle_invalid)
+
+
+class CharacterActionHandler(object):
+    """
```
Griatch: This docstring seems incomplete. :)

evennia/contrib/actions/handlers.py
```diff
@@ -0,0 +1,752 @@
...
+
+    def __init__(self, owner):
+        """
+        sets owner (Character), the character to which this handler is attached
+        """
+        self.owner = owner
+
+
+    def setup(self, override=False):
```
Griatch: This needs a docstring.

evennia/contrib/actions/handlers.py
```diff
@@ -0,0 +1,752 @@
...
+                # present movement mode (e.g. Moveout)
+
+                # The reason we are not selecting actions by key is that
+                # someone, somewhere may want to subclass MoveOut and MoveIn.
+                action['at_cancel'](action)            
+
+        self.owner.db.actions['movetype'] = value
+
+    @property
+    def movebps(self):
```
Griatch: This name suggests this is the very action of moving body parts around. The docstring  however suggests this is just where one defines the mapping between bodyparts and movement mode. Maybe something to the effect of `bodypart_movement_map` would be more instructive of its purpose?

evennia/contrib/actions/handlers.py
```diff
@@ -0,0 +1,752 @@
...
+                self.owner.actions.remove(action)
+
+        if ongoing:
+            # Then cancel all ongoing actions
+            for action in self.owner.location.actions.list:
+                if action['owner'] == self.owner:
+                    action['at_cancel'](action)
+
+
+    def unpuppet(self):
```
Griatch: Is this to be called specifically when unpuppeting a character? If so this should be mentioned in the docstring. Otherwise this name clashes with the normal meaning of "unpuppet" in Evennia.

evennia/contrib/actions/handlers.py
```diff
@@ -0,0 +1,752 @@
...
+    """
+
+    def __init__(self, owner):
+        """
+        sets owner (Room), the room to which this handler is attached
+        """
+        self.owner = owner
+
+
+    def setup(self, override=False):
```
Griatch: This requires a docstring detailing when this is called and maybe why it's here alongside `__init__`.

evennia/contrib/actions/handlers.py
```diff
@@ -0,0 +1,752 @@
...
+            self.script = self.owner.scripts.get("ActionSystemScript")[0]
+            self.script.start()
+
+    @property
+    def list(self):
+        """
+        the list of all ongoing actions in the room
+        """
+        return self.owner.db.actions['list']
+
```
Griatch: Btw, according to PEP8, only single blank space between methods in a class (double line breaks between classes).

evennia/contrib/actions/scripts.py
```diff
@@ -0,0 +1,226 @@
+from math import ceil
```
Griatch: A docstring is needed, detailing the purpose of this module.

evennia/contrib/actions/scripts.py
```diff
@@ -0,0 +1,226 @@
...
+from time import time
+from evennia import DefaultScript, DefaultCharacter
+from evennia.utils import logger
+from evennia.contrib.actions.actions import Turn
+from evennia.contrib.actions.utils import process_queue, validate, handle_invalid
+
+
+class ActionSystemScript(DefaultScript):
```
Griatch: A class doc string is missing here.

evennia/contrib/actions/scripts.py
```diff
@@ -0,0 +1,226 @@
...
+    def at_script_creation(self):
+        self.key = "ActionSystemScript"
+        self.desc = "Processes the real-time action system in the current room."
+        self.interval = 1  # seconds
+        #self.repeats = 5  # repeat only a certain number of times
+        self.start_delay = False  # wait self.interval until first call
+        self.persistent = True
+
+    def at_repeat(self):
```
Griatch: This probably belongs in the class docstring?

evennia/contrib/actions/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+from evennia.contrib.actions.typeclasses import ActionCharacter, ActionRoom, ActionExit
+from evennia.contrib.actions.commands import ActionCmdSet 
+from evennia.contrib.actions.debug import ActionDebugCmdSet
+from evennia.contrib.actions.setup import setup
+
+__all__ = [Action, ActionCharacter, ActionRoom, ActionExit, ActionCmdSet, 
+    ActionDebugCmdSet, setup]
+
+
+def setup(override=False):
```
Griatch: This code should not be repeated both here and in setup.py. I suggest moving this functioninto the utils.py module instead and removing setup.py completely.

evennia/contrib/actions/setup.py
```diff
@@ -0,0 +1,20 @@
+from evennia import ObjectDB
```
Griatch: I suggest to remove this file and move this code out from here (and from `__init__.py` into utils.py.

evennia/contrib/actions/utils.py
```diff
@@ -0,0 +1,387 @@
+from evennia import DefaultRoom, DefaultCharacter, DefaultObject, ObjectDB
```
Griatch: This requires a module docstring.

evennia/contrib/actions/utils.py
```diff
@@ -0,0 +1,387 @@
...
+from evennia import search_channel
+
+
+
+
+def same(action, other):
```
Griatch: Missing a docstring.

evennia/contrib/actions/utils.py
```diff
@@ -0,0 +1,387 @@
...
+
+            errors += "TargetNotInRoom "
+
+    if errors:
+        return errors
+    else:
+        return "Valid"   
+
+
+def handle_invalid(action, validate_result):
```
Griatch: Missing a docstring.

evennia/contrib/actions/utils.py
```diff
@@ -0,0 +1,387 @@
...
+        room + " has been found invalid with error " + validate_result)
+
+    if action['room'] and action in action['room'].actions.list:
+        action['room'].actions.remove(action)
+
+    if action['owner'] and action in action['owner'].actions.list:
+        action['owner'].actions.remove(action)
+
+
+def share_bps(action, other):
```
Griatch: Make this comment into a docstring instead.

evennia/contrib/actions/setup.py
andrei-pmbcn: I cannot put setup in utils.py without changing isinstance(self, ActionCharacter) etc. to isinstance (self, DefaultCharacter) etc, as importing things from evennia.contrib.actions.typeclasses.py into utils.py will otherwise create import errors. I don't find this surprising, as typeclasses.py imports from handlers.py, which imports from utils.py as well as from actions.py, which in turn imports from utils.py. If there is any way around this, please let me know.

Griatch: @andrei-pmbcn 

Yes, this is called a circular import error and is solved by delaying the import. It's best to break the cycle in the utils.py modules to allow utils to easily be imported from the other modules.  Here is the recommended and efficient way to do a delayed import: 

```python
_ACTION_CHARACTER = None

def function_needing_actionchar():
    global _ACTION_CHARACTER
    if not _ACTION_CHARACTER:
        from evennia.contrib.actions.typeclasses import ActionCharacter as _ACTION_CHARACTER

    # use _ACTION_CHARACTER below ...
```

andrei-pmbcn: Thanks! I'll try this solution.

evennia/contrib/actions/handlers.py
andrei-pmbcn: How do you feel about "bpsformove"? It's shorter (more python-themed) and still gets the point across. Alternatively, maybe "bpmovemap" or "bp_move_map". If you prefer "bodypart_movement_map", I'll go with that in the end. What are your thoughts on this?

Griatch: Python standard is to try to name variables to clear, descriptive names using `_` to separate words. If the variable was only used internally in some calculation or other it wouldn't matter all that much, but this appears to be a property people are expected to set and deal with from the outside. 
At least to me reading the  code without having written it, "bodyparts" over "bp" and "bodypart_movement_map" over "bpsformove" appears clearer and makes it easier to trace through the code while more directly conveying of what is going on.

andrei-pmbcn:
@Griatch, I've handled everything you requested. You should not longer get the error when trying to set a bodypart_movement_map and a movetype.

The changes I made in addition to the ones you requested involve allowing movement actions to be queued, making Turn actions issue no message when created at the start of turn-based mode (this avoids a lot of pointless spam) and providing pre_process_action and post_process_action hooks on character object methods, which should come in handy for AI-related tasks.

Griatch:
```
  File "/home/griatch/Devel/MUD/vienv/bin/evennia", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/bin/unix/evennia", line 10, in <module>
    main()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/evennia_launcher.py", line 1268, in main
    server_operation(option, service, args.interactive, args.profiler, args.logserver)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/evennia_launcher.py", line 1064, in server_operation
    error_check_python_modules()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/evennia_launcher.py", line 805, in error_check_python_modules
    imp(settings.BASE_CHARACTER_TYPECLASS)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/evennia_launcher.py", line 750, in imp
    __import__(mod, fromlist=[fromlist])
  File "/home/griatch/Devel/MUD/evennia/game/typeclasses/characters.py", line 11, in <module>
    from evennia.contrib.actions import ActionCharacter
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/__init__.py", line 1, in <module>
    from evennia.contrib.actions.actions import Action
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/actions.py", line 14, in <module>
    from evennia.contrib.actions.utils import (triage, process_queue, validate,
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/utils.py", line 175
    if isinstance(other['bodyparts'], list):
                                           ^
IndentationError: unindent does not match any outer indentation level
```
I've fixed this in my work copy to be able to test, but it should be fixed in the PR.

Griatch:
```python
  File "/home/griatch/Devel/MUD/vienv/bin/evennia", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/bin/unix/evennia", line 10, in <module>
    main()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/evennia_launcher.py", line 1268, in main
    server_operation(option, service, args.interactive, args.profiler, args.logserver)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/evennia_launcher.py", line 1064, in server_operation
    error_check_python_modules()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/evennia_launcher.py", line 805, in error_check_python_modules
    imp(settings.BASE_CHARACTER_TYPECLASS)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/evennia_launcher.py", line 750, in imp
    __import__(mod, fromlist=[fromlist])
  File "/home/griatch/Devel/MUD/evennia/game/typeclasses/characters.py", line 11, in <module>
    from evennia.contrib.actions import ActionCharacter
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/__init__.py", line 2, in <module>
    from evennia.contrib.actions.typeclasses import ActionCharacter, ActionRoom, ActionExit
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/typeclasses.py", line 14, in <module>
    from evennia.contrib.actions.handlers import (CharacterActionHandler, 
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/handlers.py", line 20, in <module>
    from evennia.contrib.actions.scripts import ActionSystemScript
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/scripts.py", line 77
    result = action['at_attempt'](action)
                                        ^
IndentationError: unindent does not match any outer indentation level
```
Another one, on starting the server.

Griatch:
And another one: 
```python
  File "/home/griatch/Devel/MUD/vienv/bin/evennia", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/bin/unix/evennia", line 10, in <module>
    main()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/evennia_launcher.py", line 1268, in main
    server_operation(option, service, args.interactive, args.profiler, args.logserver)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/evennia_launcher.py", line 1064, in server_operation
    error_check_python_modules()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/evennia_launcher.py", line 805, in error_check_python_modules
    imp(settings.BASE_CHARACTER_TYPECLASS)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/evennia_launcher.py", line 750, in imp
    __import__(mod, fromlist=[fromlist])
  File "/home/griatch/Devel/MUD/evennia/game/typeclasses/characters.py", line 11, in <module>
    from evennia.contrib.actions import ActionCharacter
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/__init__.py", line 2, in <module>
    from evennia.contrib.actions.typeclasses import ActionCharacter, ActionRoom, ActionExit
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/typeclasses.py", line 14, in <module>
    from evennia.contrib.actions.handlers import (CharacterActionHandler, 
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/handlers.py", line 20, in <module>
    from evennia.contrib.actions.scripts import ActionSystemScript
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/scripts.py", line 181
    `
    ^
IndentationError: unexpected unindent
```

Griatch:
And a final one, when the server tries to start from a completely fresh database: 

```python
  application = service.loadApplication(filename, style, passphrase)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/twisted/application/service.py", line 411, in loadApplication
    passphrase)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/twisted/persisted/sob.py", line 224, in loadValueFromFile
    eval(codeObj, d, d)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/server.py", line 481, in <module>
    EVENNIA = Evennia(application)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/server.py", line 165, in __init__
    self.run_initial_setup()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/server.py", line 257, in run_initial_setup
    initial_setup.handle_setup(0)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/initial_setup.py", line 200, in handle_setup
    setup_func()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/server/initial_setup.py", line 91, in create_objects
    nohome=True)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/utils/create.py", line 123, in create_object
    new_object.save()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/utils/idmapper/models.py", line 377, in save
    super(SharedMemoryModel, self).save(*args, **kwargs)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/db/models/base.py", line 708, in save
    force_update=force_update, update_fields=update_fields)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/db/models/base.py", line 745, in save_base
    update_fields=update_fields, raw=raw, using=using)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/dispatch/dispatcher.py", line 192, in send
    response = receiver(signal=self, sender=sender, **named)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/typeclasses/models.py", line 77, in post_save
    instance.at_first_save()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/objects/objects.py", line 879, in at_first_save
    self.at_object_creation()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/typeclasses.py", line 49, in at_object_creation
    self.actions.setup()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/handlers.py", line 68, in setup
    self.movetype = None
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/handlers.py", line 205, in movetype
    if (self.owner.actions.bodypart_movement_map and
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/handlers.py", line 241, in bodypart_movement_map
    return self.owner.db.actions['bodypart_movement_map']
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/utils/dbserialize.py", line 180, in __getitem__
    return self._data.__getitem__(key)
exceptions.KeyError: 'bodypart_movement_map'


Failed to load application: 'bodypart_movement_map
```

andrei-pmbcn:
Griatch, I apologize for these errors and am surprised I missed them. The first three must have taken place after I yank-put several lines of code. What always happens is that the first line and the last line I put have offset indentation for some reason, so I have to manually change their indentation level. I will see how to work around this.

The other bug stems from the fact that I wrote no unit-tests for this contrib and did not check whether spawning still works after I made some changes to the code. The AI contrib will definitely be fully covered by unit tests.

Once again, I apologize for all the bugs and hope my noobishness will wear off soon.

Griatch:
Things are improving! I can start the server with things running and I can look at the help files. I still cannot actually test the action functionality though. Notably, it seems like the `actions` dict on the Character is not properly initialized or set at the right time for me. 

After making sure I've reset my typeclass to the one inheriting from ActionCharacter, -Room and -Object, this is the status of the `actions` Attribute: 

```
@py self.db.actions
{'list': [], 'turnbased': False, 'movespeed': None, 'active': True, 'new': 'ignore', 'bodypart_movement_map': None, 'movetype': None}
```

Now I try to run a selection of different commands, I get a traceback that the system is looking for a key in this dictionary that is not there, leading to a traceback. Here are some examples:

- `turns` - looks for `actions["view"]` unsuccessfully
- `@@actfast/slow/larm - looks for `actions[`mode"]` unsuccessfully

The traceback looks like this: 

```python
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/commands/cmdhandler.py", line 464, in _run_command
    ret = yield cmd.func()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/debug.py", line 138, in func
    DebugActionRightArm(self.caller, self.caller.location)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/debug.py", line 154, in __init__
    duration=duration,
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/actions.py", line 172, in __init__
    self.at_creation(action)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/actions.py", line 188, in at_creation
    handler = triage(action)
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/utils.py", line 339, in triage
    if (action['room'].actions.mode == 'TB' and not action['non_turn'] and
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/handlers.py", line 433, in mode
    return self.owner.db.actions['mode']
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/utils/dbserialize.py", line 180, in __getitem__
    return self._data.__getitem__(key)
KeyError: 'mode'
```

Also `@@actsetup` gives a traceback - this one is different though and is a simple omission of a variable in the global scope: 

```python
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/commands/cmdhandler.py", line 464, in _run_command
    ret = yield cmd.func()
  File "/home/griatch/Devel/MUD/evennia/evennia-trunk/evennia/contrib/actions/debug.py", line 254, in func
    actions.setup(override=False)
NameError: global name 'actions' is not defined
```

andrei-pmbcn:
Griatch, the room was indeed not set up, and the @@actsetup command was broken because I had moved functions around across modules without fully updating their references. Unit testing would have caught this. The extensive unit tests I have built for the AI system should ensure that that particular contrib will run much more smoothly than this one.

I will test the action system myself on a clean database tomorrow.

Griatch:
@andrei-pmbcn 

Ok, will await your testing before further review.

andrei-pmbcn:
@Griatch I tested the action system on a new database, fixed the bug you pointed at and fixed another bug as well. Everything should work smoothly now.

Griatch:
Goodie, I can now run this properly and test out the action system! It's really quite nifty! I also really like that you have supplied the debug commands - the existence of those is great, both for me evaluating things and for the potential user wanting to figure out if this system appeals to them. 

Now that the system runs I have some further feedback on it as someone coming at it as a newbie of the Action system. Sorry if this is a bit long, but it's a large contrib. 

- The @@actfast command reports when it finishes while @@actslow does not. You can see the latter finishing with the @@actdebug mode, but it would be good to show this in text too. 
- In the turnbased mode, you can enter an unlimited number of @@actfoe commands out of the box (no change of the actsetting). While it's just for testing, it would be better to limit this in the same way @@actfriend is limited.
- The *actsettings* command help entry is not very clear: It is confusing to give its syntax as `actsettings <argument> <option>` when the only supported argument is `actsettings new ignore|override|queue`. I understand this is meant to be expandable, but until it's expanded the help entry should reflect what it supports. Also, if "new" is the only allowed argument, why having it at all?
- It's not strictly needed, but a good idea is to give helpful information when not giving arguments to a command that requires it. For example,  when seeing "please specify an argument for the actsetting command" it requires the player to do one extra step and read the help entry (it's not even correct since both an argument and an option is needed). A more useful return would be "Usage: actsetting new ignore|overide|queue". Or even "Usage: actsetting <argument> <option>" to at least show the correct syntax at a glance. This goes for commands in general. 
- I think it would be a good idea to group all Action-related commands (as well as the Debug commands) under one "Actions" help category. People can shuffle this around later if they want, but for trying things out it makes things more self-contained, especially since not all commands are obviously related to actions (like queue and stop). 
- The `queue` command does not seem to do anything. At least it does not report the existence of any actions. Maybe it's supposed to be removed in favor of `actions`?
- When entering turn-based mode, it feels like there should some brief help entry about it available Maybe the `actions` command output should mention the "stop" command, for example. 
- The sequential form of turn-based mode makes for very slow play, especially since you, depending on your actsetting, can spend your time queuing any number of commands that will take forever to step through. I can see a risk of abuse here. I would suggest limiting the amount of commands possible can sat during their turn set somehow - maybe only one per turn or only as many as could time-wise fit into the time slot.
- Is there some way to prematurely end your turn when you are happy with your commands?
- If Player A initiates turn-based mode, this affects Player B. But if Player B tries `tb off` the error is that they are "not in turn-based mode to begin with". Which is not an intuitive error for the user when they clearly are affected by the turn-order. All affected by turn-based mode should be noted as being in that mode (see also below). 
- Meanwhile player A can exit turn-based mode at any time, at which time real-time starts up again. This looks quite cool when it happens but I suspect it may be a possible exploit to fire turn-based mode, write something (with plenty of time to do it) and then turn it off again without giving B a chance to reply (they could trigger tb themselves but sounds kinda meta-gamey as a design). This would be more natural if all affected by tb mode would also be identified as such (see previous point). My suggestion would be that turn it off should be something all involved must agree to ("Player A wants to turn off Turn-based mode. Enter "tb off" if you agree"). In general it sounds reasonable that if all involved Players turn off their turn-based mode, it goes away immediately. To avoid one offline player locking up everyone else, maybe one must explicitly state "tb on" to keep it  or your"vote" will default after a turn (or something like that).
- In turn-based mode, movements between rooms cannot be canceled with the `stop` command: The stop command goes through, but the action is still in the list and is performed. 
- In the README, don't use triple back-ticks to mark verbatim inline text, this messes up formatting. In running text, you just use single back-ticks to mark inline stuff, like `verbatim code`.

Griatch:
Any progress on this?

Griatch:
Since it doesn't look like there's any more activity on this, I'm closing this PR now.

Issue #1042, opened 2016-09-05
EvMenu on session object generates traceback


feend78:
#### Brief summary of issue / Description of requested feature:

In Ainneve, chargen is handled by an EvMenu instance that is passed a **session** instead of a character. It displays one menu node, but crashes with any input.

The traceback points to evmenu.py, line 237, which was recently added in commit f6286ac.

#### Steps to reproduce the issue / Reasons for adding feature:

* Updated to current master of both evennia and ainneve
* Ran `pip install --upgrade -e <evennia home>`
* Deleted the previously-generated twistd.bat file and evennia.db3 file
* Ran migrations and started the ainneve server
* Connect as superuser 'Ainneve'
* Run `@ic ainneve` - chargen begins
* Enter any selection from the first menu node

#### Error output / Expected result of feature
```
2016-09-05T10:58:09-0500 [AMPProtocol,0,127.0.0.1] [::] Traceback (most recent call last):
2016-09-05T10:58:09-0500 [AMPProtocol,0,127.0.0.1] [::]   File "c:\users\dan\sandbox\evennia\evennia\commands\cmdhandler.py", line 464, in _run_command
2016-09-05T10:58:09-0500 [AMPProtocol,0,127.0.0.1] [::]     ret = yield cmd.func()
2016-09-05T10:58:09-0500 [AMPProtocol,0,127.0.0.1] [::]   File "c:\users\dan\sandbox\evennia\evennia\utils\evmenu.py", line 237, in func
2016-09-05T10:58:09-0500 [AMPProtocol,0,127.0.0.1] [::]     caller = caller.player
2016-09-05T10:58:09-0500 [AMPProtocol,0,127.0.0.1] [::] AttributeError: 'Player' object has no attribute 'player'
2016-09-05T10:58:09-0500 [AMPProtocol,0,127.0.0.1] [::] Traceback (most recent call last):
2016-09-05T10:58:09-0500 [AMPProtocol,0,127.0.0.1] [::]   File "c:\users\dan\sandbox\evennia\evennia\commands\cmdhandler.py", line 599, in cmdhandler
2016-09-05T10:58:09-0500 [AMPProtocol,0,127.0.0.1] [::]     ret = yield _run_command(syscmd, syscmd.key, sysarg)
2016-09-05T10:58:09-0500 [AMPProtocol,0,127.0.0.1] [::] ErrorReported
```

PR #1045, opened 2016-09-07
Fixed EvMenu error when caller does not have a player attribute

feend78:

Added a check that the caller has a `player` property before attempting to access it.

This issue has completely broken chargen in Ainneve.

Addresses issue #1042.

### DIFF
--- a/evennia/utils/evmenu.py
+++ b/evennia/utils/evmenu.py
@@ -234,18 +234,19 @@ class CmdEvMenuNode(Command):
             if _restore(caller):
                 return
             orig_caller = caller
-            caller = caller.player
-            menu = caller.ndb._menutree
-            if not menu:
-                if _restore(caller):
-                    return
-                caller = self.session
+            if hasattr(caller, 'player'):
+                caller = caller.player
                 menu = caller.ndb._menutree
                 if not menu:
-                    # can't restore from a session
-                    err = "Menu object not found as %s.ndb._menutree!" % (orig_caller)
-                    orig_caller.msg(err)
-                    raise EvMenuError(err)
+                    if _restore(caller):
+                        return
+            caller = self.session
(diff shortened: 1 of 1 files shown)


Griatch:
@feend78 

What *is* the `caller` entity in the Ainneve chargen then? There should only be three possibilities - an Object, a Player or a Session. If a Player, then the first case (caller.ndb._menutree) should be true. If an Object or a Session, then the .player property is also defined ...

feend78:
@Griatch 

It is a Session. If sessions have a .player property, then I'm not sure why I'm getting the traceback from issue #1042.

Here's where the EvMenu is launched for Ainneve: 

https://github.com/evennia/ainneve/blob/master/commands/chargen.py#L191

feend78:
So I dropped a debug statement in just before the line that throws an error, and it appears that at that point, `caller` is a Player object, rather than a session.

Griatch:
Oh, I understand. You are sending the Session to the EvMenu, so that's what it will use to store its internal state, but the *command* of the menu is operating with the assumption that caller is a Player. The error comes from this disconnect.

evennia/utils/evmenu.py
```diff
@@ -234,18 +234,19 @@ def _restore(caller):
             if _restore(caller):
                 return
             orig_caller = caller
-            caller = caller.player
-            menu = caller.ndb._menutree
-            if not menu:
-                if _restore(caller):
-                    return
-                caller = self.session
+            if hasattr(caller, 'player'):
```
Griatch: This looks like an incorrect indent. This check will as written run also if a menu was actually found in the first if statement above (c.f. the indent of the removed code).

Griatch:
Merged with rebase and some tweaks. Closing.

Issue #613, opened 2014-11-08
Webclient improvements


Kelketek:
I noticed that while we've talked about this frequently, we don't actually have an issue in for it yet.

The webclient works as a pretty decent proof of concept, and it can even be used by the casual user. However, there's a good deal of room for improvement, especially now that we have access to OOB.

Here are a few things that should be added:

1. ~~Share login session with Django. If you're logged into Django, the web client should log you into your player automatically.~~ Implemented
2. Play a sound (can be toggled off) and make some indication in the title that a new message has arrived. This is possibly the most important thing in terms of immediate practicality-- on the game I ran, this was the feature that allowed many users to stay with the webclient, and not bother downloading their own.
3. ~~Modularity. There should be a way to hook tabs or windows into the client so that coders can implement their own custom plugins without having to remake the GUI from scratch. This is going to be the bulk of the task.~~ Implemented by way of separate GUI lib. Defer to #614.
4. Saving user preferences. The webclient at this point will have enough features that its configuration state should be savable for the player.

Griatch:
There is some work done (not yet online) on a new webclient. This one is based on a javascript component (`evennia.js`)  that relays incoming text data to a javascript function. These functions can be simply added to (modularity) but just importing one's own javascript module and append one's function to the relevant object dictionary mapping from `evennia.js`. 

While not ready for prime time yet, the webclient component I'm toying with currently gives the dev the ability to use any js gui toolkit (I use Foundation for now) and just mark up their containers with `class=` markers that `evennia.js` recognizes. The component comes with a few convenience functions so far, like a function to relay data to a specified class (accessed with something like `msg("Text", ooc=("divclass", "textarea")` as well as single- and multiple-line input fields with mud-like history handling. 
.
Griatch

Cecelik:
I am completely new to Evennia and I have a lot of reading to do before I can confidently give precise feedback. However, if I understand Griatch right, I think he is taking this in the right direction. What you ought to do is equip web developers with something that allows them to completely paint their own experience using whatever client-side methods that they wish. The language of the web is Javascript and Javascript uses JSON for all things communication. Give us the flexibility to see just JSON and we will build you elaborate engines that let you play whatever kinds of sounds you wish. On any platform. :)

hakupaku:
Hey there, I had a brief discussion with @Griatch about the form of the web component and he asked me to contribute my thoughts to this issue.

[...]

PR #1046, opened 2016-09-07
Web Client Notifications

jrobertsgb:

Adds notifications to the webclient; changing tab title, badge on the favicon and popup tray notifications.

A common request to be able to see when people post ingame, as other clients are able to flash when activity happens.

Related to #613 
Perhaps in future a sound could be played on notification? Need to find a suitable sound.

### DIFF
--- a/evennia/web/webclient/static/webclient/js/webclient_gui.js
+++ b/evennia/web/webclient/static/webclient/js/webclient_gui.js
@@ -203,6 +203,8 @@ function onText(args, kwargs) {
     mwin.animate({
         scrollTop: document.getElementById("messagewindow").scrollHeight
     }, 0);
+
+    onNewLine(args[0], null);
 }
 
 // Handle prompt output from the server
@@ -241,6 +243,22 @@ function onBeforeUnload() {
     return "You are about to leave the game. Please confirm.";
 }
 
+var unread = 0;
+var originalTitle = document.title;
+
+function onVisibilityChange() {
+  if(!document.hidden) {
+    document.title = originalTitle;
+  }
+}
+
+function onNewLine(originator, text) {
+  if(document.hidden) {
+    unread++;
+    document.title = "(" + unread + ") " + originalTitle;
+  }
+}
+
 //
 // Register Events
 //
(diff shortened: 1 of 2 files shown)


Griatch:
Looks good. In Firefox I get only tab-notification (this looks very good though). In Chrome I get the row of notifications in the lower right, but they seem to stay up for a very long time which is kind of annoying. I almost wonder if the existing tab marker with an optional sound would not be less intrusive than those lower right popups ...

evennia/web/webclient/static/webclient/js/webclient_gui.js
```diff
@@ -239,16 +241,74 @@ function onDefault(cmdname, args, kwargs) {
...
     return "You are about to leave the game. Please confirm.";
-}
+}
+
+var unread = 0;
+var originalTitle = document.title;
+var focused = true;
+var favico;
+
+/*function onVisibilityChange() {
```
Griatch: Should this be removed?

evennia/web/webclient/static/webclient/js/webclient_gui.js
```diff
@@ -239,16 +241,74 @@ function onDefault(cmdname, args, kwargs) {
 // in Firefox, there it's a standard error.
 function onBeforeUnload() {
     return "You are about to leave the game. Please confirm.";
-}
+}
+
+var unread = 0;
```
Griatch: All these functions should have a comment block explaining wthat they are there for (c.f. other callbacks in this section).

Griatch:
As for sound, something like a pen-scratch sound always struck me as suitable for MUDs and not as intrusive as a beep. Something like a fraction-of-a-second sample out of this: https://www.freesound.org/people/OwlStorm/sounds/320151/. Just an idea. 

... But I believe we would probably need to have an option pane if we want to have sounds - some would hate it whichever one we picked and would want to turn it off. So sounds may be outside the scope of this PR (unless you are feeling adventurous).

Griatch:
To be clear, this merge is pending dealing with the few issues mentioned. I must also say that the pop-ups used are very obnoxious (at least on Chrome under Linux) so I think the popups should be removed and favor only the tab notification. 

Eventual sound must be something settable as an option, but since adding an option panel is outside the scope of this PR, sound should probably be deferred as a future project.

Griatch:
@jrobertsgb Any plans to update this further?

Griatch:
Merged with some modifications. I commented out the full notification for now, pending a webclient config pane.

Issue #1038, opened 2016-09-05
Bug: user defined aliases for channels don't appear to function


TehomCD:
#### Brief summary of issue / Description of requested feature:
I thought it might just be something I messed up in my database migration when converting between versions, but I made a test character on Horizondark and ran into the same issue - aliases added with 'addcom' don't seem to be recognized currently. You'll see the alias listed under @channels, but attempting to execute the command <alias> <message> will return a command not found error.

#### Steps to reproduce the issue / Reasons for adding feature:

1. addcom foo=public
2. @channels
3. foo testing

Griatch:
Confirmed. This appears to be unrelated to the latest Attribute/Nick changes but an effect of the inputfunc not correctly handling the input to the nickreplace.

Griatch:
Closed in PR #1055

PR #1055, opened 2016-09-11
Attempt to fix aliases for channels not working (#1038)

TehomCD:
So from what I was able to gather, when the nickname/alias was added for the channel, it didn't follow the template format specified, and was just adding the nickname/channel keys. In order to function, they need to have '$1' to specify that they take arguments. As a note, I tried using '*' initially first for both of them, but that doesn't build allowed arguments in the outgoing template, so wouldn't work.

I couldn't figure out why this was a problem specific to channel commands, so the category check for channels probably isn't the best way of doing things. It does seem to work when I tested it, but I'm going to guess there's a better way of doing this.

### DIFF
--- a/evennia/typeclasses/attributes.py
+++ b/evennia/typeclasses/attributes.py
@@ -800,9 +800,9 @@ class NickHandler(AttributeHandler):
 
         """
         if category == "channel":
-            key += " $1"
-            replacement += " $1"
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks, merged with rebase. Closing #1038.

PR #1069, opened 2016-09-20
New `msg_contents` for ContribRPRoom

feend78:

Added the ability to include {}-style named format tokens in the message, along with a `mapping` argument containing a dict of tokens mapping to objects. The updated `msg_contents` then calls `get_display_name` on the objects for each viewer.

The RPSystem relies heavily on the use of `get_display_name`. The default `msg_contents` method did not provide any way to insert display names into room messaging.

### DIFF
--- a/evennia/contrib/rpsystem.py
+++ b/evennia/contrib/rpsystem.py
@@ -1342,48 +1342,7 @@ class ContribRPRoom(ContribRPObject):
     """
     Dummy inheritance for rooms.
     """
-    def msg_contents(self, message, exclude=None, from_obj=None, mapping=None, **kwargs):
-        """
-        Emits a message to all objects inside this object.
-
-        Args:
-            message (str): Message to send.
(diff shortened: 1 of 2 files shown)


Griatch:
The code itself looks good, but looking at it, wouldn't it be better to add this behavior to the default `msg_contents` on `objects/objects.py` rather than just to the rproom? The `mapping` would need to be the last keyword in that case, but apart from that it should not mess with existing code but just extend the functionality.

BlauFeuer:
I would add it to `objects/objects.py` or at least into `ContribRPObject` simply to cover cases where objects are vehicles, or a held radio might be heard by another held item that responds to it.

But this functionality would save me a lot of code duplication as I convert to sending messages that include object names created by calls to `get_display_name`.

feend78:
I am signing off for the night. I will update this pull request and move
`msg_contents` to DefaultObject tomorrow.

Griatch:
Thanks, merged with a rebase.

PR #1070, opened 2016-09-20
added the `append` saver method

feend78:

Recently, `_SaverDequeue` was added to the dbserialize module, so I used it in Ainneve combat. However, I found that the `append` method was not present. This PR adds it.

To support the deque API.

I do not believe there is an issue for this problem.


Griatch:
Huh, I presumed the SaverMutable would supply this by inheritance but you are right that using append with the _saverdeque causes an error. Merged with a rebase, thanks!

PR #1071, opened 2016-09-21
Resolves traceback error when object is moved to a room rather than a…

TehomCD:
An object being moved from None location generated a traceback if they were moved to an object that did not have a player puppeting it, such as a room.


Griatch:
Looks good, I won't have access to my dev machine for a few more days though, so the merge of this will be a little delayed.

PR #1072, opened 2016-09-21
Adding 'blank=True' to null fields in Msg for validation in django admin

TehomCD:
Although Evennia has the admin model for Msg() turned off by default, I was using it in my old version because of the ways I'm using Msg objects in Arx: they're used for many things, from character journals, to in-game messengers, etc, all separated by their header information. But I had forgotten that I had set blank=True for all the null fields before: this is necessary in order to use django's admin changeform to change Msg objects. Without it, all null fields are required to be set for its form validation, which means that a Msg object requires you to set a tag, an external sender, a channel to exclude, a player to exclude, an object to exclude, etc.

Allow Msg to function in admin.

I don't know why django requires a migration to be created for something that doesn't actually change the database schema, but it does.


Griatch:
Migrations are done for every minor change in models. I'm pretty sure any modification of how fields are formatted (allowing blanks/null etc) could lead to potential hidden errors and differences over time and thus must be tracked, especially if wanting to backtrack later. Change looks good, will merge when I can.

Issue #1076, opened 2016-09-24
Creating a player from pre-login containing apostrophe


BlauFeuer:
#### Brief summary of issue / Description of requested feature:
Creating a character from pre-login containing apostrophe yields unexpected and varied results, requires manual intervention.

#### Steps to reproduce the issue / Reasons for adding feature:

1. `create T'Pring Vulcan` 
2. `create "T'Pau" Vulcan`
3. `create Muad’Dib Dune`

#### Error output / Expected result of feature

The first creates a player and character named `T`

The second yields:
`Usage (without <>): create <name> <password>`
`If <name> or <password> contains spaces, enclose it in quotes.`

The third yields:
`Playername can max be 30 characters or fewer. Letters, spaces, digits and @/./+/-/_ only.`

The user who attempts this receives conflicting information and no success, requiring a manual renaming to name the character object what the user wants.

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
Latest revision, Evennia, master branch

Escape input to allow a larger variety of character names, instruct that creating a character with a name containing an apostrophe requires special action - renaming in web admin after creation.

robbintt:
I tested these as well, results are the same:

```
 Usage (without <>): create <name> <password>
If <name> or <password> contains spaces, enclose it in quotes.
create Muad`dib Dune

 Playername can max be 30 characters or fewer. Letters, spaces, digits and @/./+/-/_ only.
create "T'pau" Vulcan

 Usage (without <>): create <name> <password>
If <name> or <password> contains spaces, enclose it in quotes.
create T'Pring Vulcan
A new account 'T' was created. Welcome!

You can now log with the command 'connect T <your password>'.
```

robbintt:
Submitted PR to fix this behavior by simplifying the `create` and `connect` parser.  See the bottom of https://github.com/evennia/evennia/pull/1078 for more details on future actions.

BlauFeuer:
Since the character takes on the name of the player when both are created pre-login, this also affects character names by indirectly limiting them when created this way.

If the assumption that passwords with spaces in them will be quoted then anything left of the last quoted group is the player name  (after parsing out "connect" or "create").  I'm just concerned that there might be unseen issues with player names containing the `'` character.  I know some people who will insist on trying to use it in the name, so the instructions at pre-login need to be clear.

And if there's consequences in renaming the player after the fact, I want to be sure not to do that unless I know it's safe to do so.

[...]

PR #1080, opened 2016-09-26
Removed complex parsing in favor of disallowing spaces in passwords.

robbintt:

Removed the complex parsing of doublequotes on `create` and `connect` commands in favor of a parse light system that disallows spaces in the password.

bugfix - https://github.com/evennia/evennia/issues/1076

###### Passwords with spaces:
Future goal: create/connect module to accept password in response to a `create` or `connect` command instead of on the same line.  Use a python module for securely handling passwords.

###### Single Quotes in Names:
Please submit an enhancement request to use single quotes in names.

### DIFF
--- a/evennia/commands/connection_screen.py
+++ b/evennia/commands/connection_screen.py
@@ -15,7 +15,7 @@ DEFAULT_SCREEN = \
  If you need to create an account, type (without the <>'s):
       {wcreate <username> <password>{n
 
- If you have spaces in your username, enclose it in quotes.
+ Passwords cannot contain spaces.
  Enter {whelp{n for more info. {wlook{n will re-show this screen.
 {b=============================================================={n""" \
 % (settings.SERVERNAME, utils.get_evennia_version())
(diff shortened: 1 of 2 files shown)


Griatch:
Thanks for the PR! Some points though: 

1. This contains a critical error which suggests it has not been tested before submittal.
2. While removing the space ability from passwords is interesting, it's not in itself related to resolving #1076 - this actually removes functionality to fix an edge-case bug. 
3. How will this change affects existing users that try to log back into their games with a password containing spaces? 
4. While issue #1076 needs to be fixed, using quotes to separate multi-word names and passwords is the more flexible solution. If passwords-with-spaces is not desired this is something for a custom login command to handle on a per-game basis. 
5. The password is safely handled by the django auth system, the safety check is basically a copy of the validation Django itself does (and Django allows spaces in passwords as most modern system does). 

I'm open to discussion and arguments but at least from a first glance I'm sorry to say this doesn't look really like that useful a change for the problem at hand.

robbintt:
I'll address only that issue in a separate PR and keep thinking on the password thing.

Griatch:
I think allowing spaces in passwords is a useful feature; but maybe one could improve the parsing to favor the player name rather than the password (that is, so you must put the password into quotes if you want spaces in it rather than to have to do so for the username)?

robbintt:
I see what you are saying, that would be nice.

PR #1083, opened 2016-09-26
modified `has_cmdset` to accept either a cmdset's python path or key

feend78:

The cmdsethandler's `add` and `remove` methods all will accept a python module path as their argument, but it appears that `has_cmdset` only returns true if you search on the cmdset's key. Added logic to enable it to work with either key or module path.

Ainneve's turn-based combat handler does a lot of adding and removing cmdsets to combat participants during different phases of the combat turn. I found I needed to use `has_cmdset` to ensure that a given cmdset is added no more than once. For these cmdset operations I chose to manage cmdsets by module path string and forego the use of cmdset keys, but it revealed this problem with `has_cmdset`.

This arose during Ainneve development and is a one-line fix, so I do not believe there is an existing Issue for this.

### DIFF
--- a/evennia/commands/cmdsethandler.py
+++ b/evennia/commands/cmdsethandler.py
@@ -553,7 +553,8 @@ class CmdSetHandler(object):
         if must_be_default:
             return self.cmdset_stack and self.cmdset_stack[0].key == cmdset_key
         else:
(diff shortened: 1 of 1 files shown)


feend78:
Another thought: should `has_cmdset` also accept a cmdset instance, similar to how add() and remove() work?

Griatch:
Merged with a rebase.

PR #1084, opened 2016-09-28
AttributeHandler fix when db_category or db_attrtype saved as blank string from django admin

TehomCD:
When saving Attributes in django admin, it can save the category or attrtype as a blank string rather than a null object. The attributehandler created its cache dict keys with a string conversion of the db_category field, which caused any attribute saved with a db_category value of '' to not show up where those with a db_category of None did. In addition, the queries for getting the objects originally when building the cache also checked for Null, not for empty strings, which prevented them from being found.

Fix to mismatch of db_category and db_attrtype for AttributeHandler when either were saved as blank strings from django admin


Griatch:
Merged with a rebase.

PR #1086, opened 2016-09-29
modified msg_contents to allow objects without get_display_name

feend78:

We recently merged a PR that updated the signature of DefaultObject.msg_contents to accept a mapping of objects for which it would call get_display_name(looker) for each looker in the room's contents. This PR is a minor modification to that which will allow objects without `get_display_name()` (such as plain strings) to be passed in as part of the mapping.

The original modification to `msg_contents` required that only typeclassed objects be passed in as part of the mapping. This change relaxes that requirement, and allows any object.


Griatch:
Merged with a rebase.

PR #1089, opened 2016-10-01
EvMore changes (resolves #1074)

### LINKED ISSUE #1074: Confusing behavior for players for EvMore
Because EvMore covers player commands, this includes the character look command. But what is often happening is if a player doesn't understand the syntax for EvMore, they'll get something like 'Command 'more' is not available' then hit look, and see: 'You are out-of-character (OOC). Use @ic to get back into the game.' because the player OOCLook is still available. So they'll hit @ic, and then see that they're already in-game, and be thoroughly confused, not knowing that they're stuck in the EvMore window.

1. Open an EvMore message that's 2 or more pages
2. enter 'more'
3. 'look'
4. '@ic'

The player will receive the confusing message that they're currently @ooc, while being told they're IC if they try to use @ic

Perhaps EvMore should cover the OOCLook command as well to replay the current window. I'd also suggest 'return' be replaced with 'next', because for non-ansi terminals they can't tell the N is highlighted in return, and 'return' itself is an invalid command.

TehomCD:
Some changes to the display of Evmore to make it more clear that the player is inside its menu, and how they can leave it. Sends an exit message when they leave, and look will redisplay the first page. 

Make EvMore less confusing to players, who would use 'look', receive the OOCLook command, and think they were bugged.

### DIFF
--- a/evennia/utils/evmore.py
+++ b/evennia/utils/evmore.py
@@ -42,7 +42,7 @@ _SCREEN_HEIGHT = settings.CLIENT_DEFAULT_HEIGHT
 
 _DISPLAY = \
 """{text}
-({{wmore{{n [{pageno}/{pagemax}] retur{{wn{{n|{{wb{{nack|{{wt{{nop|{{we{{nnd|{{wa{{nbort)"""
+({{wmore{{n [{pageno}/{pagemax}] {{wn{{next|{{wb{{nack|{{wt{{nop|{{we{{nnd|{{wa{{nbort)"""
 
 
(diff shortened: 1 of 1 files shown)


Griatch:
Merged with a rebase. 

This is a nice idea, you were however not actually adding the CmdEvMore to any cmdset ... ;) 

I fixed that and merged this along with a slew of other updates to EvMore since I found it didn't properly handle very long single lines. It now makes use of the new justify function by default (but it can be turned off).

PR #1090, opened 2016-10-02
Removed reference to target after target.at_desc. This allows the tar…

robbintt:

Moves the target.at_desc() call to the bottom of at_look.  If this is the last call to target, the object that is looked at can take self modifying actions such as self.delete().

This provides a clearer code pattern for verbs on consumables.  For example, if a developer wants to implement `at_drink` and `at_drunk` following the pattern of `at_look`, `at_desc`, it will work without any modification.

…get to delete itself in response to being looked at without throwing an error.


Griatch:
Thanks, it's a simple change that makes sense. Merged with a rebase.

PR #1094, opened 2016-10-04
Overload the 'last', 'first', and 'count' queryset methods

TehomCD:
Adding the last(), first(), and count() methods for the TypedObjectManager so that we can do things like Character.objects.count() or Room.objects.last() and so on and get the expected object for that typeclass, rather than the unfiltered ObjectDB object.

Currently, doing MyTypeclass.objects.first(), .last(), or count() will produce unexpected/inaccurate results, because it will return the method for the parent model (ObjectDB, PlayerDB, etc), rather than filtered by the typeclass path as expected. This change causes them to produce the expected result.

#### Other info (issues closed, discussion etc
N/A


Griatch:
Good idea! Merged with a rebase.

PR #1096, opened 2016-10-06
AI system contrib

andrei-pmbcn:

Introduces a behavior tree-based AI system to Evennia. For more information on behavior trees, check out the readme file.

Evennia could use an AI system contrib. I would have made the pull request later, after the action system was incorporated, but @feend78 wants to use the AI system for Ainneve as soon as possible.

Everything was unit-tested and should hopefully work. If any problems arise, I'll get straight to them.


evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
+## Behavior tree-based AI
+The AI this system provides is based on behavior trees, structures that evaluate
+situations and perform actions in a specified order. As of 2016, behavior trees
```
Griatch: Since you are fererring to AIObjects/AIScripts later in the tree overview without defining them, it might be a good idea to briefly explain them here. Overall, pointing out that the behavior tree is assigned to/used by such an agent is a good idea. An introductory list of features and an example of AI behavior possible with this system might also be good to put at the top, to show users why this might benefit them over a more traditional state-based AI approach.

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+child, then returns that child's return status. It stores this message in the
+node's `msg` property and a given node instance's 'msg' entry. If no message is
+provided (this is the default scenario), the EchoDecorator returns FAILURE
+without sending a message or running its child. Note that this node may only be
+used when its agent is an AIObject, not an AIScript (see the section on setting
+up the AI system for more information.)
+
+#### Blackboards
+
+Because it would be inefficient to store an instance of a tree on each agent
```
Griatch: Is "blackboard" really a standard name for this type of agent storage? At any rate it would be good for understanding to get a visual metaphor explaining why it's called "blackboard". 
It might also be useful to learn what type of entity this is - an object stored on the Character, a list, dict?

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+blackboard to be reset in order to incorporate them, thus overriding your
+manually assigned parameters. Instead, the blackboard data should only be
+modified from within the game's code, unless you are debugging that code and
+want to see what happens when the data has certain values. In all other cases, 
+you should modify the properties of the node objects themselves.
+
+Each blackboard contains the following node-related properties (you can consult
+them later and skip straight to the section on setting up the AI system if you
+prefer):
+
```
Griatch: This should probably be a list for readability.

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+class MyScript(AIScript, <other superclasses>)
+    aitree = 'fighter tree'
+```
+
+Also set up the player class of any players that you want to have access to
+the AI system commands by subclassing it from AIPlayer.
+
+```
+from evennia.contrib.aisystem import AIPlayer
```
Griatch: Sounds less likely to put AI on a Player, maybe instead use a fictive "Mob" AIObject as the example of choice here? Also, I suppose you must assign `self.ainodes.update(new_nodes)` also on the other entity types?

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+you will reference in-game when adding new nodes to your trees, and whose items
+are the respective node classes. For instance, if new\_nodes is
+`{'my_node': MyNode}, then you can add a node of class MyNode to the currently
+browsed node by using the "@aiadd my_node 'my node'" command. Node names should
+only contain alphanumeric characters and underscores, never spaces.
+
+Even if they are subclassed from this class, players will only have access to
+the AI system commands if they are wizards, immortals or the superuser.
+
+Finally, you need to load the AI system's command set into your game. One way
```
Griatch: From looking at the commands, it appears the commands are all builder-related commands. It should be pointed out that these are not needed unless for entities that should be able to modify AI states in-game.

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+    [...]
+    def at_cmdset_creation(self):
+        [...]
+        self.add(AICmdSet())
+```
+
+All newly created AIPlayers, AIObjects and AIScripts are automatically set up
+upon creation. Those that already exist in the database when you start using
+the AI system, however, need to be set up manually. To set up everything once
+you have gone through the above steps, simply enter the game and run the
```
Griatch: So `@aisetup` is basically a shortcut for manually looping over all previously existing objects and doing the ainodes.update() call? Like a globally running `@typeclass/force/reset` call?

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+```
+from evennia.contrib.aisystem import FAILURE, SUCCESS, RUNNING, ERROR
+```
+
+### Working with behavior trees in-game
+
+The following commands allow you to navigate, view and modify behavior trees,
+their nodes and these nodes' properties, as well as AI agents and their
+blackboards, from within the game itself.
+
```
Griatch: This could likely be clearer as a list.

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+@airemove - removes a node from its tree
+
+Note that, at present, there is no way to convert AI trees stored in the
+game's database into Python code. I can make that feature available on
+request, however.
+
+### Adding your own node types
+
+If you are not a coder, this section will not be useful to you.
+
```
Griatch: An example of a custom node of some particular type might be useful here.

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+The first three characters of this hash are used by the AI system's various
+commands and can be used to reference the nodes when two or more nodes in the
+tree share the same name.
+
+Each node contains the following methods, all of which may be overridden.
+With the exception of the update() method, all of them should feature a call to
+`super()`. Many of the methods have the `bb` argument, which is the blackboard
+of the agent whose AI is running.
+
+`update(self, bb)` - this method performs the procedure that occurs whenever the
```
Griatch: If this is called upon a tick, is there a reason this is not called `tick`, or inversely that one uses `node.children.update()` instead. It sounds confusing to have both `tick` and `update` used to mean the same thing?

Also, since these are specified as methods, the `self` argument should not be included, those are implicit in methods.

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+    [...]
+    child.tick(bb)
+    [...]
+```
+
+Be sure to have the update method return a status, either SUCCESS, FAILURE,
+RUNNING or ERROR. Exceptions are automatically handled by the ai system (see
+below), but you can override the system's functionality.
+
+The tick method should always have one argument, the blackboard in the update
```
Griatch: It is still unclear to me as a reader just what kind of an object a blackboard is at this point ...

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+
+Each node also has a `name` property that contains the name you assigned it,
+as well as a `hash` property, which contains three alphanumeric characters that
+are unique to the node within a given tree, followed by an underscore character
+and the id number of its tree. For instance: `H5q_1566`, `Zmr_43`, `7Fl_5425163`
+The first three characters of this hash are used by the AI system's various
+commands and can be used to reference the nodes when two or more nodes in the
+tree share the same name.
+
+Each node contains the following methods, all of which may be overridden.
```
Griatch: This could well be a new subsection, for example "Properties of the Node object".  All the methods listing within should probably go in a list for better readability in the formatted version.

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+
+```
+agent.ai.clean()
+```
+
+Finally, you should never need to overload the node's \_\_init\_\_ method.
+
+### Limitations
+
+The theoretical upper limit for the number of nodes in a tree is 238,328.
```
Griatch: I assume this is due to hash clashes, but it is best to explain this explicitly.

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+There is a risk of nodes failing to be generated as you approach that limit. 
+You should probably split your tree up into multiple trees connected by 
+Transition nodes well before this starts to happen, however.
+
+In theory, in extremely rare cases, a node might fail to generate a hash due to
```
Griatch: Is there a reason the hash is so short that this is worth mentioning?

evennia/contrib/aisystem/README.md
```diff
@@ -0,0 +1,661 @@
...
+
+The data properties of any nodes you create should not have a __call__ method.
+If they do, they will not be displayed appropriately by the @ailook command, 
+which attempts to avoid displaying the node's methods by ignoring properties
+with a __call__ method and will also therefore fail to display your own.
+
+If there is a feature you would like added to this system, or if you find a
+bug or typo that needs fixing, don't hesitate to contact me on my github
+account, andrei-pmbcn, or at my email address, andreipmbcn at yahoo dot com.
+I am Andrei Pambuccian, the maintainer of this contrib.
```
Griatch: I think it would be excellent with a full example of a small behavior tree as a final section. 

1. Defining a small tree, only 2-3 nodes small
2. Show how to add to an AIObject in code. 
3. Example of usage in code
4. Example of usage from in-game.

Griatch:
One thing, please add `*.pyc` to your `.gitignore` file - you are committing the `.pyc` equivalents for the files which does not belong in the repo. Please remove them. :)

Griatch:
@andrei-pmbcn: So, @feend78 and I have been digging into the code to see if we could wrap our heads around it as newcomers to it. It's a big piece of work - one can tell you put a lot of effort into it, kudos! I think we have a decent grasp of its workings even though I've not looked in detail at every function yet. 

I don't know if feend agrees, but this is my understanding at this point. I'm listing it here so you can check I'm not misunderstanding anything fundamental: 

1. Behavior Trees are implemented as Script subclasses. Each such Script stores the root node in an Attribute and the nodes of the tree in another Attribute.
2. A Node is in itself a normal Python object. Each Node in a given tree has a unique hash.  The Node instance is pickled directly into the Tree's `nodes` Attribute. 
3. Whereas the Tree contains the structure and code of the Tree and its Nodes, the Object/Script using the AI system instead stores what is jointly known as a  "blackboard". This is actually dictionaries `nodes` and `globals` on the `AIHandler` (`obj.ai`). Under the hood, the AIHandler persists this data as a dict to `obj.db.ai`. 
4. Each blackboard-node/global offers object-specific data and setting to a given tree by relating to its unique hash in a flat 1:1 relation. This way multiple objects could make use of the same Tree structure while retaining different options and store individual results. 
5. The Player is intended to be the builder of AI Trees rather than Characters; The AIPlayer gets the `.aiwizard` handler which is used to retain state when working with Trees. 
6. The nodes available for the AIPlayer to build with are stored in `player.ainodes`. This is required by build commands to know which nodes are possible to add in-game. For a coder to add a new Node to the pool, one must inherit from the existing types and then import and add the new Node to this dictionary on all Players in-game every restart (doing it in `at_player_creation` as suggested in the readme won't work, the `ainodes` dict is not persistent). 
7. The build commands allow for creating new trees as well as add and reorganize nodes on those nodes. 
8. For a given object to use the AISystem, it must inherit from one of the `AI*` typeclasses and add the desired Tree to it with `obj.ai.setup(tree)`. To crank the wheels the system needs to call `obj.ai.tick()` on demand or using some sort of timer. 

Does that sound about right?

Griatch:
Bump?

Griatch:
Since it doesn't look like there's any more activity on this, I'm closing this PR now.

PR #1102, opened 2016-10-18
Request for change to ChannelHandler defaults

TehomCD:
The recent change to cmdsets included some changes to ChannelHandler which prevented some functionality that I was using: it prevents players from executing channel commands if they are no longer subscribed to the channel. You might wonder how that might come up, but there's a surprising amount of use cases for that - players often stop listening to a channel to avoid spam, but still might want to occasionally check channel history or the like, and most importantly, I had '<channelname> on' be an alias for addcom, which allowed people to use defined nicknames for their channels to toggle them back on. With the ChannelHandler checking subscriptions for whether they have the command or not, it would cause the situation of a player doing something like 'chan off', seeing that they had disconnected, then 'chan on', and being told it was an invalid command. I changed the allow_duplicates to False, because it's necessary to avoid redundant channel commands without the subscription check.

To allow players to use channel commands without an active subscription, such as '<channelname> on'.

Slightly related to this, but I found players were confused by the default behavior of delcom and 'allcom off' to remove channel aliases. I just made override commands that removed the subscriptions without wiping aliases, but it might be more intuitive for the vanilla commands to not wipe aliases unless there's a switch to tell it to explicitly do so.

### DIFF
--- a/evennia/commands/default/comms.py
+++ b/evennia/commands/default/comms.py
@@ -112,7 +112,10 @@ class CmdAddCom(COMMAND_DEFAULT_CLASS):
             else:
                 string += "You now listen to the channel %s. " % channel.key
         else:
-            string += "You are already connected to channel %s." % channel.key
+            if channel.unmute(player):
+                string += "You unmute channel %s." % channel.key
+            else:
+                string += "You are already connected to channel %s." % channel.key
 
         if alias:
             # create a nick and add it to the caller.
@@ -130,10 +133,12 @@ class CmdDelCom(COMMAND_DEFAULT_CLASS):
 
     Usage:
        delcom <alias or channel>
+       delcom/all <channel>
 
     If the full channel name is given, unsubscribe from the
     channel. If an alias is given, remove the alias but don't
-    unsubscribe.
+    unsubscribe. If the 'all' switch is used, remove all aliases
+    for that channel.
     """
 
     key = "delcom"
@@ -162,13 +167,16 @@ class CmdDelCom(COMMAND_DEFAULT_CLASS):
                 self.msg("You are not listening to that channel.")
                 return
             chkey = channel.key.lower()
+            delnicks = "all" in self.switches
             # find all nicks linked to this channel and delete them
-            for nick in [nick for nick in make_iter(caller.nicks.get(category="channel", return_obj=True))
-                         if nick and nick.pk and nick.value[3].lower() == chkey]:
-                nick.delete()
+            if delnicks:
+                for nick in [nick for nick in make_iter(caller.nicks.get(category="channel", return_obj=True))
+                             if nick and nick.pk and nick.value[3].lower() == chkey]:
+                    nick.delete()
             disconnect = channel.disconnect(player)
             if disconnect:
-                self.msg("You stop listening to channel '%s'. Eventual aliases were removed." % channel.key)
+                wipednicks = " Eventual aliases were removed." if delnicks else ""
+                self.msg("You stop listening to channel '%s'.%s" % channel.key)
             return
         else:
             # we are removing a channel nick
--- a/evennia/commands/default/comms.py
+++ b/evennia/commands/default/comms.py
@@ -176,7 +176,7 @@ class CmdDelCom(COMMAND_DEFAULT_CLASS):
             disconnect = channel.disconnect(player)
             if disconnect:
                 wipednicks = " Eventual aliases were removed." if delnicks else ""
-                self.msg("You stop listening to channel '%s'.%s" % channel.key)
+                self.msg("You stop listening to channel '%s'.%s" % (channel.key, wipednicks))
             return
         else:
             # we are removing a channel nick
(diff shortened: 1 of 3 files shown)


Griatch:
I see what you mean by the use case, but it really makes no sense that you should have anything to do with a channel you no longer subscribe to. If people want to "mute" channels while still being subscribed to them, they should *not* be doing so by *unsubscribing* from the channel! The most logical way to mute a channel would be to make sure a "listen" lock is checked before sending to the channel and not send to players who fails this check. 

I agree the addcom/delcom behavior is confusing however. It would be clearer to add aliases and then be able to turn on/off/mute channels in parallel rather than having to remove all aliases before you can unsubscribe. This is MUX legacy syntax we inherited. I'd be fine with some other functionality that is more intuitive. 

So I won't accept this PR as it is, but I'd be open to changes along the lines outlined above! I'm keeping this open for eventual further discussion.

TehomCD:
Hmm, I see what you mean. How does adding 'mute' and 'unmute' methods and a mute_list property to ChannelDB, which distribute_message would check against before sending a message to a subscriber, and then having the base channel command have 'on' and 'off' arguments for calling those? 

I'm unsure if you want me to mess with delcom or not, since I don't really know what a good solution is there: while I think most people would want to keep aliases, I dunno if having an '/all' switch for delcom would be very intuitive for removing aliases.

Griatch:
Hm, I was first considering if it could be implemented with a lock somehow, but maybe having something on the channel is the more flexible way. I'll look at your suggested changes soon as I can. 

As a side note: As someone not originally familiar with MUX, I never found addcom/delcom/allcom to be very intuitive at all. I would personally be more comfortable with having a single command, like `channel public on|off|mute` for manipulating things; and then use the normal `nick` (also known as `alias`) command: `alias/channel public = pub`  to add a nick to it (which is what a channel alias already is under the hood). 

Not saying this is something necessary to cover in this PR, but it's something to discuss I think.

TehomCD:
Locks might well be better, I just had trouble thinking of what the appropriate lock would be, so a list stored in an attribute seemed more natural to me. A somewhat related problem with locks on channels is that if the lockstring is large, it causes some readability issues in the @channels display when you have a whole bunch of channels with long lockstrings. like 'send: perm(builders) or attr(someattrname, some attrvalue) or attr(otherattrname, otherattrvalue)' and so on. Maybe a niche issue for when you're doing stuff with attr checks that should really be custom lockfuncs anyway, or maybe locks shouldn't display by default? Not sure.

I like your idea about a single 'channel <dostuff>' command, since that sounds more intuitive to me, and moving more functionality to the alias command. It reminds me that at some point, @set should have the ability to change attributes on channels, and @tag should be able to tag the other types of typeclasses that can accept it. I usually just change that stuff with @py calls or whatever, but it seems like there should be @tag/<typeclass> and so on.

Griatch:
I think an eventual modification of the addcom/delcom/allcom system might be a bigger thing worthy of going into a future dev branch; I have some ideas of other changes for the channel system too that I'd like the test. Time will tell, for now though I'll likely merge your changes. 

The ability of `@set` to affect all typeclasses is a good idea. Please make a separate Feature request for it!

evennia/commands/default/comms.py
```diff
@@ -162,13 +167,16 @@ def func(self):
...
-                nick.delete()
+            if delnicks:
+                for nick in [nick for nick in make_iter(caller.nicks.get(category="channel", return_obj=True))
+                             if nick and nick.pk and nick.value[3].lower() == chkey]:
+                    nick.delete()
             disconnect = channel.disconnect(player)
             if disconnect:
-                self.msg("You stop listening to channel '%s'. Eventual aliases were removed." % channel.key)
+                wipednicks = " Eventual aliases were removed." if delnicks else ""
+                self.msg("You stop listening to channel '%s'.%s" % channel.key)
```
Griatch: This will lead to a traceback since not all `%s` formatting statements are matched.

evennia/comms/channelhandler.py
```diff
@@ -43,6 +43,8 @@ class ChannelCommand(command.Command):
     Usage:
        {lower_channelkey}  <message>
        {lower_channelkey}/history [start]
+       {lower_channelkey} off - mutes the channel
```
Griatch: I think this should really somehow be shown in the return of the `channels` command, maybe as channelname (mute) or some such.

Griatch: Or maybe in the sub collumn, replace "Yes" with "Yes (Muted)".

TehomCD:
I thought that it was a good time for refactoring '@cwho' and 'allcom who' to call a shared method in the channel typeclass that returns the who list, so I went ahead and did that also. A very frequent request of my players was that channel who only show people who are connected, but I thought a more generic/universally used version would be just highlighting subscribers who are connected and don't have the channel muted, so I went with that.

Griatch:
This looks good, I like the cwho changes too. Merged with a rebase, thanks!

PR #1103, opened 2016-10-19
get tag from list returned by get_tag

TehomCD:
get_tag returns a queryset, but the create_tag check didn't get the object from the queryset if it returned it, so generated an AttributeError when trying to call save() on a queryset

small bug fix

N/A


Griatch:
Thanks! Merged with a rebase.

PR #1106, opened 2016-10-20
Add 'justify' kwarg to EvTable

TehomCD:
A player noticed that a table that was formatted by EvTable looked odd, because the align method does not strip out whitespace when wrapping lines, so it was common to have a line start with a whitespace. Now if a 'justify=True' kwarg is passed, the align kwarg will have the text strip out the whitespace on either the left or right border, depending on alignment.

Text formatting for evtable more in line with player expectations.

N/A

### DIFF
--- a/evennia/utils/evtable.py
+++ b/evennia/utils/evtable.py
@@ -427,7 +427,6 @@ class EvCell(object):
         # alignments
         self.align = kwargs.get("align", "l")
         self.valign = kwargs.get("valign", "c")
-        self.justify = kwargs.get("justify", False)
 
         #self.data = self._split_lines(unicode(data))
         self.data = self._split_lines(_to_ansi(data))
@@ -561,7 +560,9 @@ class EvCell(object):
 
     def _align(self, data):
         """
-        Align list of rows of cell.
+        Align list of rows of cell. Whitespace characters will be stripped
+        if there is only one whitespace character - otherwise, it's assumed
+        the caller may be trying some manual formatting in the text.
 
         Args:
             data (str): Text to align.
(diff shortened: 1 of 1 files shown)


Griatch:
While the functionality is good, I'm not sure a special kwarg (your second commit) is really needed; as you say; when would you NOT want to align text evenly like this? Also 'justify' is a special type of aligning that is not what's being done here.

TehomCD:
The one situation I can think of is people wanting to handle manual indentation themselves, such as adding a bunch of tabs of whitespace between paragraphs, or something of that nature, while still being left-aligned when wrapping text: such as having an initial 'headline' being centered via manual indentation, while the rest of the body is then aligned nicely, or several such centered/indented statements throughout the text. It would be admittedly a pretty obscure situation, but I have a habit of having people ask for things I'd never personally see a case for. Maybe a kwarg named something like 'no_flush' or 'no_strip' that defaults to False that would allow someone to left/right align without stripping the whitespace?

Griatch:
The thing with people doing manual alligning like this is that it won't work as the table is resized - that indent might just as well end up partly on the line above. 

But fair enough, who knows what people want to use. The problem here is then the use of the all encompassing `strip/lstrip`, which will delete *1 or more* beginning spaces from the left. I still feel it's too detailed to offer a kwarg for this. Would it not make sense to apply the lstrip always if there is a *single* space starting the line, but not if there are 2 or more (which we would then assume to be the start of some intentional spacing/indent)?

Griatch:
By the way, unittests can be modified if the intended default changes.

TehomCD:
Checking for a single whitespace might be the best solution - offhand, I can't think of any situation in which someone would want a single whitespace that appeared from a wrap to not go away, while in theory someone might want a bunch of whitespaces for some reason. Let me try that, then.

evennia/utils/evtable.py
```diff
@@ -574,9 +575,9 @@ def _align(self, data):
         hfill_char = self.hfill_char
         width = self.width
         if align == "l":
-            return [(line.lstrip() if self.justify else line) + hfill_char * (width - m_len(line)) for line in data]
+            return [(line.lstrip() if line.startswith(" ") and not line.startswith("  ") else line) + hfill_char * (width - m_len(line)) for line in data]
```
Griatch: ... but on the other hand reverting the order of the if like this: 

```python
line if line.startswith("  ") else line.lstrip()
```
appears about twice faster than both.

Griatch: Just for fun and completeness:

```python
>>> timeit.timeit('s.lstrip() if s.startswith(" ") and not s.startswith("  ") else s', "s=' text'")
<<< 0.7753889560699463
>>> timeit.timeit('s if s.startswith("  ") else s.lstrip()', "s=' text'")
<<< 0.5325000286102295
```
So more around 30% faster.

PR #1107, opened 2016-10-22
Make portal log less noisy

TehomCD:
My portal logs have been rotating several times a day, and they're mostly filled with lines of amp factory starting and stopping, which happens every time someone connects/disconnects. I found that you can suppress these messages with the 'noisy=False' argument for factories, and I can't really think of a reason why someone would want to see them, since they make the portal logs pretty hard to read.

Make portal logs less spammy and fill up much slower.

N/A


Griatch:
Good idea! I went ahead and added the flag to the factories too; should  cut down spam a bit.

Issue #1002, opened 2016-06-29
MXP not escaping single quote character when generating link.


BlauFeuer:
#### Brief summary of issue / Description of requested feature:

The single quote can't be escaped, and isn't automatically escaped when included as part of a MXP response. Trying to substitute `&#39;` yields `&amp;#39;` as it escapes the `&` character instead.
#### Steps to reproduce the issue / Reasons for adding feature:
1. `say |lc' Aye!|lt|gAll in favor?|n|le`
2. `say |lcget lady's coin purse|lt|gpick up the item that the merchant dropped|n|le`
3.  Notice no effect when clicking on links with MXP-enabled client or webclient
#### Error output / Expected result of feature

`You say, "<SEND HREF='' Aye!'>All in favor?</SEND>"`
`You say, "<SEND HREF='get lady's coin purse'>pick up the item that the merchant dropped</SEND>"`

Expected escape of single quote as an HTML entity when within single quotes or use of double quotes when no double quote exists in the HREF:

`You say, "<SEND HREF='&#39; Aye!'>All in favor?</SEND>"`
`You say, "<SEND HREF='get lady&#39;s coin purse'>pick up the item that the merchant dropped</SEND>"`

or

`You say, "<SEND HREF="' Aye!">All in favor?</SEND>"`
`You say, "<SEND HREF="get lady's coin purse">pick up the item that the merchant dropped</SEND>"`
#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:

current

The raw links are visible by forcing `@option mxp=on` in a non-MXP client.

BlauFeuer:
I tested a possible solution in evennia/server/portal/mxp.py:

`MXP_SEND = MXP_TEMPSECURE + \`
`           "<SEND HREF=\"\\1\">" + \`
`           "\\2" + \`
`           MXP_TEMPSECURE + \`
`           "</SEND>"`

which corrects the link to:
`<SEND HREF="get lady's coin purse">pick up the item that the merchant dropped</SEND>`

and also to the parsing:

`text = text.replace("&", "&amp;") \`
`           .replace("'", "&#39;") \`
`           .replace("<", "&lt;") \`
`           .replace(">", "&gt;")`

which corrects the link to:

`<SEND HREF='get lady&#39;s coin purse'>pick up the item that the merchant dropped</SEND>`

[...]

PR #1108, opened 2016-10-22
MXP tags to use double quote instead of single

BlauFeuer:

Partially addressing issue #1002 - MXP commands and text can now contain a (simple) single quote/apostrophe without quoting issues. 

This works around the problem of having to escape single quotes (apostrophes) in MXP commands or texts, but then requires (simple) double quotes to be escaped.

Caveat - webclient does not expect simple double quotes in MXP commands or text, so this is not a full solution when trying to include a double quote in MXP commands or texts.   Escaping them for webclient may need to be a separate issue or more needs to be done in this PR to address that.

### DIFF
--- a/evennia/server/portal/mxp.py
+++ b/evennia/server/portal/mxp.py
@@ -21,7 +21,7 @@ LINKS_SUB = re.compile(r'\|lc(.*?)\|lt(.*?)\|le', re.DOTALL)
 MXP = "\x5B"
 MXP_TEMPSECURE = "\x1B[4z"
 MXP_SEND = MXP_TEMPSECURE + \
-           "<SEND HREF='\\1'>" + \
+           "<SEND HREF=\"\\1\">" + \
            "\\2" + \
            MXP_TEMPSECURE + \
            "</SEND>"
--- a/evennia/server/portal/mxp.py
+++ b/evennia/server/portal/mxp.py
@@ -38,6 +38,7 @@ def mxp_parse(text):
 
     """
     text = text.replace("&", "&amp;") \
+               .replace('"', "&quot;") \
                .replace("<", "&lt;") \
                .replace(">", "&gt;")
 
--- a/evennia/utils/text2html.py
+++ b/evennia/utils/text2html.py
@@ -234,7 +234,7 @@ class TextToHTMLparser(object):
 
     def convert_links(self, text):
         """
-        Replaces links with HTML code.
+        Replaces MXP links with HTML code.
 
         Args:
             text (str): Text to process.
--- a/evennia/utils/text2html.py
+++ b/evennia/utils/text2html.py
@@ -243,7 +243,7 @@ class TextToHTMLparser(object):
             text (str): Processed text.
 
         """
-        return self.re_mxplink.sub(r"""<a id='mxplink' href='#' onclick='Evennia.msg("text",["\1"],{}); return false;'>\2</a>""", text)
+        return self.re_mxplink.sub(r'''<a id="mxplink" href="#" onclick="Evennia.msg(&quot;text&quot;,[&quot;\1&quot;],{}); return false;">\2</a>''', text)
 
     def do_sub(self, match):
         """


Griatch:
Thanks for the work on this. It sounded in the chat like you were still looking into this, so will hold off with this one for a bit.

BlauFeuer:
This solves most of the issue but I think there's still an edge case in webclient where the MXP link becomes malformed when the command text contains a simple single quote:

`say |lc' Aye!|lt|gAll in favor?|n|le`
becomes (in HTML):
`<a id="mxplink" href="#" onclick="Evennia.msg(&quot;text&quot;,[&quot;" aye!"],{});="" return="" false;'=""></a>`

It's difficult to test fully without actually using an MXP client.   It looks like it's clickable, but doesn't yet perform the desired results.  Still working on it.

BlauFeuer:
Alright, the only remaining issue is with webclient. The mu* client now correctly is receiving `<SEND HREF="' Aye!">All in favor?</SEND>` which allows clicking to send an Aye! response.

Will try to solve the webclient issue next. Note to self: **The MXP HTML generation is in evennia/utils/text2html.py**

BlauFeuer:
This solves more of the issue.  The only problem left is to escape double quotes in MXP text and commands.

`say |lc' Aye!|lt|gAll in favor?|n|le`
now properly becomes (in HTML):
`<a id="mxplink" href="#" onclick="Evennia.msg(&quot;text&quot;,[&quot;' Aye!&quot;],{}); return false;"></a>`

Griatch:
Maybe the solution of this is to pre-sub the command/text strings with `&#39;` (') and `&quot;` (") respectively before going into the main html conversion. See the [HTML symbol table](http://www.degraeve.com/reference/specialcharacters.php).

BlauFeuer:
Single quotes don't have to be escaped now - that much is solved.  Use of double quotes inside the MXP text and command is the last remaining edge case.

I'll try per-substituting `"` with `\&quot`, which will turn into `\"` when clicked and sent back to Evennia, and then Python should interpret `\"` inside double quotes properly as a `"` character.

BlauFeuer:
Confirmed: Replacing `"` with `\&quot;` when found in `\1` and `\2` in `re_mxplink` matches solves the last edge case.  Now I have to see how to implement it in Python code or regexp.

Nothing has to be done with `'`, and since escaping a `'` in MXP by using `&#39;` is invalid in the MXP standard (e.g. MXP and HTML entity quoting is incompatible in some cases, `'` being one of them), this solution works around that potential problem, also.

BlauFeuer:
### Proposed solution:
Replacing " with \&quot; when found in \1 and \2 in re_mxplink matches solves the last edge case. I'm looking into how to use Python's `re.sub` to do this just before returning the HTML in `convert_links`.

### Reasoning:
Substituting " with \&quot; will turn into \" when clicked and sent back to Evennia, and then Python should interpret \" inside double quotes properly as a " character.

The docstring says "Replaces links with HTML code.", but it looks like this only replace MXP links with HTML code.  Also, there's a helper function `do_sub` below it that seems to process replacements by dictionary lookup.  Can we make a new case and cdict entry for the `"` to `\&quot;` case and use this method to do the substitution in `re_mxplink` in `\1` and `\2`? 

Nothing has to be done with ', and since escaping a ' in MXP by using &#39; is invalid in the MXP standard (e.g. MXP and HTML entity symbol table is incompatible in some cases, ' being one of them), this solution works around that potential problem, also.

Griatch:
@BlauFeuer Sounds like a working way to do it, sure.

Griatch:
@BlauFeuer Re-reading the question, you might not be clear just where cdict comes from. The cdict dictionary is generated from the regex itself. If you look at `re_string`:

```
re_string = re.compile(r'(?P<htmlchars>[<&>])|(?P<space> [ \t]+)|(?P<lineend>\r\n|\r|\n)', re.S|re.M|re.I)
```
Here you soo that the groups are given names `htmlchars`, `space` and `lineend`. These group matches are used in `do_sub` and extracted as `cdict` (for example `cdict["htmlchars"]`). I can't comment on if this particular issue will be resolved by including the replacement of `"`  to `\&quot;`, but you could append `|(?P(<quote>\")` to the regex string - then the key "quote" will be available for use inside `do_sub`.

Griatch:
I take it this is not quite ready for review and merge?

BlauFeuer:
You can certainly review it as-is; I've been running these changes for awhile without issue, understanding that there's an edge case for double quotes in webclient.

I'm fairly certain my solution for it works, but don't know anyone who can review it or how to implement it myself.

I still don't understand Python enough to grab the `/1` and `/2` match from `re_mxplink` and perform the last needed substitution of `"` to `\&quot;` only within the `/1` and `/2` - the issue in webclient might have to remain unresolved until someone better at Python can handle it.

BlauFeuer:
In summary, MXP links in general would be garbled when they included a single quote (even in Telnet). Webclient MXP links would be garbled when the MXP link included a single or double quote.  This PR as-is fixes all issues, _except_  double quotes still need to be sanitized in MXP link content in `text2html.py` to properly generate the MXP link in webclient.

Hopefully, I'm using the correct terminology and calling the one remaining issue an "edge case".  Maybe it's a "corner case"!  I have it cornered.

Griatch:
I have substituted the double quote to `&quot;` in text2html as suggested; Double quote in the text part works, but in the command part this creates an invalid JS call (a " too many in the JS clickable link). Will try to see if one can escape that quote one step further...

BlauFeuer:
Maybe my Python-Foo will be better next time. This issue gets complicated inside of HTML and then through JavaScript and bringing it back to Python in Evennia to parse properly is quite tricky!  Thanks for moving this PR further along.

Griatch:
@BlauFeuer I believe your Python-foo is just fine. This is not a trivial issue and I've not been able to resolve the full case either. I made it so mxp links handle quotes in the text-part of the link. But using the quote (") in the command-part will lead to a malformed JS onClick definition. The result is that such a link looks fine but nothing happens when clicking on it (it never even reaches Evennia).

The problem is that the `&quote;` gets translated into the " too soon and it closes the JS call we are trying to make. I did some tests with alternative quote-escaping specific for JS but that instead gives other errors. 

I merged your work so far, with the minor addition I did to it. I'll keep this open in case inspiration for that last edge case strikes, but we might have to chalk it up to a quirk for now.

BlauFeuer:
Now it's a very specific case: the command section of the MXP link, when presented in webclient, does not function if it contains _a double or single quote._

But the double quote can be escaped, and the command functions if prior to processing, the double quotes in the command section are escaped with `\`   i.e.  all `"` inside `|lc(...)|le` are converted to `\"`

Fixing that just leaves the single quote being unable to be used in an MXP command in the web client, which excludes me from using it in my original use case:  `say |lcTake lady's coin purse|ltSteal the item?|le`

BlauFeuer:
Here's the test case for double quotes that I am using:

`say |lcpose says, "We want it!"|lt All in favor?|le`

only works if you pre-sanitize the double quotes:

`say |lcpose says, \"We want it!\"|lt All in favor?|le`

Griatch:
The `take lady's coin purse` example works just fine with the latest code. The double quoting in the command part does not though, for reasons already known.

BlauFeuer:
Griatch, you're correct.

Alright, my test case with trying to escape double quotes doesn't work after the update, but single quotes do.  I must still have a stale .pyc file.

That last case is just going to be a quirk for now, or be a separate issue.

Issue #1114, opened 2016-10-29
Bug: TagHandler fails for Msg and HelpEntry objects


TehomCD:
#### Brief summary of issue / Description of requested feature:

Msg objects have tags, but their manager, MsgManager, does not inherit from TypedObjectManager, so it doesn't have the create_tag method, get_tag, etc. Attempting to create a new tag with msg.tags.add() will cause an attribute error. Same thing happens for HelpEntry objects, for the same reason.
#### Steps to reproduce the issue / Reasons for adding feature:
1. Create a new Msg object, then attempt to add a tag that does not currently exist to it
2. create_tag will be called, and raise an Attribute Error
3. same thing will happen with a HelpEntry object
#### Error output / Expected result of feature

File "evennia\evennia\typeclasses\tags.py", line 252, in add
    tagobj = self.obj.**class**.objects.create_tag(key=tagstr, category=category, data=data,
AttributeError: 'MsgManager' object has no attribute 'create_tag'
#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:

I'm not sure the best way to resolve this, since Msg and HelpEntry aren't TypedObjects, so I wasn't sure if having their managers just inherit from a TypedManager would cause any issues. I guess that maybe the Tag methods could be refactored out and made a mixin that go into Typed Manager, and then also into the managers for Msg and HelpEntry, something like that?

TehomCD:
I was glancing at typeclasses/manager.py, and I think I might have seen a possible issue that's not directly related to this.  https://github.com/evennia/evennia/blob/master/evennia/typeclasses/managers.py#L209 is the get_tag method for TypedManager. One of the arguments is obj, which is supposed to allow you to specify an object to search against. But if global_search is set to true, it just searches Tag.objects.filter with the ID of obj, which would match a tag by that ID rather than the object. It looks like it needs the same type of query as used below in line 219.

TehomCD:
Another thing I just noticed is that db_model doesn't appear to be populated currently for tags: adding a random tag to myself has the db_model for the tag return None. I think when I did my migration from the old version I just added the classname (eg: objectdb, playerdb, msg, etc) to all of em, but I'm unsure what the intent behind the field is - whether it's supposed to match the model of where the tag is assigned for possible use in queries, or if it's intended to eventually act as a constraint on what the tag is allowed to be attached to. In the latter case, db_model could be a comma seperated charfield in which queries would check whether the object's classname is in it for permission, for example.

[...]

Issue #1009, opened 2016-07-16
Attribute duplication via admin interface


BlauFeuer:
#### Brief summary of issue / Description of requested feature:

When using localhost:8000/admin/typeclasses/attribute/ to add or rename attributes,
the objectdb attribute is created with an empty Strvalue (instead of NULL), and editing
an attribute in-game via `@attribute` causes a duplicate attribute instead of updating
the existing one.
#### Steps to reproduce the issue / Reasons for adding feature:
1. Rename an existing attribute via admin interface.
2. `@set` the attribute in-game.
3. `examine` and observe duplicate.
#### Error output / Expected result of feature

`@set #4/desc_brief='A person on the move.'`

```
Created attribute Atrial/desc_brief = 'A person on the move.'
------------------------------------------------------------------------------
Name/key: Atrial (#4)
session(s): 29
Player: Wizard
Player Perms: <Superuser>
Typeclass: Character (typeclasses.characters.Character)
Location: Limbo (#2)
Permissions: players [Superuser]
Locks:call:false(); control:perm(Immortals); delete:perm(Wizards);
      edit:perm(Wizards); examine:perm(Builders); get:false(); hug:all();
      puppet:id(4) or pid(1) or perm(Immortals) or pperm(Immortals);
      tell:perm(Wizards); view:all()
Stored Cmdset(s):
commands.default_cmdsets.CharacterCmdSet [DefaultCharacter] (Union, prio 0)
Merged Cmdset(s):
evennia.commands.cmdset.CmdSet [_channelset] (Union, prio 120)
evennia.commands.cmdset.CmdSet [_exitset] (Union, prio 101)
evennia.commands.cmdset.CmdSet [_danceset] (Union, prio 0)
commands.default_cmdsets.PlayerCmdSet [DefaultPlayer] (Union, prio -10)
commands.default_cmdsets.SessionCmdSet [DefaultSession] (Union, prio -20)
Commands available to Atrial (result of Merged CmdSets):
@alias, @ban, @batchcode, @batchcommands, @boot, @charcreate, [...]
Persistent attributes:
desc = The invisible cloak of covering up.
prelogout_location = Limbo
combat_parry_mode = False
puzzle_clue = 0
coins = 12
desc_brief = 20000
traits = {'health': {'base': 20, 'name': 'Health', 'min': 0, 'max': 20, 'extra[...]
messages = {}
desc_brief = A person on the move.
Non-Persistent attributes:
last_cmd = None
last_location = Limbo
[...]

PR #1115, opened 2016-10-30
TagAdmin tweaks

TehomCD:

When I tried to mess around with Tags in django admin, I saw that trying to add one wouldn't work - there wasn't a search function, so I'd need to manually look for one out of 17000+ tags from aliases. I put in search fields and a list filter to allow separating aliases, permissions, and normal tags, then set the inline to mirror the appearance of AttributeInline.

Allow for search of tags to be viable, have adding tags look roughly the same as adding attributes.

Adding a tag via django admin will not presently update the object's taghandler. My guess is that I'd need to add a custom form to the Tag and Attribute admins to reload the caches of taghandler and attributehandlers of objects with the tags/attributes.

### DIFF
--- a/evennia/typeclasses/admin.py
+++ b/evennia/typeclasses/admin.py
@@ -1,14 +1,14 @@
 from django.contrib import admin
 from django.contrib.admin import ModelAdmin
-from django.core.urlresolvers import reverse
 from evennia.typeclasses.models import Attribute, Tag
 from django import forms
+from evennia.utils.picklefield import PickledFormField
+import traceback
 
 
 class TagAdmin(admin.ModelAdmin):
     """
     A django Admin wrapper for Tags.
-
     """
     search_fields = ('db_key', 'db_category', 'db_tagtype')
     list_display = ('db_key', 'db_category', 'db_tagtype', 'db_data')
@@ -17,22 +17,54 @@ class TagAdmin(admin.ModelAdmin):
 
 
 class TagForm(forms.ModelForm):
+    """
+    This form overrides the base behavior of the ModelForm that would be used for a Tag-through-model.
+    Since the through-models only have access to the foreignkeys of the Tag and the Object that they're
+    attached to, we need to spoof the behavior of it being a form that would correspond to its tag,
+    or the creation of a tag. Instead of being saved, we'll call to the Object's handler, which will handle
+    the creation, change, or deletion of a tag for us, as well as updating the handler's cache so that all
+    changes are instantly updated in-game.
+    """
     tag_key = forms.CharField(label='Tag Name')
     tag_category = forms.CharField(label="Category", required=False)
     tag_type = forms.CharField(label="Type", required=False)
     tag_data = forms.CharField(label="Data", required=False)
 
     def __init__(self, *args, **kwargs):
+        """
+         If we have a tag, then we'll prepopulate our instance with the fields we'd expect it
+         to have based on the tag. tag_key, tag_category, tag_type, and tag_data all refer to
+         the corresponding tag fields. The initial data of the form fields will similarly be
+         populated.
+        """
         super(TagForm, self).__init__(*args, **kwargs)
+        tagkey = None
+        tagcategory = None
+        tagtype = None
+        tagdata = None
         if hasattr(self.instance, 'tag'):
-            self.fields['tag_key'].initial = self.instance.tag.db_key
-            self.fields['tag_category'].initial = self.instance.tag.db_category
-            self.fields['tag_type'].initial = self.instance.tag.db_tagtype
-            self.fields['tag_data'].initial = self.instance.tag.db_data
+            tagkey = self.instance.tag.db_key
+            tagcategory = self.instance.tag.db_category
+            tagtype = self.instance.tag.db_tagtype
+            tagdata = self.instance.tag.db_data
+            self.fields['tag_key'].initial = tagkey
+            self.fields['tag_category'].initial = tagcategory
+            self.fields['tag_type'].initial = tagtype
+            self.fields['tag_data'].initial = tagdata
+        self.instance.tag_key = tagkey
+        self.instance.tag_category = tagcategory
+        self.instance.tag_type = tagtype
+        self.instance.tag_data = tagdata
 
     def save(self, commit=True):
+        """
+        One thing we want to do here is the or None checks, because forms are saved with an empty
+        string rather than null from forms, usually, and the Handlers may handle empty strings
+        differently than None objects. So for consistency with how things are handled in game,
+        we'll try to make sure that empty form fields will be None, rather than ''.
+        """
         # we are spoofing a tag for the Handler that will be called
-        #instance = super(TagForm, self).save(commit=False)
+        # instance = super(TagForm, self).save(commit=False)
         instance = self.instance
         instance.tag_key = self.cleaned_data['tag_key']
         instance.tag_category = self.cleaned_data['tag_category'] or None
@@ -42,8 +74,16 @@ class TagForm(forms.ModelForm):
 
 
 class TagFormSet(forms.BaseInlineFormSet):
+    """
+    The Formset handles all the inline forms that are grouped together on the change page of the
+    corresponding object. All the tags will appear here, and we'll save them by overriding the
+    formset's save method. The forms will similarly spoof their save methods to return an instance
+    which hasn't been saved to the database, but have the relevant fields filled out based on the
+    contents of the cleaned form. We'll then use that to call to the handler of the corresponding
+    Object, where the handler is an AliasHandler, PermissionsHandler, or TagHandler, based on the
+    type of tag.
+    """
     def save(self, commit=True):
-        print "inside TagFormSet"
         def get_handler(finished_object):
             related = getattr(finished_object, self.related_field)
             try:
@@ -58,32 +98,30 @@ class TagFormSet(forms.BaseInlineFormSet):
                 handler_name = "tags"
             return getattr(related, handler_name)
         instances = super(TagFormSet, self).save(commit=False)
+        # self.deleted_objects is a list created when super of save is called, we'll remove those
         for obj in self.deleted_objects:
             handler = get_handler(obj)
-            try:
-                tagkey = obj.tag_key
-            except AttributeError:
-                tagkey = obj.tag.db_key
-            handler.remove(tagkey)
+            handler.remove(obj.tag_key, category=obj.tag_category)
         for instance in instances:
             handler = get_handler(instance)
-            handler.add(instance.tag_key)
+            handler.add(instance.tag_key, category=instance.tag_category, data=instance.tag_data)
 
 
 class TagInline(admin.TabularInline):
     """
-    A handler for inline Tags.
-
+    A handler for inline Tags. This class should be subclassed in the admin of your models,
+    and the 'model' and 'related_field' class attributes must be set. model should be the
+    through model (ObjectDB_db_tag', for example), while related field should be the name
+    of the field on that through model which points to the model being used: 'objectdb',
+    'msg', 'playerdb', etc.
     """
     # Set this to the through model of your desired M2M when subclassing.
     model = None
     form = TagForm
     formset = TagFormSet
-    related_field = None
-    #fields = ('tag', 'key', 'category', 'data', 'tagtype')
+    related_field = None  # Must be 'objectdb', 'playerdb', 'msg', etc. Set when subclassing
     raw_id_fields = ('tag',)
     readonly_fields = ('tag',)
-    #readonly_fields = ('key', 'category', 'data', 'tagtype')
     extra = 0
 
     def get_formset(self, request, obj=None, **kwargs):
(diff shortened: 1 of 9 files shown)


Griatch:
Yes, the taghandler cache must be flushed before this will work seamlessly. There are hooks for the admin that should be able to do this on save. I'll hold on merging this until that has been implemented since it would otherwise create really weird behavior.

TehomCD:
I was thinking a bit about this. To me, the most intuitive way of doing this is overriding Tag and Attribute's save methods where they would try to update associated Tag and Attribute handlers if some conditions are met. The conditions for them doing so are actually fairly specific for when there would be issues:

- If any of the fields are changing that are used to index them in their handlers
- If there any current instances of TagHandlers/AttributeHandlers that use them

If it's a Tag that's being modified from django admin and none of the objects that are tagged have been instantiated, then there's no TagHandlers that could get out of sync, for example, so it's safe to change. If someone does a @py call to change an Attribute's name and the object that attribute is attached to isn't currently in memory, that's also safe. So what I'd propose is that each Tag and Attribute has a private attribute that is a list of the Handlers that are currently in game and using them. Then their save() method checks if any of the fields relevant to indexing them have been changed, and if so, recache each of those handlers. The advantage of this approach is that it's more general than something specific to a modeladmin class - it would fix breaking handlers with @py calls as well. Downsides - possibly memory usage? I'm in the middle of something else, but if this seems sound to you, I'd like to try it out when I have time.

Griatch:
A list of handlers on Tags/Attributes sounds like way overkill (and overhead) for the problem. The concept of overriding the Tag/Attribute save() method might have some merit though. The save() is called no matter from where the Tag is set (admin, `@py` or wherever) - maybe just have the save mechanism itself update the TagHandler cache when it fires? It would just need to have a check to avoid infinite loops (when save is called from the TagHandler itself). 

*(Btw, in github it's best to avoid writing ampersand-beginning commands like `@py` without putting them in back-ticks. The reason is that if there happened to be a user named "py" anywhere, they would get notified when you do that. ;) I've been contacted by a user named "spawn" in the past for this reason ... )*

TehomCD:
The reason I suggested registering the handlers is because a Tag could be related to an arbitrarily large number of objects that might not be in memory, and if you save a tag, you wouldn't necessarily know which handlers already exist and would need to be updated - otherwise, you would need to iterate through every single object the Tag is attached to and create their handlers, then reload them again, in order to make sure they have the correct version. While this wouldn't be a big deal if a Tag is in 5 or 6 objects, a Tag could easily be used in every single room, for example, or even every object but a handful if it was used for something like 'non-deleted' or some such. (A bad idea to use it for everything but a few exceptions, but I could easily see that being done, by way of example).  

On the other hand, since the only times handlers would run into trouble is when the fields that are used to index the tag are changed, it could potentially be a rare enough case that it wouldn't be a big deal to handle conditionally that way.

Griatch:
Been pondering this further - and are we not overthinking this? Is there really ever a realistic risk for TagHandlers/AttributeHandlers to go out of sync? Of the below, only points 4 and 5 needs addressing, I figure: 

1. When we create a new Tag in the admin (or in Python via `create_tag()` there is no need for updating any TagHandlers since the Tag is not attached to any objects yet. 
2. When we add a new Tag to an object via `@py` does so via the TagHandler.
3. When we add a new Tag to an object via Admin we know which objects we are adding it to and can reset each particular object's TagHandler directly (by resetting their `_cache` and `._catcache` properties - we can add a method for doing this in a clean way). 
4. The remaining case may be if you edit the Tag's key or category. Since the key + category is what is uniquely identifies a Tag you are actually creating a new Tag and the old one should be deleted rather than be replaced in-place. I just added a check for the `.pk` property in the TagHandler.
5. Attributes work along similar ways. Changing an Attribute's value is not a problem since this is not cached on the AttributeHandler. If you are remotely changing the Attribute's key or category you are actually creating a new Attribute and the old one should be deleted and replaced by a new one. I added a check for the `.pk` value of the AttributeHandler for this use.  

Am I missing some case?

Griatch:
Addendum: Just noticed there is already a `.reset_cache()` method on both TagHandler and AttributeHandler so that bit is already in place, at least.

TehomCD:
I think you're right that we might be overthinking it, since the vast majority of times everything will be called from the TagHandler/AttributeHandlers and work fine. I suppose the one use case I could think of would be adding or removing a Tag or Attribute from the inline manager of ObjectDB, PlayerDB, etc -  clicking to add a new relationship (whether creating a new tag or just associating an existing one) won't update their handler, so the tag or attribute wouldn't be found on them, and a removed Tag/Attribute would still be there until a reload. But as I was testing that, I noticed other issues with trying to add Attributes/Tags too - searching by Attribute throws a server error since icontains isn't permitted for it, and attaching an Attribute that already exists to another object would usually mean that the same attribute is being used in two objects, which probably isn't what someone would want. So while changing most fields of Attributes/Tags would be fine in admin, adding/creating new ones or changing their names either would require a reload to get them to show up in their handlers, and could create subtle and nasty bugs if the same Attribute is used in multiple objects (on the other hand, maybe someone would want to use the same Attribute in multiple objects - like some shared list used by multiple characters, or the like?).

Maybe the best solution might be to replace their inline add form with some custom thing that just calls the appropriate methods of the object's Tag/AttributeHandlers, I think, since that would handle all these edge cases, and then remove the Tag/Attribute categories that show up in django admin since they're not searchable anyway. I could try to figure out how to do that if that sounds like the best approach.

BlauFeuer:
This relates to #1009 ?  I am not sure what activities currently are safe regarding tags and attributes edited in the admin interface, so have to manage them in-game.

TehomCD:
Right, it does, thanks for noting that - changing a name from admin interface will cause issues because the handler won't be updated until a reload. Possibly worse, duplicating a name will cause one of the Attributes to cover up for the other when the handler's cache is rebuilt, with the other Attribute still attached/related to the object but no longer accessible, and it may not be the one someone would want.

Griatch:
Associating a new Tag with an object from the Admin *shouldn't* confuse the handler's cache, since the TagHandler will actually perform a proper query if it can't find the Tag in its cache. It *will* go out of sync if the Tag was removed from it though (and the Tag itself was not deleted).

That said, changing the Admin to directly call the relevent Tag/AttributeHandlers (and remove their current existence in the admin) would probably be the most sane solution. 

(As a side note - I'm currently looking at the Tags since it looks like the use of their `db_model` field has fallen by the wayside and I'm not sure that was a concious decision. :)  As it is now, a Tag can attach to any number of different entities - anything with a TagHandler. Which *sounds* good until you start considering how to effectively query for those objects. The Tag system was designed so that each taggable entity would have its own set of Tags, denoted by the tag's `db_model` field (like "objectdb", "scriptdb", "msg" etc). So I'll look into straightening this out, which will require a migration down the line. It should not affect the changes from this PR though (and if it does, the migration should catch it). )

TehomCD:
Just as a note, when you do that migration with the db_model being used, could you add db_model to the unique_together meta field for Tag so that we could have the same tag name/category pair also be used for different models? Like having an Event tag with the name of the event and category "event" that could be used to tag Msg objects about the event, Players that hosted or participated, etc, rather than needing to change the category to 'event_message', 'event_participant', etc.

TehomCD:
Assuming everything works, the latest change should resolve #1009 from when I was testing it

TehomCD:
The last commit fixed #1114 but I didn't set up the Msg admin object like HelpEntry, since Msg was commented out of comms admin and I wasn't sure if there was a reason for that. I probably should have commented out registering the AttributeAdmin model itself too, or set its db_key and db_category to be readonly fields, since otherwise it still has the #1009 issue.

Griatch:
There appears to be a validation problem with the adding of new Attributes. Here's how to reproduce: 

1. Log out your test user from the client (I tried with the superuser over telnet).
2. Login to the Admin panel and edit the Player's current Character object.
3. Add a new Attribute.
4. Write just a plain text and Save, prompting the "You must enter a valid text string, here's our suggestion..." error. Accept by pressing Save again.
5. Reconnect to the game.
6. The Character is not found. Using `@ic` gives the following error: 

```
You cannot become Griatch: Error ('unicode' object has no attribute 'db_location'): ('__packed_dbobj__', (u'objects', u'objectdb'), '2016:11:14-18:39:34:715315
', 2) is not a valid location.
```

Deleting the Attribute does not fix the situation, It appears the `prelogout_location` Attribute (which contains a pickled  tuple on a special form `('__packed_dbobj__', (u'objects', u'objectdb'), '2016:11:14-18:39:34:715315', 2)` describing uniquely which database object is stored here) has been converted to a string that will be de-pickled to a string rather than the tuple it needs to be. The automatic helper that converts input to strings is likely to blame here. 

(On a separate note, the `evennia.utils.dbserialize` tools could operate on the object representation and maybe display the object properly. Not sure how that would be added or manipulated in the admin though.)

TehomCD:
Thanks for catching that! Since what I was doing was populating the attribute form with the raw (and serialized) pickled values, adding them to the AttributeHandler resulted in it just handling them as the text strings they were rather than proper objects. I just had it call from_pickle on each now on each form to get the appropriate objects before adding them back in. Adding an invalid serialized object, will just wind up adding None in there, and creating that text string from hand is pretty onerous, but people can just copy/paste them from elsewhere if they already know what they're looking for (switching two objects, for example). Trying to come up with some way for people to just look up an object and have it generate the appropriate tuple for the form would be a fantastic feature, but I struggle to think of how that could be implemented - I think that level of django/javascript magic is well beyond my abilities, unfortunately.

evennia/typeclasses/admin.py
```diff
@@ -1,68 +1,325 @@
...
+    attr_value = PickledFormField(label="Value", help_text="Value to pickle/save", required=False)
+    attr_type = forms.CharField(label="Type", help_text="nick for nickname, else leave blank", required=False)
+    attr_strvalue = forms.CharField(label="String Value",
+                                    help_text="Only enter this if value is blank and you want to save as a string",
+                                    required=False)
+    attr_lockstring = forms.CharField(label="Locks", required=False, widget=forms.Textarea)
+
+    def __init__(self, *args, **kwargs):
+        """
+         If we have a tag, then we'll prepopulate our instance with the fields we'd expect it
```
Griatch: Probably copy&pasted from the Tagform; should refer to Attribute properties.

evennia/typeclasses/admin.py
```diff
@@ -1,68 +1,325 @@
...
-            return "Not yet set or saved."
-        return instance.attribute.value
+        ProxyFormset.related_field = self.related_field
+        return ProxyFormset
 
-    def strvalue(self, instance):
-        if not instance.id:
-            return "Not yet set or saved."
-        return instance.attribute.strvalue
+# class AttributeInline(admin.TabularInline):
```
Griatch: Probably ok to remove this commented-out section; Attributes never appear away from an object.

evennia/typeclasses/admin.py
```diff
@@ -1,68 +1,325 @@
...
+                            lockstring=instance.attr_lockstring)
+            except (TypeError, ValueError):
+                # catch errors in nick templates and continue
+                traceback.print_exc()
+                continue
+
 
+class AttributeInline(admin.TabularInline):
+    """
+    A handler for inline Tags. This class should be subclassed in the admin of your models,
```
Griatch: The docstring mentions Tag - should be Attribute.

Griatch:
@TehomCD 

While the converting/deconverting of objects into its representation can be handled, yeah changing the GUI to intuitively handle this doesn't sound very trivial, alas.

Griatch:
Oh, and AttributeAdmins should not be registered I don't think - they only coexist with objects. Not really sure about MsgAdmin honestly; I guess one *could* picture a case when you want to create individual Messages from the admin and use that to "send" to particular entities. So maybe there is a use-case for having that manager available.

TehomCD:
Sounds good! I went ahead and unregistered/removed the non-inline Attribute admin. I'm using the Msg admin for Arx, but that's really a consequence of all the different ways I'm using msg objects in my game. I don't think there's probably a big use-case for looking at them in the existing game, and if someone runs across it, it's easy enough for them to just register their own admin.

Griatch:
@TehomCD Alas, the error with replacing the prelogout_location object with a string as reported in my earlier post remains also in this version for me. When the auto-correct mechanism kicks in (to convert the malformed input to a string), it also converts all other Attributes on that object to strings at the same time.

TehomCD:
That seems like that should only happen if it was a string beforehand: it runs from_pickle on every string from the form before adding them back in to the AttributeHandler, so they should only remain strings if they can't convert into objects. If you convert the string into an object manually (using set griatch/prelogout_location=here, or something like that), then run through the Attribute admin, does it still happen?

Griatch:
@TehomCD 

To test, I did 

```
 @py self.db.blah = here
 @py self.attributes.get("blah")
>>> Limbo
 @py self.attributes.get("blah", return_obj=True)
>>> blah(14)
 @py self.attributes.get("blah", return_obj=True).db_strvalue
>>>  None
```
From the admin, I now create a new Attribute "test" and enter plain `test` without quotes in the value form to get the auto-fix: 

image

Note how these are both strings now. After saving in the admin: 

```
 @py self.attributes.get("blah")
>>> ('__packed_dbobj__', (u'objects', u'objectdb'), '2016:11:14-19:04:14:852988', 2)
 @py self.attributes.get("blah", return_obj=True)
>>> blah(14)
 @py self.attributes.get("blah", return_obj=True).db_strvalue
>>>  None
```

TehomCD:
Okay, the issue is that Picklefield seems to redundantly attempt to convert things into strings if any validation error is raised. To see what I mean, try to add a field with no name, hit save, get the validation error, then continue to try to save it. You'll see the 'u' being added to it each time, and that's happening to every attribute in the formset. So I need to figure out a way around that - this all appears to be happening inside evennia.utils.picklefield, and I'm not sure exactly where. I'll see if I can get a handle on it. I need to be careful, since I think stuff inside pickle could affect Attributes as a whole, not just the admin stuff.

TehomCD:
Alright, another try. Now, picklefield's validation will attempt to convert something into a string of the format "u'OriginalText'" to prevent malformed string errors when people, quite reasonably, would forget to enclose things in quotes for a string they wish to add to an attribute. But the big thing I changed is just to raise an error if there's a validation error rather than proceeding with django's default "there is an error on the page..." stuff, because of the error with how it converts all the initial fields after a validation error is raised. I wish it was a cleaner solution, but it definitely will stop people from accidentally turning all their attributes into strings, which was far worse than not having an error highlighted. If that's not good enough, I'll keep on looking, but I have no idea at all how it's messing with the initial values on the validation errors, so I don't know how long it would take me to figure it out, or how hard it would be to fix. I'd personally prefer to just merge a working but somewhat messy solution (raising a server 500 error) rather than spend god knows how long looking at that, but I might be a bit biased there...

Griatch:
A server 500 error is not acceptable (DEBUG=True is extremely memory-leaky and useless for a production server and only to be used for short times for a devel server). But I don't see it actually triggering such an error with the latest changes; instead it looks like it correctly (if quietly) converts the faulty input to a string. 
Am I missing a server 500 error here in my testing?

TehomCD:
It was only raised on a validation error, which could only happen by omitting necessary fields (like not providing a name of the attribute). Why I went that route is a validation error being raised causes the form to be prepopulated by string conversions of all the db_value fields. This happens with the current AttributeAdmin that I'm replacing as well - if you try to create a new attribute with some object as a value, it'll convert it to a string on a validation error, and keep adding a 'u' before the string each time you fail to validate, once per attempt. To see what I mean, go to the AttributeAdmin and create an attribute with a number value. Then wipe it's name and keep saving it.  For a value like 5, if you tried to save it three times and got a validation error each time, it'll now be something like 'u'u"u\'5\'"'. I'll try to figure it out.

TehomCD:
The problem with the formset values appeared to be a result of PickledObjectField.get_db_prep_value calling force_text on the value, which does cause a string of the form "u'Something'" to become "u'u\'Something\''", and this seemed to happen to every form in the formset whenever a validation error is raised. Since the comments for get_db_prep_value mentioned that it was necessary to avoid database errors in different formats that I can't easily test (avoiding problems in postgres while I use sqlite), I decided the best solution was to avoid validation errors entirely, and use default values when bad values are entered on the form. While it would be more elegant to figure out exactly how to avoid that problem in the PickledObjectField itself, I didn't want to possibly cause a problem where Attributes would stop working for different databases just to solve a relatively trivial form submission error.

So this latest version should prevent validation errors from ever being raised, at least as far as I can tell, and should prevent the problem from ever occurring.

Griatch:
Thanks for your work on this! There is a lingering issue with Tags with a category not being updated in-game but I think this may be an issue with `examine`/`tag`/`.tags.all()` rather than with the Tag form.
Merged with a rebase. Closing.

Issue #1125, opened 2016-11-24
evennia stop + ctrl-C in windows causes traceback


TehomCD:
#### Brief summary of issue / Description of requested feature:
The latest commit causes a traceback to be generated upon using evennia stop and hitting ctrl-C to bypass the repeated 'stop batch job (y/n)' prompt:
` Traceback (most recent call last):
   File "C:\arx\Lib\site-packages\twisted\application\app.py", line 395, in startReactor
     self.config, oldstdout, oldstderr, self.profiler, reactor)
   File "C:\arx\Lib\site-packages\twisted\application\app.py", line 310, in runReactorWithLogging
     reactor.run()
   File "C:\arx\Lib\site-packages\twisted\internet\base.py", line 1195, in run
     self.mainLoop()
   File "C:\arx\Lib\site-packages\twisted\internet\base.py", line 1204, in mainLoop
     self.runUntilCurrent()
 --- <exception caught here> ---
   File "C:\arx\Lib\site-packages\twisted\internet\base.py", line 798, in runUntilCurrent
     f(*a, **kw)
   File "C:\arx\Lib\site-packages\twisted\internet\base.py", line 581, in stop
     "Can't stop reactor that isn't running.")
 twisted.internet.error.ReactorNotRunning: Can't stop reactor that isn't running.` It's essentially just a display error, as far as I can tell, but hitting ctrl-C will continue to generate it each time, which can be confusing.

#### Steps to reproduce the issue / Reasons for adding feature:

1. evennia stop
2. ctrl-c
3. ctrl-c

#### Error output / Expected result of feature
The above traceback, printed each time ctrl-c is entered.

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
I'll take a look and see if I can figure it out tonight, and write a pull request if so. Probably a lot easier for me to do so since I use my windows machine for testing stuff.

TehomCD:
I poked around this for a while, but no matter where I tried to catch ReactorNotRunning in evennia_launcher, it's just consistently being raised when Evennia stops on windows. At least it's just a small display error and doesn't seem to impact anything.

Griatch:
I saw this error during testing for #1100 but not if running on a new Console with a previously fully stopped and started Evennia. I have not seen this Ctrl-C based traceback since. Worth checking.

Windows is *extremely* finicky about sending signals though so in the worst case this is something time-dependent or even Windows dependent (I tested it on Win7). 

[...]

PR #1126, opened 2016-11-24
Fix (maybe) to Windows evennia stop weirdness

TehomCD:
These changes fixed two different tracebacks that kept popping up into evennia on windows for me - when doing 'evennia stop', I would consistently see a traceback from twisted about the reactor already being stopped, which was fixed by catching the exception in evennia_runner. The other was seeing a KeyboardInterrupt exception occasionally (not always) being raised by the message that prints out that the portal/server have been stopped in evennia_launcher's kill() function. It seems completely redundant, but catching KeyboardInterrupt just for that print statement fixed it.

Fix some display error/traceback stuff on Windows. Was it just my install of Windows 7? Not sure!

I have no idea if this affected anyone else on Windows 7, since as you mentioned, it seems very finicky. Hopefully this wouldn't break anything for anyone else, though.

### DIFF
--- a/evennia/server/evennia_launcher.py
+++ b/evennia/server/evennia_launcher.py
@@ -714,12 +714,13 @@ def kill(pidfile, signal=SIG, succmsg="", errmsg="",
                 f.write("shutdown")
         try:
             if os.name == 'nt':
-                from win32api import GenerateConsoleCtrlEvent
+                from win32api import GenerateConsoleCtrlEvent, SetConsoleCtrlHandler
                 try:
                     # Windows can only send a SIGINT-like signal to
                     # *every* process spawned off the same console, so we must
                     # avoid killing ourselves here.
                     GenerateConsoleCtrlEvent(CTRL_C_EVENT, 0)
+                    SetConsoleCtrlHandler(None, True)
                 except KeyboardInterrupt:
                     pass
             else:
@@ -732,12 +733,8 @@ def kill(pidfile, signal=SIG, succmsg="", errmsg="",
                   "The PID file 'server/%(pidfile)s' seems stale. "\
                   "Try removing it." % {'pid': pid, 'pidfile': pidfile})
             return
-        try:
-            print("Evennia:", succmsg)
-            return
-        except KeyboardInterrupt:
-            print("Evennia:", succmsg)
-            return
+        print("Evennia:", succmsg)
+        return
     print("Evennia:", errmsg)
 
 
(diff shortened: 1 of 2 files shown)


evennia/server/evennia_launcher.py
```diff
@@ -732,8 +732,12 @@ def kill(pidfile, signal=SIG, succmsg="", errmsg="",
                   "The PID file 'server/%(pidfile)s' seems stale. "\
                   "Try removing it." % {'pid': pid, 'pidfile': pidfile})
             return
-        print("Evennia:", succmsg)
-        return
+        try:
```
Griatch: I think it is a dangerous assumption to put the try statement here since I suspect it's time dependent from machine to machine. What happens is that the SIGINT(-like) signal being sent will hit the launcher (since Windows can only send SIGINT to *all* of the Console's processes at the same time, yeah - that's a thing). But exactly where in the source the execution happens when the SIGINT hits is time dependent. In your case the computer happens to be at the print statement; this may not be the case for others. So if a try...except is needed in this module I fear it must encompass all of this section. Or the Windows version should, at least.

evennia/server/evennia_runner.py
```diff
@@ -185,26 +185,30 @@ def portal_waiter(queue):
...
+                continue
+
+            # normally the portal is not reloaded since it's run as a daemon.
+            if (message == "portal_stopped" and int(rc) == 0 and
+                  get_restart_mode(PORTAL_RESTART) == "True"):
+                print(PROCESS_RESTART.format(component="Portal"))
+                PORTAL = thread.start_new_thread(portal_waiter, (processes, ))
+                continue
+            break
+        except ReactorNotRunning:
```
Griatch: I would have thought the ReactorNotRunning error was emanating from one or both of the Server/Portal processes, not from the runner. But if this is enough to handle the error - great!

evennia/server/evennia_launcher.py
Griatch: Most clean may be to use the win32api call I outlined in issue #1125 to simply turn of the sensitivity to SIGINT in the launcher.

TehomCD:
I used the API call that you suggested in #1125 and it worked like a charm - not seeing any KeyboardInterrupt exceptions even after I removed the place where I was running into them before. So hopefully this will be a much more general solution that'd work for everyone.

Griatch:
Merged with a rebase. I moved the handler API call up one step to just before the firing of the event, just to be absolutely sure there is no chance the event could be caught by the launcher on a super-fast computer. ;)

PR #1129, opened 2016-11-30
Get JQuery over HTTPS to prevent browsers from blocking it

maxsond:

This does nothing more than get JQuery over HTTPS.

In the current state (getting JQuery over HTTP), the webclient raises a security error when running on a remote server and accessed in a browser (I encountered this when running Evennia on c9.io). The webclient then fails to load, stating that JQuery was unavailable. This is possible for the user to get around by opting to "run unsafe scripts" in their browser, but that's a suboptimal experience. This is a simple fix with no potential fallout which will make the user experience smoother and improve security.

I think this is a no-brainer, but interested if there's anything that needs to be discussed around it. I didn't upgrade the JQuery version while I was in there because that's outside the focus of this fix, which is simply to make the webclient get JQuery with no fussing by the user.

### DIFF
--- a/evennia/web/webclient/templates/webclient/base.html
+++ b/evennia/web/webclient/templates/webclient/base.html
@@ -21,7 +21,7 @@ JQuery available.
 
     <!-- Import JQuery and warn if there is a problem -->
     {% block jquery_import %}
-        <script src="http://code.jquery.com/jquery-2.1.1.min.js" type="text/javascript" charset="utf-8"></script>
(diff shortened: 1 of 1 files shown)


Griatch:
Makes sense; merged - thanks! One hint though, for your own sanity - don't make PRs from your Evennia fork's *master* branch - it is easy for it to go out of sync with upstream master (because it may merge with a rebase if other commits are entered before your PR gets merged). So it's highly recommended to always make your modifications in separate branches and make PRs based on those. See [here](https://github.com/evennia/evennia/wiki/Version%20Control#committing-fixes-to-evennia) for more details.

PR #1132, opened 2016-12-02
Fix to error on django admin's objectdb creation

TehomCD:
For some reason, the save_model method in ObjectDB admin doesn't seem to properly assign the typeclass when an object is being created, which causes server errors in creating the object. Additionally, it prevents the server from being reloaded, as ObjectDB.get_all_cached_instances() will fail with an ObjectDB lacking a typeclass and raising an AttributeError when at_server_reload is called.

Prevent nasty errors on attempting to create an object from django admin.

The real question is why the typeclass isn't assigned normally from `TypedObject.__init__`, and I don't have an answer there. I even tried explicitly importing and calling `TypedObject.__init__(obj)` just to see if it would work, and nothing. Apparently the `__class__` has to be specifically assigned here, and I have no idea why.

### DIFF
--- a/evennia/comms/admin.py
+++ b/evennia/comms/admin.py
@@ -6,6 +6,7 @@ This defines how Comm models are displayed in the web admin interface.
 from django.contrib import admin
 from evennia.comms.models import ChannelDB
 from evennia.typeclasses.admin import AttributeInline, TagInline
+from django.conf import settings
 
 
 class ChannelAttributeInline(AttributeInline):
@@ -69,5 +70,24 @@ class ChannelAdmin(admin.ModelAdmin):
 
         """
         return ", ".join([str(sub) for sub in obj.db_subscriptions.all()])
+
+    def save_model(self, request, obj, form, change):
+        """
+        Model-save hook.
+
+        Args:
+            request (Request): Incoming request.
+            obj (Object): Database object.
+            form (Form): Form instance.
+            change (bool): If this is a change or a new object.
+
+        """
+        obj.save()
+        if not change:
+            # adding a new object
+            # have to call init with typeclass passed to it
+            obj.set_class_from_typeclass(typeclass_path=settings.BASE_CHANNEL_TYPECLASS)
+        obj.at_init()
+
 
 admin.site.register(ChannelDB, ChannelAdmin)
--- a/evennia/objects/admin.py
+++ b/evennia/objects/admin.py
@@ -147,13 +147,8 @@ class ObjectDBAdmin(admin.ModelAdmin):
         obj.save()
         if not change:
             # adding a new object
-            # no idea why this has to be called manually, __init__ does not seem to be called
-            from evennia.utils.utils import class_from_module
-            try:
-                typeclass = class_from_module(obj.db_typeclass_path)
-            except ImportError:
-                typeclass = class_from_module("evennia.objects.objects.DefaultObject")
-            obj.__class__ = typeclass
+            # have to call init with typeclass passed to it
+            obj.__init__(typeclass=obj.db_typeclass_path)
             obj.basetype_setup()
             obj.basetype_posthook_setup()
             obj.at_object_creation()
--- a/evennia/objects/admin.py
+++ b/evennia/objects/admin.py
@@ -103,11 +103,6 @@ class ObjectDBAdmin(admin.ModelAdmin):
                            )}),
         )
 
-    def get_urls(self):
-        urls = super(ObjectDBAdmin, self).get_urls()
-        print urls
-        return urls
-
     def get_fieldsets(self, request, obj=None):
         """
         Return fieldsets.
--- a/evennia/objects/admin.py
+++ b/evennia/objects/admin.py
@@ -103,6 +103,11 @@ class ObjectDBAdmin(admin.ModelAdmin):
                            )}),
         )
 
+    def get_urls(self):
+        urls = super(ObjectDBAdmin, self).get_urls()
+        print urls
+        return urls
+
     def get_fieldsets(self, request, obj=None):
         """
         Return fieldsets.
@@ -153,6 +158,13 @@ class ObjectDBAdmin(admin.ModelAdmin):
             obj.basetype_posthook_setup()
             obj.at_object_creation()
         obj.at_init()
+
+    def response_add(self, request, obj, post_url_continue=None):
+        if '_continue' in request.POST:
+            from django.http import HttpResponseRedirect
+            from django.core.urlresolvers import reverse
+            return HttpResponseRedirect(reverse("admin:objects_objectdb_change", args=[obj.id]))
+        return super(ObjectDBAdmin, self).response_add(request, obj, post_url_continue)
 
 
 admin.site.register(ObjectDB, ObjectDBAdmin)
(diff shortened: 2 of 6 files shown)


Griatch:
Thanks for he PR! That is indeed strange. I wonder if the admin explicitly create using ObjectDB.objects.create() and this bypasses some of Evennia's creation setup somehow. That said, one would think `__init__` gets called in all circumstances. 

I need to look a bit at the fix here, especially since `__dbclass__` is not set in this fix, which could lead to other issues down the line (assuming it's not set automatically at some later stage).

Griatch:
I'm reluctant to merge this fix since this is exactly what `__init__` is supposed to do for us. `__init__` also runs a slew of other tests and sets other properties that need to be added to this PR before merger ... Which is not very efficient code-wise. So poking at this, `typeclass.__init__` *does* fire in the form's save hook. But what that hook does is to super()-call the `SharedMemoryModel.__init__`. And after that super() call,  `db_typeclass_path` is suddenly `None` and not available to process. Since `SharedMemoryModel` does not implement `__init__`, this means something's happening in core Django. Since that super() call works just fine for normal object creation it may be related to the form or the admin. Not sure which at this point.

TehomCD:
When looking at it more closely, I saw that init can take typeclass as an argument to it, and passing obj.db_typeclass_path seemed to work just fine. What's odd to me is it seems like it should fall back to the object's db_typeclass_path if that's not specified, and I don't know why that doesn't happen. Maybe I should test to see if that's working at all - if object creation is always explicitly passing the kwarg, maybe something in the elif/else statements for the fallback just isn't functioning.

Griatch:
Interestingly, I got an error on my end when explicitly setting the `__init__(typeclass=path)` kwarg. I don't remember what the error was now exactly but my impression was that it confused the form due to it thinking the model to save was suddenly changing (even thought it does not change the base model of course). You not seeing the same error is interesting.

TehomCD:
I do see some strange behavior, after more testing. I tried with various typeclasses, and the object saved without error, and it looked as if the creation hook ran, giving it all the right starting attributes, but the object's key was missing. Why is the key missing? No clue. Bizarrely, when I tried again, one error I got was my dreaded database lock error, but from the django side rather than from in-game, which kind of flies in the face of my assumption that it was always based on activity.

TehomCD:
Okay, so calling `__init__` a second time seems to cause issues (like the object's key vanishing, skipping ID numbers during creation, probably other stuff), I figured the logical thing to do would just be to refactor the class-setting stuff from `TypedObject.__init__` into a new method, and have django admin call that manually. So far this seems to work without side effects.

Griatch:
Hm that does appear to be a sensible solution to the issue. I just wish I understood why the super() call of `__init__` would clear out the field properties. Ah well, I might look at this a little more but I'll likely merge this tonight, thanks!

Griatch:
Hm. When pressing the "Save" or "Save and add another" buttons in the Admin, this will save correctly and the object appear in Limbo (if location is set). However, if pressing "Save and continue editing", the result is this traceback:

```python
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 149, in get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 147, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/contrib/admin/options.py", line 541, in wrapper
    return self.admin_site.admin_view(view)(*args, **kwargs)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/utils/decorators.py", line 149, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/views/decorators/cache.py", line 57, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/contrib/admin/sites.py", line 244, in inner
    return view(request, *args, **kwargs)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/contrib/admin/options.py", line 1437, in add_view
    return self.changeform_view(request, None, form_url, extra_context)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/utils/decorators.py", line 67, in _wrapper
    return bound_func(*args, **kwargs)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/utils/decorators.py", line 149, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/utils/decorators.py", line 63, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/utils/decorators.py", line 184, in inner
    return func(*args, **kwargs)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/contrib/admin/options.py", line 1383, in changeform_view
    return self.response_add(request, new_object)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/contrib/admin/options.py", line 1084, in response_add
    current_app=self.admin_site.name)
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/core/urlresolvers.py", line 600, in reverse
    return force_text(iri_to_uri(resolver._reverse_with_prefix(view, prefix, *args, **kwargs)))
  File "/home/griatch/Devel/MUD/vienv/local/lib/python2.7/site-packages/django/core/urlresolvers.py", line 508, in _reverse_with_prefix
    (lookup_view_s, args, kwargs, len(patterns), patterns))
NoReverseMatch: Reverse for 'typeclasses_object_change' with arguments '(7,)' and keyword arguments '{}' not found. 0 pattern(s) tried: []
```

When trying this, the object curiously appears in the location's contents, but I cannot target it, and when `@reload`:ing, it's gone (also from `@objects`). This suggests it has been initialized in the content cache but that it has not actually saved to the database.

TehomCD:
The number looks to be the ID that it's trying to pass to the view, so I'm not sure why it's not accepting that. Possibly something to do with objectadmin using a custom form, I guess? I'll try to play around with it and see if I can figure it out.

TehomCD:
Er wait. typeclasses_objects_change, not objectdb_admin_change? Is it trying to call the wrong view because its .__class__ is now different, I wonder?

TehomCD:
So I guess changing '__class__' screwed with it trying to determine what view it should reverse, but fortunately I was able to figure out how to override that behavior. That particular bug seems to be fixed now.

Griatch:
Well done, yes that resolves the problem for Objects and "Save and Continue Editing". I guess all needed now is to do the same fix for the other typeclasses modifiable from the admin.

TehomCD:
Okay, the latest version should make Players, Scripts, and Channels work as well. There might be a possible issue with initial attributes that are things like empty lists, empty sets, etc, not being handled on the initial save, and becoming None instead - I did a fallback in picklefield's clean if it encountered non-strings, but I could try to improve that if it causes problems.

Issue #1133, opened 2016-12-11
EvTable slowness caused by m_len


titeuf87:
#### Brief summary of issue / Description of requested feature:
Profiling the code of EvTable to find out where the bottlenecks are made me notice that the m_len function gets called a -lot-.
This function is only needed to calculate the length of a string if there are any MXP tags in it. Even if there are no such tags in the string, this function will still always run a regex through it.

#### Steps to reproduce the issue / Reasons for adding feature:

1. `@py/time self.execute_cmd('@objects')`
2. Change the function m_len in utils.py to just return len(target)
3. `@reload`
4. `@py/time self.execute_cmd('@objects')`

Notice how this time it runs much faster! (Twice as fast actually for me).

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
There are different ways this can be solved, I'm not sure which one is preferred:

- Have an extra arg to EvTable to disable MXP: don't call the m_len function, the table will break if the user pushes an MXP tag in it.
- Strip the MXP tag in the ANSIString: this way the stripping will only happen once and the ANSIString will cache it for us (in the cleaned_string).
- Other ideas?

Griatch:
Hm, I would be reluctant to use custom flags or to completely strip MXP in ANSIString - The ability to correctly handle the length of formatting at all times and to be able to include clickable links in tables is a very powerful feature. m_len and ANSIStrings are also used in EvMenu, for example. 
It is possible one could make `m_len` be a bit more clever about it though; for example instead of always running a regex, one could do a pre-check: `if "|lt" in string: ...` and do the MXP check then. The `in` operator is often quite fast so this could potentially lead to a defacto speedup when there are no MXP markers in the text.

Griatch:
Suggested fix implemented with #1134.

PR #1134, opened 2016-12-11
Speedup for m_len to not do the regex when not needed

titeuf87:
pull request for #1133

### DIFF
--- a/evennia/utils/utils.py
+++ b/evennia/utils/utils.py
@@ -1667,7 +1667,8 @@ def m_len(target):
     # Would create circular import if in module root.
     from evennia.utils.ansi import ANSI_PARSER
     if inherits_from(target, basestring):
-        return len(ANSI_PARSER.strip_mxp(target))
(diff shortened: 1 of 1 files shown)


evennia/utils/utils.py
```diff
@@ -1667,6 +1667,8 @@ def m_len(target):
     # Would create circular import if in module root.
     from evennia.utils.ansi import ANSI_PARSER
     if inherits_from(target, basestring):
+        if "|lt" in target:
+            return len(ANSI_PARSER.strip_mxp(target))
         return len(ANSI_PARSER.strip_mxp(target))
```
Griatch: Why would this help though? With this code, it will use `strip_mxp` if `|lt` is in `target` ... and also if it is not?

titeuf87:
My mistake, I messed up when I copied my changes to a clean fresh copy of Evennia.
Updated the PR.

PR #1135, opened 2016-12-12
Moving the default permission set on an object created with @create o…

KiwiTrinsic:
Small change to the default command set for the create and dig commands. Both of those set permissions on the created object after creation, including a reference to the 'Wizards' permission. I moved the reference to the 'Wizards' permission into an attribute on the create and dig commands so that people with alternate permission structures can easily set it to their desired permission level without having to override the entire func() procedure.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -468,6 +468,7 @@ class CmdCreate(ObjManipCommand):
     key = "@create"
     locks = "cmd:perm(create) or perm(Builders)"
     help_category = "Building"
+    created_object_permission = "Wizards" # The permission level needed to control or delete the object if not the creator
 
     def func(self):
         """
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks for the PR! I merged it with an extra addition - I moved the entire lockstring into the property instead for maximum flexibility (just insert {id} in the places you want the caller.id to be used in the lockstring). Closing this with a rebase.

> Btw don't make additions like this in your own evennia-fork's master branch. The reason is that when I merge it I, like now, merge with a rebase which means that your master's history is out of sync with the history of upstream Evennia. Best is to always do your changes in branches and make PRs off those. See the [Contributing to Evennia](https://github.com/evennia/evennia/wiki/Version-Control#committing-fixes-to-evennia) page for more info.

KiwiTrinsic:
Thanks! Got it, will use a branch next time :)

Griatch:
It works fine from our end, but it will help your own sanity. :)

PR #1136, opened 2016-12-12
Moving the default permission set on an object created with @create o…

KiwiTrinsic:

…r a room created with @dig into an attribute of the command so it can be easily overridden without having to override the entire func call


KiwiTrinsic:
Whoops, shouldn't github while half-asleep

PR #1137, opened 2016-12-13
Put in check for maximum character size in input strings

TehomCD:
Just an optional setting, MAX_CHAR_LIMIT, that lets you specify the maximum length of any string passed on from the portalsessionhandler to the server, to prevent people from obnoxiously spamming and the like.

Just a little optional anti-troll protection.

I made the default None, which leaves it uncapped, so current behavior is unaffected.

### DIFF
--- a/evennia/server/portal/portalsessionhandler.py
+++ b/evennia/server/portal/portalsessionhandler.py
@@ -18,9 +18,11 @@ _MOD_IMPORT = None
 # throttles
 _MAX_CONNECTION_RATE = float(settings.MAX_CONNECTION_RATE)
 _MAX_COMMAND_RATE = float(settings.MAX_COMMAND_RATE)
+_MAX_CHAR_LIMIT = settings.MAX_CHAR_LIMIT
 
 _MIN_TIME_BETWEEN_CONNECTS = 1.0 / float(settings.MAX_CONNECTION_RATE)
 _ERROR_COMMAND_OVERFLOW = settings.COMMAND_RATE_WARNING
+_ERROR_MAX_CHAR = settings.MAX_CHAR_LIMIT_WARNING
 
 _CONNECTION_QUEUE = deque()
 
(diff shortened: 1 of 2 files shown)


evennia/settings_default.py
```diff
@@ -201,6 +201,14 @@
 MAX_COMMAND_RATE = 80
 # The warning to echo back to users if they send commands too fast
 COMMAND_RATE_WARNING ="You entered commands too fast. Wait a moment and try again."
+# Determine how large of a string can be sent to the server in number
+# of characters. If they attempt to enter a string over this character
+# limit, we stop them and send a message. Set to None by default. To
+# change it, just set it to a number of characters - ie, 6000 to be
+# roughly two pages of text.
+MAX_CHAR_LIMIT = None
```
Griatch: I think it may be a reasonable thing to actually set a sane default max here - it's good for people that don't bother with the settings. People can always up the limit manually later if they think it's limiting. I don't know what the max would be though - it would need to be a value you rarely hit yet not so large you run into slowdowns. 6000 sounds like a lot to me, but maybe it's not ...?

evennia/server/portal/portalsessionhandler.py
```diff
@@ -18,9 +18,11 @@
 # throttles
 _MAX_CONNECTION_RATE = float(settings.MAX_CONNECTION_RATE)
 _MAX_COMMAND_RATE = float(settings.MAX_COMMAND_RATE)
+_MAX_CHAR_LIMIT = settings.MAX_CHAR_LIMIT
```
Griatch: Better convert this to an int() here, so eventual unit errors in the settings file get caught early (doing > operations can have some unexpected results when mixing types).

TehomCD:
6000 is just around the edge of the largest things I've seen people write that aren't just trying to mess with people - ascii art for item descriptions, generally speaking. When I start to see clients die is when it's probably like an order of magnitude larger than that, though that's just a ballpark estimate. If you think a smaller value would be more sane as a default, by all means! It's just what I would land on myself.

PR #1140, opened 2016-12-15
Docker support

feend78:

Adds support for developing and running Evennia-based games in Docker containers. It includes a base Dockerfile that can be set up to automatically build and push images to Docker Hub, as well as a Dockerfile and docker-compose.yml file in the game_template that can be used by individual game developers to build images that contain both Evennia and their game.

See below for more detailed instructions on how to review this PR.

This is certainly a nonessential addition to the framework, but it does provide a few benefits:

* Increased number of hosting options if a game can be deployed in a container.
* Provides an alternative to virtualenv isolation. May get newcomers up and running faster.

I can't say that it's categorically better or easier to use than a traditional python configuration, and most folks will probably want to still have their virtualenv set up so their IDEs can resolve classes, etc. However, putting this PR together has been helpful in my own learning about Docker containerization, and perhaps others may find it useful in that way too.

At present, I have published my base evennia image on docker hub as **feend78/evennia**. If this PR is accepted, I would follow up by publishing it at **evennia/evennia** and setting up automated builds out of github.

#### Setting Up

First, install [Docker](docker.com). 

Download the image from docker hub with

`docker pull feend78/evennia`

Create a new Evennia game under the current directory:

`docker run -it -v ${PWD}:/usr/src/game feend78/evennia evennia --init mudgame`

Change to the newly created **mudgame** directory and build a runnable game image with the Dockerfile in the template. 

`cd mudgame`
`docker build -t feend78/mudgame .`

Perform migrations

`docker run -it -v ${PWD}:/usr/src/game feend78/mudgame evennia migrate`

The last step of setup is to start the container once in interactive mode to create the superuser.

`docker run -it -p 8000-8001:8000-8001 -p 4000:4000 -v ${PWD}:/usr/src/game feend78/mudgame`

The above command line is recommended for use in development. It runs the container and exposes ports 8000, 8001, and 4000 on it to the host machine and maps the game directory to `/usr/src/game` so that any changes made to the source files on the local machine appear within the container and will be picked up with a standard `@reload`.

Entering all of those command line parameters each time you start your game server container becomes tiresome. Docker Compose is an orchestration tool most often used to control complex multi-container applications. Here, we use it to pre-set those command line options for development. It is installed automatically with docker for Windows and MacOS; and installation instructions for others can be found [here](https://docs.docker.com/compose/install/).

With Docker Compose installed, starting the game container with the same options as above is just:

`docker-compose up`

The container can be stopped by running `@shutdown` in-game or by opening a second shell and issuing

`docker-compose down`

However, this currently results in an unclean shutdown, as evidenced by pid files that remain in the mudgame/server/ directory.


Griatch:
Hm, I'm not convinced this is something that need to go into the core evennia repo. There exist one docker image from before, gtaylor's https://github.com/gtaylor/evennia-docker, which essentially does the same as this but using git clone to get the image setup. I think it would probably make more sense to host this in a separate repo in the same way. We could certainly link to those resources more prominently though. 

As for the unclean shutdown, maybe setting up the docker with the recently created [Screen & init.d scripts](https://github.com/evennia/evennia/wiki/Start-Stop-Reload#optional-server-startup-script-linux-only) might make this a smoother experience.

Griatch:
@feend78  o before merging this, I think the lone Docker file would probably do well with having a brief explanation to what it is and the install instructions embedded directly inside it as an initial comment. 
Also a reminder: Once merged it also needs a wiki page with more detailed instructions.

feend78:
I have added the explanatory comment as requested. It includes a link to the proposed page on the wiki which does not yet exist. I am prepared to post instructions for use there ASAP after this is merged and we can set up the automatic build process.

Griatch:
@feend78 Ok, thanks for the update! Merged with a rebase. Looking forward to the wiki page. ;)

Issue #1105, opened 2016-10-20
Task: Add unit tests to all contribs


Griatch:
#### Brief summary of issue / Description of requested feature:

All contribs in `evennia/contribs` should have full (edit: as full as feasible?) unittest coverage.
#### Steps to reproduce the issue / Reasons for adding feature:

This is a sub-task of #120 (add more unittests to Evennia), arguably with a higher priority. This is because contribs are optional and thus errors in them or changes relative core may go undetected longer. 

Having existing contribs under unittesting will also make it more natural to demand unittests for new contribs as well.

gtaylor:
whew, full coverage? There is no universal truth with testing, but covering the right 20-40% is often as good as 100%. It's a matter of making sure that you've got the right bases covered. I fear that you're going to see less contrib activity if there's a 100% requirement.

Griatch:
Maybe my definition of "full" is too imprecise. I'd be happy to just have each contrib have unittests ... any at all at this point. :)

gtaylor:
Fair enough!

Griatch:
Added coverage testing with [coverall](https://coveralls.io/github/evennia/evennia) which should help a little with this.

Griatch:
Added unit tests for the contrib barter system in b12a850 (forgot to reference this issue from it).

Griatch:
This is moved to development towards Evennia 0.7, since finishing this will lead to us being able to demand unittest coverage for all new contribs.

Griatch:
All contribs are now covered by unittests to some degree and this can be closed. We will henceforth request that new contribs also come with unit tests. Removed devel branch tag since this is already live in master. Closing.

PR #1145, opened 2016-12-20
mail.py

grungies1138:
In-game mail system.

### DIFF
--- a/evennia/contrib/mail.py
+++ b/evennia/contrib/mail.py
@@ -22,24 +22,21 @@ class CmdMail(default_cmds.MuxCommand):
     """
     Commands that allow either IC or OOC communications
 
-    Usage:    
-        {w@mail{n
-                - Displays all the mail a player has in their mailbox
+    Usage:
+        @mail       - Displays all the mail a player has in their mailbox
 
-        {w@mail <#>{n
-                - Displays a specific message
+        @mail <#>   - Displays a specific message
 
-        {w@mail <players>=<subject>/<message>{n
+        @mail <players>=<subject>/<message>
                 - Sends a message to the comma separated list of players.
 
-        {w@mail/delete <#>{n
-                - Deletes a specific message
+        @mail/delete <#> - Deletes a specific message
 
-        {w@mail/forward <player list>=<#>[/<Message>]{n
+        @mail/forward <player list>=<#>[/<Message>]
                 - Forwards an existing message to the specified list of players,
                   original message is delivered with optional Message prepended.
 
-        {w@mail/reply <#>=<message>{n
+        @mail/reply <#>=<message>
                 - Replies to a message #.  Prepends message to the original
                   message text.
     Switches:
@@ -47,12 +44,6 @@ class CmdMail(default_cmds.MuxCommand):
         forward - forward a received message to another object with an optional message attached.
         reply   - Replies to a received message, appending the original message to the bottom.
 
-    Status Legend:
-        '|wU|n' = Unread/New message
-        '|wO|n' = Read/Old message
-        '|wR|n' = Replied to
-        '|wF|n' = Forwarded
-        
     Examples:
         @mail 2
         @mail Griatch=New mail/Hey man, I am sending you a message!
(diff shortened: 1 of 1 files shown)


evennia/contrib/mail.py
```diff
@@ -0,0 +1,194 @@
...
+SUB_HEAD_CHAR = "-"
+
+class MailCommand(default_cmds.MuxCommand):
+    """
+    Commands that allow either IC or OOC communications
+
+    Usage:
+        {w@mail{n
+                - Displays all the mail a player has in their mailbox
+
```
Griatch: Excellent docstring, but please make it in the style of the Evennia [Command documentation policy](https://github.com/evennia/evennia/wiki/Using%20MUX%20As%20a%20Standard#documentation-policy) - basically make a summary of the Switches and move the examples to the end. This is not just me being picky - doing so will help the API2md generator automatically create a nice API wiki page for the contrib.

evennia/contrib/mail.py
```diff
@@ -0,0 +1,194 @@
...
+                table.reformat_column(3, width=13)
+                table.reformat_column(4, width=7)
+
+                self.caller.msg(HEAD_CHAR * 78)
+                self.caller.msg(table)
+                self.caller.msg(HEAD_CHAR * 78)
+            else:
+                self.caller.msg("Sorry, you don't have any messages.  What a pathetic loser!")
+
+    def get_all_mail(self):
```
Griatch: This needs a docstring in Google format. See the [Evennia code style guide](https://github.com/evennia/evennia/blob/master/CODING_STYLE.md#function--method-docstrings).

evennia/contrib/mail.py
```diff
@@ -0,0 +1,194 @@
...
+        mail_messages = Msg.objects.get_by_tag(category="mail")
+        messages = []
+        messages = [m for m in mail_messages if self.caller.player in m.receivers]
+        return messages
+
+    def send_mail(self, recipients, subject, message, caller):
```
Griatch: This needs a docstring in Google format. See the [Evennia code style guide](https://github.com/evennia/evennia/blob/master/CODING_STYLE.md#function--method-docstrings).

evennia/contrib/mail.py
```diff
@@ -0,0 +1,194 @@
...
+Installation:
+    Import this module into the default Player or Character commandset.
+"""
+
+from evennia import default_cmds, search_object
+from evennia.utils import create, utils, evtable, evform
+from evennia.comms.models import Msg
+
+
+HEAD_CHAR = "|015-|n"
```
Griatch: Since these global variables are not meant to accessible from the outside or listed as contents of the module, a good idea is to start their names with `_`, so `_HEAD_CHAR` etc.

Griatch:
Also, we don't currently request that contribs come with unit tests, but we are slowly adding ones to the contrib folder as per [Issue 1105](https://github.com/evennia/evennia/issues/1105) so those will eventually have to be written also for this. For now though, I won't require it for merging. :)

evennia/contrib/mail.py
```diff
@@ -0,0 +1,220 @@
...
+_HEAD_CHAR = "|015-|n"
+_SUB_HEAD_CHAR = "-"
+_WIDTH = 78
+
+class CmdMail(default_cmds.MuxCommand):
+    """
+    Commands that allow either IC or OOC communications
+
+    Usage:    
+        {w@mail{n
```
Griatch: The `{` were not corrected, but looking at it, I think it's probably better to ditch the coloration in the help entry. It will not work well with the API docs.

evennia/contrib/mail.py
```diff
@@ -0,0 +1,220 @@
...
+                        subject, body = self.rhs.split("/", 1)
+                    else:
+                        body = self.rhs
+                    self.send_mail(self.lhslist, subject, body, self.caller)
+                else:
+                    message = self.get_all_mail()[int(self.lhs) - 1]
+
+                    messageForm = []
+                    if message:
+                        messageForm.append(HEAD_CHAR * _WIDTH)
```
Griatch: HEAD_CHAR should be _HEAD_CHAR (several more examples in the file - best do a search & replace)

evennia/contrib/mail.py
```diff
@@ -0,0 +1,220 @@
...
+    def func(self):
+        subject = ""
+        body = ""
+        if self.switches or self.args:
+            if "delete" in self.switches:
+                if not self.lhs:
+                    self.caller.msg("No Message ID given.  Unable to delete.")
+                    return
+                else:
+                    if self.get_all_mail()[int(self.lhs) - 1]:
```
Griatch: All the cases where you do int(self.lhs) is susceptible to a traceback if the user enters something other than a number. For example `@mail blah` currently gives a traceback. There are many such crashes in the code below due to invalidated input.

evennia/contrib/mail.py
```diff
@@ -0,0 +1,220 @@
...
+_SUB_HEAD_CHAR = "-"
+_WIDTH = 78
+
+class CmdMail(default_cmds.MuxCommand):
+    """
+    Commands that allow either IC or OOC communications
+
+    Usage:    
+        {w@mail{n
+                - Displays all the mail a player has in their mailbox
```
Griatch: Much better docstring, but it's so long it doesn't fit in a normal-sized terminal. Would it be possible to shorten it down?

Griatch:
Actually; scratch removing coloring if you don't want to; this is an optional contrib after all.

evennia/contrib/mail.py
```diff
@@ -0,0 +1,220 @@
...
+        
+        Returns:
+            messages (list): list of Msg objects.
+        """
+        mail_messages = Msg.objects.get_by_tag(category="mail")
+        messages = []
+        messages = [m for m in mail_messages if self.caller.player in m.receivers]
+        return messages
+
+    def send_mail(self, recipients, subject, message, caller):
```
TehomCD: As a small idea, you may want to do `Msg.objects.filter(db_receivers_players=self.caller.play).get_by_tag(category="mail")` rather than doing the `in m.receivers` check in the list comprehension - having a for loop check membership and hit the database each time could make the command run really slowly if you have a lot of objects, while a filter by player name would only hit it once. I ran into a similar situation using Msg objects for bulletin boards and checking receivers to see read posts.

Griatch:
Nitpick: "update mail.py" is not a very informative commit message. Please try to make commit messages useful for your own sanity and ours. :)

grungies1138:
My bad.

PR #1146, opened 2016-12-22
Change gender neutral pronouns from 'it' to the more commonly used 'they'

Torild:
'They' is much more commonly used as a gender neutral pronoun, so I thought it'd be more appropriate.

### DIFF
--- a/evennia/web/website/templates/website/base.html
+++ b/evennia/web/website/templates/website/base.html
@@ -15,6 +15,9 @@
     <link rel="stylesheet" type="text/css" href="{% static "website/css/prosimii-print.css" %}" media="print" />
     <link rel="icon" type="image/x-icon" href="/static/website/images/evennia_logo.png" />
 
+    <!-- Custom CSS -->
+    <link rel="stylesheet" type="text/css" href="{% static "website/css/app.css" %}">
+
     {% block header_ext %}
     {% endblock %}
 


BlauFeuer:
Please don't take "it" away.  #1151

As I understand it, the use of "they" is intended to provide a substitute between making the choice of he or she (making the reference to the character's gender "ambiguous"), but there are cases where the character may want to identify as neither (making the reference to the character's gender as "neutral"); that is when "it" is the better choice,

As well-intentioned as it is, the current intended change removes a legitimately neutral choice and replaces it with a choice that's socially neutral, yet ambiguous.  The loss of the neutral pronoun isn't progress, nor is "they" always a more appropriate pronoun than "it", no matter how much more commonly used it is.

Griatch:
Considering the gendersub contrib is a quick demo I did to show how it could be done, it is seeing a lot of attention. :) 
While I personally prefer to use "they" for people-gender neutrality, I agree with Blue that "it" is used for different things here. Maybe we should instead have a separate tag for the ambiguous case (they)?

(Also, this contrib needs to change to use |-markers instead of {-markers since the latters are deprecated).

Griatch:
Any further discussion or opinion on this?

arumford:
Seems to me that 'They' tends to refer to an ambiguous sentient.  Where as 'it' refers to inanimate objects.

Kelketek:
@Griatch Only comment I have is suggesting that one look at what Collab does with the pronouns. Something like that might be the better method of handling this. If pronouns can be overwritten more easily per-game and per-character, the problem is side-stepped and developers can use their preference.

You can be sure that if we get converts from MUCK-land, the hermaphrodite wolves will have their say.

Griatch:
@Kelketek Not sure I like the use of this contrib as an import target for collab actually. ;) Contribs are meant to be stand-alone things you can plug in or not. But anyway, a user wanting to add their own genders would copy this simple little contrib and modify `GENDER_PRONOUN_MAP` to their liking. 

All: As for the defaults available in this PR, I think it probably works best to instead of replacing "neutral" "it" with "they" we should add an "ambiguous" key and use "they" for that. It seems to offer the most flexibility and options.

@Torild 
The website additions for mobile use seem nice but are not covered by your PR request. Are they tested and are intended to be added? Don't recall off hand if something like this was already merged.

Kelketek:
@Griatch Other contribs are interdependent (such as other ones relying on menu), so this is not unprecedented. Collab is absolutely going to need gender handling because of the amount of customization that will be afforded to individual users.

Griatch:
EvMenu is a core utility these days actually. Anyway, let's not discuss contrib here, that's for when it has a PR of its own again. :)

Griatch:
There seems to be an issue with this PR in the sense that it comes from an "unknown repository" and that the patch file is apparently empty. So I'll have to cherry-pick from here instead.

Griatch:
Thanks for the PR, in the end, since the patch didn't seem to work I implemented this using the "ambiguous" gender type as a new default, keeping the "neutral" one as a fourth alternative.

BlauFeuer:
All robot-kind thanks you. "It"s are not always inanimate objects.

PR #1148, opened 2016-12-26
fix evmore sending messages to stale session object

TehomCD:
If a player is inside an EvMore page and disconnect then reconnect, they will only get blank messages from the pager, because it tries to send to a session object that is no longer attached to them. Changed display method to check for this and hook it back up to the correct session if necessary.

Fix for players wondering why they can't see anything if they reconnect from when they were inside an EvMore pager.

N/A


Griatch:
Thanks! The only think I might be looking at for the future is that you are assuming the first session is the one to use - not sure how well that will play with multisession modes. Anyway, merged with a rebase.

PR #1149, opened 2016-12-27
Fix an issues were exit aliases weren't being check when interating o…

Phayte:
The create_exit method for CmdOpen uses None as a default value for exit_aliases. Within the body of the code, exit_aliases is iterated over in two parts of the body. Neither verify if it contains a valid value. If exit_aliases is left as None, a NoneType exception is thrown. 

CmdOpen will never throw an exception because it always passes an array (even if empty) to the create_exit method. However, anyone unknowing inheriting this class and extending it might try to reuse the command and run into the error. If the method defaults the parameter to a None value, then it's the responsibility of the method to error check this. The alternatives would be to default exit_alises to an empty array (which I believe isn't recommended due to how scoping and manipulations work), or making it a required variable.

N/A

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -1240,7 +1240,8 @@ class CmdOpen(ObjManipCommand):
                 if old_destination.id != destination.id:
                     # reroute the old exit.
                     exit_obj.destination = destination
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks! Merged with a rebase.

Issue #614, opened 2014-11-08
Webclient extensions


Kelketek:
This issue tracks the overall task of extending the webclient beyond the basic telnet-like default.
### Default extended features

~~1. Tab Notifications when not in window.~~ Implemented in #1046.
~~2. Make a webclient configuration/settings pane (saves to the Player)~~
~~3. Give option of notifications with sound.~~
~~4. Scroll management - option to auto-scroll/not with new data~~
~~5. Option to split the GUI into panes, redirect different data to different panes by OOB marker or manually with regex (settable in option pane). A suggestion for a lightweight JS library that could be used for this is [Split.js](https://nathancahill.github.io/Split.js/) which also allows the user to themselves adjust the size of the panes created.~~

### Further ideas, Possibly optional

1. Private message window (relates to multipane support)
2. Notes scratchpad
3. GUI editor, for things like descriptions
4. Tab completion in input window

Any others?

Griatch:
@zero9k 

A clear thing that could be improved with that example (while certainly pretty) is to split out the room look info into its separate window as well (update it in-place), to avoid repeated looks spamming the main window and pushing actual content (such as emotes or combat updates) out of sight. The more you can focus only on the new stuff in the main window, the better. 
.
Griatch

BattleJenkins:
@Griatch - I've personally always wanted that feature in every MUD I've played, so I would absolutely love to see that as an option!

Griatch:
With the changes to the webclient, the server communication has been separated out away from the GUI-component. it's would now be possible to plug in things like this by simply having the plugin hook into and listen to any custom events the server feels like sending. One can even to replace the entire webclient_GUI. It would require to modify the GUI js code but I suppose one could have it load an empty js file that you can replace and fill with whatever glue code you need for your functionality ... 

Any ideas on this, now that we have a good baseline client to work from?

Griatch:
This issue now tracks extensions to the webclient.

titeuf87:
I started working on a settings window which can be found on my [webclient branch](/titeuf87/evennia/tree/webclient).

So far it suppots the following:

- Simple popup dialog that can be opened by the player
- Retrieves settings stored on the server
- Stores settings back to the server
- Monitor for settings changes

It looks like this:
![screen shot 2016-12-31 at 13 31 11-fullpage](https://cloud.githubusercontent.com/assets/2372338/21577542/7fe567e0-cf5e-11e6-949f-5e436dbd054d.png)

[...]

PR #1152, opened 2016-12-28
Send an OOB message to the client once a player logs in.

titeuf87:
This sends an OOB message to the client once the player logs in.

This is useful so that the client knows it can retrieve character/player specific data (like player avatar, player description, client settings, ...).
The idea is that, once the client gets this message, it sends OOB messages back to the server to retrieve whatever it needs.

I think this is better than the other way around (as soon as the player logs in, the server sending the data the client wants straight away).
This makes making different clients easier (for instance a desktop webclient and another phone webclient): the server doesn't need to know the difference between the different clients as they each are responsible for retrieving the data they need.

I plan on using this as part of #614: this step will allow the client to start monitoring/retrieving the saved settings from the server.

### DIFF
--- a/evennia/players/players.py
+++ b/evennia/players/players.py
@@ -725,6 +725,10 @@ class DefaultPlayer(with_metaclass(TypeclassBase, PlayerDB)):
         if session and protocol_flags:
             session.update_flags(**protocol_flags)
 
+        # inform the client that we logged in through an OOB message
+        if session:
+            session.msg(logged_in={})
+            
         self._send_to_connect_channel("|G%s connected|n" % self.key)
         if _MULTISESSION_MODE == 0:
             # in this mode we should have only one character available. We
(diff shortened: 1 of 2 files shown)


Griatch:
This sounds like a useful addition. I have tested it with a few telnet clients without seeing any ill effects. The only question is if GMCP/MSDP already has standard names for this operation and it should thus be translated to the correct form (something like "Event.Login"). Will need to look at that before merging.

titeuf87:
For both GMCP and MSDP I can't really find anything.
GMCP has a `Char.Login` but it's used the other way around: for the client to login a character.
MSDP I didn't really see anything relevant.
I looked at the documentantion found [here](https://github.com/keneanung/GMCPAdditions) for GMCP and [here](http://tintin.sourceforge.net/msdp/) for MSDP.

On an unrelated note: I changed the webclient_gui.js file. This should work fine for everyone that has not overriden that one. But people that did change it will need to manually add the logged_in event listener or the webclient will show an error line when it receives that message...

Griatch:
Yes, I'll make a mention of the webclient change on the mailing list. You are correct, it doesn't look GMCP/MSDP handle this direction of send. We'll see if any legacy telnet clients freak out over this though, in that case we may have to restrict this only to the webclient. Merged with rebase.

PR #1155, opened 2016-12-30
Fixes error messages on the boot command

titeuf87:
Two small fixes for the `@boot` command:

- No traceback if the booted player is not found
- Send a message to the correct player if the lockfunc fails


Griatch:
Looks good, merged with a rebase.

PR #1158, opened 2017-01-02
Update format_output on CmdExamine to display obj.email.

jrsteensen:
Updated format_output on CmdExamine to add a check for email attribute. If true, it displays the email address.

Several other contribs ask for an email , and setting up the superuser adds this attribute in stock Evennia. Spent several hours trying to figure out why it wasn't displaying in exam before @Days4121 and I realized that it was set as a direct attribute, and not a db.attribute. Made it display in exam.


Griatch:
The email is optional during superuser setup (you can leave it blank) and not used anywhere in the default system. So this technically belongs as a contrib. But as a two-line addition I guess it's okay. Merged with a rebase.

jrsteensen:
Appreciate it. The reason why I submitted it as a standalone, was that the create.create_player already had an argument for email, it just wasn't asked for during standard player_create, but it still jammed in a default value.

PR #1159, opened 2017-01-05
Fix get_all_scripts to not return [None]

titeuf87:
This is a small fix for get_all_script to return [] instead of [None] when searching for a script by `#dbref` and that script does not exist.

When using `@script/stop #nonexistendid` the user would get a python traceback. Now they'll get a nice error message saying the script can't be found.

### DIFF
--- a/evennia/scripts/manager.py
+++ b/evennia/scripts/manager.py
@@ -85,7 +85,7 @@ class ScriptDBManager(TypedObjectManager):
             script = []
             dbref = self.dbref(key)
             if dbref or dbref == 0:
-                script = [self.dbref_search(dbref)]
+                script = filter(None, [self.dbref_search(dbref)])
             if not script:
                 script = self.filter(db_key=key)
             return script


Griatch:
Thanks! Merged with a rebase.

evennia/scripts/manager.py
```diff
@@ -85,7 +85,7 @@ def get_all_scripts(self, key=None):
             script = []
             dbref = self.dbref(key)
             if dbref or dbref == 0:
-                script = [self.dbref_search(dbref)]
+                script = filter(None, [self.dbref_search(dbref)])
```
Phayte: I know this is closed, but I just noticed, should this be *self.filter* and not just filter or where is filter coming from? I don't see it defined or in imports.

titeuf87: `filter` is a builtin function in Python:

    filter(...)
    filter(function or None, sequence) -> list, tuple, or string
    
    Return those items of sequence for which function(item) is true.  If
    function is None, return the items that are true.  If sequence is a tuple
    or string, return the same type, else return a list.

So there are no imports needed for this.

For what it's worth, Griatch changed this to a list comprehension instead of using filter for compatibility with Python3 (filter there returns an iterator and not a list).

Phayte: Ah ok, cool. Good to know.

PR #1162, opened 2017-01-13
When calling EvMenu as a player, the session is not passed through wh…

Phayte:
When calling EvMenu as a player, the session is not passed which causes issues for cmd_on_exit commands that have player_caller = true. EvMenu executes cmd_on_exit via execute_cmd function, but leave session empty. In the parse() of the MuxCommand, it attempts to retrieve a player's puppet via the session but has no check to see if it's None. The change just opens up an optional keyword argument to pass the session through to avoid this issue.

Alternatively, the MuxCommand parse() could have defensive measure to check to see if a session not none before attempting to retrieve a puppet. (https://github.com/evennia/evennia/blob/master/evennia/commands/default/muxcommand.py#L206) I'm conflicted on which one would make more sense as a player should always have a session and by putting a None check, the code reads as though it's possible to have a player without a session.

I'm honestly not sure how much is to be gained fixing this bug because I'm not sure who else would actually run into the issue because I do realize that have a menu on a player level may cause a multitude of issues. I'll leave it for 3rd parties to comment on the usefulness.

If this change is accepted,  a note on the wiki should probably be added to document the addition of the new keyword parameter to specify that if the caller is a player and the cmd_on_exit command is a player_caller = true command, this field needs to be set.

### DIFF
--- a/evennia/utils/evmenu.py
+++ b/evennia/utils/evmenu.py
@@ -510,7 +510,10 @@ class EvMenu(object):
             startnode_input (str, optional): Send an input text to `startnode` as if
                 a user input text from a fictional previous node. When the server reloads,
                 the latest visited node will be re-run using this kwarg.
-
+            session(Session, optional): If the caller is a player and specifies
+                a player command (a command with the player_caller set to true)
+                for the cmd_on_exit value, the session needs to be set in to 
+                execute properly.
         Kwargs:
             any (any): All kwargs will become initialization variables on `caller.ndb._menutree`,
                 to be available at run.
--- a/evennia/utils/evmenu.py
+++ b/evennia/utils/evmenu.py
@@ -420,7 +420,8 @@ class EvMenu(object):
                  options_formatter=evtable_options_formatter,
                  node_formatter=underline_node_formatter,
                  input_parser=evtable_parse_input,
-                 persistent=False, startnode_input="", **kwargs):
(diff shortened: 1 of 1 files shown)


Phayte:
While this change fixed my immediate issue, the more I look at the code, the more I dislike the change for the following reasons:

- In persistent mode, I assume sessions remain open so that if the lambda referencing the passed session would still remain valid. If this is not the case, this would not be a change we should make given we don't want to reference a dangling/non-existent session.
- In the case a user would like to pass a session into the menu via the **kwargs, there would be a naming conflict. Assuming the change is acceptable, would it be valid as well to store a reference of the session inside the menu (self._session = session).

Griatch:
As you say, this solution is not future-proof with persistent menus. Session is a pretty fleeting thing and not persistent across shutdowns. So thanks for the PR but I don't think I will merge this, but try with another solution:  

Let CmdEvMenuNode do `caller.ndb._menutree._session = self.session` and then use the stored session in the `execute_cmd` call by accessing it on the menu. This solves this particular Issue fully. 
By adding `session` keywords to all msg() calls inside Evmenu, it will also correctly redirect the text to Players in multisession modes > 2, *except* for the very first node before the command was called for the very first time. For that one case a `session` keyword like suggested here is needed.

Closing the PR. I pushed the alternative solution, see if that works for your issue.

PR #1168, opened 2017-01-18
Current RSS & IRC bot functionality recreated with Inputfunc.

CloudKeeper1:
The implementation of the RSS and IRC bot has been changed to utilise inputfunc to match current Evennia  input conventions. Cleaned up the bot commands no longer in use.

### DIFF
--- a/evennia/server/inputfuncs.py
+++ b/evennia/server/inputfuncs.py
@@ -80,6 +80,41 @@ def text(session, *args, **kwargs):
     session.update_session_counters()
 
 
+def bot_data_in(session, *args, **kwargs):
+    """
+    Text input from the IRC and RSS bots.
+    This will trigger the execute_cmd method on the bots in-game counterpart.
+
+    Args:
+        text (str): First arg is text input. Other arguments are ignored.
+
+    """
+
+    text = args[0] if args else None
+
+    # Explicitly check for None since text can be an empty string, which is
+    # also valid
+    if text is None:
+        return
+    # this is treated as a command input
+    # handle the 'idle' command
+    if text.strip() in _IDLE_COMMAND:
+        session.update_session_counters(idle=True)
+        return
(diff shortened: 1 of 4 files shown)


evennia/server/inputfuncs.py
```diff
@@ -80,6 +80,41 @@ def text(session, *args, **kwargs):
     session.update_session_counters()
 
 
+def bot_data_in(session, *args, **kwargs):
+    """
+    Text input from the IRC and RSS bots.
+    This will trigger the execute_cmd method on the bots in-game counterpart.
+
+    Args:
```
Griatch: The `session` argument is missing from the argument list documentation.

evennia/server/inputfuncs.py
```diff
@@ -80,6 +80,41 @@ def text(session, *args, **kwargs):
...
+    # this is treated as a command input
+    # handle the 'idle' command
+    if text.strip() in _IDLE_COMMAND:
+        session.update_session_counters(idle=True)
+        return
+    if session.player:
+        # nick replacement
+        puppet = session.puppet
+        if puppet:
+            text = puppet.nicks.nickreplace(text,
```
Griatch: Is nick replacement at this level really something the bot needs?

Griatch:
@CloudKeeper1 
Thanks for the contribution! Merged it. A hint though - don't make additions like this from your Evennia clone's `master` branch. In this case it works fine since upstream has no other changes to merge. But very often there would be other changes happening from the time you make the PR and it can be merged. In that case I would rebase your changes into master which means you would have a conflict with your changes to master and upstream master when you pull next. So for your own sanity, keep your master "clean" and always make additions in new branches (that you can then throw away when the change has merged). See [here](https://github.com/evennia/evennia/wiki/Version%20Control#committing-fixes-to-evennia) for more info.

PR #1171, opened 2017-01-20
Set another factory to not be noisy

TehomCD:
I was wondering why I still had so much factory spam in my portal logs after we set other factories to have `noisy = False`, and finally found where the last one was.

Make portal logs readable.

Continuation of #1107


Griatch:
Merged with a rebase, thanks!

PR #1178, opened 2017-01-27
Wilderness area contrib

titeuf87:
This adds a new contrib that adds a wilderness area.
The idea comes from Griatch from the [mailing list](https://groups.google.com/forum/#!searchin/evennia/ocean%7Csort:relevance/evennia/MGI6VYwMP2g/CTuyIksCDAAJ)

This can be used by others so they have something to base their work on without having to reinvent the wheel.

I hope that the comments/docstrings are clear!

### DIFF
--- a/evennia/contrib/wilderness.py
+++ b/evennia/contrib/wilderness.py
@@ -1,0 +1,723 @@
+"""
+Wilderness system
+
+Evennia contrib - titeuf87 2017
+
+This contrib provides a wilderness map. This is an area that can be huge where
+the rooms are mostly similar, except for some small cosmetic changes like the
+room name.
+
+Usage:
+
+    This contrib does not provide any commands. Instead the @py command can be
+    used.
+
+    A wilderness map needs to created first. There can be different maps, all
+    with their own name. If no name is provided, then a default one is used.
+
+    @py from evennia.contrib import wilderness; wilderness.create_wilderness()
+
+    Once created, it is possible to move into that wilderness map:
+    @py from evennia.contrib import wilderness; wilderness.enter_wilderness(me)
+
+    All coordinates used by the wilderness map are in the format of (x, y)
+    tuples. x goes from left to right and y goes from bottom to top. So x = 0
+    is on the left and y = 0 is at the bottom of the map.
+
+
+Customisation:
+
+    The defaults, while useable, are meant to be customised. When creating a
+    new wilderness map it is possible to give a "map provider": this is a
+    python object that is smart enough to create the map.
+
+    The default provider, WildernessMapProvider, just creates a grid area that
+    is unlimited in size.
+    This WildernessMapProvider can be subclassed to create more interesting
+    maps and also to customize the room/exit typeclass used.
+
+    There is also no command that allows players to enter the wilderness. This
+    still needs to be added: it can be a command or an exit, depending on your
+    needs.
+
+Customisation example:
+
+    To give an example of how to customize, we will create a very simple (and
+    small) wilderness map that is shaped like a pyramid. The map will be
+    provided as a string: a "." symbol is a location we can walk on.
+
+    Let's create a file world/pyramid.py:
+
+    ```python
+    map_str = \"\"\"
+         .
+        ...
+       .....
+      .......
+    \"\"\"
+
+    from evennia.contrib import wilderness
+
+    class PyramidMapProvider(wilderness.WildernessMapProvider):
+
+        def is_valid_coordinates(self, wilderness, coordinates):
+            x, y = coordinates
+            try:
+                lines = map_str.split("\n")
+                # The reverse is needed because otherwise the pyramid will be
+                # upside down
+                lines.reverse()
+                line = lines[y]
+                column = line[x]
+                return column == "."
+            except IndexError:
+                return False
+
+        def get_location_name(self, coordinates):
+            x, y = coordinates
+            if y == 3:
+                return "Atop the pyramid."
+            else:
+                return "Inside a pyramid."
+    ```
+
+    Now we can use our new pyramid-shaped wilderness map. From inside Evennia:
+
+    ```
+    @py from world import pyramid as p; p.wilderness.create_wilderness(mapprovider=p.PyramidMapProvider())
+
+    @py from evennia.contrib import wilderness; wilderness.enter_wilderness(me, coordinates=(4, 1))
+    ```
+
+Implementation details:
+
+    When a character moves into the wilderness, they get their own room. If
+    they move, instead of moving the character, the room changes to match the
+    new coordinates.
+    If a character meets another character in the wilderness, then their room
+    merges. When one of the character leaves again, they each get their own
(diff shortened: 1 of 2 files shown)


evennia/contrib/wilderness.py
```diff
@@ -0,0 +1,490 @@
...
+            # location
+
+            location = itemlocs[moved_obj]
+            # Setting the location to None is important here so that we always
+            # get a "fresh" room
+            moved_obj.location = None
+            self.wilderness.move_obj(moved_obj, location)
+        else:
+            # This object wasn't in the wilderness yet. Let's add it.
+            itemlocs[moved_obj] = self.ndb.active_location
```
Griatch: While I can see later in the code that self.ndb.active_location is set by the wilderness script, this is the first time we see this property appear, so it would help readability to mention where it comes from, or even mention this in the class docstring.

evennia/contrib/wilderness.py
```diff
@@ -0,0 +1,490 @@
...
+        self.ndb.active_location = new_location
+        rooms[self.ndb.active_location] = self
+
+        # Every obj inside this room will get its location set to None
+        for item in self.contents:
+            if not item.destination or item.destination != item.location:
+                item.location = None
+        # And every obj matching the new location will get its location set to
+        # this room
+        for item, location in self.wilderness.db.itemlocations.iteritems():
```
Griatch: The performance of this will scale linearly (O(N)) with the number of the objects in the wilderness. Which may be fine; for a larger set one could consider using Tags for the coordinates. Those scale as O(1) due to database indexing and would eventually be considerably more effective. Not something for this PR, but maybe worth to make a comment on for those having bigger performance requirements.

evennia/contrib/wilderness.py
```diff
@@ -0,0 +1,490 @@
...
+class WildernessScript(DefaultScript):
+
+    def at_script_creation(self):
+        self.persistent = True
+
+        # Store the location of every item that is inside the wilderness
+        # Key: object, Value: (x, y)
+        self.db.itemlocations = {}
+
+        # Store the rooms that are used as views into the wilderness
```
Griatch: Does this mean you can insert "fixed" rooms into the dynamic grid too (like a dungeon entrance or town)? Considering new "dynamic" rooms are only taken from the `unused_rooms` list, this sounds like an easy additional feature to add, but would require a helper method to specify the fixed room and at which coordinate it should appear.

evennia/contrib/wilderness.py
```diff
@@ -0,0 +1,490 @@
...
+        y -= 1
+    if direction in ("northwest", "west", "southwest"):
+        x -= 1
+    if direction in ("northeast", "east", "southeast"):
+        x += 1
+
+    return (x, y)
+
+
+class WildernessScript(DefaultScript):
```
Griatch: This requires a docstring detailing what the purpose of it is.

evennia/contrib/wilderness.py
```diff
@@ -0,0 +1,490 @@
...
+
+    Once created, it is possible to move into that wilderness map:
+    @py from evennia.contrib import wilderness; wilderness.enter_wilderness(me)
+
+
+Customisation:
+
+    The defaults, while useable, are meant to be customised. When creating a new
+    wilderness map it is possible to give a "map provider": this is a python
+    object that is smart enough to create the map.
```
Griatch: You should give a brief example of a Map Provider here, especially which methods the wilderness system expects it to have.

evennia/contrib/wilderness.py
```diff
@@ -0,0 +1,490 @@
...
+This contrib provides a wilderness map. This is an area that can be huge where
+the rooms are mostly similar, except for some small cosmetic changes like the
+room name.
+
+Usage:
+
+    This contrib does not provide any commands. Instead the @py command can be
+    used.
+
+    A wilderness map needs to created first. There can be different maps, all
```
Griatch: It would be useful with a very brief example of creating such a map, here or at the end of this docstring.

evennia/contrib/wilderness.py
titeuf87: That's a good idea! Currently that's not possible as the rooms would be moved away to storage when they're not needed anymore. But adding a "keep" flag could fix that.

titeuf87:
I think I fixed all various docstrings/style issues. 
Added an example too on how to customise it.
Changed location to coordinates (which is indeed clearer)

Let me know if I missed something!

evennia/contrib/wilderness.py
```diff
@@ -0,0 +1,723 @@
...
+
+        traversing_object.at_after_move(None)
+        return True
+
+
+class WildernessMapProvider(object):
+    """
+    Default Wilderness Map provider.
+
+    This is a simple provider that just creates an infite large grid area.
```
Griatch: infite -> infinite

evennia/contrib/wilderness.py
```diff
@@ -0,0 +1,723 @@
...
+        """
+        x, y = coordinates
+        if x < 0:
+            return False
+        if y < 0:
+            return False
+
+        return True
+
+    def get_location_name(self, coordinates):
```
Griatch: Only a get_location_name method seems very limited for the kind of use this will see. The average developer will have to at least overload the description per-room, based on the coordinate. Would it not be suitable to also supply something like `at_select(self, coordinates, caller, room)` that gets called with the room object and the character moving when the room is being prepared and allows the user to assign data to the room based on its coordinate?

evennia/contrib/wilderness.py
```diff
@@ -0,0 +1,723 @@
...
+
+        traversing_object.location.msg_contents("{} arrives from {}".format(
+            traversing_object.key, current_coordinates),
+            exclude=[traversing_object])
+
+        traversing_object.at_after_move(None)
+        return True
+
+
+class WildernessMapProvider(object):
```
Griatch: Since this is sort of a default input to the other entities, it should probably be at the very end, after the WildernessScript, which is not intended to be modified as far as I understand.

Griatch:
I tested the example commands and could move around on the default infinite wilderness just fine! I added some usability comments.

titeuf87:
That `at_select` was a good idea. I renamed it to `at_prepare_room` though to make it clearer what it does. Also fixed your other comments.

Griatch:
I merged with a rebase. One thing struck me though when looking at this. There are dicts for storing the position of objects in the wilderness, but I don't see any examples for a developer to actually manage objects and which methods they would use.

titeuf87:
What exactly do you mean by managing the objects inside the wilderness?

The only issue I can think of right now is that while you can add any object in the wilderness (using the `enter_wilderness`, it is not actually possible to remove random objects back out of the wilderness unless they just happen to be at the same coordinates than a player is standing in.

Griatch:
Aha, you use enter_wilderness to put stuff in the wilderness. That makes sense in retrospect, but it was not obvious to me from the doc.

PR #1182, opened 2017-01-29
Change taghandler clear to delete through models, not Tags themselves

TehomCD:
`Taghandler.clear()` calls delete on a queryset of Tags, not the through model, resulting in deleting tags for all objects rather than just the object trying to clear the tags. It was easy to miss because of caching making the other tags appear to remain until a `@reload` wiped them.

Fix other tags disappearing if tags are cleared.

N/A


Griatch:
Ooh, good catch!

PR #1186, opened 2017-02-01
Change default at_pre_puppet for characters

BlauFeuer:
No longer move character "back onto the grid" when not in None location.
This allows for Character objects to be moved while unpuppeted and remain
where placed when puppeted again.

Also updates markup in code within this module.

Prevent default behavior of storage while unpuppeted from moving character when it has already been moved, or has deliberately been moved back into the game to a different location.

Only affects behavior if character object has been moved since last being stored in None.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -1347,18 +1347,18 @@ class DefaultObject(with_metaclass(TypeclassBase, ObjectDB)):
             if con.destination:
                 exits.append(key)
             elif con.has_player:
-                users.append("{c%s{n" % key)
+                users.append("|c%s|n" % key)
             else:
                 things.append(key)
         # get description, build string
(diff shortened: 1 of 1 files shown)


BlauFeuer:
This refactoring should also address #1189, but I'm wondering if ...

the `self.location.at_object_receive(self, self.location)`  (trigger the location's reception hook.) should occur conditionally only when the first session puppets the character and triggers the object out of storage, not when successive sessions join, as the object technically does not enter the room in the latter case.

BlauFeuer:
After giving some time for comment and thought, I decided that...

a second session shouldn't move the character again, and triggering at_object_receive only when room actually receives character and accurately reporting that character came from None seems most reasonable, then at_object_receive can decide if appearance of character from None location is actionable.

Griatch:
Seems reasonable. Merged with a rebase.

PR #1189, opened 2017-02-03
Fix for multisession mode 1 switching locations on login

TehomCD:
In `MULTISESSION_MODE = 1`, logging in another session of the character would switch their location. Prevented that with the check.

Prevent snapping back to `_prelogout_location` on login of another session.

N/A

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -1494,7 +1494,7 @@ class DefaultCharacter(DefaultObject):
             session (Session): Session controlling the connection.
 
         """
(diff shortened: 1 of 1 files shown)


BlauFeuer:
Does #1186 fix this, already? It was one of the annoying behaviors I was trying to change.

TehomCD:
Oh whoops, sorry! I should have checked if there was a pull request already for this. I turned on `MULTISESSION_MODE` 1 at player request, so when I saw this behavior, I just made the PR without looking.

PR #1193, opened 2017-02-03
Attempt at making bulk deletion clear out handlers

TehomCD:
So right now, while `delete()` works fine for clearing out the handlers if you delete a single object, doing something like `MyTypeclass.objects.filter(db_tags__db_key="soft deleted").delete()` will leave you with orphaned Attributes, because bulk deletion on a queryset doesn't call the individual model's `delete()` method.

Attempt to remove the orphaned Attribute problem when using bulk deletion.

The suggested approach I read was to use pre_delete or post_delete signals. I thought post_delete might not work because the handlers use the through model to find what objects to delete, and those may not exist when post_deletion is called. So I used a suggestion I found on how to attach a signal to the subclasses of an abstract class like `TypedObject`, and a print statement confirms that it connects to the four subclasses of `TypedObject`. However, I learned that proxy classes do not actually process signals connected to the classes they attach to. I finally saw your example of using the metaclass to connect signals, and that did the trick.


Griatch:
This looks like a good addition. You must however supply migrations if you are modifying the meta properties of a model. Use `evennia makemigrations <app>` on the apps in question and make sure to sanitize the migration - I find that the makemigrations system sometimes thinks it should remove all proxy models like DefaultObject, DefaultCharacter etc just because they are not used explicitly in the game folder. 

Let me know if you have issues with setting up the migrations and I can do it.

TehomCD:
It doesn't seem to want to make a migration around adding `app_label = "typeclasses"` to the Meta for tags/attributes - the migration it creates is just around deleting the proxy models like you mentioned. Let me just remove the `app_label` part - I only added that so that I could use a custom `apps.py` class for attaching signals to the parent classes, which wound up not working due to being proxy models anyway.

Griatch:
Looks good, merged with a rebase!

PR #1195, opened 2017-02-04
Add a method to customize the help message of a command

vlegoff:

This pull request adds a method 'get_help' to all commands.  By default, this method simply returns the command's docstring, but one can create a command that still uses 'auto_help', but displays a customizable message.

The 'auto_help' system is great, but it's either on or off.  This pull request offers an alternative: a way to customize the help of commands without rejecting their docstring.

Example: I have a set of commands with sub-commands (separated by spaces), like "room" (master command), "room list" (see the list of rooms in the area), "room name" (change the name of the room) and so on.  When one enters "help room", he/she should see the other sub-commands available.  Updating the "room" docstring could increase the potential for errors.  But the help system could auto-generate help for this command (and sub-commands) very easily.

### DIFF
--- a/evennia/commands/command.py
+++ b/evennia/commands/command.py
@@ -127,7 +127,11 @@ class Command(with_metaclass(CommandMeta, object)):
 
     (Note that if auto_help is on, this initial string is also used by the
     system to create the help entry for the command, so it's a good idea to
-    format it similar to this one)
+    format it similar to this one).  This behavior can be changed by
+    overriding the method 'get_help' of a command: by default, this
+    method returns cmd.__doc__ (that is, this very docstring, or
+    the docstring of your command).  You can, however, extend or
+    replace this without disabling auto_help.
     """
 
     # the main way to call this command (e.g. 'look')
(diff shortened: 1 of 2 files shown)


evennia/commands/command.py
```diff
@@ -423,3 +427,22 @@ def get_extra_info(self, caller, **kwargs):
...
+    def get_help(self, caller, cmdset):
+        """
+        Return the help message for this command and this caller.
+
+        By default, return self.__doc__ (the docstring just under
+        the class definition).  You can override this behavior,
+        though, and even customize it depending on the caller.
+
+        Args:
+            self: the command itself.
```
Griatch: See the Evennia style guide [here](https://github.com/evennia/evennia/blob/master/CODING_STYLE.md#function--method-docstrings). 
 - `self` should not be included as an argument
 - All arguments should list its type within parentheses. it comes to Evennia entities, I usually use `Object` rather than `ObjectDB` for clarity, so for example `caller (Object or Player)`. The `cmdset` should have `CmdSet` as type. 
 - Since the return is not implicit, the `Returns` block should be listed, something like 
    ```
    Returns:
        docstring (str): The help text to return for this command. 
    ```

Griatch:
This looks like a good idea. See my note about the Evennia policy on docstring syntax.

Griatch:
Merged with a rebase.

PR #1197, opened 2017-02-05
Add the /edit switch to the @help command

vlegoff:

This just adds a /edit switch to the @help command, allowing to edit the text of the help entry in an EvEditor.

It's easier for some users to manage long help entries within an editor, rather than handling it through the other switches.

### DIFF
--- a/evennia/commands/default/help.py
+++ b/evennia/commands/default/help.py
@@ -12,6 +12,7 @@ from evennia.utils.utils import fill, dedent
 from evennia.commands.command import Command
 from evennia.help.models import HelpEntry
 from evennia.utils import create, evmore
+from evennia.utils.eveditor import EvEditor
 from evennia.utils.utils import string_suggestions, class_from_module
 
 COMMAND_DEFAULT_CLASS = class_from_module(settings.COMMAND_DEFAULT_CLASS)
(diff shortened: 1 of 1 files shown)


evennia/commands/default/help.py
```diff
@@ -228,14 +229,35 @@ def func(self):
         self.msg(self.format_help_entry("", "No help entry found for '%s'" % query, None, suggested=suggestions))
 
 
+def loadhelp(caller):
```
Griatch: These should probably be made into private functions (starting with `_`); then they also don't need the full docstrings (which would not help readability in this case).

Griatch:
I like the idea but this does not seem to work for me. Entering `@sethelp/edit foo` or `@sethelp/edit foo = testtext` both tell me that topic "foo" could not be found.

vlegoff:
@Griatch: perfectly true, sorry.  The switch works when the topic exists, to edit its content, but not to create it.  I will update the code.

Griatch:
Thanks! I made only some small further usability changes, like allowing for `@sethelp/edit topicname` without having to give the the right-hand-side text as input (and appending it to the buffer if you do). Merged with a rebase.

PR #1198, opened 2017-02-05
Fix a bug when @reload-ing while in EvEditor

vlegoff:

When reloading Evennia while editing, the buffer is restored as saved, it is not saved by default and will fail if asked to be saved without any additional modification.

The bug could be reproduced as follow:

1. Open a text editor (like @desc/edit).
2. Add some changes, like a line of text.
3. Reload Evennia (this may be done by another character).
4. Try to quit the editor.  It works and doesn't save the modifications before the @reload.

There is now an additional "_eveditor_unsaved" flag which is stored on the caller to remember whether the text was modified before @reload-ing or not.


Griatch:
Merged with a rebase, thanks for the PR!

PR #1201, opened 2017-02-08
Changes to Bot handling to allow custom Bot classes.

CloudKeeper1:
- Messages are no longer dressed by the portal side bot. Instead text and all relevant kwargs are passed through inputfunc through to the server side bot for handling.
-The @irc2chan command now accepts an optional class path as an argument to use as the bot class.

- Opening Evennia up to custom bots without having to change the Evennia folder.


Griatch:
Merged with a rebase. Thanks for your work on this. I made some minor changes to the irc2chan command to match syntax in the rest of Evennia, but otherwise, seems to work very well!

PR #1202, opened 2017-02-09
Cache non-existent attributes too to avoid extra queries

titeuf87:
When trying to retrieve an attribute (using `MyObject.db.attrname` and that attribute does not exist then a query gets sent to the database to try to retrieve it.
If that attribute is also not present in the database None gets returned as expected. But the next time that that attribute gets called again, instead of immediately returning None, a new query fires.

This PR changes that behavior to cache non-existent attributes too so that only one query gets fired.

This can be a big speed win in some scenarios. In my case, I draw a map of a few rooms. Those rooms can have an optional attribute set to change how they look like. If that attribute is not set, then a default rendering is used. Because maps can get large, and most rooms don't have that optional attribute set, without that extra caching, it would create a lot of queries, causing a noticeable delay.

It is possible to fix that speed issue without this change, by making sure that every room has that attribute set, but I don't feel that works as nicely: having an attribute explicitly set to None or not having that attribute at all would cause different behavior (one doesn't query, the other does) while both look the same.
`_quell` will also benefit from this: this attributes gets checked in `is_superuser` so that check always causes a query to fire for an unquelled player. There are probably other similar cases but that is one I saw.

### DIFF
--- a/evennia/typeclasses/attributes.py
+++ b/evennia/typeclasses/attributes.py
@@ -222,7 +222,7 @@ class AttributeHandler(object):
                                        attr.db_category.lower() if attr.db_category else None),
                             attr) for attr in attrs)
         self._cache_complete = True
-
+    
     def _getcache(self, key=None, category=None):
         """
         Retrieve from cache or database (always caches)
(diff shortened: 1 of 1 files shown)


Griatch:
Clever! Merged with a rebase.

PR #1208, opened 2017-02-10
Allow user-defined extensions of ansi mappings in ansi parser

TehomCD:
The mention of the deprecation of the '{' format of ansi codes made me think that a painless way to handle this might just be to have a setting in which a game can set as many additional mappings as they like, so they can continue to support any mappings that are deprecated/removed without the need for pre-parsing in a msg override or the like. This change just has ansi.py look in settings for a list of additional mappings to add to itself, but otherwise will be unchanged.

Allow for easy customization of the ansi parser's mappings with a user-defined list in `settings.py`.

An alternate implementation that seems trickier but is a little more consistent with the rest of the settings would be to have a text pathname to an ansi-parser module, with the default being the current. I personally felt that aside from the mappings, there wouldn't be much demand to change the actual behavior of the ansi parser itself, so that degree of customization wouldn't be desired.

### DIFF
--- a/evennia/settings_default.py
+++ b/evennia/settings_default.py
@@ -319,7 +319,10 @@ PROTOTYPE_MODULES = ["world.prototypes"]
 # Module holding settings/actions for the dummyrunner program (see the
 # dummyrunner for more information)
 DUMMYRUNNER_SETTINGS_MODULE = "evennia.server.profiling.dummyrunner_settings"
-
+# Additional mappings used to extend the existing Evennia ANSI mappings.
+# All mappings should be a tuple of a string to one of the constants provided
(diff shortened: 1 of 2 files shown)


BlauFeuer:
I don't think `{` markup interpretation is going away any time soon, but your ANSI mapping seems to cover the end game, so I'm proposing #1212 to move toward having no depreciated markup in code strings, and a transition path that's smooth and deliberate. Tossing it @Griatch to see if it's the path he wants to take.

Griatch:
Merged with a rebase and some renaming. @BlauFeuer: I'll comment on #1212 separately.

PR #1209, opened 2017-02-10
EvEditor with code handling

vlegoff:

This PR adds a `code` keyword to the `EvEditor` constructor, to activate a few options to handle code.

I would need review and help on this PR, there's still something that doesn't work (see below).

#### What already works

One can now create an `EvEditor` with the `code` keyword argument set to `True`.  This adds a few automatic options in the editor.  By default, indentation is handled automatically.  If you write a statement beginning by `if` or `try` for instance, the next line will be indented.  Keywords like `else` or `elif` or `except` will look for matching keywords and close indentation.  The system isn't perfect to this day, but it's not that bad, from my tests.

In code mode, the options `:-`, `:+`, and `:=` are active.  The first one allows to decrease indentation by one level.  The second one increases it (it shouldn't be necessary to increase it, but you never know).  The third one disables automatic indentation, so that you can enter "normal" code with spaces or tabs at the beginning.

#### Work-in-progress

Once the manual indentation is off, the user is expected to enter indentation manually.  This is useful, for instance, if the user wants to paste code into the editor.  Leading spaces are automatically deleted from the command, however, and it seems it's an automatic Evennia feature that I haven't been able to work around.  Thus, manual indenting for the time being cannot be used.

### DIFF
--- a/evennia/utils/eveditor.py
+++ b/evennia/utils/eveditor.py
@@ -376,18 +376,15 @@ class CmdLineInput(CmdEditorBase):
 
         # add a line of text to buffer
         line = self.raw_string
-        print "Entered", repr(line)
         if not editor._code:
             if not buf:
-                buf = self.args
+                buf = line
             else:
-                buf = buf + "\n%s" % self.args
+                buf = buf + "\n%s" % line
         else:
             # if automatic indentation is active, add spaces
             if editor._indent >= 0:
                 line = editor.deduce_indent(line, buf)
-            else:
-                line = self.args
 
             if not buf:
                 buf = line
--- a/evennia/utils/eveditor.py
+++ b/evennia/utils/eveditor.py
@@ -375,7 +375,7 @@ class CmdLineInput(CmdEditorBase):
         buf = editor.get_buffer()
 
         # add a line of text to buffer
-        line = self.raw_string
+        line = self.raw_string.strip("\r\n")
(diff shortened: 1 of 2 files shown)


vlegoff:
The `@py/edit` switch works well, it seems.  The bug regarding leading spaces in the `EvEditor` was due to the fact that `self.args` was being used, while `self.raw_string` should have been while adding new lines.  I've switched to this behavior even when `EvEditor` is not in code mode.

`@py/edit` opens a code `EvEditor` where one can enter one or several lines of code.  If quitting with `:wq`, the input is executed in `exec` mode with the same variables as the `@py` command.  Automatic and manual indentation seems to work well.

evennia/utils/eveditor.py
```diff
@@ -602,6 +652,41 @@ def func(self):
             # set echoing on/off
             editor._echo_mode = not editor._echo_mode
             caller.msg("Echo mode set to %s" % editor._echo_mode)
+        elif cmd == ":-":
```
Griatch: Since the EvEditor mimcs VIM, the increase indent should be `:>` and decrease indent `:<`. The `:=` is just fine though.

Griatch:
Merged with a rebase. I did some further tweaks to the code editor to make it a little more flexible for possible future code uses, as outlined in 7453cf3.

Issue #1176, opened 2017-01-25
Task: Reduce number of alerts found with LGTM Static Analysis Tool


robbintt:
This tool might be helpful for starting to secure and harden Evennia.

https://lgtm.com/projects/g/evennia/evennia/alerts/

Griatch:
Thanks for the link, looks interesting. I rephrased this issue into something actionable and made it a long-term task.

Griatch:
While this task is essentially endless, I think us bringing down the number of alerts from hundreds and hundreds to a mere 20, most of which might be invalid as alerts, seems like a good spot for considering this task complete.

robbintt:
Cool!!  I'm glad this helped.

PR #1211, opened 2017-02-10
Target LGTM's "String concatenation in loop alert" in building.py

BlauFeuer:
Just a refactor:
Collect string concatenations in a result list for output of command summary message to user.
Also removes a redundant string strip, removing one line of code.

For what it's worth, optimizing string concatenations in python loops by collecting them in a list is the recommended way; though probably doesn't amount to much difference, it should remove this particular topical alert in building.py.

Attempts to help with #1176


Griatch:
Thanks for the fixes! Please make more informative commit messages than "Update building.py" in the future though, that's not useful when looking back to the commit history. Merged with a rebase.

PR #1213, opened 2017-02-12
Add a setting to handle absolute game time (assuing a standard calendar)

vlegoff:

The goal of this pull request is to add a way to handle absolute game time.  By "absolute", I mean Evennia is aware of the current game time, not just the number of virtual seconds.  The information is displayed in the `@time` command.

The current `gametime` module handles relative virtual game: while being aware of the number of seconds since the time was last reset, it doesn't handle absolute game time.  This is definitely open for debate, as the functions I have added may not be useful, for instance, if one wants to use a specific calendar.  However, it provides some basic settings and handles all calculations in seconds, which means even a custom calendar could easily be applied on top of the absolute time system.

#### This PR adds

- A function `virtual_epoch` in the `gametime` module, to return the virtual number of seconds since the MUD started.
- A function `abs_gametime` in the `gametime` module, to return the absolute number of game seconds.
- A setting `TIME_VIRTUAL_START` to set the virtual epoch (timestamp).
- A few new lines in the `@time` command to display this information.

### DIFF
--- a/evennia/commands/default/system.py
+++ b/evennia/commands/default/system.py
@@ -601,11 +601,17 @@ class CmdTime(COMMAND_DEFAULT_CLASS):
 
     def func(self):
         "Show server time data in a table."
+        virtual_epoch = datetime.datetime.fromtimestamp(
+                gametime.virtual_epoch())
+        virtual_current = datetime.datetime.fromtimestamp(
+                gametime.abs_gametime())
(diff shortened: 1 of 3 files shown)


Griatch:
Merged with a rebase. This was a good incentive for me to go back and review the gametime module in general. I ended up reworking it based from your input and also to break out some of its game-specific bits into a contrib.

PR #1217, opened 2017-02-13
Add support for gametime scheduling

vlegoff:
#### Short description

In the `convert_gametime` contrib are now functions/scripts to handle gametime scheduling.  The system is not working perfectly because I don't see how to change script delays at runtime.  If you can give me a hint here that would be great!

The `time.delay` function or more complicated scripts don't interact very easily with user-defined gametime.  Since not all users work with a custom calendar, this new script is kept in the dedicated contrib.

The most useful function for what I have added in this module is `schedule()`.  It takes a game time (as units) and a callback.  It works like this:

    schedule(func, repeat=False, min=30, sec=0)
    # This will call func each hour at x:30:00 (gametime).

It works with custom calendars having defined other units as well.  The "every" unit is taken from the higher, not defined unit.  For instance, should you call `schedule` with `hour=3, min=0, sec=0``, the higher, not-mentioned unit is day.  Therefore, if you set `repeat` to `True`, the callback will be executed everyday at 3:00:00.  By default, `repeat` is set to `False`, so the callback is scheduled next time (using the same context, if it's already more than 3:00, it will pick the following day).

#### An error persists

More tests ought to be done, along with some unittest whenever I'll have the time to write them.  The biggest problem for the time being is that `interval` in scripts can't change at runtime (at least, it seems so).  My script created by `schedule()` checks the next gametime, in case of a reload or shutdown that would push it back and forth a few seconds each time.  It changes its `interval` in `at_repeat`, for instance, but this modification isn't acknowledged by the underlying loop.

To simplify testing, I've kept the `dummy()` callback function.  This will simply send a message to all rooms.  You can test the system like this:

    @py/edit
    from evennia.contrib.convert_gametime import *
    self.msg(str(schedule(dummy, repeat=True, sec=30)))

What this should do is set the `dummy()` callback for execution every minute at XX:30.  It works the first time.  But the next executions depend on the original intervals, and they are not updated.  If you call this while it's XX:25 in game, for instance, the script will run every 5 seconds (or less, due to TIME_FACTOR).  I'm not sure how to fix this... but the PR cannot be merged as is, I think.  If you can redirect me on the right method, I'll try to update it quickly.

Thanks!

### DIFF
--- a/evennia/contrib/convert_gametime.py
+++ b/evennia/contrib/convert_gametime.py
@@ -28,23 +28,23 @@ TIMEFACTOR = settings.TIME_FACTOR
 # when defining in-game events. The words month, week and year can  be
 # used to mean whatever units of time are used in your game.
 SEC = 1
-MIN = 60          # seconds per minute
-HOUR = MIN * 60   # minutes per hour
-DAY = HOUR * 24   # hours per day
-WEEK = DAY * 7    # days per week
-MONTH = WEEK * 4  # weeks per month
-YEAR = MONTH * 12 # months per year
-UNITS = {
-    "sec": SEC,
-    "min": MIN,
-    "hr": HOUR,
-    "hour": HOUR,
-    "day": DAY,
-    "week": WEEK,
-    "month": MONTH,
-    "year": YEAR,
-    "yr": YEAR,
-}
+MIN = getattr(settings, "SECS_PER_MIN", 60)
+HOUR = getattr(settings, "MINS_PER_HOUR", 60) * MIN
+DAY = getattr(settings, "HOURS_PER_DAY", 24) * HOUR
+WEEK = getattr(settings, "DAYS_PER_WEEK", 7) * DAY
+MONTH = getattr(settings, "WEEKS_PER_MONTH", 4) * WEEK
+YEAR = getattr(settings, "MONTHS_PER_YEAR", 12) * MONTH
+UNITS = getattr(settings, "TIME_UNITS", {
+        "sec": SEC,
+        "min": MIN,
+        "hr": HOUR,
+        "hour": HOUR,
+        "day": DAY,
+        "week": WEEK,
+        "month": MONTH,
+        "year": YEAR,
+        "yr": YEAR,
+})
 
 
 def time_to_tuple(seconds, *divisors):
--- a/evennia/contrib/convert_gametime.py
+++ b/evennia/contrib/convert_gametime.py
@@ -23,17 +23,28 @@ from django.conf import settings
 # The game time speedup  / slowdown relative real time
 TIMEFACTOR = settings.TIME_FACTOR
 
-# Game-time units, in real-life seconds. These are supplied as a
+# Game-time units, in game time seconds. These are supplied as a
 # convenient measure for determining the current in-game time, e.g.
 # when defining in-game events. The words month, week and year can  be
 # used to mean whatever units of time are used in your game.
-
+SEC = 1
 MIN = 60          # seconds per minute
 HOUR = MIN * 60   # minutes per hour
 DAY = HOUR * 24   # hours per day
 WEEK = DAY * 7    # days per week
 MONTH = WEEK * 4  # weeks per month
 YEAR = MONTH * 12 # months per year
+UNITS = {
+    "sec": SEC,
+    "min": MIN,
+    "hr": HOUR,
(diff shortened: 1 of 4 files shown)


Griatch:
You can modify the interval at runtime, but no, you cannot do so by modifying the `interval`. What you do is call `Script.restart`: ´Script.interval(interval=None, repeats=None, start_delay=None)` which will stop the running script (calling the `at_stop` hook) and then start it again with the new settings (calling the `at_start` hook). One can modify the callcount with `Script.reset_callcount(value=0)` (maybe not so relevant in this case) and use `Script.force_repeat()` to cut short a delay.

vlegoff:
@Griatch: thanks for your answer.  I'm using `Script.restart()` now and it works well, when not reloading or restarting the server.  When the server is reloaded, if this script is running, it seems something odd occurs:  since the script needs to restart on startup (to update its interval), I get several scripts in the pool (well, there's still only one script, but the number of messages per minute increases, one of them isn't timed properly).

What I have done follows this process:

1. When the script starts, it checks whether it needs to be "reset", that is, whether its interval must be updated.  This is not the case by default, so it does nothing.
2. Repeat is as before, except that the script restarts each time to make sure it follows the right schedule.  This might be updated in the future if it's too much on performances.
3. If the server reloads or shutdown, the script is updated as needing a reset.  It means, next time the `Script.start` method is called, the script will restart with a new, updated interval.  It is necessary, because after a reload, the game time always shifts a bit, and it will be worse for shutdown of course.

So my current problem is that, if you launch the script with the `dummy()` function (see above), everything works perfectly.  The message is displayed everywhere, every game minute at X:30.  But when using `@reload`, the number of messages per minute increases.  It's hard to tell how much, and it seems to be related to the "old" script before restarting.  I'm not sure about the cause of the problem, so I can't even think of a solution.  Perhaps pausing the script before reloading/shutting down, if there's a way to unpause it (or restarting it) properly afterward.  One more time, if you can help me figure things out, that would be great.

Thanks,

vlegoff:
Error fixed!  It came from the call to `script.restart` in `at_start`.  That didn't exactly please Evennia.  As far as I could understand, the reason was that `restart` would create a new task to handle the new interval.  When restarting, the "old" task is usually deleted, but when reloading the game, it seems that the `at_start` method is called before the task is created.  In short, when the server reloads (or restarts), if there's a `restart` in the `at_start` method, two tasks are created with different delays to handle the same script, which creates some hard-to-debug situations.

The solution for me was to add a check in `script._start_task`: the task is only created if another task isn't already there.  It might not be the best of behaviors:  it does work for my use case, but I don't know if it breaks compatibility for others.  I don't think it would, but I don't know the code as much as you do.

If this fix seems good to you, I think the PR can be merged (that is, if you don't see anything else).  The `convert_gametime` contrib is stable.  I think I should write the `schedul` function and a script to handle standard calendar, for the behavior is slightly different, but I'm not yet positive.

vlegoff:
The system finally works!  I have undone the correction on scripts, and the custom script is much lighter in behavior and code.  It still keeps on restarting at repeat, though, that's a necessity to avoid drifting (the first repeat after a `@reload` or `@shutdown` might be off by one or two seconds).  The difference is obviously more perceptible if the game time rapidly spins, and if a long-time schedule has been declared, but overall I believe the difference to be negligible.

I have fixed the bug that created a drift in gametime when the game was reloaded or restarted.  I added a similar interface in `utils.gametime` and `contrib.convert_gametime`.  Both have a `schedule` function, for instance.  The former (in `utils.gametime`) handles tasks with a standard calendar (it therefore uses `datetime`, `timedelta` and `calendar`).  The latter (in `contrib.convert_gametime`) uses a custom calendar that the user can set up, down to different units.

If you want to read the small tutorial that I destine to users, you can find it here: https://github.com/evennia/evennia/wiki/Gametime-Tutorial .  Obviously, for the time being, it isn't connected to the tutorial page, it will be when you find the code can be merged into the master branch.  The choice to handle standard/custom calendar separately may not be aligned with what you had imagined:  I believe it is necessary, but you'll see what sounds preferable.

Griatch:
Thanks! Merged with a rebase. Note that I did some changes to the name of the module and the settings variables, so please tweak the wiki page accordingly before linking it to the tutorial index (further fixes to the wiki page can happen in the context of that page). Closing.

PR #1222, opened 2017-02-18
nicks cmd fix

TehomCD:
The recent change to have None matches be cached broke the nicks/del command when executed on nicks that did not exist: the `has` check would be executed on a None key and return all cached results by category, which would be a list of None results and resolve to true.

Add check for nick existing.

Minor bug fix to nick command.

N/A


Griatch:
Merged with a rebase, thanks!

PR #1226, opened 2017-02-19
Another small fix to nickname command

TehomCD:
Last fix didn't seem to over all the cases where the nick command would return a traceback. Put in another check to fix another case where it seemed an attribute error was being caused when displaying nicknames.

Small check to prevent an attribute error being occasionally raised in nicknames.

N/A

### DIFF
--- a/evennia/commands/default/general.py
+++ b/evennia/commands/default/general.py
@@ -123,6 +123,7 @@ class CmdNick(COMMAND_DEFAULT_CLASS):
         nicktypes = [switch for switch in switches if switch in ("object", "player", "inputline")] or ["inputline"]
 
         nicklist = utils.make_iter(caller.nicks.get(return_obj=True) or [])
+        nicklist = [nick for nick in nicklist if hasattr(nick, 'value')]
 
         if 'list' in switches or self.cmdstring in ("nicks", "@nicks"):
 


Griatch:
This fix makes no sense, at least it tries to fix a symptom rather than a cause. Why would we get nicks (=Attributes) without a value? If the problem is that we are returning `None`-objects as nicks, this is something that must be fixed in the nickhandler.

TehomCD:
I think this is because it has cached None values, and it's trying to access the None values as if they were Attributes. Or something like that.

Griatch:
Hm, the original change to the nickhandler does indeed cache non-matches. But that's a optimization to lower database queries. Does it really make sense to actually return those None matches later?

TehomCD:
Hmm. Maybe omit them when we're doing a category match? That'd probably match the previous non-cached behavior, and I don't think we really want to see fake positives for a category if we're trying to retrieve all the Attributes a player has for a given category.

Griatch:
I pushed a change that makes it so that even though the non-found Attribute is cached as key:None, that None value will never be returned by the cache. Without this correction it would seem the cached return and the full query return return different values which is a big no-no. Please check if this resolves your issue so we can close this.

TehomCD:
Your fix seems to work. Thanks!

Griatch:
Ok, good. :)

Issue #1212, opened 2017-02-11
Task:  Convert deprecated {- markup in code/contribs/tutorials toward full deprecation.


BlauFeuer:
#### Brief summary:
Use of `{` in markup (or rather, Evennia responding to it) causes unintentional markup when someone is using the braces to demonstrate code or for other reasons.

An ongoing task should be to update markup in modules that are edited or receive PRs, and eventually to any remaining depreciated markup in isolated modules.

#### Reasons for making a task of converting:
Use of depreciated markup in code/contribs/tutorials perpetuates further use, which slows the migration to the current markup.   A task would encourage contributors to update markup in submitted code, and remind me to update modules as I have the time.

#### Error output / Expected result of feature
At some future point, Evennia should stop interpreting `{` as markup by default.

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
Perhaps first step should be use of a setting to opt out of responding to depreciated markup by 0.7.0 (Ability with setting to choose to turn off responding to `{`), then off by default in 0.8.0 (Ability with setting to choose to turn it on), and finally only by user-defined extension by 0.9.0. This progression would be slow and deliberate.

Griatch:
Actually I would suggest a (potentially) faster process: Once the core Evennia is stripped of the use of deprecated `{`-type color tags, we leverage Tehom's color-plugin system and move the `{`- type color codes into a contrib. From there people can just plug it back in if they want it. We could even add `%c` type tags in the same contrib for those that want that.

BlauFeuer:
#1228 should be the last of the old markup changeover.  IRC bot still has old markup in `IRC_COLOR_MAP' dictionary, but Tehom's color-plugin system should handle allowing alternate markup after full deprecation occurs, which potentially could take place in the 0.7.0 dev

Griatch:
@BlauFeuer Ok, excellent! I'm closing this and opening a new task for the moving of the deprecated markup into a contrib for use with @TehomCD 's new color settings (that one will go into devel).

PR #1230, opened 2017-02-21
Markup, LGTM, PEP 8 updates (To complete #1212)

BlauFeuer:
Markup and PEP 8 updates to `commands/default/batchprocess.py`

Markup, PEP 8 and ~~convert Prettytable to EvTable usage in `@bans` helper function of `commands/default/admin.py` and various tables in `commands/default/player.py`~~

LGTM alert fixes and PEP 8 whitespace updates in `players/players.py`

Per #1212, #1176,  ~~[Remove prettytable from evennia/utils](https://github.com/evennia/evennia/projects/6) directive for 0.7 devel branch~~ and generally keeping to PEP 8 style

The first two files were in my local repository, but did not get committed with the others before closing #1212.

~~**As a result of editing two of the files submitted for pull:
_No more import of prettytable was found, so there should be no more usage of prettytable in current code._**~~

### DIFF
--- a/evennia/commands/default/admin.py
+++ b/evennia/commands/default/admin.py
@@ -9,7 +9,7 @@ import re
 from django.conf import settings
 from evennia.server.sessionhandler import SESSIONS
 from evennia.server.models import ServerConfig
-from evennia.utils import prettytable, search, class_from_module
+from evennia.utils import evtable, search, class_from_module
 
 COMMAND_DEFAULT_CLASS = class_from_module(settings.COMMAND_DEFAULT_CLASS)
 
@@ -40,7 +40,7 @@ class CmdBoot(COMMAND_DEFAULT_CLASS):
     help_category = "Admin"
 
     def func(self):
-        "Implementing the function"
+        """Implementing the function"""
         caller = self.caller
         args = self.args
 
@@ -86,7 +86,7 @@ class CmdBoot(COMMAND_DEFAULT_CLASS):
         # Carry out the booting of the sessions in the boot list.
 
         feedback = None
-        if not 'quiet' in self.switches:
+        if 'quiet' not in self.switches:
             feedback = "You have been disconnected by %s.\n" % caller.name
             if reason:
                 feedback += "\nReason given: %s" % reason
@@ -108,13 +108,12 @@ def list_bans(banlist):
     if not banlist:
         return "No active bans were found."
 
-    table = prettytable.PrettyTable(["{wid", "{wname/ip", "{wdate", "{wreason"])
+    table = evtable.EvTable("|wid", "|wname/ip", "|wdate", "|wreason")
     for inum, ban in enumerate(banlist):
-        table.add_row([str(inum + 1),
-                       ban[0] and ban[0] or ban[1],
-                       ban[3], ban[4]])
-    string = "{wActive bans:{n\n%s" % table
-    return string
+        table.add_row(str(inum + 1),
+                      ban[0] and ban[0] or ban[1],
+                      ban[3], ban[4])
+    return "|wActive bans:|n\n%s" % table
 
 
 class CmdBan(COMMAND_DEFAULT_CLASS):
@@ -174,7 +173,7 @@ class CmdBan(COMMAND_DEFAULT_CLASS):
 
         if not self.args or (self.switches
                              and not any(switch in ('ip', 'name')
-                             for switch in self.switches)):
+                                         for switch in self.switches)):
             self.caller.msg(list_bans(banlist))
             return
 
@@ -202,7 +201,7 @@ class CmdBan(COMMAND_DEFAULT_CLASS):
         # save updated banlist
         banlist.append(bantup)
         ServerConfig.objects.conf('server_bans', banlist)
-        self.caller.msg("%s-Ban {w%s{n was added." % (typ, ban))
+        self.caller.msg("%s-Ban |w%s|n was added." % (typ, ban))
 
 
 class CmdUnban(COMMAND_DEFAULT_CLASS):
@@ -223,7 +222,7 @@ class CmdUnban(COMMAND_DEFAULT_CLASS):
     help_category = "Admin"
 
     def func(self):
-        "Implement unbanning"
+        """Implement unbanning"""
 
         banlist = ServerConfig.objects.conf('server_bans')
 
@@ -240,14 +239,14 @@ class CmdUnban(COMMAND_DEFAULT_CLASS):
         if not banlist:
             self.caller.msg("There are no bans to clear.")
         elif not (0 < num < len(banlist) + 1):
-            self.caller.msg("Ban id {w%s{x was not found." % self.args)
+            self.caller.msg("Ban id |w%s|x was not found." % self.args)
         else:
             # all is ok, clear ban
             ban = banlist[num - 1]
             del banlist[num - 1]
             ServerConfig.objects.conf('server_bans', banlist)
             self.caller.msg("Cleared ban %s: %s" %
-                                    (num, " ".join([s for s in ban[:2]])))
+                            (num, " ".join([s for s in ban[:2]])))
 
 
 class CmdDelPlayer(COMMAND_DEFAULT_CLASS):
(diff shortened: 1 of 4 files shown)


BlauFeuer:
@Griatch ... the test coverage decreased, perhaps from removing the prettytable import which may have been under test.  Sorry, now we're below 50% again.

BlauFeuer:
@Griatch ... the test coverage decreased, perhaps from removing the prettytable import which may have been under test.  Sorry, now we're below 50% again.

TehomCD:
One very slight potential issue with the removal of PrettyTable and converting to EvTable exclusively - while EvTable is definitely far superior for readability for almost all situations when the table size is small, it runs noticeably slower if you have a fairly large table. If a table has, say, a hundred rows, there's actually a perceptible delay for building it, while PrettyTable runs about 5 times as fast or so. I actually reverted to using PrettyTable in a few instances whenever the number of rows for a table was over 20 to avoid complaints of it feeling unresponsive. Keeping it as a contrib would certainly be fine for it as an edge case, I'd just caution against removing it entirely.

Griatch:
@TehomCD Are these tables with color tags in them? Prettytable generally cannot handle ansi code markers as I recall. Do you have an example of code you use for building a (slow) EvTable?

TehomCD:
Here's a sample that I quickly converted from PrettyTable to EvTable for testing:
```
    def list_actions(self):
        from server.utils.prettytable import PrettyTable
        if "old" in self.switches:
            qs = CrisisAction.objects.filter(sent=True)
        else:
            qs = CrisisAction.objects.filter(sent=False)
        if "needgm" in self.switches:
            qs = qs.filter(story__exact="")
        if "listquestions" in self.switches:
            qs = qs.filter(questions__isnull=False, questions__answers__isnull=True)
        if self.args:
            qs = qs.filter(Q(crisis__name__iexact=self.args) |
                           Q(dompc__player__username__iexact=self.args))
        # table = PrettyTable(["{w#{n", "{wCrisis{n", "{wPlayer{n", "{wAnswered{n", "{wQuestions{n", "{wDate Set{n"])
        # for ob in qs:
        #     date = "--" if not ob.crisis.end_date else ob.crisis.end_date.strftime("%x")
        #     questions = "{rYes{n" if ob.questions.filter(answers__isnull=True) else "{wNo{n"
        #     table.add_row([ob.id, ob.crisis.name, str(ob.dompc), "{wYes{n" if ob.story else "{rNo{n", questions, date])
        table = EvTable("{w#{n", "{wCrisis{n", "{wPlayer{n", "{wAnswered{n", "{wQuestions{n", "{wDate Set{n")
        for ob in qs:
            date = "--" if not ob.crisis.end_date else ob.crisis.end_date.strftime("%x")
            questions = "{rYes{n" if ob.questions.filter(answers__isnull=True) else "{wNo{n"
            table.add_row(ob.id, ob.crisis.name, str(ob.dompc), "{wYes{n" if ob.story else "{rNo{n", questions, date)
        self.msg(table)
```
For 80 rows, when it's PrettyTable, it executes in 350 ms or so. When it's EvTable, it's 2200ish ms or so. If the rows have cells that need to wrap, such as if I made date's `strftime("%x %M:%D")`, for example, it'd increase the time proportionally based on the number of additional rows it adds, it seems. Say to around 2800ms or so. Conversely, while it breaks PrettyTable's display and makes it look awful, there's no meaningful difference in performance for row size.

I do think this is probably an edge case because it only happens if you're dealing with very large tables, but it might possibly come up if people use `EvMore` to limit the display, since they'd still incur the cost of building the table that is passed to EvMore the first time they generate it. I only noticed it at all because one of the tables returned by a staff command was monstrous, something like 300 rows, and it was a noticeable delay - I was surprised to see that it was EvTable rather than the query that was the bottleneck.

Griatch:
@TehomCD Ok, will check that. The main expense I expect to be the `add_row` operation in that case; EvTable will need to recalculate the row size and possibly adjust the table (EvTable allows adding non-symmetric rows/columns and expect the table to automatically fill in with empty slots, something prettytable doesn't do at all). I'm not sure if this happens for every new addition but I do recall this being the slower mode. I imagine creating the table in one go with EvTable(table=[[...], [...], [...]]) will be a lot faster for this use case.

TehomCD:
I just tried a version with:
```
rows = []
        for ob in qs:
            date = "--" if not ob.crisis.end_date else ob.crisis.end_date.strftime("%x %M:%H")
            questions = "{rYes{n" if ob.questions.filter(answers__isnull=True) else "{wNo{n"
            #table.add_row(ob.id, ob.crisis.name, str(ob.dompc), "{wYes{n" if ob.story else "{rNo{n", questions, date)
            rows.append([ob.id, ob.crisis.name, str(ob.dompc), "{wYes{n" if ob.story else "{rNo{n", questions, date])
        table = EvTable("{w#{n", "{wCrisis{n", "{wPlayer{n", "{wAnswered{n", "{wQuestions{n", "{wDate Set{n",
                        table=rows)
        self.msg(table)
```
And still took around 2500ms. I couldn't get it to let me set width to 78 - if I tried, it returned an error saying:
```
File "c:\arx\evennia\evennia\utils\evtable.py", line 1527, in __str__
    return  str(unicode(ANSIString("\n").join([line for line in self._generate_lines()])))
  File "c:\arx\evennia\evennia\utils\evtable.py", line 1333, in _generate_lines
    self._balance()
  File "c:\arx\evennia\evennia\utils\evtable.py", line 1250, in _balance
    raise Exception("Cannot shrink table width to %s. Minimum size is %s." % (self.width, cwmin))
Exception: Cannot shrink table width to 78. Minimum size is 321.
```
Not sure if I was just doing something wrong there with the formatting of the table or if that's a bug.

Griatch:
@TehomCD I'm pretty sure you must add columns, not rows, this way.

TehomCD:
Ah, okay. I see now. I think I did it properly this time, with
```
        col1 = []
        col2 = []
        col3 = []
        col4 = []
        col5 = []
        col6 = []
        rows = [col1, col2, col3, col4, col5, col6]
        for ob in qs:
            date = "--" if not ob.crisis.end_date else ob.crisis.end_date.strftime("%x %M:%H")
            questions = "{rYes{n" if ob.questions.filter(answers__isnull=True) else "{wNo{n"
            #table.add_row(ob.id, ob.crisis.name, str(ob.dompc), "{wYes{n" if ob.story else "{rNo{n", questions, date)
            #rows.append([ob.id, ob.crisis.name, str(ob.dompc), "{wYes{n" if ob.story else "{rNo{n", questions, date])
            col1.append(ob.id)
            col2.append(ob.crisis.name)
            col3.append(str(ob.dompc))
            col4.append("{wYes{n" if ob.story else "{rNo{n")
            col5.append(questions)
            col6.append(date)
        table = EvTable(table=rows, width=78)
        self.msg(table)
```
and was still around 2700 ms.

Griatch:
@TehomCD 
Ok, could you please write this up as a feature request for speeding up EvTable? I'm not sure we can do all the processing EvTable does any faster, but we migth add a "fixed" mode if you don't need all the dynamic load balancing and accepts that it will print stuff just the way you enter it.

BlauFeuer:
Since the prettytable to EvTable has been shown not to be an optimization at this time, and the original PR was intended to release the final two files to satisfy #1212, I've moved the prettytable->EveTable edits in two module to another PR.

Griatch:
@BlauFeuer I'd say, keep the prettytable removal in. The slowdown @TehomCD is talking about is not in the commands you converted to EvTable necessarily, but in his custom commands using much bigger tables.

BlauFeuer:
@Griatch The `@bans` command in evennia/commands/default/admin.py could reach 100 or more entries, and the who list in evennia/commands/default/player.py regularly does on Arx.  I only tested with 1 or 2 entries on both and didn't visually notice a difference.

If the last two PR commits, ab12965 and f17e42b, are not pulled, then this PR includes the prettytable to EvTable conversion. Without the last two, Coveralls will report test coverage dropping below 50%.

I could submit another PR against devel, keeping the prettytable conversion for 0.7, or it could be rolled in with the feature request to add a less-flexible speed mode to EvTable, or I could commit an undoing of the undoing of the prettytable conversion in this PR...

I'm open to all options.

Griatch:
Merged with a rebase. I chose to keep your removal of prettytable; we'll see if it has an impact or not.

PR #1233, opened 2017-02-23
Fix the `characters` property on players having multiple characters

vlegoff:

This PR is supposed to fix the bug one can get from querying the `characters` property on a player.

The `characters` property was not working, an AttributeError was raised.

This PR seems to fix the issue, although it doesn't seem to be tested right now.

### DIFF
--- a/evennia/players/models.py
+++ b/evennia/players/models.py
@@ -106,10 +106,10 @@ class PlayerDB(TypedObject, AbstractUser):
 
     # alias to the objs property
     def __characters_get(self):
-        return self.objs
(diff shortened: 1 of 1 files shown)


Griatch:
Alas, this is not a correct fix; the `.db._playable_characters` dict is not a core feature of the model, but something only used by the default puppeting commands -- and then only for certain MULTISESSION_MODEs. In fact, the `.characters` property is a deprecated layover from a different time and will be removed. Use ´player.sessions` to get the sessions currently active on the player - from them you can then get the character each puppet.

PR #1237, opened 2017-02-26
BodyFunctions.py - Adjusted suggested use so that it works.

CloudKeeper1:
Previously gave an incomplete path leading to an importerror.


Griatch:
Nice catch. Merged with a rebase.

PR #1240, opened 2017-02-28
Created Whisper command.

CloudKeeper1:
Created a whisper command which sends private messages to objects in the same location.

### DIFF
--- a/evennia/commands/default/general.py
+++ b/evennia/commands/default/general.py
@@ -439,11 +439,15 @@ class CmdWhisper(COMMAND_DEFAULT_CLASS):
             caller.msg("Usage: whisper <player> = <message>")
             return
            
-        receiver = caller.search(self.lhs,
-                                 nofound_string="'%s' cannot be found." % self.lhs)
+        receiver = caller.search(self.lhs)
+
+        if not receiver:
+            caller.msg("Usage: whisper <player> = <message>")
+            return
+           
         if caller == receiver:
-            caller.msg("You can't whisper yourself.")
-           return
+            caller.msg("You can't whisper to yourself.")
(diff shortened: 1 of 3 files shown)


evennia/commands/default/general.py
```diff
@@ -417,6 +417,44 @@ def func(self):
...
+        caller = self.caller
+
+        if not self.lhs or not self.rhs:
+            caller.msg("Usage: whisper <player> = <message>")
+            return
+           
+        receiver = caller.search(self.lhs,
+                                 nofound_string="'%s' cannot be found." % self.lhs)
+        if caller == receiver:
+            caller.msg("You can't whisper yourself.")
```
Griatch: This looks like a wrong indent. Also should it not be "You can't whisper TO yourself" ?

evennia/commands/default/general.py
```diff
@@ -417,6 +417,44 @@ def func(self):
...
+    def func(self):
+        """Run the whisper command"""
+
+        caller = self.caller
+
+        if not self.lhs or not self.rhs:
+            caller.msg("Usage: whisper <player> = <message>")
+            return
+           
+        receiver = caller.search(self.lhs,
```
Griatch: There should be no need for the `nofound_string` here. Also, there doesn't appear to be a catcher if receiver is not found or is a multimatch, at which time `receiver` will be `None` and the function should simply return.

Griatch:
Looking good, apart for some comments. Since this is to go into the default command set, a unittest is also required, in `evennia/commands/default/tests.py`.

Griatch:
Thanks, merged with a rebase.

PR #1246, opened 2017-03-07
Inclusion of type kwarg for common communications

CloudKeeper1:
Including a type entry in the option kwarg for the say, whisper and pose commands allows the receivers of those messages to treat them differently depending on type in code swiftly. IMO This is a wide enough concern that it goes beyond a game specific requirement. The inclusion will not disrupt any existing systems of Evennia.

### DIFF
--- a/evennia/commands/default/general.py
+++ b/evennia/commands/default/general.py
@@ -414,8 +414,8 @@ class CmdSay(COMMAND_DEFAULT_CLASS):
 
         # Build the string to emit to neighbors.
         emit_string = '%s says, "%s|n"' % (caller.name, speech)
-        caller.location.msg_contents(emit_string, exclude=caller,
-                                     from_obj=caller, options={"type": "say"})
+        caller.location.msg_contents(text=(emit_string, {"type": "say"}), 
+                                     exclude=caller, from_obj=caller)
 
 
 class CmdWhisper(COMMAND_DEFAULT_CLASS):
@@ -457,7 +457,7 @@ class CmdWhisper(COMMAND_DEFAULT_CLASS):
 
         # Build the string to emit to receiver.
         emit_string = '%s whispers, "%s|n"' % (caller.name, speech)
-        receiver.msg(emit_string, from_obj=caller, options={"type": "whisper"})
+        receiver.msg(text=(emit_string, {"type": "Whisper"}), from_obj=caller)
 
 
 class CmdPose(COMMAND_DEFAULT_CLASS):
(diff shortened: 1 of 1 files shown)


Griatch:
This is a bigger change than suggested, since it establishes "type" as a new standard keyword for certain commands. The idea is not bad, but putting the type info in the options dict is the wrong place for this info to be made available - the only place where the options dict is parsed is in the Portal sender function. While I realize this has importance for one particular protocol - the IRC bot, it is not generally useful and even wasteful for other protocols. It also sets a dangerous precendent - once you start doing this for say and whisper you may want to do the same for page or +job etc. Adding options this way will require need the user to also make a plugin-modification in the Portal for it to be useful. Not the way outputfuncs are intended to be used. 

So I won't accept this as-is. Rather I suggest the following modification: For this to be relevant to the receiver (such as a client over OOB), this should IMO be a custom kwarg to the "text" outputfunc, so sent as `receiver.msg(text=("mytext", {"type": "whisper"}))`. That way it can still be parsed by the protocol if so needed, but it is also be passed on so the client can do custom stuff with it if it can.

BlauFeuer:
The text is in the tuple. It should be easier to parse. Trying to determine communication type before this change is messy - so I've added this traceback as an example and reference even though it's only partially relevant. http://pastebin.com/Zb4nV0aN

CloudKeeper1:
Made the changes as suggested.

evennia/commands/default/general.py
```diff
@@ -457,7 +457,7 @@ def func(self):
 
         # Build the string to emit to receiver.
         emit_string = '%s whispers, "%s|n"' % (caller.name, speech)
-        receiver.msg(text=(emit_string, {"type": "Whisper"}), from_obj=caller)
+        receiver.msg((emit_string, {"type": "Whisper"}), from_obj=caller)
```
BlauFeuer: Should type be "Whisper" capitalized? That's different from "say" and "pose"

CloudKeeper1:
Where would I be without you Blau.

BlauFeuer:
The msg_contents might need a tweak, too:
```
>>> self.location.msg_contents(('Test message {num}', {'type': 'Test'}),
                                                 from_obj=self, mapping=dict(num='3'))
  File "/home/evennia/evennia/evennia/objects/objects.py", line 596, in msg_contents
    obj.msg(message.format(**substitutions), from_obj=from_obj, **kwargs)
AttributeError: 'tuple' object has no attribute 'format'
```
The short of it is that tuple-texting and mapping kwarg aren't playing well together.

Griatch:
Thanks! Merged with some modifications (such as lowercase variable names). I also extended msg_contents to support the normal tuple-form of the text outputcommand. Closing.

PR #1248, opened 2017-03-12
Correct the IRC reset mapping to use 0x0F (15)

BlauFeuer:

Mapping mIRC's reset character 0x0F to /015 instead of /017, which is incorrect.

Without the correct mapping, Evenna passes along the raw mIRC reset character (0x0F) without properly converting it to Evennia markup.

Using the correct mapping causes expected text line to be terminated with Evenna's reset markup, then correctly converted to other Evennia protocol clients, such as Telnet.

### DIFF
--- a/evennia/server/portal/irc.py
+++ b/evennia/server/portal/irc.py
@@ -18,7 +18,7 @@ from evennia.utils import logger, utils, ansi
 
 IRC_BOLD = "\002"
 IRC_COLOR = "\003"
-IRC_RESET = "\017"
+IRC_RESET = "\015"
 IRC_ITALIC = "\026"
 IRC_NORMAL = "99"
 IRC_UNDERLINE = "37"


Griatch:
As you say, the mIRC  "reset" is `0x0f´, which is really equal to `\017`. So this cannot be "incorrect" as stated. By contrast ´\015` is the `\r` carriage return symbol which is not what mIRC wants. This fix would appear to make the Evennia IRC protocol inconsistent with the definition. 

If there is a problem with converting IRC-colors -> Evennia input I would think the fix need to be in that converter rather than here? I would need to see an example of the problem solved with this I think.

BlauFeuer:
Hmm, Seems like \017 is octal (not decimal, which I first thought), which is 15 in decimal.

A good example of the effect of the problem is seen regularly in Evennia's demo server's public channel where Telnet-based mu* clients are being sent 0x0f from the IRC bots.  If your particular client does not display the character, then it can be seen with a debug mode, like found in Potato client - but the character is being sent untranslated to webclient and Telnet, when it should be translated.

Should I make an issue for this?  I'm sorry for my math in the wrong base.

Griatch:
Might be best to make an issue yes. But if the problem is the little "note" character in Potato, it's unrelated to this, but due to Evennia sending `\r\n` as specified by telnet and Potato not properly understanding it. 
Closing.

PR #1249, opened 2017-03-12
Add a new style 4 for the time_format utility

vlegoff:

This PR adds a style (4) to the `time_format` utility.  It also adds new unittests for this function.

The style 4 is useful to have a short but readable version of a time in seconds.  It only displays the highest unit of time, but do so in a human-readable way.  It remains close to the style 1, although it uses long unit names.

Examples:

```python
time_format(1, 4) # Returns "a second"
time_format(10, 4) # Returns "10 seconds"
time_format(70, 4) # Returns "a minute"
time_format(3600, 4) # Returns "an hour"
# ... it also includes days, months and years
```

This can be handy if you want to display a short version of, say, an updated time: this thing was updated 3 hours ago or 2 days ago or 9 months ago.

This PR also contains new unittests for the `time_format` utility, which wasn't tested before.

It also uncovered some potential bugs in style 3:

- A comma is used as a separator between day and hour, and hour and minute, but not between minute and seconds.
- Sometimes, it seems that "other units 0 seconds" is returned, whereas "0 seconds" should not.

I have left these (potential) errors untouched, waiting to see if there are really bugs.  You'll find the strange outputs in the `test_style_3` method.

### DIFF
--- a/evennia/utils/tests.py
+++ b/evennia/utils/tests.py
@@ -411,3 +411,98 @@ class TestEvForm(TestCase):
         # note that in a msg() call, the result would be the  correct |-----,
         # in a print, ansi only gets called once, so ||----- is the result
         self.assertEqual(unicode(evform.EvForm(form={"FORM":"\n||-----"})), "||-----")
+
+class TestTimeformat(TestCase):
+    """
+    Default function header from utils.py:
+    time_format(seconds, style=0)
+
+    """
+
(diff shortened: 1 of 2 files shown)


Griatch:
@v-legoff Merged with a rebase. Those additional things do look like bugs; might be worth an issue or another PR.

PR #1250, opened 2017-03-13
Add URL for BSD 3-Clause license

BlauFeuer:
Suggested to me by a user, since the URL names the license type, it serves as a direct replacement for the text.

Updates the text in the license field of the `@about` command.

Provides link to resource of license type

### DIFF
--- a/evennia/commands/default/system.py
+++ b/evennia/commands/default/system.py
@@ -613,7 +613,7 @@ class CmdAbout(COMMAND_DEFAULT_CLASS):
          |cEvennia|n %s|n
          MUD/MUX/MU* development system
 
-         |wLicence|n BSD 3-Clause Licence
+         |wLicence|n https://opensource.org/licenses/BSD-3-Clause
          |wWeb|n http://www.evennia.com
          |wIrc|n #evennia on FreeNode
          |wForum|n http://www.evennia.com/discussions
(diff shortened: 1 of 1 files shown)


Griatch:
We can add the link, for sure. But for reference it does not serve as a replacement per se - the LICENSE.txt contains the copyright "publisher" info that should be copied along with Evennia. A simple thing, but that's how the license is formulated.

Griatch:
Merged with a rebase.

BlauFeuer:
I didn't mean to imply the URL was a replacement for the LICENSE.txt, but ...

instead, it serves as a replacement for the representation of it in the ABOUT display, meaning a human reading "BSD 3-Clause Licence"  (where License is misspelled), is nearly identical to a human seeing the URL `https://opensource.org/licenses/BSD-3-Clause`

Since I made this change in my own version of `@about`, I thought I'd offer it here.   If the license literally says that the product must contain "BSD 3-Clause Licence" then I understand and can revert mine to comply.  It was the suggestion of a user to place the URL in place of the text to reference the LICENSE.txt file more directly and of course designate the license type.

Griatch:
No it's fine, what must be included with redistribution is the file LICENSE.txt which has the Evennia-specific part of the license - and that is not explained in the old version of About either (and doesn't need to be - the important bit is that it's BSD after all).

Issue #1174, opened 2017-01-24
Feature Request: Rename "Player" to "Account"


Griatch:
#### Description of requested feature:

Rename the `PlayerDB` database model to `AccountDB`. Rename its child typeclass from `DefaultPlayer` to `DefaultAcccount` and `Player` to `Account`. 

#### Reasons for adding feature:

A common source of confusion in Evennia is the separation between "Player" and "Character". The problem is that "Player" is usually what our "Character" is called in other code bases. 

The "Player" is an OOC entity whose main purpose is to act as the central account for the connected user - it holds the password info and any global settings. While we won't rename Character, renaming Player to Account should hopefully make it very clear how this relationship works, with no confusion as to what the Account level is or that multiple Characters are sharing a single Account instance. 

An effect of this change is the renaming of the use of "player" in other places of Evennia to "account", such as on Commands. 

#### Extra information

This is such a major API change that it belongs in a development branch leading up to Evennia 0.7. It must also come with database migrations and detailed help on how to update a game to use the new naming scheme.

Griatch:
It would be good with more people trying to migrate their databases to see the process is stable across many different uses. Here's the process (it will be documented more fully when devel is released for real):

0. First make sure you have pulled and migrated everything to the latest on `master` branch. Make a copy of your entire `mygame` dir just to be safe!
1. From your evennia repo, run `git checkout -b devel origin/devel` - this will download and place you on Evennia's current devel branch (which is not yet ready for production!)
2. Run `pip install -e .` This will fetch the latest dependencies needed by Evennia 0.7. 
3. In `evennia/bin/` is a program `project_rename.py`. Copy this to the top of your `mygame` folder.
4. `git checkout player-account-step1` - this will place you in an "detached HEAD state", at an earlier point of the migration history, where both `player/` and `account/` still co-existed (the rename program is not included yet at this point so make sure you copied it befor this).
4. We must rename all mentions of "Player" in your game dir to "Account". Go to your `mygame` folder and launch the rename program you copied like this: `python project_rename.py -i 'a player' -o 'an account' -i 'player' -o 'account' -x '*/migrations/*' -x '*/mssp.py' -r .` This starts an interactive program I've written that interactively renames all occurrences of "Player" in all your files. You should generally accept all renames unless you know you have added your own stuff separate from Evennia or is using `Player` in some other context (such as in a doc string). Use `h` in the program for help. If you don't want interactivity, add `-a` as a flag to the command.
[...]

PR #1262, opened 2017-03-20
Fixed bug in get_all_mail.

grungies1138:
db_receivers_players=self.caller needed to be changed to self.caller.player

### DIFF
--- a/evennia/contrib/mail.py
+++ b/evennia/contrib/mail.py
@@ -87,7 +87,7 @@ class CmdMail(default_cmds.MuxCommand):
         """
         # mail_messages = Msg.objects.get_by_tag(category="mail")
         # messages = []
-        messages = Msg.objects.get_by_tag(category="mail", raw_queryset=True).filter(db_receivers_players=self.caller)
+        messages = Msg.objects.get_by_tag(category="mail", raw_queryset=True).filter(db_receivers_players=self.caller.player)
         return messages
 
     def send_mail(self, recipients, subject, message, caller):


Griatch:
This change causes it to fail its unittest.

eldritchsemblance:
The test failure is almost identical to the issue I fixed in PR #1265, where the evennia making an assumption about whether the account is ooc or puppeting.  Maybe that determination needs to be baked in at a higher level in the code?

BlauFeuer:
Here are the tests for this contrib at https://github.com/evennia/evennia/blob/master/evennia/contrib/tests.py#L540

It's helpful to know which test failed and what error was given, so I ran
`evennia test evennia`

Running the test without patching passes:
``
Ran 200 tests in 67.216s

OK
```
But then adding the PR and re-running:

```
ERROR: test_mail (evennia.contrib.tests.TestMail)
  File "/home/evennia/evennia/evennia/contrib/tests.py", line 542, in test_mail
    self.call(mail.CmdMail(), "2", "'2' is not a valid mail id.", caller=self.player)
  File "/home/evennia/evennia/evennia/commands/default/tests.py", line 68, in call
    cmdobj.func()
  File "/home/evennia/evennia/evennia/contrib/mail.py", line 209, in func
    message = self.get_all_mail()[int(self.lhs) - 1]
  File "/home/evennia/evennia/evennia/contrib/mail.py", line 90, in get_all_mail
    messages = Msg.objects.get_by_tag(category="mail", raw_queryset=True).filter(db_receivers_players=self.caller.player)
AttributeError: 'DefaultPlayer' object has no attribute 'player'

Ran 200 tests in 66.805s

FAILED (errors=1)
```

eldritchsemblance:
This error can happen anywhere a command that is available both ooc and puppeting and something on the Player (or Account, post #1174) needs to be accessed.  Checking for the AttributeError exception every time .player is accessed seems to be not very DRY.

BlauFeuer:
Confusion about `self.caller` goes away when code uses `self.character` and `self.player` instead.  None of my code ever uses `self.caller` to assume anything more than "the object that initiated the command".

eldritchsemblance:
@BlauFeuer If that's the case I'm going to revisit #1265

BlauFeuer:
The default commands inherit from evennia.commands.default.muxcommand.MuxCommand, and so the caveat is ...

 [this code](https://github.com/evennia/evennia/blob/22e6cb4f8fe347ebce169577b056cbffac7af1c8/evennia/commands/default/muxcommand.py#L135) only engages when `self.player_caller == True`

eldritchsemblance:
So the fix for this should just be to use self.player then.

BlauFeuer:
@eldritchsemblance If you changed line 91 to filter on `db_receivers_players=self.player` it would still pass unit tests, but wouldn't be a functional mail system when used as a player or a character command as written.

If you just add `self.player_caller = True` to CmdMail, then, just like applying the PR, it works as a player or a character command, but doesn't pass unit tests.

This leads me to believe that a successful PR would have to also update the unit tests.

eldritchsemblance:
@BlauFeuer @grungies1138 The change as is simply changes the point of failure from while puppeting to while ooc.  The same change I applied to #1265 before @BlauFeuer advice works here, all unit tests pass, and the mail system works both ooc and while puppeting:

`try:
            player = self.caller.player
        except AttributeError:
            player = self.caller
        messages = Msg.objects.get_by_tag(category="mail", raw_queryset=True).filter(db_receivers_players=player)`

Griatch:
The regression was fixed by @eldritchsemblance in #1289. Closing.

PR #1266, opened 2017-03-21
Only parse as switches when directly after mux command

BlauFeuer:
Only parse as switches when directly after mux command, to enforce the syntax
 **command[ possibly several words][/switch[/switch..]] arg1[,arg2,...] [[=|,] arg[,..]]**

Allows mux switch character (`/`) to be used as the first character of an argument when whitespace is provided between the command and the argument, but parsed normally when whitespace does not separate the command and intended switch.

Example:
`say /Tomorrow/, not today!`
should not take the first word as a switch, because it is separated from the command, not flush against it.

`say/ooc /ooc is the switch to speak out of character.`
will parse `ooc` as a switch, but the second `/ooc` is parsed as part of the argument as expected.

### DIFF
--- a/evennia/commands/default/muxcommand.py
+++ b/evennia/commands/default/muxcommand.py
@@ -99,7 +99,7 @@ class MuxCommand(Command):
 
         # split out switches
         switches = []
-        if args and len(args) > 1 and args[0] == "/":
+        if args and len(args) > 1 and raw[0] == "/":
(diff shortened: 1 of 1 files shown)


Griatch:
I can buy that. Merging.

Griatch:
Merged with a rebase.

PR #1267, opened 2017-03-22
Adding at_give() hook to the default object

BattleJenkins:
Adds an at_give() hook to the default object, which is called by the default give command and passes the giving and receiving objects as arguments.

I'm working on a simple contrib for wearable objects, and wanted clothing objects that are worn to have their 'worn' status automatically removed when dropped or given away - there's a built-in hook for the former, but not the latter.


Griatch:
Thanks! Looks good. Merged with a rebase.

Issue #1268, opened 2017-03-23
Feature Request: Testing contents of command return


vlegoff:
#### Brief summary of issue / Description of requested feature:

It would be great to be able to test a message returned by a command without a specific syntax.  Check, for instance, that a line is present in a table, but not to duplicate all this table.

#### Motivation for this issue

As far as I can see, the `evennia.commands.default.tests.CommandTest` class with its `call` method allows to test a single command and compare its message.  This seems to be a direct, absolute comparison.  With commands that return a lot of information (like a table), it would be a matter of creating a new table in the test, knowing that should a minor, aesthetic modification occur, the test would be invalid.

I suggest to return, or at least give access to the mocked msg method in order to retrieve all messages sent to it and test a single line of a table, or make sure an information is present without specifying the exact syntax.

In other words, either:

```python
text = self.call(...)
self.assertIn(" 3 days ago", msg.splitlines())
```

Or something even more subtle to test a tiny portion of the returned message.

Griatch:
As it currently stands, the `CommandTest.call()` tester will compare the beginning of the Command return, not counting any markup or ascii chars like table line edges. So you can supply as short a snippet as you like, as long as it matches the first part of the return it will pass. 
We could by all means add a new kwarg to the `call` command for instead returning the value verbatim and leave the checking up to the caller to handler.

vlegoff:
I saw. But isn't there a way to get the returned value and compare it effectively?  For instance, in my case, my command returns an entire table.  I don't care about how it's formatted, I want the second column on every row to contain 0.  It would be easier in this case to get it like:

```python
table = self.call(...)
lines = table.splitlines()
self.assertTrue(all(line.split("|")[1].strip() == "0" for line in lines))
```

vlegoff:
The class wouldn't change its behavior and still work as expected.  Except if no expected message is specified to `call()` it could return the message it received, which would allow for greater comparison without breaking anything in the current tests.

Griatch:
@vlegoff Sure. I'm not arguing against you, it's a simple enough thing to add.

vlegoff:
@Griatch: I would have opened a PR but I wanted to check with you that it was okay.  Do you want me to do it?  As you said, it's just a return to add and a few lines to write in the doc, but I can do that, at least.

vlegoff:
Okay, that's done.  I tested it to and it doesn't seem to break unittesting for the current 200 unit tests in Evennia.

Griatch:
[...]

PR #1271, opened 2017-03-23
The TestCommand.call() now returns the received message

vlegoff:

The `TestCommand.call` method now returns the message received by the caller, useful for comparisons.

This is in answer to issue #1268, which would explain a bit more of the interest.


Griatch:
Merged with a rebase.

PR #1272, opened 2017-03-23
CONTRIB: Gendersub help fix

eldritchsemblance:
Made the gendersub contrib package work with messages that don't include gender information, such as those coming from the help system.  Made gender substitution respect escape sequences
When the gendersub contrib package is activated, it is impossible to access the help system.
Also, when that package is used, it was not respecting escaped pipes, causing undesired substitution to occur.
This issue was not tracked.  Regex was changed to include negative-lookaround substitution.


Griatch:
Thanks, looks reasonable. Merged with a rebase.

Issue #1251, opened 2017-03-14
Searches by DBREF allowed to ALL


vlegoff:
#### Brief summary of issue / Description of requested feature:

It seems that one doesn't need to have builder privileges in order to search the database using DBREF.  One easy example:

```look #1```

This allows to look at the super-user, even without builder privileges, even if the super-user is in a distant room.

#### Steps to reproduce the issue / Reasons for adding feature:

1. Enter a command with search features and specify a DBREF after a '#' sign.

#### Error output / Expected result of feature

No error.  Looking in the database without proper privileges is still allowed.  One can actually look through the entire universe using the 'look' command.  Okay, that's not an '@examine', but that sounds like a spoiler alert if not a high risk in security.

Possible fixes would be to make sure searching with DBREF is disabled to users who don't have builder right (permissions above players', at least).  Evennia's default commands could adapt to this, although it might be preferable to update the 'search' method one can find in `evennia.objects.objects`.

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:

- Evennia version: 0.6.0

**Comment by BlauFeuer** [comment]
I mentioned this issue perhaps a year ago.  The resolution was to add an option that disallows searches by #db, but that option is not incorporated in the default commands, as you've noticed.

When making your own custom commands, it is an important consideration.

**Comment by vlegoff** [comment]
There's a use_dbref parameter to caller.search.  In principle it should be aligned on privileges, I believe (while builders should use it, players shouldn't).  That's doable, but perhaps that's not the right way to do it.

**Comment by BlauFeuer** [comment]
In principle, it _should_ be aligned on builder or higher privilege, and you can decide how you want that particular command to act for builders/non-builders by variable call parameters.  In general, that's left to the dev to decide, but having said that, you could PR the default commands to act in a more appropriate manner as an example of how it should be done.  I typically just wholesale change the default commands, anyway with whatever nuance I think they are lacking.

Other than `look`, are there any other commands that seem to have unintended elevation of privilege that you feel shouldn't?  Certainly I can PR/patch them as a coding exercise to resolve the issue, but I think forcing search to align to builder or higher privilege would reduce flexibility as a framework for devs.

**Comment by Griatch** [comment]
This is indeed something that was discussed some time ago, and there was even work done on it but since there was issue for this particular thing it got forgotten. I would be fine with this being added to default commands as

```python
self.caller.search(query, use_dbref=self.caller.locks.check_lockstring("Builders"))
```

[...]

PR #1273, opened 2017-03-24
CmdLook use_dbref by Builders perm group #1251

BlauFeuer:
Only allow referencing targets to view by #dbid if viewer is in Builders permission group

Solves specific issue in #1251 with default CmdLook

Does not resolve the general issue, which could be left to devs to call search with their specific check.

The reason I never made an issue of this is because devs can easily create their own custom replacements, which I did, but having the test in the default CmdLook (with this PR) provides an example of how it can be done.

### DIFF
--- a/evennia/commands/default/general.py
+++ b/evennia/commands/default/general.py
@@ -60,16 +60,17 @@ class CmdLook(COMMAND_DEFAULT_CLASS):
         """
         Handle the looking.
         """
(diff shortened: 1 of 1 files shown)


Griatch:
Merged with a rebase. The search functionality is affecting all things though I'd think, like get and give, so will not close issue #1251 just yet.

PR #1275, opened 2017-03-25
Fix error when viewing help while OOC and not puppeting while using webclient

eldritchsemblance:
Check whether the caller has the "player" attribute when viewing help in the webclient.  Webclient feature "help in popup" requires checking the db usually found on a Player object.  However, when not puppeting, it was still assuming the caller was a character and tried to access the callers "player" field.  .  Only KeyError exceptions were caught from the try block, thus resulting in an unhandled exception.
It is impossible to access the help system while OOC using the webclient.
This issue was introduced when the pop-up help window was added to the webclient.
This issue was not open in the tracker
Resubmissionof #1265  from different branch

### DIFF
--- a/evennia/commands/default/help.py
+++ b/evennia/commands/default/help.py
@@ -60,7 +60,7 @@ class CmdHelp(Command):
 
             if self.session.protocol_key in ("websocket", "ajax/comet"):
                 try:
-                    options = self.caller.player.db._saved_webclient_options
+                    options = self.player.db._saved_webclient_options
                     if options and options["helppopup"]:
                         usemore = False
                 except KeyError:


BlauFeuer:
For reference, the traceback before the PR is:
```
  File "/home/evennia/evennia/evennia/commands/cmdhandler.py", line 509, in _run_command
    ret = yield cmd.func()
  File "/home/evennia/evennia/evennia/commands/default/help.py", line 228, in func
    self.msg_help(formatted)
  File "/home/evennia/evennia/evennia/commands/default/help.py", line 63, in msg_help
    options = self.caller.player.db._saved_webclient_options
AttributeError: 'Player' object has no attribute 'player'

An untrapped error occurred.
(Traceback was logged 2017-03-25 03:57:52-0000).
```

Griatch:
Cool, thanks. Merged with a rebase.

Issue #1282, opened 2017-03-27
Feature request: Implement details as full objects


xelaadryth:
PR: https://github.com/evennia/evennia/pull/1283

#### Brief summary of issue / Description of requested feature:
It's currently very difficult to implement details with the `contrib`-supplied `ExtendedRoom` and such. Making details a built-in feature of Evennia that is easy to use for such a common request would be nice.

#### Steps to reproduce the issue / Reasons for adding feature:

1. Commonly used
2. Easy to implement

#### Error output / Expected result of feature
I can create an object and then `@lock my object = notice:false()` to disable seeing it in object contents (mostly room contents, but also works with objects in chests or on players since they're all just objects)

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
By default add the lock `notice:all()` to all `DefaultObject`s. In `DefaultObject.return_appearance`, add an extra condition for room contents:
```
        # get and identify all objects
        visible = (con for con in self.contents if con != looker
                   and con.access(looker, "view")
                   and con.access(looker, "notice"))
```
Problem with this implementation: All existing Objects in the game world will be missing the `notice:all()` lock by default, unless there's a way to implement default behavior when a lock is undefined.

Alternatively, can make the lock default to True when not present.
```
        # get and identify all objects
        visible = (con for con in self.contents if con != looker
                   and con.access(looker, "view")
                   and con.access(looker, "notice", default=True))
```

Alternatively can add a negative lock called `hidden` that when lock is enabled, disable viewing. Seems unintuitive because of the name "lock" though.
```
        # get and identify all objects
        visible = (con for con in self.contents if con != looker
                   and con.access(looker, "view")
                   and not con.access(looker, "hidden"))
```

If this conflicts with a custom-defined lock, however, there may be conflict issues.

eldritchsemblance:
This contrib package simulates the typical MUSH/MUX +view code, in which the intent was to provide simple extended text descriptions that did not rely on an extra database object.  What you want to do is something different than that.

PR #1283, opened 2017-03-27
Adding a notice lock to hide room contents.

xelaadryth:
Able to easily hide objects in rooms but have them be fully interactable.
Because it's cool.
Full context:
https://github.com/evennia/evennia/issues/1282
May introduce conflicts if user defines a lock called "notice" as well.


eldritchsemblance:
@xelaadryth This feature is cool enough on its own.  It actually could replicate the MUX/MUSH "Dark" flag nicely, be tied in with darkness/vision systems, etc.  Maybe make it a contrib?

PR #1285, opened 2017-03-27
Default email domain to example.com per RFC2606

eldritchsemblance:
Changes the default email addressed assigned to an account from "dummy@dummy.com" to "dummy@example.com"
dummy.com is an actual domain that can receive email.  IETF RFC 2606 reserves the special domain "example.com" specifically for this use case
this was not a tracked issue


Griatch:
Good catch, merged with a rebase.

PR #1287, opened 2017-03-28
Update Object.announce_move_* to take advantage of mapping

vlegoff:

Two additions in the `announce_move_from` and `announce_move_to` hooks:

- Support to override the default message without re-coding both methods.
- Support mappings in string.

The `announce_move_from` or `announce_move_to` messages weren't easy to override without copy/pasting the initial code.  Moreover, it still used a system to send individual objects the message, rather than relying on `room.msg_contents` while specifying mapping.  Both characteristics made it a bit hard to update.

`location.msg_contents` (or `destination`) is now used with custom mapping.  It is possible to override these messages very simply:

```python
    def announce_move_from(self, destination):
        """..."""
        super(Character, self).announce_move_from(destination,
                msg="{character} leaves {exit}, toward {destination}.")
```

I have also added the `exit` mapping, which may be "somewhere" if the exit wasn't found.  `{character}`, `{origin}` and `{destination}` are self-explanatory.  Notice, however, that I didn't use `location` in the mapping, as I believe it's more clear to have mappings with `origin` and `destination` rather than `location` which isn't always straight-forward depending on the movement.  This is open to debate, of course.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -1159,10 +1159,10 @@ class DefaultObject(with_metaclass(TypeclassBase, ObjectDB)):
         You can override this method and call its parent with a
         message to simply change the default message.  In the string,
         you can use the following as mappings (between braces):
-            character: the character who is moving.
-            exit: the exit from which the character is moving (if found).
-            origin: the location of the character before the move.
-            destination: the location of the character after moving.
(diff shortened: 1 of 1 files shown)


Griatch:
Closed with a rebase.

PR #1288, opened 2017-03-28
Add useful hooks for the say and whisper commands

vlegoff:

The say and whisper commands now have hooks to control and display messages, which makes updating the behavior of these commands quite simple without overriding them.

The say command already used a hook: this hook, `object.at_say`, was called before the say and was used to change the content of the message to be displayed.  However, it didn't allow to display the content in another format.  Moreover, it used `%s` symbols instead of relying on mappings, which meant it didn't use `get_display_name` on objects.

There's now two different hooks for the say command:

- `object.at_before_say`: it can be used much like `object.location.at_say` (notice that this change invalids the latter `at_say` hook).  It can be used to modify the text to be said in the location.  It can also prevent the say by returning a None value.
- `object.at_after_say`: this hook is responsible for displaying the said message in the location.  The say command doesn't do anything but calling these hooks, which means changing the format of the say command is quite easy without re-writing the say command.  This second hook allows for simple overriding of messages to display, mapping (with `get_display_name`) and simple extensions by users' sub-classing.

The whisper command adds these two hooks on objects too, since it didn't use any hook at all.

#### A warning on compatibility

This PR breaks compatibility of the `at_say` hook.  The hook isn't used anymore.  Beforehand, it was used on the location (that is, more likely than not, on the room of the character speaking).  This, however, didn't seem to allow the greatest flexibility, and the `at_before_say` hook, like the `at_after_say` one, are put on the object speaking, not on its location.

Therefore, there's a deprecated warning in the `at_say` hook.  I'm not sure whether it should be removed or simply re-mapped to `at_before_say`.  If the latter, then it can be somewhat misleading to users, because they don't do the exact same thing, but it depends on Evennia's policy on hook compatibility.

### DIFF
--- a/evennia/commands/default/general.py
+++ b/evennia/commands/default/general.py
@@ -452,13 +452,15 @@ class CmdWhisper(COMMAND_DEFAULT_CLASS):
             return
 
         speech = self.rhs
+        # Call a hook to change the speech before whispering
+        speech = caller.at_before_whisper(receiver, speech)
 
-        # Feedback for the object doing the talking.
-        caller.msg('You whisper to %s, "%s|n"' % (receiver.key, speech))
+        # If the speech is empty, abort the command
+        if not speech:
+            return
 
-        # Build the string to emit to receiver.
-        emit_string = '%s whispers, "%s|n"' % (caller.name, speech)
-        receiver.msg(emit_string, from_obj=caller)
+        # Call the at_after_whisper hook for feedback
+        caller.at_after_whisper(receiver, speech)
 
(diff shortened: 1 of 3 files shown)


vlegoff:
The `at_say` hook has been removed from `DefaultObject`.  At least the API change is clear.

Griatch:
While useful, I think this is an backwards-incompatible API change that probable is best done towards devel branch and Evennia 0.7, where we are several big API changes anyway. So holding off on this one for now.

PR #1289, opened 2017-03-29
CONTRIB: mail.py: check if ooc or puppeting before finding player object during get_all_mail

eldritchsemblance:
Checks to see if OOC or Puppeting during get_all_mail
Trying to use the mail system fails when OOC
This fixes the test failure and the bug introduced in PR #1262


vlegoff:
Changes seem fine to me, from what I could see.  If nothing else the change of the default message was welcome.  Checked for Python compliance.

PR #1294, opened 2017-03-30
Added contrib module for turn-based battle system

BattleJenkins:
This adds a new contrib file, a module that provides a simple framework for a tabletop style turn based battle system. From the module itself:

This is a framework for a simple turn-based combat system, similar
to those used in D&D-style tabletop role playing games. It allows
any character to start a fight in a room, at which point initiative
is rolled and a turn order is established. Each participant in combat
has a limited time to decide their action for that turn (30 seconds by
default), and combat progresses through the turn order, looping through
the participants until the fight ends.

Only simple rolls for attacking are implemented here, but this system
is easily extensible and can be used as the foundation for implementing
the rules from your turn-based tabletop game of choice or making your
own battle system.

Adds contrib module for a simple turn-based battle system.

Combat systems are a near-universal feature for MUDs, and currently no contrib module provides a framework for them - this system provides a framework that resembles those of common tabletop role playing games and can be readily extended to suit the coder's needs.

I've tested this thoroughly on my local system and have fixed all the bugs I could find, but it hasn't been tested by multiple users yet.

### DIFF
--- a/evennia/contrib/turnbattle.py
+++ b/evennia/contrib/turnbattle.py
@@ -26,20 +26,16 @@ instead of the default:
 
     class Character(BattleCharacter):
         
-Next, import the combat commands into your default_cmdsets.py module:
+Next, import this module into your default_cmdsets.py module:
 
-    from evennia.contrib.turnbattle import CmdFight, CmdAttack, CmdRest, CmdPass, CmdDisengage
+    from evennia.contrib import turnbattle
 
-And add the commands to your default command set:
+And add the battle command set to your default command set:
 
     #
     # any commands you add below will overload the default ones.
     #
-    self.add(CmdFight())
-    self.add(CmdAttack())
-    self.add(CmdRest())
-    self.add(CmdPass())
-    self.add(CmdDisengage())
+    self.add(turnbattle.BattleCmdSet())
 
 This module is meant to be heavily expanded on, so you may want to copy it
 to your game's 'world' folder and modify it there rather than importing it
@@ -494,6 +490,22 @@ class CmdRest(Command):
         You'll probably want to replace this with your own system for recovering HP.
         """
 
+class BattleCmdSet(default_cmds.CharacterCmdSet):
+    """
+    This command set includes all the commmands used in the battle system.
+    """
+    key = "DefaultCharacter"
+
+    def at_cmdset_creation(self):
+        """
+        Populates the cmdset
+        """
+        self.add(CmdFight())
+        self.add(CmdAttack())
+        self.add(CmdRest())
+        self.add(CmdPass())
+        self.add(CmdDisengage())
+        
 """
 ----------------------------------------------------------------------------
 SCRIPTS START HERE
--- a/evennia/contrib/turnbattle.py
+++ b/evennia/contrib/turnbattle.py
@@ -68,18 +68,17 @@ def roll_init(character):
     Notes:
         By default, does not reference the character and simply returns
         a random integer from 1 to 1000.
+        
+        Since the character is passed to this function, you can easily reference
+        a character's stats to determine an initiative roll - for example, if your
+        character has a 'dexterity' attribute, you can use it to give that character
+        an advantage in turn order, like so:
+        
+        return (randint(1,20)) + character.db.dexterity
+        
+        This way, characters with a higher dexterity will go first more often.
     """
     return randint(1,1000)
-    """
-    Since the character is passed to this function, you can easily reference
-    a character's stats to determine an initiative roll - for example, if your
-    character has a 'dexterity' attribute, you can use it to give that character
-    an advantage in turn order, like so:
-    
-    return (randint(1,20)) + character.db.dexterity
-    
-    This way, characters with a higher dexterity will go first more often.
-    """
     
 def get_attack(attacker, defender):
     """
@@ -96,14 +95,13 @@ def get_attack(attacker, defender):
     Notes:
         By default, returns a random integer from 1 to 100 without using any
         properties from either the attacker or defender.
+        
+        This can easily be expanded to return a value based on characters stats,
+        equipment, and abilities. This is why the attacker and defender are passed
+        to this function, even though nothing from either one are used in this example.
     """
     # For this example, just return a random integer up to 100.
     attack_value = randint(1, 100)
-    """
-    This can easily be expanded to return a value based on characters stats,
-    equipment, and abilities. This is why the attacker and defender are passed
-    to this function, even though nothing from either one are used in this example.
-    """
     return attack_value
     
 def get_defense(attacker, defender):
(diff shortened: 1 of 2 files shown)


evennia/contrib/turnbattle.py
```diff
@@ -0,0 +1,520 @@
...
+        This performs the actual command.
+        """
+        here = self.caller.location
+        fighters = []
+        
+        if not self.caller.db.hp: # If you don't have any hp
+            self.caller.msg("You can't start a fight if you've been defeated!")
+            return
+        for thing in here.contents: # Test everything in the room to add it to the fight.
+            if thing.db.HP: # If the object has HP...
```
Griatch: This is a nice touch, it means you could fight a table or door assuming it has HP. :)

evennia/contrib/turnbattle.py
```diff
@@ -0,0 +1,520 @@
...
+        """
+        for fighter in self.db.fighters:
+            combat_cleanup(fighter) #Clean up the combat attributes for every fighter.
+    
+    def at_repeat(self):
+        """
+        Called once every self.interval seconds.
+        """
+        currentchar = self.db.fighters[self.db.turn] # Note the current character in the turn order.
+        self.db.timer -= 1 # Count down the timer by one second.
```
Griatch: Ticking every second is a a simple but wasteful way to handle a countdown. The only times that the script cares about are after 20s (when the warning happens) and 30s (when the next round is force-started). You could therefore get away with ticking the script only every 10 seconds instead. You use the scripts' `restart` method whenever you want to prematurely start the next round on-demand. You can simply have your action-set commands check in to the script whenever an action is performed, so it knows when the actions are used up.

I'm not necessarily requiring you to rewrite this, but doing things on-demand is the recommended way to handle timeouts in an asynchronous system like Evennia. Checking things over and over for change is generally a bad practice.

evennia/contrib/turnbattle.py
```diff
@@ -0,0 +1,520 @@
...
+        for fighter in self.db.fighters:
+            if fighter.db.Combat_LastAction != "disengage": # If a character has done anything but disengage
+                DisengageCheck = False
+        if DisengageCheck == True: # All characters have disengaged
+            self.obj.msg_contents("All fighters have disengaged! Combat is over!")
+            self.stop() # Stop this script and end combat.
+            return
+        
+        # Check to see if only one character is left standing. If so, end combat.
+        DefeatedCharacters = 0
```
Griatch: Python PEP8 standard (and thus Evennia's style guide) dictates that local variables are always written in lowercase with underscores. CamelCase is only ever used for class names.

evennia/contrib/turnbattle.py
```diff
@@ -0,0 +1,520 @@
...
+    
+    
+
+"""
+----------------------------------------------------------------------------
+COMBAT FUNCTIONS START HERE
+----------------------------------------------------------------------------
+"""
+def roll_init(character):
+    """
```
Griatch: As elsewhere, google-style docstrings are needed.

Griatch:
@BattleJenkins The concept of a unit test is simple - they are simply example calls of the individual components of a system. making sure a given function actually returns the result you expect to a given input. These are run automatiucally and makes sure your combat system keeps working even when other things inside Evennia change. If we were to change something in the API that the combat system relies on, the unit test will hopefully fail and tell us that we must update the combat system to match. The wiki page on unit tests should explain most of what you need to know, but don't be shy to ask if you have further questions!

evennia/contrib/turnbattle.py
```diff
@@ -0,0 +1,663 @@
...
+    Returns:
+        initiative (int): The character's place in initiative - higher
+        numbers go first.
+        
+    Notes:
+        By default, does not reference the character and simply returns
+        a random integer from 1 to 1000.
+    """
+    return randint(1,1000)
+    """
```
Griatch: docstrings are usually at the top of the function/method/class etc. Probably better to put this info in the Notes section of the docstring instead or, if you want to have it in the  code, make it a normal comment with # markers.

evennia/contrib/turnbattle.py
```diff
@@ -0,0 +1,663 @@
...
+        if self.db.timer == 10: # 10 seconds left
+            currentchar.msg("WARNING: About to time out!")
+        
+        # Force current character to disengage if timer runs out.
+        if self.db.timer <= 0:
+            self.obj.msg_contents("%s's turn timed out!" % currentchar)
+            spend_action(currentchar, 'all', action_name="disengage") # Spend all remaining actions.
+            
+            
+    def initialize_for_combat(self, character):
```
Griatch: Needs the `Args:` block.

Griatch:
@BattleJenkins 

I ran the combat module with the latest fixes, seems to work fine. There are a few remaining issues:
1. When your turn times out, the other player only sees a message about the other timing out, there is no mention that it is now your turn. This is pretty clear on two players but could be a bit confusing for a larger fight. 
2. When trying to use an exit while in combat you get a traceback (this is a simple error, `self` on the object is the object itself, not the command):
       ```
 "/home/griatch/Devel/Home/evennia/evennia/evennia/objects/objects.py", line 683, in move_to
h     if not self.at_before_move(destination):
   File "/home/griatch/Devel/Home/evennia/evennia/evennia/contrib/turnbattle.py", line 308, in at_before_move
     self.caller.msg("You can't exit a room while in combat!")
 AttributeError: 'Character' object has no attribute 'caller'
        ```
3. It might be an idea to wrap your combat commands in a small cmdset. That way your install instructions could just be `from evennia.contrib import turnbattle` and then `self.add(turnbattle.BattleCmdSet)`.
4. I wonder if the combat should not offer some sort of custom help command. I reflexively entered help when going into combat and I figure many newbies would too - maybe help will just show a list of the available combat commands while help all and help <specific topic> will still show the normal help, something like that?

Griatch:
I also see some `[EE] Script Combat Turn Handler(#1) of type 'TurnHandler': at_repeat() error ''NoneType' object has no attribute 'cancel''.` errors in the logs. I wonder if this is an issue in the TurnHandler jor in the scripting mechanism itself though.

BlauFeuer:
The lgtm analysis doesn't seem to always run.  I don't think it's directly related to your code.  I've seen PRs accepted with the lgtm "could not build the merge commit" message, so wouldn't worry much about it.

Griatch:
Yeah, as Blau said, the LGTM PR analysis seems to itself sometimes fail. Travis one is the important one for stability (once the unittests are in).

Griatch:
@BattleJenkins Cool, at a quick glance it looks really nice now! I'll re-review it when the unit tests are in place. 

Btw, I turned off the LGTM Pull Request integration for now, it fails to load way too often on PRs for it to be useful.

Griatch:
Looks good now, thanks for the contribution - merged with a rebase.

PR #1298, opened 2017-04-04
Add automatic escaping of color codes in the EvEditor

vlegoff:

This PR simply adds automatic escaping of color codes (like `|w` or `{y`) in the EvEditor.

When using the EvEditor in code mode, the fact that Evennia should interpret symbols as color codes is a problem:

```python
character.msg("{word} would sound high and clear.")
```

This is problematic for code, but is it problematic for standard descriptions?  I hesitated a few days before deciding even for room descriptions, it would be more comfortable seeing plainly the color codes, that would make modifications easier (not mentioning more straightforward for us who can't see colors but are willing to use them).  Still, the decision on whether this should be added only for code editing or for any EvEditor isn't mine to make.

### DIFF
--- a/evennia/utils/eveditor.py
+++ b/evennia/utils/eveditor.py
@@ -49,6 +49,7 @@ import re
 from django.conf import settings
 from evennia import Command, CmdSet
 from evennia.utils import is_iter, fill, dedent, logger, justify, to_str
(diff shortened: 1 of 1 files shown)


Griatch:
It's a good idea I think. Merged!

PR #1302, opened 2017-04-06
Adjust docstring to resolve #1300 in shipped command.py

### LINKED ISSUE #1300: inconsistency with docstring in shipped command.py file
In the command.py file that is shipped out, in the docstring for the Command class it says:

    Inherit from this if you want to create your own command styles
    from scratch.  Note that Evennia's default commands inherits from
    MuxCommand instead.

    Note that the class's `__doc__` string (this text) is
    used by Evennia to create the automatic help entry for
    the command, so make sure to document consistently here.

    Each Command implements the following methods, called
    in this order (only func() is actually required):
        - at_pre_command(): If this returns True, execution is aborted.
        - parse(): Should perform any extra parsing needed on self.args
            and store the result on self.
        - func(): Performs the actual work.
        - at_post_command(): Extra actions, often things done after
            every command, like prompts.

But the evennia docs say that the name is actually supposed to be 'at_pre_cmd', and 'at_post_cmd'.
MuxCommand has the right ones according to the docs.

BlauFeuer:
Adjust docstring to resolve #1300
Also realigns code indent to multiple of 4 after `# ` is removed to uncomment

Reference the correct names of BaseCommand methods in the docstring and realign line comments if/when uncommented.

PyCharm line comments/uncomments by adding/removing `# `, so to have correct indentation, the single space was added to the line commented code.

### DIFF
--- a/evennia/game_template/commands/command.py
+++ b/evennia/game_template/commands/command.py
@@ -21,17 +21,17 @@ class Command(BaseCommand):
 
     Each Command implements the following methods, called
     in this order (only func() is actually required):
-        - at_pre_command(): If this returns True, execution is aborted.
+        - at_pre_cmd(): If this returns True, execution is aborted.
         - parse(): Should perform any extra parsing needed on self.args
             and store the result on self.
         - func(): Performs the actual work.
-        - at_post_command(): Extra actions, often things done after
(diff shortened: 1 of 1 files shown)


texdarkstar:
Hah, I was struggling with the block comments earlier today.... Lol.

PR #1303, opened 2017-04-10
Adds new 'clothing' contrib module

BattleJenkins:
Adds a new contrib module, providing a system for wearable clothing objects that are appended to one's description. From the module itself:

Items of clothing can be used to cover other items, and many options are provided to define your own clothing types and their limits and behaviors. For example, to have undergarments automatically covered by outerwear, or to put a limit on the number of each type of item that can be worn. Characters can also specify the style of wear for their clothing - I.E. to wear a scarf 'tied into a tight knot around the neck' or 'draped loosely across the shoulders' - to add an easy avenue of customization. The system as-is is fairly freeform - you can cover any garment with almost any other, for example - but it can easily be made more restrictive, and can even be tied into a system for armor or other equipment.

The unit tests have not been put in for this module yet, but I thought I would share it first to get people's thoughts!

Adds new 'clothing' contrib module.

A system for clothing didn't exist as a contrib module before - I came up with this system for clothing to use in a future game, and thought I'd offer it as a contrib while I was at it.

### DIFF
--- a/evennia/contrib/clothing.py
+++ b/evennia/contrib/clothing.py
@@ -51,7 +51,7 @@ inherit from ClothedCharacter in your game's characters.py file:
 
     class Character(ClothedCharacter):
         
-And do the same with the ClothedCharacterCmdSet in your game'same
+And do the same with the ClothedCharacterCmdSet in your game's
 default_cmdsets.py:
 
     from evennia.contrib.clothing import ClothedCharacterCmdSet
--- a/evennia/contrib/clothing.py
+++ b/evennia/contrib/clothing.py
@@ -1,0 +1,646 @@
+"""
+Clothing - Provides a typeclass and commands for wearable clothing,
+which is appended to a character's description when worn.
+
+Evennia contribution - Tim Ashley Jenkins 2017
+
+Items of clothing can be used to cover other items, and many options
+are provided to define your own clothing types and their limits and
+behaviors. For example, to have undergarments automatically covered
+by outerwear, or to put a limit on the number of each type of item
+that can be worn. Characters can also specify the style of wear for
+their clothing - I.E. to wear a scarf 'tied into a tight knot around
+the neck' or 'draped loosely across the shoulders' - to add an easy
+avenue of customization. The system as-is is fairly freeform - you
+can cover any garment with almost any other, for example - but it
+can easily be made more restrictive, and can even be tied into a
+system for armor or other equipment.
+
+To install, import this module and have your default character
+inherit from ClothedCharacter in your game's characters.py file:
+
+    from evennia.contrib.clothing import ClothedCharacter
+
+    class Character(ClothedCharacter):
+        
+And do the same with the ClothedCharacterCmdSet in your game'same
+default_cmdsets.py:
+
+    from evennia.contrib.clothing import ClothedCharacterCmdSet
+
+    class CharacterCmdSet(default_cmds.CharacterCmdSet):
+        
+From here, you can use the default builder commands to create clothes
+with which to test the system:
+    
+    @create a pretty dress : evennia.contrib.clothing.Clothing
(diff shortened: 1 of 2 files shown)


eldritchsemblance:
There goes the contrib package I was working on!  I should have been faster!

evennia/contrib/clothing.py
```diff
@@ -0,0 +1,678 @@
...
+        # Append worn clothes.
+        if worn_string_list:
+            string += "|/|/%s is wearing %s." % (self, list_to_string(worn_string_list))
+        else:
+            string += "|/|/%s is not wearing anything." % self
+        return string
+        
+"""
+----------------------------------------------------------------------------
+HELPER FUNCTIONS START HERE
```
Griatch: Common coding standards suggest the helper functions go above the classes that use them - this makes the code easier to read.

evennia/contrib/clothing.py
```diff
@@ -0,0 +1,678 @@
...
+
+    from evennia.contrib.clothing import ClothedCharacterCmdSet
+
+    class CharacterCmdSet(default_cmds.CharacterCmdSet):
+        
+From here, you can use the default builder commands to create clothes
+with which to test the system:
+    
+    @create a pretty dress : evennia.contrib.clothing.Clothing
+    @set dress/clothing_type = 'body'
```
Griatch: Would not `wear_location` be a more suitable Attribute name for this?

Griatch:
Cool system, nice contrib! I tested it out and it appears to work well. There was one edge case I found which I think is not quite working right. The way I understand it, certain garments are "auto-covering" others as defined in the options. The order of wearing appears to override this though. Example:

```
@create a shining suit of armor : evennia.contrib.clothing.Clothing
> @set suit/clothing_type = body
> wear suit
> look self
Griatch is wearing a suit of shining armor
> @create a pair of blue underpants : evennia.contrib.clothing.Clothing
> @set suit/clothing_type = underpants
> wear underpants
> look self
Griatch is wearing a suit of shining armor and a pair of underpants.
```
Which does not appear to be correct since the 'body' location is set to always cover the 'underpants'  in `CLOTHING_TYPE_AUTOCOVER`. On the other hand, the `cover` command works fine to manually have the suit cover the underpants correctly, so I'm not sure if wear order mattering is expected behavior or not (I guess one could argue that the absent-minded knight should be able to put the underpants outside his armor if he really wants to ...)

Griatch:
Also, the `give` command raises a traceback. The unittests seem unnecessarily shallow since all of them only test the single case of having no input (thus missing the traceback when actually having an argument to the command).

evennia/contrib/clothing.py
BattleJenkins: I'd say "clothing type" is better, as this system allows you to specify where a garment is worn - say, perhaps, a scarf worn around the waist, around the head, or around the neck - but I think the confusion comes from 'body' being the name of the clothing types for dresses and bodysuits is pretty misleading. I'll either rename that slot or give a different example.

Griatch:
If the auto-cover works as intended, then that's fine, just some comments about when the `CLOTHING_TYPE_AUTOCOVER` kicks in and that you can override it by the order you wear things would be enough to clarify any confusion. An alternative usability addition would be to have the `wear` command make a simple note of when you are wearing something in an order contradicting the `CLOTHING_TYPE_AUTOCOVER`.

Griatch:
The unit tests look fine now, for future reference the EvenniaTest suite will already create a set of utility objects available in your test, those are available as self.char, self.char2, self.room, self.room2 and so on. There is also self.player with correctly mocked session data etc. Those can often be used to test things in a simple environment. You can see this all set up in evennia/util/test_resources.py.

I merged with a rebase after some further fixes to comply with PEP8 standard. Thanks for the contrib! :D

Issue #1306, opened 2017-04-14
IndexError in telnet.py


TehomCD:
#### Brief summary of issue / Description of requested feature:
While looking at the portal log, I saw a traceback pop up frequently:

```
2017-04-14T10:44:37-0400 [AMPProtocol,client] [::] Traceback (most recent call last):
2017-04-14T10:44:37-0400 [AMPProtocol,client] [::]   File "/home/tehom/arx/evennia/evennia/server/portal/portalsessionhandler.py", line 420, in data_out
2017-04-14T10:44:37-0400 [AMPProtocol,client] [::]     getattr(session, funcname)(*cmdargs, **cmdkwargs)
2017-04-14T10:44:37-0400 [AMPProtocol,client] [::]   File "/home/tehom/arx/evennia/evennia/server/portal/telnet.py", line 340, in send_text
2017-04-14T10:44:37-0400 [AMPProtocol,client] [::]     linetosend = ansi.parse_ansi(_RE_N.sub("", text) + ("|n" if text[-1] != "|" else "||n"),
2017-04-14T10:44:37-0400 [AMPProtocol,client] [::] IndexError: string index out of range
```

It seems to pop up fairly regularly, say a few times an hour. No players have reported seeing anything strange, so I'm guessing that it'd be an empty string from the IndexError that just isn't sent when the exception is raised.

#### Steps to reproduce the issue / Reasons for adding feature:

1. Seems to happen randomly but somewhat frequently.

#### Error output / Expected result of feature
Traceback in portal log.

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
I guess just check for empty string before trying to parse it, but I have no idea if that'd cause any issues.

Griatch:
This is due to the recent correction of the end-point color tag. Easy to fix.

PR #1308, opened 2017-04-17
Fix to color normalization end tags

BlauFeuer:
Account for empty strings sent in Telnet and SSH

Addresses a potential problem similar to IndexError in telnet.py #1306

Initial fix in #1243 does not account for empty string.


BlauFeuer:
#1243 (the cause of #1306) was my doing. Thanks @TehomCD and @Griatch for catching and fixing it.

Griatch:
@BlauFeuer Good catch, this is a more complete fix. Merged with a rebase. Closing.

PR #1314, opened 2017-04-24
More thorough explanation of clothing types

BattleJenkins:
Added some further explanation of clothing types and their purpose. I felt like it wasn't clear enough that you could change the given clothing types or even elect not to use clothing types at all in your own implementation of the system.


Griatch:
Merged with rebase.

PR #1317, opened 2017-05-04
Systemd unit

eldritchsemblance:
Added a sample script for running Evennia as a systemd unit
Using newer distributions, especially Ubuntu and Debian, using initd and screen with a system user is not supported.
The documentation on the wiki would need to be updated to reflect systemd

### DIFF
--- a/bin/unix/evennia.service
+++ b/bin/unix/evennia.service
@@ -1,12 +1,18 @@
+# Evennia systemd unit script
+# Place in the appropriate directory for a system or user account
+# Then, register with systemctl
 [Unit]
 Description=Evennia Server
 
 [Service]
 Type=forking
+# Change to your location.
 WorkingDirectory=/var/evennia/evennia
+# These paths must be absolute.  Change to your location.
 ExecStart=/bin/bash -c 'source /var/evennia/pyenv/bin/activate ; evennia start'
 ExecStop=/bin/bash -c 'source /var/evennia/pyenv/bin/activate ; evennia stop'
 ExecReload=/bin/bash -c 'source /var/evennia/pyenv/bin/activate ; evennia reload'
+# restart on all failures, wait 3 seconds before doing so.
 Restart=on-failure
 RestartSec=3
 
--- a/bin/unix/evennia.service
+++ b/bin/unix/evennia.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Evennia Server
+
+[Service]
+Type=forking
+WorkingDirectory=/var/evennia/evennia
+ExecStart=/bin/bash -c 'source /var/evennia/pyenv/bin/activate ; evennia start'
+ExecStop=/bin/bash -c 'source /var/evennia/pyenv/bin/activate ; evennia stop'
+ExecReload=/bin/bash -c 'source /var/evennia/pyenv/bin/activate ; evennia reload'
+Restart=on-failure
+RestartSec=3
+
+[Install]
+WantedBy=default.target


Griatch:
A good idea, but I think this needs more documentation in the file itself, explaining what it is and why it may be useful also for Linux newbies (we get people using Linux in order to use Evennia after all). This would need documentation examples of how to install for, say, an Ubuntu system and how to register it with systemd. Finally examples on how to use it in practice.

Griatch:
Bump?

chainsol:
I have a concern as well - won't this encourage users to run Evennia as root? That is probably not a good idea, if so. You could set the user to some default, perhaps "evennia"?

BlauFeuer:
Properly documented and configured, I'd say it won't be a problem, since root on startup or manually, can run a daemon with a non-root UID, as is best practice.  Just how that's done depends on OS and distribution. In the case with more modern distributions, the UID also shouldn't be a system user, so file groups might need to be employed for maintenance on files.

> I have a concern as well - won't this encourage users to run Evennia as root? That is probably not a good idea, if so. You could set the user to some default, perhaps "evennia"?

chainsol:
@BlauFeuer Perhaps I didn't look close enough, but as-is, wouldn't this systemd configuration run Evennia as root?

BlauFeuer:
@chainsol I'm a bit rusty as an admin, but I believe it depends on who own the script, even if `root`  executes it.  Our confusion about this is probably the reason for **Griatch**'s comment that it needs more documentation.

See [the documentation file for the PR](https://github.com/evennia/evennia/pull/1317/commits/d99ae7a7b2bd48affded329dc00a2b7bf804cea8#diff-8e1aeae3d9350624c2beb6222b5b4021R2) - the script should be `chown`ed and set executable, etc.

eldritchsemblance:
Sorry I haven't been around to update this.

This unit runs as whoever owns the script.  I keep it in the proper folder for my evennia user. 
@chainsol actually your reasoning is backwards "Run as user" = no bueno, because it encourages people to put it in /etc/systemd with the script owned as root.  However, this should be owned by a nonprivileged user and invoked with systemctl --user and have it located in ~/.config/systemd instead

chainsol:
@eldritchsemblance Well then, thanks for clearing that up! I do agree with @Griatch, documentation that the userbase of Evennia can follow would be appreciated then - including perhaps how one can securely use this, where the file should be put, and how to register with systemd, purely because many people running Evennia might be using Linux for the first time. Perhaps you've already written this, even!

Griatch:
@eldritchsemblance I'd be interested in adding this, but as said, this requires more documentation. Your header shows the minimum, but as noticed in this thread, its functionality is not obvioulsy clear to everyone. Also phrases like `add to systemctl` is correct but too vague for the average user and will just lead to support questions down the line. So elaborating a little for the benefit of a relative linux newbie is needed here I think.

Griatch:
@eldritchsemblance Bump.

tsal:
I'd really like to see this too.  If we don't hear from @eldritchsemblance soon, I might see if I can document this for him in a separate PR.

Griatch:
@tsal If you are interested in documenting, go ahead, it seems like @eldritchsemblance is not coming around.

Griatch:
@tsal Still interested in documenting this?

Griatch:
Note: Before this can merge, it needs to be updated to upcoming evennia 0.8 (in `develop` branch) since that launches its processes in a different way.

tsal:
@Griatch missed the notification from Feb because of bad settings in github on my part, and then the one from a couple of weeks ago as I've been in and out of the hospital - I'll see if I can pull something together over the weekend now that I'm healed up.

friarzen:
PR #1667 submitted as a newer version of this based on the latest portal changes.

Griatch:
@tsal Sorry to hear about your hospitalization, but good to hear you are feeling better! This one was resolved in the meantime though, so closing it. Hope you didn't spend much time on it!

PR #1321, opened 2017-05-09
Remove the over-incrementing task ID

vlegoff:

The event task ID was always increasing, which would create long IDs if the system was often used.

With a system that creates lots of chained events, the task ID would have gone up pretty quickly.  Now, IDs can be re-used.  This doesn't create conflicts, as ID are free when the task is executed anyway.

### DIFF
--- a/evennia/contrib/events/scripts.py
+++ b/evennia/contrib/events/scripts.py
@@ -45,7 +45,6 @@ class EventHandler(DefaultScript):
         self.db.locked = []
 
         # Tasks
-        self.db.task_id = 0
         self.db.tasks = {}
 
     def at_start(self):
@@ -559,8 +558,12 @@ class EventHandler(DefaultScript):
         """
         now = datetime.now()
         delta = timedelta(seconds=seconds)
-        task_id = self.db.task_id
-        self.db.task_id += 1
+
+        # Choose a free task_id
+        used_ids = list(self.db.tasks.keys())
+        task_id = 1
+        while task_id in used_ids:
+            task_id += 1
 
         # Collect and freeze current locals
         locals = {}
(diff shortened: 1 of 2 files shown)


evennia/contrib/events/tests.py
```diff
@@ -58,7 +58,6 @@ def test_start(self):
         self.assertEqual(self.handler.db.to_valid, [])
         self.assertEqual(self.handler.db.locked, [])
         self.assertEqual(self.handler.db.tasks, {})
-        self.assertEqual(self.handler.db.task_id, 0)
```
Griatch: Why remove this unittest? Should not the sequence of task IDs be fully deterministic?

evennia/contrib/events/scripts.py
```diff
@@ -559,8 +558,12 @@ class methods, objects inside an instance and so on will
         """
         now = datetime.now()
         delta = timedelta(seconds=seconds)
-        task_id = self.db.task_id
-        self.db.task_id += 1
+
+        # Choose a free task_id
+        used_ids = list(self.db.tasks.keys())
```
Griatch: Converting this to a list seems unnecessary and even inefficient. For a large list, querying content is an `O(N)` operation while querying a dict remains constant, that is an `O(1)` behavior. So removing the `list()` conversion here sounds like the way to go.

vlegoff:
@Griatch okay, I'll see if it still works without convertion.  The unittest is unnecessary now because the `task_id` isn't stored as a permanent attribute anymore.  Rather it is calculated each time a task is created and will just find the lowest, non-used ID.

Griatch:
Ok. I would have thought it would be useful the id system works the way you think, for example to make sure task id's are made available correctly. But if that's not an issue, fair enough.

PR #1322, opened 2017-05-14
Fixed a bug in which creating a SimpleDoor caused the next exit creat…

Blinkity:
…ed to fail silently.

I discovered a bug in the existing SimpleDoor contrib. You can repro it as follows.
- Dig a new room.
- Create a simpledoor to the new room.
- Create a second new room.
- Create an exit to the second new room.
The second exit will never be created, and no output will be produced.

The bug is due to the "return_exit_already_created" flag never being removed. Prior to my fix, this flag would cause the subsequent exit to not be created at all, rather than preventing the second exit from causing an infinite recursion as intended. This recursion was instead prevented by simply making two basic exits and never calling the given class's create_exit method again. I've fixed this by changing the super().create_exit call to self.create_exit, as I believe was originally intended.

It's a bug fix! Now people can use SimpleDoors without breaking their building process.

I didn't post an issue for this prior to the fix. I'm not sure if I should have.

### DIFF
--- a/evennia/contrib/simpledoor.py
+++ b/evennia/contrib/simpledoor.py
@@ -101,20 +101,20 @@ class CmdOpen(default_cmds.CmdOpen):
         """
         Simple wrapper for the default CmdOpen.create_exit
         """
+        # create a new exit as normal
+        new_exit = super(CmdOpen, self).create_exit(exit_name, location, destination,
+                                                    exit_aliases=exit_aliases, typeclass=typeclass)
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks for the PR! Normally an issue should be opened but since you discovered the bug, describe it and fix it in the same PR then that's fine.  :) Looks good, will merge when I get to a computer.

Blinkity:
That's great, thanks Griatch! First OS contribution. :-p

PR #1324, opened 2017-05-21
LogFile class for rotating channel logs

### LINKED ISSUE #1253: Feature request: auto-rotate of channel log files
Although log files like `portal.log` and `http_requests.log` will automatically rotate when over a 1MB threshold, the channel logs do not appear to do so, and seem to grow unbounded in size. It might be nice to have something that would automatically rotate log files when they pass over a certain length, while keeping the most recent lines in order to not empty the channel's `/history`.

1. Rotation of channel logs to prevent them from growing too huge

Older log files becoming something like `public.log.1` while a new `public.log` is created that has the most recent messages once it passes some threshold defined in settings (or as an attribute in the channel, possibly?)

I know that twisted has something to automatically rotate logs, but I'm unsure if it'd be best to use that or just something custom for evennia.

TehomCD:
I added a logfile class based on Twisted's rotating logfiles to use for the filehandles for channel logs. I overrode its `rotate()` method in order to write the tail end of the previous log file to the new one after rotation, so that people don't see the channel history vanish whenever it rotates. Size of rotation and how many lines to rotate are both customizable with settings.

To allow for channel logs to rotate rather than ominously growing unbounded in the background, growing ever more powerful, biding their time to strike. Resolves #1253 

I tested this a bit, but since my familiarity with twisted's logging consists of an afternoon of staring at their logging API and tinkering, I'm hoping that I didn't miss anything horrific.

### DIFF
--- a/evennia/settings_default.py
+++ b/evennia/settings_default.py
@@ -126,9 +126,9 @@ LOCKWARNING_LOG_FILE = os.path.join(LOG_DIR, 'lockwarnings.log')
 # loose log info.
 CYCLE_LOGFILES = True
 # Number of lines to append to rotating channel logs when they rotate
-NUM_LOG_TAIL_LINES = 20
+CHANNEL_LOG_NUM_TAIL_LINES = 20
 # Max size of channel log files before they rotate
-LOG_ROTATE_SIZE = 1000000
+CHANNEL_LOG_ROTATE_SIZE = 1000000
 # Local time zone for this installation. All choices can be found here:
 # http://www.postgresql.org/docs/8.0/interactive/datetime-keywords.html#DATETIME-TIMEZONE-SET-TABLE
 TIME_ZONE = 'UTC'
--- a/evennia/settings_default.py
+++ b/evennia/settings_default.py
@@ -125,6 +125,10 @@ LOCKWARNING_LOG_FILE = os.path.join(LOG_DIR, 'lockwarnings.log')
 # file sizes down. Turn off to get ever growing log files and never
 # loose log info.
 CYCLE_LOGFILES = True
+# Number of lines to append to rotating channel logs when they rotate
+NUM_LOG_TAIL_LINES = 20
+# Max size of channel log files before they rotate
+LOG_ROTATE_SIZE = 1000000
 # Local time zone for this installation. All choices can be found here:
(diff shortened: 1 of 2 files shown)


evennia/settings_default.py
```diff
@@ -125,6 +125,10 @@
 # file sizes down. Turn off to get ever growing log files and never
 # loose log info.
 CYCLE_LOGFILES = True
+# Number of lines to append to rotating channel logs when they rotate
+NUM_LOG_TAIL_LINES = 20
```
Griatch: This is only for channel logs, so that should be reflected in the names, such as `CHANNEL_LOG_FILE_TAILNUM` and `CHANNEL_LOG_MAXSIZE`.

Griatch:
@TehomCD Looks like a good idea! The implementation looks ok to me, just some comments on settings-naming conventions. 
(As a side thought, I wonder how one would go about making a unittest for a mechanism like this?)

evennia/utils/logger.py
```diff
@@ -153,6 +154,25 @@ def log_dep(depmsg):
 
 # Arbitrary file logger
 
+class EvenniaLogFile(logfile.LogFile):
```
Griatch: This needs docstrings.

TehomCD:
Good question on the unit test. Maybe there might be something for mocking a log file of arbitrary size?

evennia/utils/logger.py
TehomCD: Added, let me know if anything else needs to be improved.

evennia/utils/logger.py
```diff
@@ -19,8 +19,9 @@
 import time
 from datetime import datetime
 from traceback import format_exc
-from twisted.python import log
+from twisted.python import log, logfile
 from twisted.internet.threads import deferToThread
+from django.conf import settings
```
Griatch: There is a reason the settings file was not loaded in the header of the logger module - the reason is to keep the logger as separate as possible, so that it can start even in states when django is not yet initiated.

Griatch:
Looks better, I just noticed the settings file is imported in the logger header. I would like to avoid that as mentioned in the comment.

evennia/utils/logger.py
TehomCD: Oh! That makes sense. I changed it to be local imports then.

PR #1329, opened 2017-05-29
Change portal to use Evennia's Website class

TehomCD:
Currently, http logging is still done for every request because while the server uses Evennia's Website class that suppresses these logs unless `DEBUG` is on, the portal does not. This just changes it to use the class as well.

To prevent logging of every single request done to the portal unless DEBUG mode is on so I don't have `http_requests.log.53` waiting for me after a week of not looking.

Other portal versions of the webserver don't use Evennia's classes - for example, the `TCPServer` from portal doesn't use Evennia's threaded webserver variant. Is that an issue at all? I don't really know if that matters at all.

### DIFF
--- a/evennia/server/portal/portal.py
+++ b/evennia/server/portal/portal.py
@@ -294,6 +294,7 @@ if SSH_ENABLED:
 
 
 if WEBSERVER_ENABLED:
+    from evennia.server.webserver import Website
 
     # Start a reverse proxy to relay data to the Server-side webserver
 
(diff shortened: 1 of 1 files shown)


Griatch:
I'm not sure it's a good idea to use the web root for the portal side - the reason the webserver uses the custom class is because we need to do a wsgi hack to serve websites over the threadpool. The portal end is just a dumb forwarding proxy that doesn't need any of that (you could also put a load balancing proxy like ngnx or gunicorn in front of the webserver port). But I need to look at it a bit more in detail.

TehomCD:
Unless I'm missing something, I think in this case the Website class is just site.Site with logging turned off if _DEBUG isn't set and noisy set to False, while it's the web_root resource that's the super customized proxy forwarding thing.

Griatch:
@TehomCD True, I thought of the other one. Merged, thanks!

PR #1331, opened 2017-05-30
Server hang fix (Resolves #1128)

### LINKED ISSUE #1128: Very rarely server hangs indefinitely on 'evennia reload' from console
This is a very rare bug that I have no idea how to reproduce. It's only happened 3 times now out of reloading hundreds of times. What happens is when starting a reload, the only output you'll receive is 'received SIGINT, shutting down'. And then absolutely nothing else will happen - you won't be told ports are closing, no attempt to restart, etc. Attempting to do evennia start will then tell you that the server/portal process is already running, while stopping will tell you that both are stopped, but you'll be informed the server is still running when you try to start. The server will continue until it's wiped with kill -9, and the pid file is removed. It occurs to me that I've never seen this happen with reloading from inside the game, only with attempting to reload with 'evennia reload' from the console.

1. 'evennia reload' and then maybe, just maybe, you might get very unlucky. That's the best I have to go on at the moment.

'Server received SIGINT, shutting down', and then nothing.

The only reason I was thinking of this was all the craziness with the Windows reload process, and wondering if there was any strangeness happening with specifically the console version of reload. Since I have no idea how to reproduce it and it seems very rare, I just wanted to mention it here in case anyone else runs into it and might be able to shed more light on what might cause it to happen.

TehomCD:
So issue #1128 is caused by trying to stop the threadpool when there are current requests, so I added a list to the web resource that keeps track of all current pending requests - a list of deferreds from their `notifyFinish` methods. During server shutdown, it checks a new method that returns a DeferredList from this list, which is then required to clear out before the rest of shutdown can proceed. This resolves crashing from long web requests that are currently pending when shutdown starts.

First attempt to resolve #1128 .

Although I'm reasonably certain this works for pending requests that were there at the start of shutdown, I'm less certain this works for requests made during the final `deferLater` call. I **think** it does, from trying to queue in a bunch of requests at the end for testing, but I'm not positive.

### DIFF
--- a/evennia/server/server.py
+++ b/evennia/server/server.py
@@ -397,8 +397,15 @@ class Evennia(object):
             # this will also send a reactor.stop signal, so we set a
             # flag to avoid loops.
             self.shutdown_complete = True
-            # kill the server
-            reactor.callLater(1, reactor.stop)
+            if WEBSERVER_ENABLED:
+                # Just to be extra sure, get all pending requests that might have occurred after we started
+                d = self.web_root.get_pending_requests()
+                d.addCallback(lambda _: reactor.stop())
+                from twisted.internet import task
+                yield task.deferLater(reactor, 1, d.callback, None)
+            else:
+                # kill the server
+                reactor.callLater(1, reactor.stop)
 
         # we make sure the proper gametime is saved as late as possible
         ServerConfig.objects.conf("runtime", _GAMETIME_MODULE.runtime())
--- a/evennia/server/server.py
+++ b/evennia/server/server.py
@@ -151,7 +151,8 @@ class Evennia(object):
         sys.path.insert(1, '.')
 
         # create a store of services
-        self.services = service.IServiceCollection(application)
+        self.services = service.MultiService()
+        self.services.setServiceParent(application)
         self.amp_protocol = None  # set by amp factory
         self.sessions = SESSIONS
         self.sessions.server = self
@@ -348,6 +349,10 @@ class Evennia(object):
         from evennia.server.models import ServerConfig
         from evennia.utils import gametime as _GAMETIME_MODULE
 
+        if WEBSERVER_ENABLED:
+            # finish all pending web requests. Otherwise stopping threadpool will cause deadlock.
+            yield self.web_root.get_pending_requests()
+
         if mode == 'reload':
             # call restart hooks
             ServerConfig.objects.conf("server_restart_mode", "reload")
@@ -533,11 +538,13 @@ if WEBSERVER_ENABLED:
     # recognized by Django
     threads = threadpool.ThreadPool(minthreads=max(1, settings.WEBSERVER_THREADPOOL_LIMITS[0]),
                                     maxthreads=max(1, settings.WEBSERVER_THREADPOOL_LIMITS[1]))
+
     web_root = DjangoWebRoot(threads)
     # point our media resources to url /media
     web_root.putChild("media", static.File(settings.MEDIA_ROOT))
     # point our static resources to url /static
     web_root.putChild("static", static.File(settings.STATIC_ROOT))
+    EVENNIA.web_root = web_root
 
     if WEB_PLUGINS_MODULE:
         # custom overloads
--- a/evennia/server/webserver.py
+++ b/evennia/server/webserver.py
@@ -28,6 +28,7 @@ from evennia.utils import logger
 _UPSTREAM_IPS = settings.UPSTREAM_IPS
 _DEBUG = settings.DEBUG
 
+
 #
 # X-Forwarded-For Handler
 #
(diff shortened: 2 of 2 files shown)


evennia/server/server.py
```diff
@@ -151,7 +151,8 @@ def __init__(self, application):
         sys.path.insert(1, '.')
 
         # create a store of services
-        self.services = service.IServiceCollection(application)
+        self.services = service.MultiService()
+        self.services.setServiceParent(application)
```
TehomCD: Oh I forgot about this part. When I was tinkering with services to see if there was anything wrong there, I just changed that to mirror a suggestion in some stack overflow post about threadpools hanging. I think it's equivalent code and doesn't make a difference, but didn't seem to hurt anything anyway.

TehomCD:
Added another check to try to protect against more requests made while we were in the process of stopping the reactor, not just from immediately before. Seemed to work when I started the reload during a long request, then kept trying to make a bunch of requests while some were already finishing. Maybe there might still be a period of vulnerability in there during the deferLater part, not sure, but **hopefully** it's now as resilient as it can get.

Griatch:
This sounds promising. I'll run tests as well to check (but probably not until the weekend).

TehomCD:
I was tinkering with it a bit later, and I think I still don't have a great handle on how the whole callback chain works. If I removed the first get_pending_requests that are yielded in inline callbacks, since it seems redundant, the reactor.stop call will never seem to be called at all at the end of the other deferred chain, no matter how you try to add it. No clue why. The requests will finish processing, but reactor stop just never seemed to fire. So maybe I'm missing something obvious there that could make it more elegant, but at least this version seems to function so far.

Griatch:
@TehomCD Hm, this does *not* appear to help with the issue as far as I can tell. I still get a deadlock when trying a long-running request. Ways to reproduce: 

- In `mygame/web/` add a file `myview.py` like this: 
    ```python
    from django.template.loader import get_template
    from django.http import HttpResponse
    import time
    def myview(request):
        template = get_template('myview.html')
        html = template.render()
        time.sleep(20)
        return HttpResponse(html)
    ```
- In `mygame/web/urls.py`, import `myview` and add it as `url(r'myview', myview, name='myview')` to the `custom_patterns` list.
- In `mygame/web/template_overrides/website`  add a file `myview.html` with some placeholder content:
    `<html> <body> <H1>Hello world!</H1> </body> </html>` 
- reload 
- Go to webpage, then go to `/myview`
- While it sleeps (20 seconds delay), go and reload the server. This gives me a deadlock also with this code (I have to kill the server with `kill -9` to get it back). 
Do you see a different result? Do you have some different example that gets solved by this?

The use of `sleep(20)` is extreme but I picture it emulates an extremely expensive synchronous database operation.

I experimented with a different version that modifies the threadpool itself, not allowing it to shutdown without having an empty queue. This does resolve the

Kelketek:
@Griatch I'm not entirely sure I understand everything you're doing there, but on that last point-- could you solve it with an interrupt/signal?

Griatch:
@Kelketek Alas, no the problem lies in the threadpool itself reporting when it has no more pending requests inside it - and since the threadpool is called deep inside the WSGI wrapper there is no obvious way to wait or query this. So @TehomCD's solution is the better one. 

@TehomCD I'm doing some changes based on your code, including adding the ability to lock the threadpool from accepting additional requests. It got it working for my example now, at least for when doing `@reload` in-game. When doing `evennia reload` from outside the game I still have issues; I'll see if I can figure out why. 

Btw, I think the reason you had issues with getting the second `get_pending_requests` to work is probably because the `EVENNIA.shutdown` method is called differently depending on how the server shuts down. It could well be the reactor is already shutting down when reaching this, at which point `_reactor_stopping` will be `True` and your second statement will not be reached.

Griatch:
I think I see what the issue is when trying reloading from the command line when the server is waiting for a request (or is otherwise preoccupied). The issue appears to be that if the program is occupied, the "before shutdown" signal will not fire correctly (this is what calls `EVENNIA.shutdown` which does all the cleanup. The reactor will then not stop, but end up in some sort of weird state ... maybe.

TehomCD:
How I tested it was using the django debug toolbar, which makes most requests take between 5-20 seconds with all the processing it does, and just doing a `@reload` from inside game immediately after starting a request to get the deadlock. The `yield get_pending_requests()` call seemed to resolve that situation, so not sure why time.sleep() would be different. Let me see if I can duplicate that.

Griatch:
@TehomCD I can see it working with `@reload` from inside the game now, but I think I'll have to replace the SIGINT-handler of the Twisted reactor in order to handle the case of `evennia reload` outside the game.

Griatch:
@TehomCD I merged this branch with some reworking/refactoring (based on your work, great research/effort there) as well as support for reloading from the cmdline. It passes my test with a 20s sleep(), please check it out to see if it still behaves as it should for you. If not reply in here. :)

PR #1333, opened 2017-06-03
Collab

### LINKED ISSUE #525: Feature Request: Collaborative Building Tools
We have a large number of people coming over from MUCKs and MUSHes that are used to a more collaborative building environment, where user content is king. The current build commands and configuration aren't conducive for this sort of environment. As Evennia's intended purpose is more of a MUD development kit, it doesn't make sense to make collaborative building the default. However, a well constructed contrib could be made to easily transform Evennia into a collaborative worldbuilding toolkit.

A few things would be needed for this to work right. The first is going to be a concept of object ownership. The standard locks work well for determining if someone should have access to something, but leave something to be desired in terms of 'default powers' in relationship to the object's creator. If someone makes a room, they should be able to change the locks on it, etc, even if the current locks might otherwise say they're unable. It is also useful for keeping track of what objects were created by who. It also shouldn't be possible for a normal user to be able to create something that an immortal is barred from modifying.

This will also inevitably bring up the issues of 'softcode'. Full blown softcode, like MUSHcode, or MPI, is undesirable-- more complex logic should be in the Python code, and managed with merge requests and what have you. But even some basic logic could be added without too much difficulty if we used something like [Jinja](http://jinja.pocoo.org/) to provide for a templating language with conditionals and loops and variable access. This would, effectively, allow for creativity while making the path for more complex interaction clear-- edit the Python code. Jinja supports custom tags, which would allow teams to expand the functionality of the templating system as desired.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2838122-feature-request-collaborative-building-tools?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Kelketek:
This is an updated version of the collab contrib. It is still not quite done-- but the meat of the project is complete. Sandbox building for untrusted users, as well as templating that provides (or can provide, since more filters could be implemented for cases not covered) 90+% of what Softcode is used for without the remaining heavy lifting work we'd like to see in Python.

MUCK/MUSH servers where users can build their own custom areas and objects, as well as customize their characters, are the biggest segment of text MU*s still in use today. In order to have the functionality this market segment expects, the ability to build sandbox-style is required.

This PR would close #525.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -199,11 +199,21 @@ class CmdCopy(ObjManipCommand):
     locks = "cmd:perm(copy) or perm(Builders)"
     help_category = "Building"
 
+    def check_target_location(self, target_location):
+        """
+        Check from object to make sure that it is alright to copy.
+
+        Used for overriding in subclassed commands, and is responsible for displaying
+        its own error messages.
+        """
+        return True
+
     def func(self):
         "Uses ObjManipCommand.parse()"
 
         caller = self.caller
         args = self.args
+        self.copied_obj = None
         if not args:
             caller.msg("Usage: @copy <obj> [=new_name[;alias;alias..]][:new_location] [, new_name2...]")
             return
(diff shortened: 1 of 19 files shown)


Griatch:
@Kelketek So what's the status of this at this time?

Kelketek:
@Griatch Unless someone intends to pick up the torch, I anticipate not continuing at least for a large section of next year.

I have a plan to make it in my monetary interest to continue (as crazy as that sounds) but I have a few things I have to build first before the market for it exists. If it's bugging you, feel free to close the PR again for now.

chainsol:
What work still has to be done on this? I'm happy to add it to the plate, as long as that's agreeable...Just have to know where it's at.

chainsol:
@Kelketek OK, if this is still something that Evennia's userbase would like - I know it's something I think is important - I'd love to build off what you have here.

@Griatch I have a lot on my list, but I'm happy to put this at the top - seems like a huge advantage for those looking for a MUSH server.

Kelketek:
@chainsol I'd be happy to do a knowledge transfer session with you. When would you be available for it?

chainsol:
Haven't forgotten about this, by the way! Holidays got away from me, but I've been working on this and the webclient changes!

PR #1334, opened 2017-06-04
Fix AttributeError when webserver isn't enabled.

TehomCD:
Just a quick fix to make it check if `WEBSERVER_ENABLED` during shutdown, so it doesn't get an AttributeError.

Small fix for people who don't have the webserver enabled.

N/A


Griatch:
Ah, I planned to put that in but then promptly forgot about it. Thanks!

PR #1335, opened 2017-06-05
Change IAC + GA to key off of MUDPROMPT @option

Packetdancer:
Like it says on the tin; add a MUDPROMPT @option and key the IAC + GA on the end of output blobs off of that @option. :)

Several clients interpret IAC + GA as a linefeed, in order to support MUDs that always end a line with them.  This causes Evennia to spew empty lines everywhere if IAC + GA is always on.  Giving the option to toggle it seems a good solution.

### DIFF
--- a/evennia/commands/default/player.py
+++ b/evennia/commands/default/player.py
@@ -560,6 +560,7 @@ class CmdOption(COMMAND_DEFAULT_CLASS):
                       "CLIENTNAME": utils.to_str,
                       "ENCODING": validate_encoding,
                       "MCCP": validate_bool,
+                      "MUDPROMPT": validate_bool,
                       "MXP": validate_bool,
                       "NOCOLOR": validate_bool,
                       "NOPKEEPALIVE": validate_bool,
--- a/evennia/server/inputfuncs.py
+++ b/evennia/server/inputfuncs.py
@@ -193,7 +193,8 @@ def client_options(session, *args, **kwargs):
                            "UTF-8", "SCREENREADER", "ENCODING",
                            "MCCP", "SCREENHEIGHT",
                            "SCREENWIDTH", "INPUTDEBUG",
-                           "RAW", "NOCOLOR"))
+                           "RAW", "NOCOLOR",
+                           "MUDPROMPT"))
         session.msg(client_options=options)
         return
 
--- a/evennia/server/portal/telnet.py
+++ b/evennia/server/portal/telnet.py
@@ -221,7 +221,9 @@ class TelnetProtocol(Telnet, StatefulTelnetProtocol, Session):
         # escape IAC in line mode, and correctly add \r\n
         line += self.delimiter
         line = line.replace(IAC, IAC + IAC).replace('\n', '\r\n')
-        return self.transport.write(mccp_compress(self, line + IAC + GA))
+        if self.protocol_flags.get("MUDPROMPT", False):
+            line = line + IAC + GA
+        return self.transport.write(mccp_compress(self, line))
 
     # Session hooks
 


Griatch:
Hah, I *knew* some clients would handle this differently. I find the name `MUDPROMPT` to not be a very informative name though. I don't have time to add this right now, but maybe we could brainstorm some better `@option` names for this? `NOGOAHEAD` or somthing?

Packetdancer:
Yeah, it causes Atlantis and Mudlet to display extra newlines, and breaks Duckclient and some of the Android clients that aren't aware of IAC GA/EOR badly.

I'm fine calling it something else, honestly, but IAC GA or IAC EOR _is_ technically the MUD prompt protocol.  Hence, having MUDPROMPT set to True to send it seemed appropriate when making the patch... and having it off by default seemed less destructive.  

(Duckclient, for instance, doesn't even know how to parse it, so just reads it as part of each successive line, causing accented characters.)

TehomCD:
I was wondering why someone was reporting an ü at the start of each line. That makes sense.

Packetdancer:
I mean, arguably, an unknown IAC should never be parsed as content if Duckclient supports Telnet IAC _at all_, but that's a different matter.

We probably want to also add support for IAC DO/DONT SUPPRESS-GO-AHEAD in order to turn the field on/off.  At least two of the clients affected by this would have been unaffected if it were sent and honored, so maybe that's a good extension of the patch.  I may work on that tonight.

Griatch:
@Packetdancer I was going to look into this, but if you are adding more things to this PR soon I'll hold off on it for a little longer.

Packetdancer:
I'd say it might be worth adding the Telnet IAC DO/DONT SUPPRESS-GO-AHEAD, since that would've resolved the issue for at least two clients.  But that should be pretty straightforward, as all it will do is manually toggle that protocol flag.

Your call on whether or not it's worth it; I wouldn't be able to get to it for another 7-8 hours due to day-job work.

Griatch:
@Packetdancer I renamed the `@option` to `NOGOAHEAD` and implemented to `SUPPRESS-GO-AHEAD` telnet option. I did the following handshake: 
  Evennia sends `WILL SUPPRESS-GO-AHEAD` - this might catch clients that don't themselves send this but looks out for it.
   Evennia will then monitor any reply to this from the client (`WILL/WONT SUPPRESS-GO-AHEAD`), as well as new sends `DO/DONT SUPPRESS-GO-AHEAD` and abide with whatever the client instructs it to do. 

Testing it with Tintin++, this leads to Tintin++ accepting Evennia's `WILL SUPPRESS-GO-AHEAD` and accepts suppression whereas Mudlet plain replies with `WONT SUPPRESS-GO-AHEAD` as well as sends its own `DONT SUPPRESS-GO-AHEAD` just to be sure (which turns on GA for Mudlet as it should). And the user can of course also change the `@option` and save their custom setting. So this should hopefully cover most bases.

Packetdancer:
That will work fine; Atlantis will say WILL or DO SGA, and anything that doesn't understand that Telnet option probably doesn't support GA anyway, so will be unharmed (and arguably benefited) by GA being suppressed.

Griatch:
... Heh, of course we already have reports that Duckclient apparantly will print that initial `WILL SUPPRESS-GO-AHEAD` from Evennia as a verbatim character. Sigh.

Packetdancer:
Duckclient is not... particularly Telnet compliant, I'm realizing.

Packetdancer:
(That said, I'd argue one extraneous character at initial login is probably less destructive than most other options.)

Griatch:
Just one of many, unfortunately. From the perspective of a server-dev, adding *any* telnet extras will always mess up clients that you never heard of before or which are ancient, but which a lot of people still turn out to be using. :)

Packetdancer:
Plus side, I _am_ working on Atlantis 2.0, and I'm targeting Evennia and MUX2 as the primary 'test' servers. :)

Griatch:
@Packetdancer That is, of course, an excellent choice. ;)

PR #1338, opened 2017-06-06
Docstring typo fixes in evennia_runner and evennia_launcher

BigJMoney:
I didn't open an issue for these, I'm just submitting for pull.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -184,7 +184,8 @@ class CmdCopy(ObjManipCommand):
     copy an object and its properties
 
     Usage:
-      @copy[/reset] <original obj> [= new_name][;alias;alias..][:new_location] [,new_name2 ...]
+      @copy[/reset] <original obj> [= <new_name>][;alias;alias..][
+      :<new_location>] [,<new_name2> ...]
 
     switch:
       reset - make a 'clean' copy off the object, thus
@@ -205,7 +206,8 @@ class CmdCopy(ObjManipCommand):
         caller = self.caller
         args = self.args
         if not args:
-            caller.msg("Usage: @copy <obj> [=new_name[;alias;alias..]][:new_location] [, new_name2...]")
+            caller.msg("Usage: @copy <obj> [=<new_name>[;alias;alias..]]["
+                       ":<new_location>] [, <new_name2>...]")
             return
 
         if not self.rhs:
@@ -446,7 +448,7 @@ class CmdCreate(ObjManipCommand):
     create new objects
 
     Usage:
-      @create[/drop] objname[;alias;alias...][:typeclass], objname...
+      @create[/drop] <objname>[;alias;alias...][:typeclass], <objname>...
 
     switch:
        drop - automatically drop the new object into your current
@@ -684,9 +686,9 @@ class CmdDig(ObjManipCommand):
     build new rooms and connect them to the current location
 
     Usage:
-      @dig[/switches] roomname[;alias;alias...][:typeclass]
-            [= exit_to_there[;alias][:typeclass]]
-               [, exit_to_here[;alias][:typeclass]]
+      @dig[/switches] <roomname>[;alias;alias...][:typeclass]
+            [= <exit_to_there>[;alias][:typeclass]]
+               [, <exit_to_here>[;alias][:typeclass]]
 
     Switches:
        tel or teleport - move yourself to the new room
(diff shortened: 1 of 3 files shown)


BigJMoney:
Aha, I added another commit, for evennia/commands/default/builder.py.  There were a couple of typos but I also made the command help strings more consistent with each other.  I started doing the same for the docstrings and then realized that was probably pointless, but I completed it anyway since I started it.

evennia/commands/default/building.py
```diff
@@ -1558,7 +1562,7 @@ class CmdTypeclass(COMMAND_DEFAULT_CLASS):
     set or change an object's typeclass
 
     Usage:
-      @typclass[/switch] <object> [= <typeclass.path>]
+      @typeclass[/switch] <object> [= typeclass_path]
```
BlauFeuer: I think the dot was to imply that the typeclass path can have a tree-like path with sections separated by dots and be rather long, like `evennia.contrib.tutorial_world.rooms.WeatherRoom` for example.  I can't think of a better way to denote that, though. Your changes here are rather helpful.

Griatch: As @BlauFeuer points out, `typeclass.path` is intentional here, to denote a dot-separated Python path, in the same way that you would write something like `python/path/to/file` if you were referring to a Unix path. I think the `<...>` are also correct in this case.

evennia/commands/default/building.py
```diff
@@ -718,9 +720,10 @@ def func(self):
         caller = self.caller
 
         if not self.lhs:
-            string = "Usage: @dig[/teleport] roomname[;alias;alias...][:parent] [= exit_there"
+            string = "Usage: @dig[/teleport] <roomname>[;alias;alias...][" \
```
Griatch: Whereas leaving the trailing `[` on the line above is a good idea to delimit Python code, I feel it might be confusing the eye when doing it for Command-syntax like this. I much prefer the way you do it in the `@dig` version above, when including the full `[ ... ]` on the line, but indenting the next line. Makes it easier to see what belongs to which without having to resort to counting brackets.

evennia/commands/default/building.py
```diff
@@ -868,7 +871,8 @@ def func(self):
         "Implements the tunnel command"
 
         if not self.args or not self.lhs:
-            string = "Usage: @tunnel[/switch] <direction> [= roomname[;alias;alias;...][:typeclass]]"
+            string = "Usage: @tunnel[/switch] <direction> [= <roomname>[" \
+                     ";alias;alias;...][:typeclass]]"
```
Griatch: Same here, I think the lone `[` belongs on the next line in the case of usage strings. This goes for all cases where this happens.

evennia/commands/default/building.py
```diff
@@ -205,7 +206,8 @@ def func(self):
         caller = self.caller
         args = self.args
         if not args:
-            caller.msg("Usage: @copy <obj> [=new_name[;alias;alias..]][:new_location] [, new_name2...]")
+            caller.msg("Usage: @copy <obj> [=<new_name>[;alias;alias..]]["
+                       ":<new_location>] [, <new_name2>...]")
```
Griatch: See below; I believe the lone `[` belongs on the next line in the case of a user command string example.

Griatch:
Bump. :)

evennia/commands/default/building.py
BigJMoney: Thanks, this was to make it consistent with the doc in line 484 (CmdCreate usage string).  Want me to make them both 'typeclass.path'?

evennia/commands/default/building.py
BigJMoney: Thanks

BigJMoney:
I'm ready to commit the new PR once a see a response about line 484.  I ran a search regex with \[" \\\n and other similar patterns to ensure there aren't any more of those trailing brackets in strings.

evennia/commands/default/building.py
Griatch: Sure, why not!

Griatch:
@BigJMoney  You don't need to commit a new PR, just push to this one. :) I also answered the 484-question.

BigJMoney:
I've committed all my updates but this PR still says "changes requested".  I'm not sure if there's a switch I'm supposed to flick or that just means Griatch will come back in at some point and review again to make sure I addressed them.  Thanks.

Griatch:
That means the reviewer (me, in this case) has not yet changed the status - you cannot change it. I'll try to look this over when I get access to a computer again (is currently on vacations so have limited access to the internet).
.
Griatch

PR #1346, opened 2017-06-15
New Feature: Sekizai integration for expanded template blocks.

DamnedScholar:
By default, Django does not support extending templates adding CSS of their own. Style sheets have to be declared within the `<head>`, so the user has the choice of overriding the default files (in our case, `templates/base.html` and the files in `static/css/`) or not being able to make style tweaks at all. This isn't too bad for game developers, who can copy `base.html` and add new lines to it, but it makes it impossible for developers of modular packages for Evennia (such as [AuroraChain/paxboards](https://github.com/AuroraChain/paxboards)) to include style rules without requiring a file be manually created. However, I believe there to be a better way. There's a package created for just this purpose, [sekizai](https://github.com/ojii/django-sekizai), which expands the Django template language to allow for individual template files to add content to blocks to be rendered elsewhere in the page. The main use of this is to allow CSS files to be added anywhere and rendered up at the top. [I have forked](https://github.com/DamnedScholar/paxboards/tree/sekizai-integration) paxboards to figure it out, and [it works](http://45.33.51.230:8010/boards/) (it also doesn't look like much, but it runs).

# Effects
This change allows for additional tags to be used in extending templates. It will have no effect on users who aren't making use of its features.


Griatch:
Looks interesting, but I'll delay considering this until we have moved to Django 1.11. If you have the time to test this against the `djangoupgrade` Evennia branch it might speed up things for me later.

Griatch:
@chainsol How does this Feature Request gel with your changes to the website?

chainsol:
@Griatch It shouldn't affect anything, I had this in mind when writing the new templates. Just has to be moved around in the templates, I think!

Griatch:
In that case I have no trouble adding support for this so people can easier can make extensions for Evennia with their own styles. @DamnedScholar, do you think you could update this for inclusion into devel branch? You can re-point this PR to devel branch with the instructions [here](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request).

DamnedScholar:
Done. There are some conflicts, so I'll work on resolving those when I get a chance.

Griatch:
@DamnedScholar Are the final conflicts of this close to being resolved soon? Otherwise I'll remove this from inclusion with the `devel` branch.

DamnedScholar:
Because I'm a git noob, I got my repo to a place I wasn't sure how to get out of and nuked it. I have a much more hygienic setup now, and the changes are submitted from the good repo in PR #1434.

PR #1347, opened 2017-06-16
Add the InterruptCommand exception to cleanly interrupt commands

vlegoff:

This pull request adds an `evennia.commands.command.InterruptCommand` exception to terminate commands before the `func()` execution.

For the time being, it's not possible to interrupt a command, let's say, in its `parse` method.  If the parser encounters an error (like a lacking option), `func()` has to handle it.  The `func()` method is called regardless.  This new exception allows to interrupt the command at any moment (in the `parse` method will be the most frequent, I think) before its `func()` is even called.  Its other hooks are therefore not called either.

None

### DIFF
--- a/evennia/__init__.py
+++ b/evennia/__init__.py
@@ -46,6 +46,7 @@ Command = None
 CmdSet = None
 default_cmds = None
 syscmdkeys = None
+InterruptCommand = None
 
 # search functions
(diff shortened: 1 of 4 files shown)


vlegoff:
Good point!  It's now added in the flat API.

PR #1348, opened 2017-06-17
Add the UnixCommand, to parse unix-like command options

vlegoff:

This pull request adds the `UnixCommand` class in `evennia.contrib.unixcommand`.  This class can be used as a parent for commands using Unix syntax.

It can be interesting to have other alternatives for command parsing.  Though the Unix command cannot, obviously, be chosen as a default choice, nor would it be very comfortable for players, it can definitely be useful to builders.

### DIFF
--- a/evennia/contrib/unixcommand.py
+++ b/evennia/contrib/unixcommand.py
@@ -8,7 +8,7 @@ options.
 
 The UnixCommand can be ovverridden to have your commands parsed.
 You will need to override two methods:
-- The `init` method, which adds options to the parser.
+- The `init_parser` method, which adds options to the parser.
 - The `func` method, called to execute the command once parsed.
 
 Here's a short example:
@@ -30,7 +30,7 @@ class CmdPlant(UnixCommand):
 
     key = "plant"
 
-    def init(self):
+    def init_parser(self):
         "Add the arguments to the parser."
         # 'self.parser' inherits `argparse.ArgumentParser`
         self.parser.add_argument("key",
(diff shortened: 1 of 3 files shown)


evennia/contrib/unixcommand.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+class ParseError(Exception):
+
+    """An error occurred during parsing."""
+
+    pass
+
+
+class EvenniaParser(argparse.ArgumentParser):
```
Griatch: This belongs *above* the functions/classes that use it.

evennia/contrib/unixcommand.py
```diff
@@ -0,0 +1,211 @@
...
+        try:
+            self.opts = self.parser.parse_args(shlex.split(self.args))
+        except ParseError as err:
+            msg = str(err)
+            if msg:
+                self.msg(msg)
+            raise InterruptCommand
+
+
+class ParseError(Exception):
```
Griatch: This all belongs at the top of the module, before the functions/classes that use it.

evennia/contrib/unixcommand.py
```diff
@@ -0,0 +1,211 @@
...
+        if self.command:
+            self.command.msg(self.format_usage().strip())
+
+    def print_help(self, file=None):
+        """Print the help to the caller."""
+        if self.command:
+            self.command.msg(self.format_help().strip())
+
+
+class HelpAction(argparse.Action):
```
Griatch: Again, belongs above the class in which it is used. Best to make a section up top with the support resources.

evennia/contrib/unixcommand.py
```diff
@@ -0,0 +1,211 @@
...
+                conflict_handler='resolve', add_help=False, **kwargs)
+        self.command = command
+        self.post_help = epilog
+        def n_exit(code=None, msg=None):
+            raise ParseError(msg)
+
+        self.exit = n_exit
+
+        # Replace the -h/--help
+        self.add_argument("-h", "--hel", nargs=0, action=HelpAction, help="display heeeelp")
```
Griatch: Is it intentional to write this as "heeeelp" or did the finger slip? :)

evennia/contrib/unixcommand.py
```diff
@@ -0,0 +1,211 @@
...
+    def format_usage(self):
+        """Return the usage line."""
+        return raw(super(EvenniaParser, self).format_usage())
+
+    def format_help(self):
+        """Return the parser help, including its epilog."""
+        autohelp = raw(super(EvenniaParser, self).format_help())
+        return "\n" + autohelp + "\n" + self.post_help
+
+    def print_usage(self, file=None):
```
Griatch: These methods does not appear to be used. If they are included for convenience, they should have full doc strings. What does `file` do, how can it be used?

evennia/contrib/unixcommand.py
```diff
@@ -0,0 +1,211 @@
...
+    """An error occurred during parsing."""
+
+    pass
+
+
+class EvenniaParser(argparse.ArgumentParser):
+
+    """A parser just for Evennia."""
+
+    def __init__(self, prog, description="", epilog="", command=None, **kwargs):
```
Griatch: This may be a variant on `ArgumentParser` but it still requires a docstring for those reading the code to easy see what is going on.

evennia/contrib/unixcommand.py
```diff
@@ -0,0 +1,211 @@
...
+        # Create the empty EvenniaParser, inheriting argparse.ArgumentParser
+        lines = dedent(self.__doc__.strip("\n")).splitlines()
+        description = lines[0].strip()
+        epilog = "\n".join(lines[1:]).strip()
+        self.parser = EvenniaParser(None, description, epilog, command=self)
+
+        # Fill the argument parser
+        self.init()
+
+    def init(self):
```
Griatch: I can see the need for separating the configuration of the parser from the `parse` step, but I think the name `init_parser` may be a clearer name for this method than just `init`.

evennia/contrib/unixcommand.py
```diff
@@ -0,0 +1,211 @@
...
+    - `func`: this method is called to execute the command, but after
+      the parser has checked the arguments given to it are valid.
+      You can access the namespace of valid arguments in `self.opts`
+      at this point.
+
+    The help of UnixCommands is derived from the docstring, in a
+    slightly different way than usual: the first line of the docstring
+    is used to represent the program description (the very short
+    line at the top of the help message).  The other lines below are
+    used as the program's "epilog", displayed below the options.  It
```
Griatch: I believe it's spelled `epilogue`, but I also think this is not a very good name for it. Something like "args_help" or "args_help_string" would be more representative IMO.

evennia/contrib/unixcommand.py
```diff
@@ -0,0 +1,211 @@
...
+    means in your docstring, you don't have to write the options.
+    They will be automatically provided by the parser and displayed
+    accordingly.  The `argparse` module provides a default '-h' or
+    '--help' option on the command.  Typing |whelp commandname|n will
+    display the same as |wcommandname -h|n, though this behavior can
+    be changed.
+
+    """
+
+    def __init__(self, **kwargs):
```
Griatch: I think it would be easier to read if the info about the possible arguments were moved from the class docstring above and into the docstring for `__init__` instead. That way autodocs will be easier to read.

evennia/contrib/unixcommand.py
Griatch: Not sure the `EvenniaParser` is a good name for this, it confuses it with the *actual* evennia parser in `evennia/commands/parser.py`. Maybe something like `UnixCommandParser` instead?

evennia/contrib/unixcommand.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def print_help(self, file=None):
+        """Print the help to the caller."""
+        if self.command:
+            self.command.msg(self.format_help().strip())
+
+
+class HelpAction(argparse.Action):
+
+    """Override the -h/--he.p."""
```
Griatch: It would be useful with some more explanation of what this does and why it is necessary.

vlegoff:
Okay I updated what I could.  There are a few questions/answers however, and I'm not sure how to create in-line comments.

- The class `UnixCommandParser` now has a proper docstring.  However I don't detail arguments there, for they are the same as its parent, there are MANY of them, and I don't recommend creating one directly.  That being said, if you believe it's best to have these arguments documented too, I'll do it.
- I kept the epilogue`""because it's used in the parent parser, as describing the help text below the list of options.  I have to treat it separately because of the support of color codes.
- You mentioned readability for the `UnixCommand` docstring to be moved in `__init__`.  Is it all the documentation?  You refer to arguments, but actually I don't document any (there are command arguments, I didn't check them at all, I must admit).

The rest of suggestions has been addressed in the recent commit, I hope anyway.  Thanks for the review!

Griatch:
@vlegoff  Inline comments are done by selected by starting a review (don't think you can do that on your own code). You can however reply directly on each inline comment; there is a reply button for each but not sure how accessible it is. 

- You don't need to duplicate the docstring for all sub-classes. `ArgumentParser` has indeed a lot of options, so you should either include a selection of them or at least give a link to the Python docs mentioning that's where you need to look for more options.
- Aha, `epilog` is actually what it is called in Argparser itself. Fair enough then. 
- `UnixCommand.__init__` should probably mention that additional kwargs will be available as properties on the command at run-time. This is seldomly used, but if you codedly create command instances you can use this to customize dynamic command constructs. You can just copy the doctring from the Command parent [here](https://github.com/evennia/evennia/blob/master/evennia/commands/command.py#L164).

vlegoff:
@Griatch: that should all be fixed, I hope.  Thanks for the review.

Griatch:
Thanks for the contrib! Looking good! I updated the contrib README and also took the liberty to add your name to the top of the unixcommand header. :)

PR #1349, opened 2017-06-18
Add kwargs for hooks to move_to

TehomCD:
`announce_move_from` and `announce_move_to` allow you to pass mappings and other useful kwargs to them in order to override their behavior, but `move_to` itself does not take these arguments and does not pass them to the hooks. I just added them to `move_to` arguments and pass them along to the hooks.

Allow `announce_move_to` and `announce_move_from` hooks to take advantage of their mapping and msg arguments.

N/A

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -614,7 +614,8 @@ class DefaultObject(with_metaclass(TypeclassBase, ObjectDB)):
             obj.msg(text=(outmessage, outkwargs), from_obj=from_obj, **kwargs)
 
     def move_to(self, destination, quiet=False,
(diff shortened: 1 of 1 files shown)


Griatch:
This is a good idea, but I think we may be increasing the complexity of these hooks while still keeping them limited to future expansion. I suggest that instead of adding these explicit keywords we should pass on **kwargs throughout the structure. That way we can look for the `mapping` and `map` and use those if available but an endeavouring dev can easily just pass extra stuff in too if they override, say, `at_move` too.

TehomCD:
Good point. Done!

PR #1350, opened 2017-06-20
Rename the event system into in-game Python system

vlegoff:

This pull request simply renames the "event" contrib to "ingame_python".

It's probably much clearer this way.  The documentation has also been updated.

#### Process to upgrade existing events

For those of you already running the event system in your game, there are a few steps to follow to still have access to your callbacks:

1. Backup your database before merging this pull request (or pulling from master).
2. Before doing an "evennia start", run an "evennia shell" instead.  The path of the event handler has changed, we need to update it, otherwise, nobody will be able to connect.

In the `evennia shell`, enter the following lines:

```python
import evennia
script = evennia.ScriptDB.objects.get(db_key="event_handler")
script.swap_typeclass("evennia.contrib.ingame_python.scripts.EventHandler", run_start_hooks=None)
```

> This change the typeclass_path of the event handler.  In theory, the same thing should be done with EventTimeScript, but for some reason, the system doesn't appreciate us using `swap_typeclass` on these scripts.  For them, it seems it will be necessary to stop them and re-create the callback, though perhaps another strategy can be used.

Check that you have all your callbacks.  The `swap_typeclass` shouldn't clear information in this mode with no running hooks, but better safe than sorry.

### DIFF
--- a/evennia/contrib/ingame_python/README.md
+++ b/evennia/contrib/ingame_python/README.md
@@ -1,9 +1,9 @@
-# Evennia event system
+# Evennia in-game Python system
 
 Vincent Le Goff 2017
 
-This contrib adds the system of events in Evennia, allowing immortals (or other trusted builders) to
-dynamically add features to individual objects.  Using events, every immortal or privileged users
+This contrib adds the system of in-game Python in Evennia, allowing immortals (or other trusted builders) to
+dynamically add features to individual objects.  Using custom Python set in-game, every immortal or privileged users
 could have a specific room, exit, character, object or something else behave differently from its
 "cousins".  For these familiar with the use of softcode in MU`*`, like SMAUG MudProgs, the ability to
 add arbitrary behavior to individual objects is a step toward freedom.  Keep in mind, however, the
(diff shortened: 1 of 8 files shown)


Griatch:
This looks good. I'm not sure about the inability to swap the scripts though and don't have time to check it at the moment. Is it a bug, what kind of error do you get?

vlegoff:
@Griatch You can merge it nevertheless, I think.  We are only two users of the event system (in-game Python system).  I can't test again because I lost my database file with the old system, but here's the culprit, if you're curious:

```python
# in evennia/typeclasses/models.py, line 451
    def swap_typeclass(self, new_typeclass, clean_attributes=False,
                       run_start_hooks="all", no_default=True, clean_cmdsets=False):
        # ...
        if inherits_from(self, "evennia.scripts.models.ScriptDB"):
            if self.interval > 0:
                raise RuntimeError("Cannot use swap_typeclass on time-dependent " \
                                   "Script '%s'.\nStop and start a new Script of the " \
                                   "right type instead." % self.key)
```

That's the error for sure.  The EventHandler could be swapped because it has no time dependency, but `TimeScriptEvent` do have time dependency.  Not a big deal though, as I was saying.

chainsol:
Me being the second, I can handle swapping it over, so I have no complaints.

Griatch:
Ah, right - yes that is a safety check that need to be there. Merging then. :)

PR #1351, opened 2017-06-21
Fix weird lag issue with db_model (Resolves #1173)

### LINKED ISSUE #1173: Bug: Lag with newest change of Tag/Aliashandler
After updating with the latest changes, the game immediately began to have crippling lag issues where anything with a sizable number of aliases/tags were concerned. For example, a 'look' command might take 20-30 seconds or the like. Speculation is this is due to the sheer number of tags that were in my database, which is in the tens of thousands range. Watching in `htop` would see the CPU usage for the server process continually spiking to 100%, and commands would print out lines from their return with noticeably delays between each line. (such as an `evtable` taking a quarter of a second between each line of the table displaying).

1. Compare 'look' or other commands that reference tags to others that don't

Commands where they check tags/aliases are vastly slower for no apparent reason.

Latest change to the tags/aliashandler. SQLite database. 17746 tags in the database. Test room had around 80 or so aliases between the different objects.

Additional note: On my test server, I initially saw this behavior for around a half hour, and then it just spontaneously stopped. So it might be related to loading the AliasHandler/TagsHandler, possibly? I had 13 other players on my live server when I saw this behavior, so it could have just been enough people put enough strain on it that the situation continued indefinitely, but might have eventually resolved on its own.

Another edit: I could partially reproduce this on my test environment when doing a `@tag/search <tagname>` that would return a few hundred matches and then immediately doing a `look` - I'd see a noticeable slowdown for several seconds. It seemed to resolve itself immediately after. Maybe the login process for characters is producing something similar, but with a much greater duration?

TehomCD:
So while playing around with the incredibly weird bug that only seemed to affect me (issue #1173 ), where db_model added to a query with db_attrtype and/or db_category would produce game-stopping lag, I randomly read somewhere about LIKE statements possibly interfering with the performance of an index in some types of databases, and saw django's notes about the __iexact/exact flags using LIKE. So just as a random experiment, I tried adding __iexact to db_model, and the lag issue resolved instantly. Now, this surely seems like a bizarre bug somewhere, and I have zero idea **why** this should work, but here we are.

Fix crippling lag issue that cropped up in one instance of an SQLite database. Still don't know root cause.

Resolves #1173

### DIFF
--- a/evennia/typeclasses/attributes.py
+++ b/evennia/typeclasses/attributes.py
@@ -219,7 +219,7 @@ class AttributeHandler(object):
     def _fullcache(self):
         """Cache all attributes of this object"""
         query = {"%s__id" % self._model: self._objid,
-                 "attribute__db_model": self._model,
+                 "attribute__db_model__iexact": self._model,
                  "attribute__db_attrtype": self._attrtype}
         attrs = [conn.attribute for conn in getattr(self.obj, self._m2m_fieldname).through.objects.filter(**query)]
         self._cache = dict(("%s-%s" % (to_str(attr.db_key).lower(),
@@ -273,7 +273,7 @@ class AttributeHandler(object):
                     return []  # no such attribute: return an empty list
             else:
                 query = {"%s__id" % self._model: self._objid,
-                         "attribute__db_model": self._model,
+                         "attribute__db_model__iexact": self._model,
                          "attribute__db_attrtype": self._attrtype,
                          "attribute__db_key__iexact": key.lower(),
                          "attribute__db_category__iexact": category.lower() if category else None}
@@ -298,7 +298,7 @@ class AttributeHandler(object):
             else:
                 # we have to query to make this category up-date in the cache
                 query = {"%s__id" % self._model: self._objid,
-                         "attribute__db_model": self._model,
+                         "attribute__db_model__iexact": self._model,
                          "attribute__db_attrtype": self._attrtype,
                          "attribute__db_category__iexact": category.lower() if category else None}
                 attrs = [conn.attribute for conn


TehomCD:
I mean, you'd think that if it was some issue with `LIKE` statements causing lag, it'd be caused by something being a `LIKE` rather than being `=`, right? But no. Here, turning something into a `LIKE` from a `=` resolved it. To make it even more weird, the `=` rather than a `LIKE` is just fine provided the `db_category` and `db_attrtype` checks weren't there. So I have zero clue how it all fell together that way, and whether it's only some error in my specific database, or what.

TehomCD:
Reading a little in the [SQLite](http://www.sqlite.org/optoverview.html) page to see if I could get any hint what was going on, I read `Each table in the FROM clause of a query can use at most one index ... When faced with a choice of two or more indices, SQLite tries to estimate the total amount of work needed to perform the query using each option. It then selects the option that gives the least estimated work.` So what I'm wondering now is that for some reason, it made a terrible error when selecting which index to use when given those choices, and switching to an __iexact match removed that index as a possibility, causing it to choose the correct index to use for the query.

TehomCD:
I also finally saw an explanation for why this only cropped up in my test environment for when I upgraded past 3.6 so that I could use WAL mode: they changed to [The Next Generation Query Planner](http://www.sqlite.org/queryplanner-ng.html) in 3.8, so queries suddenly were planned differently. So it seems logical that this was some bug in query planning that exists in the NGQP but not in a previous version. Maybe I could have resolved it with the steps that they list there on that page, possibly, like running `ANALYZE` on the attributes table, if that would have been used by the query planner then to build a better query. But I'm happy that the problem can be solved as easily as just setting `db_model` to use `__iexact` so that it defaults to a better index, if that's what's going on.

Griatch:
Excellent work tracking this one down, I sure could never reproduce it.

PR #1359, opened 2017-07-02
Allow to change suggestion_cutoff and suggestion_maxnum in help

vlegoff:

This PR offers to override the `suggestion_cutoff` and `suggestion_maxnum` in help, without re-writing the entire command.  They now are class variables one can either override from a distance, or just redefine in a class, adding the command as usual.

This allows to simply change the number of suggestions, and the accuracy of them.  `suggestion_maxnum` can be set to 0, which means no suggestion is offered at all.


Griatch:
Thanks, it's a good change. :)

PR #1361, opened 2017-07-13
Default to Bootstrap for Evennia Website

chainsol:
Adds Bootstrap v4 as the CSS framework for the default website of Evennia

Ease of use for people new to Evennia/Django, allows responsive-by-default designs.

I have only one complaint about this: Bootstrap v4 is alpha. That's why I've included the files rather than use one of the free and numerous CDN options that host Bootstrap, so we can have a particular version being used.

See in action: [dev.perfectdisorder.net](http://dev.perfectdisorder.net)

### DIFF
--- a/evennia/web/website/static/website/css/app.css
+++ b/evennia/web/website/static/website/css/app.css
@@ -44,8 +44,8 @@ html {
 }
 
 body {
-  /* Margin bottom by footer height */
-  margin-bottom: 90px;
+  /* Margin bottom by footer height plus padding against main content*/
+  margin-bottom: 92px;
   background-color: #fefefe;
 }
 
@@ -60,13 +60,16 @@ body {
 }
 
 .navbar-brand-logo {
-    height: 50px;
-    width: 50px;
+    height: 64px;
+    width: 64px;
 }
 
 /* Default Colors */
 .main-content {
     margin-top: 32px;
-    margin-bottom: 32px;
     background-color: white;
+}
+
+.navbar {
+  background-color: #3d5c7a;
 }
--- a/evennia/web/website/static/website/css/app.css
+++ b/evennia/web/website/static/website/css/app.css
@@ -56,7 +56,7 @@ body {
   /* Set the fixed height of the footer here */
   height: 60px;
   line-height: 60px; /* Vertically center the text there */
-  background-color: #f5f5f5;
+  background-color: #7590ae;
 }
 
 .navbar-brand-logo {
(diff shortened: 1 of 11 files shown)


Griatch:
Thanks for the PR! I have so far only glanced at the code, but I like the amount of comments and helpful info in there - looks very good so far, and it's certainly responsive too! 

I have a few first points: 

- I understand your reasoning for including the bootstrap-v4 here, I know because I once had similar ideas about including just the right version of `jquery` for the webclient. I have however come to the conclusion that it's not a good idea to add 14 000 lines of code we have no idea what it is and have no intent of ever modifying or maintaining. Should we ever find ourself unable to access this data it's time to look around for another js lib. 
- The website look is simple and functional - so simple that at first glance it looks like this is an unformatted html page with no CSS at all ;). I like the clean-ness of it overall, but maybe make that header block have a darker shade of gray or blue, just to show it's there? 
- On my screen at least (which is a 13 inch laptop), the server stats appear under the scroll. The old layout had everything on the same page on most screens. Maybe move the default content about Evennia (the second paragraph) into a second right-shifted column on the side, or remove it completely in the worst case?
.
Griatch

chainsol:
- I can use the Bootstrap CDN, then. It'll strip all those lines, and the CDN is unlikely to go away. The code won't be likely to have breaking changes.
- Yes, styling is good. :] I'll add some nicer default styles - I'll stick with the old site's color scheme
- I can make the flow change, yes. That's the nice thing about a setup like this - it's a matter of changing one css class on that element.
I'll push those changes so you can take a look again.

chainsol:
Griatch, take a gander, I've addressed those concerns.

Griatch:
@chainsol It looks really nice on my phone I like the colour scheme and flow! I'll look at it more when getting to a computer (on vacations, so has more limited access to internet); but looks very good as far I can see!

Griatch:
@chainsol This looks good! The website was long overdue an overhaul with something custom-made for Evennia! So the new dependencies are tether and bootstrap if I read it right (or was tether already used)? Anyway, the result looks fine to me and the code seems easy to read. 

Mulling it over, I think I will start with merging this into devel branch and incorporate it into Evennia 0.7 though - to have it appear along with other big changes. So it will be a little while still before this goes into general use. When it does, would you be willing to document in the wiki how a beginner could extend the website (with a new tutorial or extending an existing one maybe)?

chainsol:
Yup, Bootstrap and Tether. Tether is actually a dependency of Bootstrap, though.

And of course! I'd be happy to write a tutorial on how to use Bootstrap. If you were to use one of the existing tutorials this change won't mess with it at all - everything would still work.

chainsol:
@Griatch, I'm going to close this now that it's implemented, but I also pushed another PR to fix a bug the account/player change introduced. #1374 fixes said bug.

Issue #1341, opened 2017-06-11
Feature request: a persistent utils.delay()


vlegoff:
#### Description of requested feature:

`utils.delay()` is great, but it gets erased when the game reloads or shuts down.  I suggest a persistent `utils.delay()` to solve this.  If accepted I would be happy to work on this to integrate to Evennia vanilla.

#### Suggested implementation:

The event system offers a similar approach when it creates differed tasks: it will call a callback in a specified time with specified arguments.  If the game reloads or shuts down, it will just call the callback whenever possible, keeping track of time.

This system is less simple than `utils.delay()`, which simply create a Twisted differ (actually, the event system uses `utils.delay()` for its task system).  Having a permanent `utils.delay()` would mean storing both the callback and arguments in the database, with the usual problems regarding persistence:

- Not all callbacks can be saved.  Top-level functions or instance methods should be okay.
- Not all arguments can be saved, following the usual strategy for attributes.

In other words, we could have something like this:

1. When calling the persistent `utils.delay()`, an entry will be stored in database containing the callback and the arguments if it has been able to pickle them.
2. The normal `utils.delay()` will then be called to create the differed task.
3. When the server is started (after a reload or shutdown), the entries in the database would be checked.  If they should have fired but haven't, then a normal `utils.delay()` will be called again.

#### Questions:

Assuming this implementation is accepted, there are some questions to consider:

1. Where to store the entries?  A script could be the solution but it would require to start the script in order to use the feature.  Not an "out-of-the-box" feature.
2. How to call it?  We must keep `utils.delay()` as it is and just add a layer.  Probably not a keyword (`persistent=False`) like on a lot of features, since `utils.delay()` accepts random keyword arguments, and that might break compatibility for some users.  Besides, having a non-persistent `utils.delay()` makes sense, again, not to break compatibility.  So another function in `utils.py` might be the solution, not sure how to call it.

Thank you for your feedback,

Griatch:
A persistent `utils.delay` fulfilling your requirements is called a *Script* and had been in Evennia much longer than `utils.delay` ever was. Is there a reason you can't use `create_script` for this use?

vlegoff:
[...]

PR #1371, opened 2017-07-20
Add the persistent utils.delay

vlegoff:

This PR allows a `persistent` keyword argument to be set in the `evennia.utils.delay` function.  This allows to create persistent tasks that are stored in the `ServerConfig` and will be executed even after a reload or restart.

Motivations for this pull request are detailed in the [issue #1341](https://github.com/evennia/evennia/issues/1341).

#### Notes

- I didn't include tests.  How can we test it?  Can we track deferred and force them to execute somehow?
- It could add a bit more security.  If the object is destroyed in between, I'm not sure what happens.
- It uses a "tweak" to allow saving of instance methods.  When an instance method is detected in a callback, it will save a tuple (the object, the method name) and will restore it in loading the configuration.  I was wondering if perhaps it could be useful to add it in the serializer itself, but maybe not practical.

- (Issue #1341)[https://github.com/evennia/evennia/issues/1341]


Griatch:
@v-legoff This looks good wi. I am not sure about the need for the `utils.persistent` module though. I'd be interested in hearing your arguments for having it over letting the `delay()` function handle things.

Griatch:
After discussion in IRC, it was agreed the persistent-class better be reworked into a new `TaskHandler` to  go into `evennia/scripts/taskhandler.py`. The `utils.delay(... persistent=True)` then just becomes a wrapper for calling this handler, which also opens up the handler for other uses by other systems.

vlegoff:
@Griatch: I have moved the persistent mechanism in the task handler, which is called for persistent and non-persistent tasks by `utils.delay` now.

evennia/scripts/taskhandler.py
```diff
@@ -0,0 +1,191 @@
...
+    def create_delays(self):
+        """Create the delayed tasks for the persistent tasks.
+
+        Note:
+            This method should be automatically called when Evennia starts.
+
+        """
+        now = datetime.now()
+        for task_id, (date, callbac, args, kwargs) in self.tasks.items():
+            seconds = (date - now).total_seconds()
```
Griatch: Minor quibble: The standard way to make sure a value is bigger or equal to zero is to do `seconds = max(0, (data - now).total_seconds())`. You can then remove the `if` statement.

vlegoff:
Fixed!  Thganks for the trick, I learned something.

PR #1373, opened 2017-07-22
Add the generator contrib

vlegoff:

This pull request adds the `generator` contrib, to generate pseudo-random strings following user-defined rules.  This generator system could be useful to generate phone number, license plate numbers, non-sensitive passwords and such.

When one wants to generate pseudo-random strings that shouldn't repeat, it's necessary to add some code to make sure repetition doesn't happen.  Storing tags on objects (and obtaining the list of tags afterward) is a good solution, and the generator contrib doesn't rule it out.  The `Generator` class will just take care of pseudo-random generating unique strings.  The developer can do whatever she wants with this string afterward, create a tag or store it as an attribute on the object.

The `Generator` class functions by rules, defined in regular expressions.  Fair examples are given in the code itself, and the contrib documentation.


evennia/contrib/generator.py
```diff
@@ -0,0 +1,340 @@
...
+        Returns:
+            strings (list of strr): the list of strings that are already
+            used.  The strings that were generated first come first in the list.
+
+        """
+        script = self._get_script()
+        generated = list(script.db.generated.get(self.name, []))
+        return generated
+
+    def generate(self, store=True, keep_trying=True):
```
Griatch: This should better be called `get` to be in line with other handler-type structures in Evennia.

evennia/contrib/generator.py
```diff
@@ -0,0 +1,340 @@
...
+                epoch += 1
+                seed(epoch)
+                result = self.generate(store=False, keep_trying=False)
+
+        if store:
+            generated.append(result)
+
+        return result
+
+    def free(self, element):
```
Griatch: This should be named `remove` to be in line with other handler-style systems in Evennia.

evennia/contrib/generator.py
```diff
@@ -0,0 +1,340 @@
...
+    def clear(self):
+        """
+        Clear the generator of all generated strings.
+
+        """
+        script = self._get_script()
+        generated = script.db.generated.get(self.name, [])
+        generated[:] = []
+
+    def _get_script(self):
```
Griatch: These helper methods belong at the top of the class. Put the non-private methods that makes use of these private methods (like `get`, `remove` etc) *after* the helpers that they use.

evennia/contrib/tests.py
```diff
@@ -985,3 +985,24 @@ def test_failure(self):
         lines = ret.splitlines()
         self.assertTrue(any(l.startswith("usage:") for l in lines))
         self.assertTrue(any(l.startswith("dummy: error:") for l in lines))
+
+
+from evennia.contrib import generator
+
+SIMPLE_GENERATOR = generator.Generator("simple", "[01]{2}")
+
+class TestGenerator(EvenniaTest):
```
Griatch: Nice unit test!

vlegoff:
It should all be fixed.  I tend to put private methods at the bottom in an attempt to hide their content, which isn't always pretty (the `_find_eleents` and `_find_literal` in particular).  Users would just have to browse a bit to see the interesting methods, that's no problem.  The method names have been updated as well.

evennia/contrib/generator.py
```diff
@@ -0,0 +1,340 @@
...
+# Will return a list of all currently-used phone numbers
+phone_generator.remove("555-981-2207")
+# The number can be generated again
+```
+
+To use it, you will need to:
+
+1. Import the `Generator` class from the contrib.
+2. Create an instance of this class taking two arguments:
+   - Tje name of tje gemerator (like "phone number", "license plate"...).
```
j-f1: The?

Griatch:
@vlegoff The changes look good! But as said above, I have one main beef with it - the name. generator.py means nothing to me. It is a random string generator, so why not just call it random_string_generator.py?

evennia/contrib/random_string_generator.py
```diff
@@ -0,0 +1,340 @@
...
+phone_generator.remove("555-981-2207")
+# The number can be generated again
+```
+
+To use it, you will need to:
+
+1. Import the `Generator` class from the contrib.
+2. Create an instance of this class taking two arguments:
+   - The name of tje gemerator (like "phone number", "license plate"...).
+   - The regular expression representing the expected results.
```
Griatch: Here it could be good to give a link to the Python regex documentation: `https://docs.python.org/2/library/re.html`

evennia/contrib/random_string_generator.py
```diff
@@ -0,0 +1,340 @@
...
+        # Analyze the regex if any
+        if regex:
+            self._find_elements(regex)
+
+    def __repr__(self):
+        return "<evennia.contrib.generator.Generator for {}>".format(self.name)
+
+    def _get_script(self):
+        """Get or create the script."""
+        if type(self).script:
```
Griatch: This is a nice trick, I had to test for myself just what the effect of this would be, interesting!

evennia/contrib/random_string_generator.py
```diff
@@ -0,0 +1,340 @@
...
+        generated = list(script.db.generated.get(self.name, []))
+        return generated
+
+    def get(self, store=True, keep_trying=True):
+        """
+        Generate a pseudo-random string according to the regular expression.
+
+        Args:
+            store (bool, optional): store the generated string in the script.
+            keep_trying (bool, optional): keep on trying if the string is already used.
```
Griatch: This should be renamed to `unique` I think, to be in line with other uses.

evennia/contrib/random_string_generator.py
```diff
@@ -0,0 +1,340 @@
...
+        Remove a generated string from the list of stored strings.
+
+        Args:
+            element (str): the string to remove from the list of generated strings.
+
+        Note:
+            The specified string has to be present in the script (so
+            has to have been generated).  It will remove this entry
+            from the script, so this string could be generated again by
+            calling the `get` method.
```
Griatch: Add a `Raises:` documentation block since this method raises `ValueError` if string is not found.

vlegoff:
@Griatch: I believe I fixed it all.  I also updated the tests.  Thanks for the review!

PR #1374, opened 2017-07-23
Fix Player -> Account in Website Templates

chainsol:
After being moved to the Devel branch, the new base templates still referenced Players instead of Accounts. This PR fixes those references, adding back in the player list and connected/registered counts.

See above - fixes bug introduced via the merge into Devel


Griatch:
This is indeed due to the rename program not going into `.html` files. Should fix that for those that have modified their websites.

PR #1377, opened 2017-07-25
Allow contrib.ingame_python.scripts.EventHandler.get_events to work w…

vlegoff:
…ith typeclasses

The in-game Python global script can now work with objects and typeclasses.

This allows to test the events of a typeclass without needing its objects.  This is useful to explore the system itself without relying on individual objects.

### DIFF
--- a/evennia/contrib/ingame_python/scripts.py
+++ b/evennia/contrib/ingame_python/scripts.py
@@ -10,7 +10,7 @@ import traceback
 
 from django.conf import settings
 from evennia import DefaultObject, DefaultScript, ChannelDB, ScriptDB
-from evennia import logger
+from evennia import logger, ObjectDB
 from evennia.utils.ansi import raw
 from evennia.utils.create import create_channel
 from evennia.utils.dbserialize import dbserialize
@@ -101,7 +101,7 @@ class EventHandler(DefaultScript):
         Return a dictionary of events on this object.
 
         Args:
-            obj (Object): the connected object.
+            obj (Object or typeclass): the connected object or typeclass.
 
         Returns:
             A dictionary of the object's events.
@@ -115,7 +115,11 @@ class EventHandler(DefaultScript):
         events = {}
         all_events = self.ndb.events
         classes = Queue()
-        classes.put(type(obj))
+        if isinstance(obj, ObjectDB):
+            classes.put(type(obj))
+        else:
+            classes.put(obj)
+
         invalid = []
         while not classes.empty():
             typeclass = classes.get()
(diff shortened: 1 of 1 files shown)


evennia/contrib/ingame_python/scripts.py
```diff
@@ -115,7 +115,11 @@ def get_events(self, obj):
         events = {}
         all_events = self.ndb.events
         classes = Queue()
-        classes.put(type(obj))
+        if isinstance(obj, ObjectDB):
```
Griatch: `isinstance` will only detect if you pass the actual `ObjectDB` class in there. Use `utils.utils.inherits_from` instead, that will catch any class inheriting at any distance from `ObjectDB`.

evennia/contrib/ingame_python/scripts.py
```diff
@@ -101,7 +101,7 @@ def get_events(self, obj):
         Return a dictionary of events on this object.
 
         Args:
-            obj (Object): the connected object.
+            obj (Object or typeclass): the connected object or typeclass.
```
Griatch: This seems inaccurate. Only the connected object or an ObjectDB instance is allowed as-is.

evennia/contrib/ingame_python/scripts.py
```diff
@@ -115,7 +115,11 @@ def get_events(self, obj):
         events = {}
         all_events = self.ndb.events
         classes = Queue()
-        classes.put(type(obj))
+        if isinstance(obj, ObjectDB):
+            classes.put(type(obj))
+        else:
+            classes.put(obj)
```
Griatch: I think a confusing error would be to put a Player typeclass in here. Maybe one could handle or warn about that more gracefully?

vlegoff:
This PR has been fixed in documentation.  However, `inherits_from` doesn't seem to do the trick here: I use `inherits_from` with `DefaultObject` or `ObjectDB` as a second argument, and get `True` if I pass in a typeclass (class) or an object (instanciated).  This behavior is described in the documentation of `inherits_from` and is probably expected.  However, in my case, I need to know if the argument is a class or an object.  I modified the code a little, however, to avoid using `isinstance` on `ObjectDB` itself.

I might not have described the interest of this PR, and it's a bit limited: the point is to be able to retrieve events tied to a typeclass (since that's where they are tied first) or an object.  Only the latter was supported in the previous version.  This allows to send `Room` to the `get_events` method.  Getting events on the typeclass can be useful.

Griatch:
@vlegoff In my comment I mis-remembered the fact that `isinstance` also does handle child classes. So don't spend work on rewriting it to use `inherits_from` just for the sake of it, if `isinstance` was enough already.

Issue #1200, opened 2017-02-06
Errors in Commands cause default cmdsets to not load


j3b4:
#### Brief summary of issue / Description of requested feature:
I have noticed that errors in some original commands I'm creating, will apparently prevent any commands from loading. 

#### Steps to reproduce the issue / Reasons for adding feature:

1.  Introduce a syntax error into a command that is part of the CharacterCmdSet
2.  `@reload`
3.  Enter any command. 
> `@reload`
> help
> quit

#### Error output / Expected result of feature
```
SyntaxError encountered when loading cmdset 'commands.default_cmdsets.CharacterCmdSet'.
(Traceback was logged 2017-02-06 16:27:32-0000)
Command '@reload' is not available. Type "help" for help.
Command 'help' is not available. Type "help" for help.
Command 'quit' is not available. Type "help" for help.
```

Since no commands work, the player cannot quit without closing their client.  In practice I've had to go back to my pyenv and run > evennia reload after fixing the error.

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:

IMHO an error in a single command ought not have such a devastating effect it would be better if only the command in question failed to run... or perhaps at least limit it to a single command set.

EDIT: My opinion has changed to agree with Griatch explanation below.  Now I believe that it would be better for evennia to refuse to load at all, rather than let user log into a crippled mud.

Griatch:
Limiting this to one Command is not possible, because what happens when you make a critical syntax error at the module level is that you have created an invalid Python module - this means the module cannot be imported at all and will take all it contains with it. Running `evennia reload` (or `evennia start`) is not an extraordinary thing, you should generally be ready to do so if you create a situation where the server cannot start. 
That said, it was many years since I looked at this since, so it may be worth to examine further.

Griatch:
@j3b4 Actually, introducing a syntax error generally does not allow the server to come back at all after a `@reload` - you must fix it and run `evennia start`. Could you give an example of a syntax error that allows you to `@reload` and leads to the loss of cmdsets?

j3b4:
I know it happened one or more times previously with different errors, but I didn't really question it.  This last error is reproducible by copy-pasting a non-ASCII character into a string literal in a world file that gets imported into one of my commands.

This is the server log:
``
17-02-07 00:30:57-0000)
2017-02-07 00:30:57+0000 [AMPProtocol,0,127.0.0.1] [::] Traceback (most recent call last):
[...]

PR #1381, opened 2017-07-28
Create fallback for default cmdsets that fail to load

TehomCD:
Checks the path of any cmdset that fails to load against a dict in settings that can provide a fallback to try instead, making syntax errors in code much more forgiving.

In Evennia chat, @DamnedScholar brought up that cmdset failure isn't very forgiving - any syntax error in an imported file is likely to wreck the entire process and leave them in a state where they can't even `@reload` the server. While I discussed the workarounds, such as catching exceptions in import statements, this isn't particularly newbie friendly. An automatic fallback of the default cmdsets seemed like a good solution to me.

N/A

### DIFF
--- a/evennia/commands/cmdsethandler.py
+++ b/evennia/commands/cmdsethandler.py
@@ -355,9 +355,12 @@ class CmdSetHandler(object):
                                 # If a cmdset fails to load, check if we have a fallback path to use
                                 fallback_path = settings.CMDSET_FALLBACKS.get(path, None)
                                 if fallback_path:
+                                    logger.log_err("Error encountered for cmdset at path %s. Replacing with: %s" % (
+                                        path, fallback_path))
                                     cmdset = self._import_cmdset(fallback_path)
                                 # If no cmdset is returned from the fallback, we can't go further
                                 if not cmdset:
+                                    logger.log_err("Fallback path '%s' failed to generate a cmdset." % fallback_path)
                                     continue
                             cmdset.permanent = cmdset.key != '_CMDSET_ERROR'
                             self.cmdset_stack.append(cmdset)
--- a/evennia/commands/cmdsethandler.py
+++ b/evennia/commands/cmdsethandler.py
@@ -351,6 +351,14 @@ class CmdSetHandler(object):
                     elif path:
                         cmdset = self._import_cmdset(path)
                         if cmdset:
+                            if cmdset.key == '_CMDSET_ERROR':
+                                # If a cmdset fails to load, check if we have a fallback path to use
+                                fallback_path = settings.CMDSET_FALLBACKS.get(path, None)
+                                if fallback_path:
+                                    cmdset = self._import_cmdset(fallback_path)
+                                # If no cmdset is returned from the fallback, we can't go further
+                                if not cmdset:
+                                    continue
                             cmdset.permanent = cmdset.key != '_CMDSET_ERROR'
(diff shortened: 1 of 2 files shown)


Griatch:
Sounds like a good idea! I'm away from a computer at the moment (on vacation) so will look into this when I get back.

Kelketek:
I would suggest that the fallback commandset should be extremely minimal, containing only communication items like say, pose, and page, alongside reload. This is because if it is not obvious that such a failure has occured, someone may use a database editing command that behaves differently than expected.

evennia/commands/cmdsethandler.py
```diff
@@ -351,6 +351,14 @@ def update(self, init_mode=False):
                     elif path:
                         cmdset = self._import_cmdset(path)
                         if cmdset:
+                            if cmdset.key == '_CMDSET_ERROR':
+                                # If a cmdset fails to load, check if we have a fallback path to use
+                                fallback_path = settings.CMDSET_FALLBACKS.get(path, None)
```
Griatch: I like the look of this, in the sense that we won't wreck the reload process on a faulty cmdset string. I think we are swallowing the error here though; if we are falling back to a default cmdset the user may think all works fine yet they will get the fallback version of commands without realizing it. See for example the problems people have with us falling back to the `DefaultObject` when a typeclass fails to load - even though we log it. I think that at a minimum we need to use `logger.log_errmsg()` to forward which cmdset failed to load here.

j-f1: Maybe there should be a warning message sent to players above a certain class (maybe Builders?) each time they try to use a fallback command.

TehomCD: There's some error printing out in the `self._import_cmdset(path)` call that sets it to be `_CMDSET_ERROR` - do you think that's sufficient, or we should have more? When I have in-game errors on, I actually get the stack trace twice for some reason, not sure why yet.

Griatch: The `cmdset.key = "_CMDSET_ERROR"` is important btw - that is caught higher up in the chain and used for reporting in-game. For reporting to the user, the `cmdset.errmessage` is echoed to the user whenever a `cmdset.key == "_CMDSET_ERROR` is seen. There is nothing stopping us from doing additional logging though - especially if we quietly swap out the cmdset as done here. 
@TehomCD  We do not send tracebacks to a channel, this is true. I could see your point in doing that. I guess that could be a further use of the mudinfo channel maybe (or make a new one `syslog` for just errors and anomalies). Relaying to a channel should in that case just be an optional kwarg (with a default you can change from settings) of the `logger.log_err` function and its siblings. That would be a separate PR though. For now, this should just get logged.

TehomCD: Alrighty, I added logging. I wasn't sure if it'd be appropriate to send some message to the object if `_IN_GAME_ERRORS` is set too. If that sounds like a good idea, I could add that in as well as doing the `logger.log_err` call.

Griatch: @TehomCD Using _IN_GAME_ERRORS is a good idea. I was already working on integrating this PR when I read that, so I took the opportunity to add that as well when I was at it, thanks!

TehomCD:
Oh, this resolves #1200 also. I didn't notice that issue until I was looking at older issues to find something unrelated.

PR #1385, opened 2017-07-31
Add the plural contrib to display plural names

vlegoff:

This light contrib offers a very simple mixin to add pluralization to all typeclasses.

It's a common question I find from time to time on IRC:

> How to have "You see: 2 rocks" instead of "You see: rock, rock"?

This contrib is here to answer this question.  It allows to override the `return_appearance` method and group objects by name.  Pluralization is done by placing the quantity after an `X` sign, if there are more than one.  This default behavior can be overridden by defining a `get_plural_name` method in the desired typeclass.

The contrib documentation explains how to install it.  It might not be detailed enough, using the `get_plural_name` method does require to read the code.  I added a unittest to make sure the contrib is working properly as well.

### DIFF
--- a/evennia/contrib/plural.py
+++ b/evennia/contrib/plural.py
@@ -1,0 +1,190 @@
+"""
+Light contrib to group objects by name and display plural names.
+
+Evennia contrib - Vincent Le Goff 2017
+
+This contrib allows to group object names (in 'look' for instance)
+using their singualr names.  You could have a display varying from:
+    You see: a rock, a rock
+To:
+    You see: a rock X 2
+You could even customize it to have:
+    You see: 2 rocks
+
+To install, you simply need to import and inherit from the `PluralNames`
+mixin in your typeclasses.  Be sure to put it first in your inheritance
+tree.  You could do it for all your typeclasses, or just your rooms for
+instance.
+
+```python
+# Example with 'typeclasses/rooms.py'
+from evennia import DefaultRoom
+
+from evennia.contrib.plural import PluralNames
+
+class Room(PluralNames, DefaultRoom):
+
+    # ...
+```
+
+The `PluralNames` is a very light mixin that will override your
+`return_apperance` hook.  You can, however, write your own `return_appearance`
+hook and use the `group_objects` directly (see below for more information).
+
+To change the way your objects will display plural names, you can
+override the `get_plural_name` method (see the default implementation
+below).  You can do so with all your typeclasses, if you don't like the default.
+
+"""
+
+def group_objects(objects, attrname="key"):
+    """
+    Return the grouped list of objects, using their singular names.
+
+    Args:
+        objects (list ob Object): list of objects to be grouped.
+        attrname (str, optional): the name of the singular attribute.
+
+    Returns:
+        grouped (list): the same list of objects grouped by names.
+        The list will contain tuples: the number (int),
+        the object (Object) and the list of all objects of the same name.
+
+    Notes:
+        In the case of grouping (if an entry has more than 1 object),
+        only one of these objects will be displayed for grouping.
+        This is usually the expected behavior.
+
+        Grouping will occur based on the `attrname` argument.  By
+        default, the objects' `key` are used as singular names.  You
+        can provide a different attribute names.  It could be linked
+        to a property, or an Evennia attribute (just give it
+        "db.<name of attribute>" as value).
+
+    Example:
+        >>> group(rock1, tree, rock2)
+        [
+            (2, <Rock 1>, [<Rock 1>, <Rock 2>]),
+            (1, <Tree>, [<Tree>]),
+        ]
+
+    """
+    numbers = []
+    group_names = {}
+    for obj in objects:
+        # Try to get the `attrname` property
+        if "." in attrname:
+            name = obj
+            for part in attrname.split("."):
+                name = getattr(name, part)
+        else:
+            name = getattr(obj, attrname)
+
+        matches = group_names.get(name, [])
+        matches.append(obj)
+        group_names[name] = matches
+
+        # Add only if number is 1
+        if len(matches) == 1:
+            numbers.append((obj, name))
+
+    # Create the grouped list of objects
+    grouped = []
+    for obj, name in numbers:
+        matches = group_names.get(name, [])
+        number = len(matches)
+        grouped.append((number, obj, matches))
+
+    return grouped
+
+def _get_plural_name(obj, number, looker, matches):
+    """Helper function to return the plural name of an object."""
+    name = obj.key + " X " + str(number)
+    # If looker is a builder, show the IDs
+    if obj.locks.check_lockstring(looker, "perm(Builders)"):
+        name += "(" + ", ".join("#{}".format(match.id) for match in matches) + ")"
+
+    return name
+
+
+class PluralNames(object):
+
+    """
+    Mix-in to group object names by plural.
+
+    This mix-in will override the methods if used:
+        return_appearance: to return grouped objects.
+        get_plural_name: get the plural name for this object.
+
+    It means you could change the rule of pluralization by overriding
+    the `get_plural_name` method.  By default, it will just append
+    ' X {number}' to the objects' name.
+
+    """
+
+    def return_appearance(self, looker):
+        """
+        This formats a description. It is the hook a 'look' command
+        should call.
+
+        Args:
+            looker (Object): Object doing the looking.
+        """
+        if not looker:
+            return ""
+        # get and identify all objects
+        visible = (con for con in self.contents if con != looker and
+                   con.access(looker, "view"))
+        exits, users, things = [], [], []
+        for con in visible:
+            key = con.get_display_name(looker)
+            if con.destination:
+                exits.append(key)
+            elif con.has_player:
+                users.append("|c%s|n" % key)
+            else:
+                things.append(con)
+        # get description, build string
+        string = "|c%s|n\n" % self.get_display_name(looker)
+        desc = self.db.desc
+        if desc:
+            string += "%s" % desc
+        if exits:
+            string += "\n|wExits:|n " + ", ".join(exits)
+        if users or things:
+            grouped = group_objects(things)
+            things = []
+            for number, obj, matches in grouped:
+                if number == 1:
+                    name = obj.get_display_name(looker)
+                else:
+                    # Call `get_plural_name` on `obj` if exist
+                    if hasattr(obj, "get_plural_name"):
+                        name = obj.get_plural_name(number, looker, matches)
+                    else:
+                        name = _get_plural_name(obj, number, looker, matches)
+
+                things.append(name)
+
+            string += "\n|wYou see:|n " + ", ".join(users + things)
+        return string
+
+    def get_plural_name(self, number, looker, matches):
+        """
+        Return the plural name for self.  A plural name is displayed if
+        there is more than one object of the same name.
+
+        Artgs:
+            number (int): the number of objects (> 1) of that name.
+            looker (Object): the object looking.
+            matches (list of Object): list of objects with the same name.
+
+        Regurns:
+            name (str): the plural name of self in this quantity.
+
+        Note:
+            The `matches` argument is provided to have the list of
+            all objects (including self) sharing the same singular name.
+
+        """
+        return _get_plural_name(self, number, looker, matches)
--- a/evennia/contrib/tests.py
+++ b/evennia/contrib/tests.py
@@ -985,3 +985,35 @@ class TestUnixCommand(CommandTest):
         lines = ret.splitlines()
         self.assertTrue(any(l.startswith("usage:") for l in lines))
         self.assertTrue(any(l.startswith("dummy: error:") for l in lines))
+
+
+# Plural contrib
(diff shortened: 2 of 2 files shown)


evennia/contrib/plural.py
```diff
@@ -0,0 +1,190 @@
...
+    for obj, name in numbers:
+        matches = group_names.get(name, [])
+        number = len(matches)
+        grouped.append((number, obj, matches))
+
+    return grouped
+
+def _get_plural_name(obj, number, looker, matches):
+    """Helper function to return the plural name of an object."""
+    name = obj.key + " X " + str(number)
```
j-f1: Maybe use `×` instead (`\xD7`)?

BigJMoney:
Without looking at the code, could we make this feature a part of Evennia
instead of a contribution?

vlegoff:
This is definitely open to debate. Let's see what other users think.  It wouldn't be hard to install in the main code, but would other users ressent this change?

chainsol:
I really love this feature but I think it's best as a contrib. Granted, I don't see a game not using it, but it's usually best to let each designer choose. :]

@vlegoff, I'm glad to see this in here because I'd made something similar! Not as nice, so I'll be switching to this.

BigJMoney:
Could we add to Evennia including an option?  I.e. A configuration setting
that allows a Dev to choose between the two.

evennia/contrib/plural.py
Griatch: @j-f1 While fine for Evennia itself, this would be problematic for clients not supporting UTF-8. So probably not a good idea to have as a default.

Griatch:
This one is indeed, as said, open to debate. I would normally vote to keep this as a contrib (as presented). But it's also true that I had the same reluctance about adding the `get_display_name` hook, and that has certainly turned out to be a very worthwhile thing to have in core. 
So on one hand, this looks like a clean and nice contrib that is trivial for users to add, but on the other hand it is also a feature on which other contribs could depend. Then there is of course always the argument against adding too much stuff to core.  I can see both sides to that.

evennia/contrib/plural.py
```diff
@@ -0,0 +1,190 @@
...
+below).  You can do so with all your typeclasses, if you don't like the default.
+
+"""
+
+def group_objects(objects, attrname="key"):
+    """
+    Return the grouped list of objects, using their singular names.
+
+    Args:
+        objects (list ob Object): list of objects to be grouped.
```
BlauFeuer: Typo fix: objects (list of Object)

BlauFeuer:
Adding a `get_plural_name` hook into Evennia for default objects seems reasonable: It can be overwritten (like `get_display_name`) to generate plurals in whatever format a game would want, defaulting to the safest <name> x <number> format.

If `group_objects` was a helper or utility, I'd use it, because I could apply it to inventory, containers/vehicles, and room appearances instead of only being able to use it on rooms.

As is, it's a good contrib, though - I'd just use it as an example.

evennia/contrib/plural.py
BlauFeuer: If the session that the string is to be presented to is known, the string could be conditional on the session encoding, but just sending to the viewing character, it gets complicated in multi-session modes where there might be more than one session.

vlegoff:
@BlauFeuer: actually this contrib offers threee things: a mixin for typeclasses (rooms are a straight example but it could apply to all, just replacing the `return_appearance`), a helper function that you could use anywhere (that's why I kept them separate) and a private function that shouldn't be changed since it just holds the default value for plural_names.  If this is adopted as a main feature (again, open to debate), this private function will obviously disappear.  I'm not sure if the public function in the contrib would be a utility, but that's how I designed it.

BlauFeuer:
@vlegoff Sorry, I was thinking aloud when I commented about a get_plural_name hook as part of Evennia's default object code, with your helper function calling the private one as the default if no specific plural form was given.  Plural form for object names is something I've seen supported elsewhere.

It's sometimes a balance between server core, contrib code, and in-game code, especially when in-game code is written before a contrib appears, or the core changes to support something already written into game code, so I'm not sure what to think, other than to develop what's being requested by users and adapt as the framework grows and changes.  It's really fun, and I wish I could put more time into it instead of pointing out little details here and there, or popping into the dev chat for only 5 minutes at a time.

I guess it's my turn to go on vacation.  See you around next week!

evennia/contrib/plural.py
```diff
@@ -0,0 +1,190 @@
...
+        attrname (str, optional): the name of the singular attribute.
+
+    Returns:
+        grouped (list): the same list of objects grouped by names.
+        The list will contain tuples: the number (int),
```
Griatch: Indent one step for subsequent lines.

evennia/contrib/plural.py
```diff
@@ -0,0 +1,190 @@
...
+    Notes:
+        In the case of grouping (if an entry has more than 1 object),
+        only one of these objects will be displayed for grouping.
+        This is usually the expected behavior.
+
+        Grouping will occur based on the `attrname` argument.  By
+        default, the objects' `key` are used as singular names.  You
+        can provide a different attribute names.  It could be linked
+        to a property, or an Evennia attribute (just give it
+        "db.<name of attribute>" as value).
```
Griatch: Actually, best is to reverse this syntax since it will then match up with the way prototypes are defined. That is, start your name with `db_` to refer to a property, (like `db_key`), otherwise it is considered an Attribute.

evennia/contrib/plural.py
```diff
@@ -0,0 +1,190 @@
...
+    # Create the grouped list of objects
+    grouped = []
+    for obj, name in numbers:
+        matches = group_names.get(name, [])
+        number = len(matches)
+        grouped.append((number, obj, matches))
+
+    return grouped
+
+def _get_plural_name(obj, number, looker, matches):
```
Griatch: Regardless of if we decide to move this to core in the future or not, this here does not need to be its own function. Put this directly in `get_plural_name` on the typeclass.

vlegoff:
@Griatch: I updated these lines.  The fact that `group_names` was using `attrname` in this way allowed to use properties on the object, if you want to override the plural name without creating an attribute in `db`.  This is a small concern, but perhaps worth mentioning.

Regarding whether this should be added to main Evennia, I'm a bit shared: on one side I would find it normal since I don't know many games without this feature.  On the other, letting the choice to developers might make more sense (it could, as @BlauFeuer suggested, be something to set in the settings).  The current system is probably slowing down a little the display of objects, if there are a lot, but probably not that much either.

FlutterSprite:
My vote is for integrating plural names into core! If we're doing votes, that is. It definitely feels like one of those things that any game that uses objects is going to want - to me, NOT having objects grouped into plurals strikes me as the niche option.

TehomCD:
I lean towards contrib because I think there's a number of ways to try to address the plural issue. While I think this is a cool idea, it doesn't address the alias issue - 'look oxen', 'look guardsmen', 'get <other name for a grouping here that doesn't match object aliases>'. That's the tricky part of any sort of pluralizing so I don't blame Vincent for not wanting to get into it, but I think having a suggestion to use a method like get_plural_name without a solution for it not matching any aliases might be confusing.

vlegoff:
I'm ready to improve it if necessary.  I understand both points of integration VS non-integration.  As it is, keeping as a contrib and installing it is pretty simple through the mixin.  On the other hand, if you don't want it, or do want to do things differently as @TehomCD suggests, that might be tricky if it's integrated into the core.  That's one reason why a contrib might make more sense.  On the other hand, we'll never get something that suit everyone (that is the basis of any game engine).  `get_display_name` was perhaps questioned at first but it came in handy in the end.  And it doesn't allow everything I'd like in my own game, for a start.  So we'll never get it done perfectly as we want.  Still, the contrib idea might work much better, only because integrating it in the game core would limit freedom a bit too much in the end.

@TehomCD: I didn't think about aliases and pluralization at all, but you have a valid point.  If you think of any way to address this issue, I would be happy to try and add it in the contrib, letting users to choose what to do with it.

TehomCD:
Maybe something like the call to `get_plural_name` could add an alias of the plural name to the matches if it's not the same as their key? The problem is that you then need to wipe plural name aliases when they're no longer valid, and I'm not sure when the best time to do that would be - if it was in get_appearance, the aliases could be stale for other commands before someone does another `CmdLook`. Possibly in the movement hooks?  Check when the objects move, and if they were plural before, but no longer are, remove aliases that are no longer needed?

vlegoff:
After some thinking, I tend on a much more simple solution: using currently existing object aliases to add pluralization even for singular names.  The advantage of using plurals even for singulars is that you don't have to worry about the quantity anymore and these are usable straight out-of-the-box in default (and non-default) commands.  That includes, of course, the default search.

    ox.aliases.add("oxen")

Of course, since it might include all objects, it would be a bit of work without some kind of automation here.  And this solution doesn't work if you enter `look 3 oxen`, obviously.  But we can't look at several objects at once anyway, so the aliases in this case would be just a convenient shortcut.  If you see `3 oxen are here`, you can `look oxen`, though it's a plural.  Just convenient.  I do not know if it's a good solution.

vlegoff:
I've done some thinking on it.  You might not call it "deep thinking".  The problem of alias is real but perhaps not such an issue.  Here's my theory.

Looking to oxen is surely great.  When having two objects whose singular name is "ox" and plural name is "oxen" would display 2 oxen in the room.  So far so good.  Assuming a player wants to "look oxen", then what?  A solution would be to put "oxen" as an alias on both objects.  Why not?  A player cannot be expected to look at two objects at once, can she?

If we solve (don't solve) the issue this way, the real problematic is porting the plural alias on every object, or, perhaps, have an AliasHandler that automatically look up the plural name.  Both have disadvantages but dealing with aliases this way would greatly simplify the future code.  I don't know if it really makes sense to add, say, the "oxen" alias to both objects... once the second ox walks into the room, and remove them both afterward.  It doesn't seem like the perfect solution to me, either.  I'm more than open to suggestions, however, and I would be glad to have a working contrib to do that.

Griatch:
I ended up adding this functionality to the core `DefaultObject` in the `develop` branch, it would be good with some eyes on it from those who have followed this PR (which will thus not be merged - but thanks a lot to @v-legoff for his work on this - it helped a lot). My implementation uses a new `get_numbered_name` hook, which is so named since it is expected to spit out both the singular form of the key to display (an apple, a sword) as well as the count for the pluralized form (two apples, three swords). I added the procedural `inflect` package to automate the pluralization in English. It's not perfect on irregular words but it's decent. I tried hard to gather all the inflection in the hook to make it easy to replace the whole mechanism if needed (of course, it's only called from the `return_appearance` hook, which you can also override of you want). The default hook will auto-create aliases for the relevant forms to make things available. The efficiency of this method can be discussed, it seemed to be the easiest and most dynamic way to do it. 

Relevant commits (in develop branch) are:
https://github.com/evennia/evennia/commit/dfd2d3897ab5f8118732907835be389bfd2da0cb
https://github.com/evennia/evennia/commit/cc1bc92c3dc65a7a5cd7258350feb0d8937f3bdd

BlauFeuer:
After pulling this and installing inflect, I can no longer get unit tests for commands (mygame/commands/tests.py) to pass or fail - they error instead. 

https://pastebin.com/ZJHsdkwt

The test code is just a simple:
```python
        self.call(CmdSay(), "I'm testing things.",
                  '''Char(#6) says, "I'm testing things."''')
```

What can I do to track this down to find the source of the problem?  I'm not certain it's the fault of this commit, but since this PR is still open, I mention it here.

Griatch:
@BlauFeuer To try to figure out what's going on it would be interesting to see what `key` is for the 
`plural = _INFLECT.plural(key, 2)` call in your traceback. Just to rule out this coming as something other than a string for some reason or similar.

vlegoff:
It looks like `infect.plural` isn't returning what it should.  How did you install it?  I guess the best way to do it (and make sure every package is up-to-date within your python environment) is to do:

    pip install --upgrade -e evennia

(Being in the parent folder with the Python environment activated).

In order to debug this strange error, however, I would recommend using a debugger as [PDB](https://github.com/spiside/pdb-tutorial) (or something similar, depending on the IDE you favor using).  `pdb` will definitely be easy to launch in the test suite and could give you interesting information, even if you'll have to step a few times in order to see the exact error.

HTH,

BlauFeuer:
The offending key turns out to be '|cChar|n' in the unit test.

vlegoff:
Perhaps we could force to strip colors before calling `plural`.  I don't think it's a good idea to have color codes in the keys themselves but then, that's just me.  Does this error occur on a test you do or in the Evennia test suite itself?

BlauFeuer:
unit tests for commands (mygame/commands/tests.py)

Griatch:
@v-legoff The color is not in the `.key` property in this case but is (correctly) added at the `get_display_name` hook, which is where the `get_numbered_name` gets the name from. It should be ok to add color in the `get_display_name` hook (but we also need to call that hook since the name might be different e.g. in recog-systems) so probably one needs to strip the colors and then put them back afterwards or something like that. It's potentially tricky to make this general enough though.

Griatch:
Closing this PR. I moved the remaining color problem to a new issue [here](https://github.com/evennia/evennia/issues/1572).

Issue #1383, opened 2017-07-30
Feature Request: More security-related info sent to console/logged by the network portal


BlauFeuer:
#### Brief summary of issue / Description of requested feature:
As mentioned in #1375 , failed login attempts and temporary account lock-outs are not sent to console or logged, nor are connections that never reach the login authentication point.

#### Reasons for adding feature:
Security in the form of malicious activity detection/documentation, and getting a better idea of who/how many connections never result in any login attempt, failed or otherwise.

#### Error output / Expected result of feature
Similar to the entry for a successful login:
`2417-04-30T06:33:20+0000 [AMPProtocol,0,127.0.0.1] [..] Logged in: Golem(player 17) 8.282.39.12 (1 session(s) total)`
Perhaps something like:
`2417-04-30T06:33:20+0000 [AMPProtocol,0,127.0.0.1] [..] Failed to log in: Golem(player 17) 8.282.39.12 (1 session(s) total)`
and
`2417-04-30T06:33:20+0000 [AMPProtocol,0,127.0.0.1] [..] Connection established with 8.282.39.12 (1 session(s) from that address total)`

Finally, something similar for a temporary account lockout on multiple login failures:
`2417-04-30T06:33:20+0000 [AMPProtocol,0,127.0.0.1] [..] Locked out: Golem(player 17) 8.282.39.12 (5 failed login attempts total in last 2 minutes)`

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:

master/devel branch as appropriate, though this doesn't seem like a major change.

Griatch:
This sounds like a good idea .

strikaco:
This is definitely sorely needed. It would be great if we could capture the time, IP address and player object (whether accessed or denied), and effective character (where applicable) as CSV for things like:

* Login events
* Login timeouts
* Login failure
* Logout events
* Lockout events (too many failed password attempts)
* Ban/unban events
* Puppet events + target character name and dbref
* (optional, but useful for exploit detection and object auditing) All raw commands submitted by the players

This would make it easy to write a dashboard command (`@audit`) that allows wizard-class users to review and filter security events related to a player name, IP address, character, or even object (though aliases might skew this). Questions such as:

* "Who was puppeting this character on this date?"
* "Show me the unique IP addresses that have accessed this player account."
* "Show me all player accounts that this IP address is associated with."
* "Show me all the characters this player has ever puppeted, and when."
* "Show me all players who have puppeted this character."
[...]

Issue #1375, opened 2017-07-23
IP Address/LastSite in Player/Character db Attributes


ghost:
Not just because it's a common feature in all the classic MUSH/MUX platforms but because it's a good one:

When a player connects to an Evennia server, it would be nice if their IP Address was database-stored on their player object (and maybe any character they interact with/puppet too) in an attribute titled Lastsite.  Ideally this would store the last 2-3 dozen addresses successfully connected from, with a sister attribute called Badsite which would store the last 2-3 dozen addresses unsuccessfully connected from (say somebody tried to connect to a player object with a bad password attempt, their IP would be logged in Badsite rather than Lastsite).

At the present time there doesn't seem to be any DB storage of user-connected IP Addresses at all, so if somebody gets @booted (or just disconnects) it's not possible to check other player objects and see if that user was linked to any other player objects and if so, find out what they were up to on the MU.

This is a security/backtrace feature request.

chainsol:
This would be an excellent feature. +1

BlauFeuer:
This request seems a bit over-reaching in that it's asking for all attempts to access DB objects by a session to be IP and timestamped.yes last successful and last unsuccess

chainsol:
Hm, I can see why this should be a feature of Evennia, but I can also see why it could be game-specific. It would be easy enough to store the connection IP from the at_post_login hook on the player, correct?

tekmunkey:
I'm pretty confused how this went from me suggesting, "record the last IP address that connected to a player, and record the IP address of the last player that puppeted a character," to BlauFeuer saying, "all attempts to access DB objects by a session to record IP and timestamp on the affected objects and stored in the database for security/backtrace reasons with a history depth estimated to be in the dozens."

What I suggested is that, at one point post-successful-login, a db attribute should be appended to on a player object, or at one point post-unsuccessful-login, another should be.  Or that, at one point post-successful-puppeting, the last value from that attribute should be appended to a similar db attribute on a character object.  What BlauFeuer is arguing against is recording a log entry for every single stupid little thing a person does while connected to a game.

That's not even 2nd language level misinterpretation, it's outright taking what was said and turning around three times and stomping your feet and clapping your hands and screeching that rape is wrong and anyone who agrees that making a single point-of-contact record on a single object during a single session is a rapist.

TehomCD:
[...]

PR #1390, opened 2017-08-11
Implements user input auditing.

strikaco:
Implements auditing of all strings submitted to the ServerSessionHandler by connected players to monitor user behavior. Where possible, the session IP address, player object#dbref, character object#dbref, and room occupied#dbref are resolved and included inline.

Tested in MULTISESSION_MODE = 3 to address even the wackiest of deployments.

Disabled by default; can be enabled in settings.

The log format is admittedly rough; currently working on standardization for easier parsing.

No auditing functionality currently exists.

Attempts to provide a historical record by which to address the concerns raised in #1375 (being able to "check other player objects and see if that user was linked to any other player objects and if so, find out what they were up to on the MU") while declining to adopt the proposed implementation.

This does NOT attempt to address the concerns raised in #1383, as it does not provide any insight into network activity.

### DIFF
--- a/evennia/server/serversession.py
+++ b/evennia/server/serversession.py
@@ -13,8 +13,10 @@ import importlib
 import time
 from django.utils import timezone
 from django.conf import settings
+from evennia import settings as ev_settings
 from evennia.comms.models import ChannelDB
 from evennia.utils import logger
+from evennia.utils.logger import log_err, log_file
 from evennia.utils.utils import make_iter, lazy_property
 from evennia.commands.cmdsethandler import CmdSetHandler
 from evennia.server.session import Session
@@ -149,7 +151,59 @@ class NAttributeHandler(object):
 # -------------------------------------------------------------
 # Server Session
 # -------------------------------------------------------------
+class ServerAuditor(object):
+    """
+    Logs select auditing-related events occurring within the
+    ServerSessionHandler.
+    """
+    def __init__(self, session):
+        self.session = session
+    
+    def log(self, msg, **kwargs):
+        # Mask any passwords included in a connection attempt!
+        tokens = msg.strip().split(" ")
+        cmd = tokens[0].lower()
+        
+        if cmd in ("connect", "create", "@password"):
+            token_len = 2
+            # If @password was issued, mask everything after the command
+            if cmd == "@password":
+                token_len = 1
 
+            msg = " ".join(tokens[:token_len]) + " ********"
+            
+        # If this is a user command, flag it via kwarg to
+        # prevent users from spoofing log events in-game.
+        if kwargs.get('user'):
+            msg = "RawCmd( %s )" % msg
+        
+        # Get the current session's IP address
+        ip = self.session.address
+        
+        # Log both the Player name and its dbref
+        player = self.session.get_player()
+        if player: player = "%s#%i" % (player.name, player.id)
+        
+        # Log both the Character name and its dbref
+        character = None
+        char_obj = self.session.get_puppet()
+        if char_obj: character = "%s#%i" % (char_obj.name, char_obj.id)
+        
+        # Log both the Room name and its dbref
+        room = None
+        if char_obj:
+            room = char_obj.location
+            if room: room = "%s#%i" % (room.key, room.id)
+        
+        # Log the IP, Player (if available), Character (if available),
+        # Room (if available) and what they're trying to do
+        message = "%s :: %s\n" % (unicode(tuple(x for x in (ip, player, character, room) if x)), msg)
+        
+        # Rotate this log; too voluminous to make optional.
+        date = timezone.now().strftime("%Y-%m-%d")
+        log_file(message, '%s_audit.log' % date)
+        return True
+        
 class ServerSession(Session):
     """
     This class represents a player's session and is a template for
@@ -166,6 +220,8 @@ class ServerSession(Session):
         self.player = None
         self.cmdset_storage_string = ""
         self.cmdset = CmdSetHandler(self, True)
+        
+        self.audit = ServerAuditor(self)
 
     def __cmdset_storage_get(self):
         return [path.strip() for path in self.cmdset_storage_string.split(',')]
@@ -375,6 +431,9 @@ class ServerSession(Session):
                 for the protocol(s).
 
         """
+        # Be careful about enabling auditing of responses here--
+        # a single broadcast-type message will create one log entry per recipient!
+        
         self.sessionhandler.data_out(self, **kwargs)
 
     def data_in(self, **kwargs):
@@ -391,6 +450,20 @@ class ServerSession(Session):
             the client. It should usually always end by sending
             this data off to `self.sessionhandler.call_inputfuncs(self, **kwargs)`.
         """
+        if ev_settings.INPUT_AUDITING:
+            try:
+                text = unicode(kwargs.get('text')[0][0]).strip()
+            except IndexError as e:
+                text = kwargs.get('text', '')
+                if text: text = unicode(text)
+                
+            # Log commands only if they are greater than 2 characters long--
+            # We don't want to log every movement in a cardinal direction.
+            if len(text) > 2:
+                # Make sure to denote that this is a user-entered command,
+                # otherwise users can spoof log messages
+                self.audit.log(text, user=True)
+
         self.sessionhandler.call_inputfuncs(self, **kwargs)
 
     def msg(self, text=None, **kwargs):
--- a/evennia/settings_default.py
+++ b/evennia/settings_default.py
@@ -219,6 +219,13 @@ MAX_CHAR_LIMIT_WARNING = "You entered a string that was too long. Please break i
 # debugging. OBS: Showing full tracebacks to regular users could be a
 # security problem -turn this off in a production game!
 IN_GAME_ERRORS = True
+# If this is true, all strings entered by all players will be logged to disk
+# (in the audit.log file) regardless of whether or not the string resulted 
+# in a valid or successful command being executed.
+# Passwords are masked but if mistyped, could still get logged
+# in cleartext. Be ethical, and understand the danger concerning this file
+# getting stolen.
+INPUT_AUDITING = False
 
 ######################################################################
 # Evennia Database config


DamnedScholar:
Sorry to take some of the thunder from this PR, but what if this were implemented differently? The same goal could be achieved by having each command call a global hook function and pass its own object to that function. If the command's `__dict__` includes `"audit": "false"`, for example, then the command would be excluded from the log. The implementation of a global "on command invocation" hook function would allow for a wide variety of other features to be built on the same foundation, such as having some commands interrupt others (I know @Treatz - streetwitch - is trying to implement a command that can be directly interrupted by other commands).

BlauFeuer:
In #1375, I proposed this could be done in game code in `commands/command.py` using the `at_pre_cmd` hook.

At this point, the session, "player" account, and puppeted object is known, along with the command string entered.  It can be pushed into a log and/or displayed in the console, and can be expanded as needed without having to build auditing into the server even if it's not used by default.

As for the use case of a command interrupted by another command, that would work similarly as a game-specific change where the command used is written on the puppeted object to allow for others to trigger it to be interrupted.

Griatch:
@DamnedScholar: Command interruption is already possible (see several tutorials dealing with this), I don't think we need another hook only for this. 

@BlauFeuer: Ah, right, using the `at_pre_cmd` of a command parent would make for a very easily over-ridable hook location too, with the additional advantage of being able to use classes to further customize eventual audit for different command groups. Hm. For Commands, using `at_pre_cmd` gives more flexibility. strikaco's centralized logger offers the potential ability able to catch *all* outgoing data though (OOB, script echoes etc), for those that would want that.

DamnedScholar:
@Griatch Which tutorials? Searching the wiki for the word "interrupt" isn't turning up anything.

BlauFeuer:
All outgoing data auditing, yes I agree that would be useful for debugging in addition to security.   I don't disagree with this PR and I applaud  strikaco's efforts.  Currently I log all no match commands for debugging and determining which command aliases would be most beneficial, along with specific admin commands similar to the 'audit': False flag idea. My main point is that expanding the framework to allow game code to fill in the gaps for specific needs is the key to making Evennia extensible and broadening its user base.

Griatch:
@DamnedScholar https://github.com/evennia/evennia/wiki/Command-Duration#abortable-commands. Also https://github.com/evennia/evennia/wiki/Command-Duration could be useful.

strikaco:
@Griatch 
> I think however that something like this must probably be implemented as something the user can override. The easiest way would be to call a hook in data_in. What that hook calls is defined in settings and defaults to calling the audit logger as presented here. Alternatively the audit logger could be a contrib (not sure about that though, having default loggin capabilities is not a bad idea, that's something worth discussing). The setting could remain and be used to bypass the hook call completely.

Agreed about the hook idea, I felt uncomfortable as-is forcing output to logfiles in a format of my own dictation. Switching this to use hooks would not only allow devs to determine what they want to capture, but also how they want to store it (this would be a good entrypoint to an ELK pipeline).

> This is unfortunately not a valid way to identify "unsafe" command names. You have hard-coded default command names here, whereas there is no telling these will be used by a given user. In fact, nothing in the core can depend on a specific command name.
One way around this would to have the blacklist of "unsafe" command names in settings, so that users can edit that as needed. It would be easy to cause an problematic security leak here though - it would be an easy-to-forget, obscure setting that would be easy to forget to update. A second and probably better option would be to introduce a special class variable no_log to commands like @password that you don't want to ever log.

This is an unfortunately sticky issue. The blacklist idea is likely the most practical, even with the potential of security leaks. Given the position at which I've chosen to MITM, I'm dealing with raw strings, not command objects, so I don't see how I'd be privy to class-level variables (and even then, whoever is writing a new sensitive-data-handling command would still have to be aware of the no_log variable and set it accordingly--- the leak still exists, now it's just somewhere else).

> The RawCmd( ... ) takes up a lot of space in the log. Would it not be better to clearly mark when user is not set and this might indeed be a spoof?

Yep. That was a quick and dirty solution on my part.

> I can see the motivation for not logging short 1-2 character inputs, but this can also lead to easy circumvention of the logging. For example users could use the @nick command to create a 1-2 character shortcut for a whole sequence of commands. The nick assignment would then be visible in the log but the actual invocation would not, making it hard to track events. I think it would be better to use some sort of blacklist for this as well, or a no_log variable as suggested above.

Ouch. I wasn't aware of `@nick`. Thanks for pointing that out!

@DamnedScholar 
> Sorry to take some of the thunder from this PR, but what if this were implemented differently? The same goal could be achieved by having each command call a global hook function and pass its own object to that function. If the command's __dict__ includes "audit": "false", for example, then the command would be excluded from the log. The implementation of a global "on command invocation" hook function would allow for a wide variety of other features to be built on the same foundation, such as having some commands interrupt others 

It's all good.

I did originally explore doing auditing at the command object level. Whether this is the right way to do it depends on what you want to get out of the resultant logs-- I decided against it because I want greater visibility into all strings users are throwing against the server, not just the ones that were successfully interpreted as commands. I'd like to know if users are trying to inject code or enumerate privileged commands that may or may not exist, for example, which is something that (if unsuccessful) would likely get filtered out as garbage before it ever reaches the command level.

But it could totally be done either way. Auditing at the command level would certainly be cleaner and allow much more selective flexibility.

@BlauFeuer 
> I think auditing is a good idea, but it's not something everyone wants implemented in the database or needs implemented in the server. The type of security needed will drastically vary with scope and deployment of the game, so much so that for the last ten years, nobody's addressed this need specifically.

Fair enough. One thing I'm seeing in this thread are two different use cases for logging (you all seem to be interested in different subsets of the data than me).

Auditing isn't sexy. This is a game development platform for making fun; auditing is actual work. Nobody wants to deal with it until a crisis happens, which leads to all sorts of desperate hacks to implement it post-incident. My goal here is to provide one easy-to-set-up, comprehensive option ahead of time.

@Griatch 
> strikaco's centralized logger offers the potential ability able to catch all outgoing data though (OOB, script echoes etc), for those that would want that.

Actually, it doesn't. I only capture inbound data. I did implement capture of outbound data too but it didn't work as I'd hoped-- a message broadcasted to all of a room's recipients creates one log entry per recipient. It's technically correct (each inhabitant did receive a copy of the same message) but I felt it would create too much of a logistical mess to be usable with the core toolset.

---

Thanks for all the feedback!

At this point I think the way to go might be adding a hook to the session handler and adding a hook to the base command object (`at_pre_cmd`) in core. This at least creates taps for whoever wants data from either source and doesn't force my particular implementation on everyone whether or not it's used.

Everything that follows as far as what to do with the data, or how to store it, could/should go in a contrib.

BlauFeuer:
It has gone without saying, and I want to be sure it's clear to everyone, that we have **tekmunkey** (Not going to `@` this time) to thank for bringing the topic to the forefront.

My intent in all my discourse was to shape how this evolved, not to suppress or steal and reclaim ideas.  While part of what he wanted to do is possible now with https://github.com/evennia/evennia/wiki/Coding-FAQ#store-last-used-session-ip-address , other parts are not, and depending on your need or use case, **as strikaco** says so well, the right taps for hooks in the framework allows flexibility, and "doesn't force [any] particular implementation."

DamnedScholar:
> I want greater visibility into all strings users are throwing against the server, not just the ones that were successfully interpreted as commands. I'd like to know if users are trying to inject code or enumerate privileged commands that may or may not exist, for example, which is something that (if unsuccessful) would likely get filtered out as garbage before it ever reaches the command level.

This sounds like a universal spying system. I feel like a lot of people would take issue with that.

BlauFeuer:
@DamnedScholar I took issue with it on a particular MUSH where every command anyone enters is broadcast on a channel that some or all could subscribe to.  There's nothing stopping anyone from implementing it in their own game for Evennia.   Hooks for this exist now.  But there is a real legitimate need for being able to implement further security needs and debugging hooks.

DamnedScholar:
There's absolutely nothing stopping anyone with shell access from monitoring everything that goes on in the game. But player trust will decrease if they know that there's omnibus auditing that's being monitored, without some statement about its usage and intent (the source code being public also helps).

BlauFeuer:
Out of the box, no - I had to add my own code to see when someone was typing a command that went through NOMATCH, so when I looked at console (viewed from the shell), I could then in-game respond to the complaint of a thing not working with "You have to spell the command correctly, like so...."

Every game should have a privacy notice and something that mentions how long logs are kept and what's being logged, and the users decide (like I did with that particular MUSH) if they want to continue using such a system.  That's why I believe framework hooks should be in place to allow games to be user-extensible, but auditing shouldn't really be the core focus of the game or Evennia's framework, itself.

Believe me, if I thought this PR or that feature request was over-reaching, I would say something very much, until and after people were tired of hearing it.

strikaco:
@DamnedScholar 

With regards to at_pre_cmd, I now remember my chief grievances against it-- too easy for a dev to override the hook and not call the parent, which would completely avoid the auditing hook, and the fact that command-spamming protection can also prevent the string from reaching this point. 

Beyond that I think it would still capture the raw command string, and you'd be working with proper objects instead of strings so it is definitely a more elegant solution.

> This sounds like a universal spying system. I feel like a lot of people would take issue with that.

I agree, but as any sort of server administrator you do have the right to log user activity. Just because a service doesn't say they aren't watching your every move doesn't mean they aren't actually doing it. Every GET request you make is logged by the host without disclosure. Hell, even services that market themselves as being anonymous and untraceable still monitor you and sell your information to China. It's an unfortunate reality of the world we live in today.

But just because you're logging everything doesn't mean you're compelled to spy on individuals. Just be ethical about it. Disclose up front that activity may be monitored, don't snoop and don't sell the data.

The shell access bit was addressed in the tekmunkey thread, and his counterpoint was valid-- that means you'd have to give moderators/GMs/whatever shell access. The goal is to provide an abstraction from the shell that moderators can access, ideally in-game.

BlauFeuer:
Considering I don't need that level of auditing, I would be glad to be able to overwrite the auditing hook.  And if I can't overwrite it, I hope I can configure it to turn off in some setting somewhere.

When command spamming limiter is engaged, that's another thing that should be logged in #1383.  Silently thwarting attack attempts by brute login or command saturation is something that should appear in a log, and especially in auditing, should one feel the need for auditing.

DamnedScholar:
My position is just that there are a lot of people in this community who have seen abuses in the past. Most of them know that it's ultimately a matter of trust, and they're willing to trust the server runner, but the server runner has to *seem* transparent. If nobody knows about a logging system, that's fine. If people know about a logging system and there's an explicit norm around its usage, that's usually fine. If it comes out that there's a logging system but the circumstances around it are secretive, that's when people stop trusting you.

I don't think that anybody in this discussion has indicated that they're intending to do anything bad or be less than transparent with their players, but I feel like the *subject matter itself* requires a bit of pre-emptive establishing of ethics.

Kelketek:
@DamnedScholar There's not a way to establish ethics for the use of code you do not control. Once someone installs Evennia, we're no longer in control of it. We, at best, can provide a tool. It is up to those who use it how they use it. No amount of coding will either prevent or require anyone who uses this codebase to be a spy. The only thing in particular I would suggest is having sane defaults.

It's reasonable, in my opinion, to have some debug mode logger that can be enabled or disabled as needed. If you've got a bunch of friends beta testing, that'd be more what I'd use this sort of thing for. But anything which can be used to debug can be used to spy.

CloudKeeper1:
Not entirely relevant, but the method I log inputs (and in fact asked for a specific change in Evennia to use this method) is through the [data_in and data_out methods](https://github.com/evennia/evennia/blob/master/evennia/server/serversession.py#L367) in the [serversession](https://github.com/evennia/evennia/blob/master/evennia/server/serversession.py). I use my own serversession and used [the log utility](https://github.com/evennia/evennia/blob/master/evennia/utils/logger.py#L242). Each player has a log file and it even captures the input and output prior to logging in. I essentially can see a players whole play through this way.

Griatch:
@strikaco Bump. As said, I believe this belongs in a contrib and would be happy to see this PR converted to that.

strikaco:
It's been a while since I've visited this. Since then I've made some modifications and run a modified core ServerSession that passes dicts containing the raw string and metadata for all strings inbound and/or outbound to a custom method specified in settings. Object references to the account, character, room and session are appended for the sake of those who wish to manipulate any of them directly.

Then I have a game-specific function that receives those dicts, drops the keys I don't need, masks sensitive commands, serializes the rest as json and logs it to disk. This is the simplest and most ideal solution in my opinion-- when enabled, core provides a firehose of data for the admin/analyst; how much of it is consumed and how it gets stored is game-specific. Compared to my previous PR this no longer dictates implementation of format, storage or content so this should be much more in line with project principles.

Here's a truncated gist of the modifications I made; it's the addition of an audit() function to ServerSession that gathers metadata and some checks at init, data_in and data_out to see if we need to be logging either direction. If it's acceptable I'll submit a proper pull request.

https://gist.github.com/strikaco/c12e860f74d918275712918e3bf203f0

Griatch:
@strikaco Since the `ServerSession` is pluggable with `settings.SERVER_SESSION_CLASS`, your variant would work fine as a contrib I think!

Griatch:
Closing, this is instead handled in the contrib proposed with #1535.

PR #1399, opened 2017-08-19
Remove list wrapper from value passed to Django

### LINKED ISSUE #1398: Command "evennia migrate --database=second_database" fails with TypeError
When attempting to set up Evennia with a second database, running the migration fails with an exception:

```
File "/home/user/.virtualenvs/multi_db/local/lib/python2.7/site-packages/django/core/management/commands/migrate.py", line 67, in handle
    connection = connections[db]
  File "/home/user/.virtualenvs/multi_db/local/lib/python2.7/site-packages/django/db/utils.py", line 206, in __getitem__
    if hasattr(self._connections, alias):
TypeError: hasattr(): attribute name must be string
```
This prevents one from using two databases with Evennia - since the migrations can not be run on the second db.

Regardless of whether using multiple databases is a good idea or not, it's a feature one would expect to work, since Evennia is built on Django.

This problem was pointed out to me by Roger on the IRC - I don't believe he has a GitHub account.

1. Create a new app with "evennia startapp myapp"
2. Set up a second database connection and router as described by [Django's docs](https://docs.djangoproject.com/en/1.11/topics/db/multi-db/)
2. Run "evennia makemigrations myapp"
3. Run "evennia migrate" and "evennia migrate --database=mynewdb"

Result: Django crashes with a TypeError
Expectation: Django creates and populates the second database

This is caused by Evennia's handling of arguments to be passed to Django, in [evennia_launcher.py](https://github.com/evennia/evennia/blob/master/evennia/server/evennia_launcher.py#L1329).
For some reason, we pass the value as a list:
```
kwargs[arg.lstrip("--")] = [value]
```
Instead, we should consider:
```
kwargs[arg.lstrip("--")] = value
```

Unless there is a reason to pass a list, which I was unable to find.

chainsol:
Remove list wrapper from arguments passed to Django from evennia_launcher.py

Fixes #1398, allows multi-database use of Evennia, as well as any other Django command requiring an argument to be passed with "--", for instance, "evennia migrate --database=myapp" or "evennia dbshell --database=myapp"

Closes #1398 
Tested - Does not seem to interfere, but @Griatch, you might want to check my reading - I couldn't find any reason to pass a list to Django.

Unittesting - I'm unsure if we unittest the evennia_launcher and I'm unsure of how to do so - perhaps a future thing to do.

### DIFF
--- a/evennia/server/evennia_launcher.py
+++ b/evennia/server/evennia_launcher.py
@@ -1326,7 +1326,7 @@ def main():
                         arg, value = [p.strip() for p in arg.split("=", 1)]
                     else:
                         value = True
-                    kwargs[arg.lstrip("--")] = [value]
+                    kwargs[arg.lstrip("--")] = value
                 else:
                     args.append(arg)
         try:


Griatch:
I believe the list were needed at some point but I wrote this bit many years ago. It may well be something changed in Django that we just didn't notice and didn't change since it worked fine for most situations. Merged now, thanks!

PR #1402, opened 2017-08-22
Add more clear error message to _SaverMutable

### LINKED ISSUE #1400: SaverMutable will raise ValueError on operations if underlying Attribute removed
SaverLists, SaverDicts, etc, all store the underlying Attribute as its `._db_obj` attribute, and set the Attribute's .value appropriately upon assignment, allowing for dict and list operations to save their state without needing to be reassigned to the Attribute. I ran into a strange edge case where if you're not careful, you could get a confusing error. If the underlying Attribute is deleted while the mutable still exists, you'll see the following traceback:
```
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]   File "/home/tehom/arx/evennia/evennia/commands/cmdhandler.py", line 584, in _run_command
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]     ret = cmd.func()
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]   File "./commands/commands/jobs.py", line 447, in func
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]     invite_dict[targ] = ticket
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]   File "/home/tehom/arx/evennia/evennia/utils/dbserialize.py", line 118, in save_wrapper
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]     self._save_tree()
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]   File "/home/tehom/arx/evennia/evennia/utils/dbserialize.py", line 145, in _save_tree
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]     self._db_obj.value = self
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]   File "/home/tehom/arx/evennia/evennia/typeclasses/attributes.py", line 152, in __value_set
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]     self.save(update_fields=["db_value"])
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]   File "/home/tehom/arx/evennia/evennia/utils/idmapper/models.py", line 382, in save
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]     super(SharedMemoryModel, self).save(*args, **kwargs)
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]   File "/home/tehom/arx/local/lib/python2.7/site-packages/django/db/models/base.py", line 708, in save
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]     force_update=force_update, update_fields=update_fields)
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]   File "/home/tehom/arx/local/lib/python2.7/site-packages/django/db/models/base.py", line 736, in save_base
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]     updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]   File "/home/tehom/arx/local/lib/python2.7/site-packages/django/db/models/base.py", line 792, in _save_table
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::]     raise ValueError("Cannot force an update in save() with no primary key.")
2017-08-18T16:07:17-0400 [AMPProtocol,0,127.0.0.1] [::] ValueError: Cannot force an update in save() with no primary key.
```

1. my_dict = self.db.some_dict_attr
2. self.attributes.remove('some_dict_attr')
3. my_dict[some_key] = some_value

The above traceback.

It's a wacky edge case that I only ran into when I was popping keys from a dictionary stored in an Attribute, removing the Attribute if the dict was empty, and then a player requested that I re-add the previous key on some failure condition. While it's easy to avoid, I was trying to think if there'd be some elegant way to have the SaverMutables handle this, but not sure if there's a good solution. Are the fields of a deleted object still accessible after it's been deleted? If so, maybe could create a new Attribute and save it.

TehomCD:

Adds a more clear error message for the edge case of someone removing an Attribute and then modifying the mutable that was stored there.

To prevent confusion in anyone as easily confused as I am.

Resolves #1400

### DIFF
--- a/evennia/utils/dbserialize.py
+++ b/evennia/utils/dbserialize.py
@@ -153,6 +153,12 @@ class _SaverMutable(object):
         if self._parent:
             self._parent._save_tree()
         elif self._db_obj:
+            if not self._db_obj.pk:
+                cls_name = self.__class__.__name__
+                non_saver_name = cls_name.lstrip("_Saver")
+                err_msg = "%s %s has had its root Attribute deleted." % (cls_name, self)
+                err_msg += " It must be cast to a %s before it can be modified further." % non_saver_name
+                raise ValueError(err_msg)
             self._db_obj.value = self
         else:
             logger.log_err("_SaverMutable %s has no root Attribute to save to." % self)


evennia/utils/dbserialize.py
```diff
@@ -153,6 +153,12 @@ def _save_tree(self):
         if self._parent:
             self._parent._save_tree()
         elif self._db_obj:
+            if not self._db_obj.pk:
+                cls_name = self.__class__.__name__
+                non_saver_name = cls_name.lstrip("_Saver")
```
Griatch: It's a small thing so I'll just fix this before I merge, but if you are not aware, this is *not* the way to strip a fixed string from the beginning of another string. What `lstrip` (and `rstrip` as well as `strip`) takes for argument is a *sequence of characters* to remove. The order of them does not matter, nor how many that should be removed. To see what I mean, try what you get if you do 
```python
"testtsettttsseset".lstrip("test")
```

Unless you want to use indexing (`cls_name[5:]`) I think the commonly recommended to strip an initial string is to use split: 
```python
non_saver_name = cls_name.split("_Saver", 1)[1]
```

Issue #1388, opened 2017-08-06
Tests failed for in-game python contrib


TehomCD:
#### Brief summary of issue / Description of requested feature:
I'm finally trying to start writing unit tests for some of my own code, and was trying to get the test runner working right for my own settings file. It threw syntax errors on the unit tests for the in-game python contrib with an error I'm really unfamiliar with:

```
======================================================================
ERROR: test_accept (evennia.contrib.ingame_python.tests.TestCmdCallback)
Accept an callback.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\arx\evennia\evennia\contrib\ingame_python\tests.py", line 252, in setUp
    "evennia.contrib.ingame_python.scripts.EventHandler")
  File "c:\arx\evennia\evennia\utils\create.py", line 178, in create_script
    typeclass = class_from_module(typeclass, settings.TYPECLASS_PATHS)
  File "c:\arx\evennia\evennia\utils\utils.py", line 1357, in class_from_module
    mod = import_module(testpath, package="evennia")
  File "C:\Python27\Lib\importlib\__init__.py", line 37, in import_module
    __import__(name)
SyntaxError: unqualified exec is not allowed in function 'call' it contains a nested function with free variables (scripts.py, line 448)
```

The above error repeated for each unit test.

Looking at this [stackoverflow post](https://stackoverflow.com/questions/9195309/any-explanation-of-execs-behavior), it seems that it's a python 2.x problem - in 3.x, exec() is a function, which is how it was used there, but previously it needs to be qualified with the 'in' syntax if there's generators used in the function it's called.

#### Steps to reproduce the issue / Reasons for adding feature:

1. I ran `evennia test --settings settings.py evennia`
2. Saw above error

#### Error output / Expected result of feature
Above traceback for each unit test.

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
Some sort of compatibility fix so it runs in both python 2.7 and 3.x? I'm not sure what a good solution is there.

Griatch:
I have asked @v-legoff and he'll take a peek since he's the author of the contrib.

vlegoff:
[...]

PR #1413, opened 2017-09-01
Put older syntax for the exec statement and fix can_say and say events

vlegoff:

This PR fixes issue #1388 (changing the syntax of the `exec` statement to support older Python versions) and fix the `can_say` and `say` events.

- Fixing the issue #1388 (although it's already closed).
- Fixing the `say` commands to allow `can_say` and `say` events to still be called.

### DIFF
--- a/evennia/contrib/ingame_python/scripts.py
+++ b/evennia/contrib/ingame_python/scripts.py
@@ -454,7 +454,7 @@ class EventHandler(DefaultScript):
                 continue
 
             try:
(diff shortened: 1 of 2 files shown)


Griatch:
I will merge the contrib fixes; the exec() syntax should stay as it is, that's also the way future versions of `exec` will have to be called. We'll tell people to upgrade if they run into this issue.

Griatch:
Lifted only the contrib changes into devel. Closing.

PR #1414, opened 2017-09-01
Add a confirmation to the default @destroy command

vlegoff:

The default `@destroy` command now displays a confirmation before deleting.  This behavior can be bypassed in several ways.

- Using the `@destroy/force` switch, no confirmation is displayed.
- Modifying the `confirm` class variable in the `CmdDestroy` class, no confirmation will ever be displayed.

Destroying objects has to be efficient, but lots of damage can be done by a change in letter.  The default behavior is a bit more secure: it will display a confirmation message (with the list of keys and IDs) before destroying.  The user has to enter `y` or `yes` in order to continue.  If the user is really sure about what she does, the `/force` switch has been added to the command.  If an admin wishes to go back to the previous behavior (no confirmation), a simple class variable on the command can be changed.

Since default commands are meant to present default code to use in the commands, this is a great opportunity to have a default command using `yield` to require user input.

#### Testing

Testing this feature is difficult.  The tests written on `@destroy` have been slightly altered in order to bypass confirmation.  It means that the confirmation itself is not tested.  It might be worth adding tests to see if the confirmation effectively works, but I do not know how to write them.

Additionally, I have added some lines to the `CommandTest.call` which didn't support testing commands with `yield` statements.

None

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -614,35 +614,37 @@ class CmdDestroy(COMMAND_DEFAULT_CLASS):
     switches:
        override - The @destroy command will usually avoid accidentally
                   destroying account objects. This switch overrides this safety.
+       force - destroy without confirmation.
     examples:
        @destroy house, roof, door, 44-78
        @destroy 5-10, flower, 45
+       @destroy/force north
 
     Destroys one or many objects. If dbrefs are used, a range to delete can be
-    given, e.g. 4-10. Also the end points will be deleted.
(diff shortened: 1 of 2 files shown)


Griatch:
I merged this with an addition - a default choice, which allows you to accept the deletion by just pressing Enter. I think this makes for a good intermediary between the current no-safety check and having to explicitly answer write "y/es" every time.
I set the default with another class variable `confirm_default` (so you could easily change it to default to 'no' if you wanted).

PR #1423, opened 2017-09-12
Contrib/slots handler

DamnedScholar:
#### Brief overview of PR additions
This contrib introduces SlotsHandler, a system for registering an arbitrary number of slots on a typeclassed object. It's intended to be used as a storage system by game system commands, where any sort of Python object can be recorded in each slot. The basic use case of this is to provide a limiter for entities like magic items and spells, but it is as generic as possible, so it should be useful for a broad array of systems.

The behavior required for slot-based mechanics isn't strongly supported, and this handler serves as a wrapper that performs management operations to allow for a developer to easily create/delete slots and fill/empty them.

### DIFF
--- a/evennia/contrib/slots_handler/readme.md
+++ b/evennia/contrib/slots_handler/readme.md
@@ -1,0 +1,84 @@
+# Slots Handler
+It's pretty common for RPG systems to have mechanics where a character is limited to, or must choose, a specific number of items. This contrib is designed to be adaptable to handle as many permutations of this style of mechanic as possible.
+
+## How to use
+### Slotted object example
+The `@lazy_property` declaration below can be put on any typeclass parent. The functionality of `SlotsHandler` relies on the `AttributesHandler`, so any deviations from the standard typeclass system have to at least include that. This document will assume that you're using the name `slots` for the property.
+```python
+from evennia.objects.objects import DefaultObject
+
+class SlottedObject(DefaultObject):
+    "This is a test object for SlotsHandler."
+
+    @lazy_property
+    def slots(self):
+        return SlotsHandler(self)
+```
+### Slottable object example
+```python
+from evennia.objects.objects import DefaultObject
+
+class SlottableObject(DefaultObject):
+    "This is a test object for SlotsHandler."
+
+    def at_object_creation(self):
+        self.slots = {"addons": ["left"]}
+```
+### Concepts
+**Slots format:** In this contrib, all arguments and attributes labeled slots are intended to come in one of two formats:
+```python
+["magic items", "spells", "stunts"]
+```
+This represents a list of categories, and is used for targeting every slot in each of the named categories. There's no point in using this format for `add()`, but it can be used to delete a whole category, list specific categories, or attach/drop in broad strokes.
+```python
+{"spells": [2, "class bonus", "racial bonus"], "stunts": ["athletics", "weaponry"]}
+```
+This format is used to target specific slots. If numerical values are included, they will be treated in specific ways. For `add()` and `delete()`, the numerical values will be summed and an equal number of numbered slots will be added to or deleted from the category. For `attach()` and `drop()`, they will be summed and that number of slots will be used or freed up. For `check()`, the first *n* numbered slots will be returned.
+
+**Slot category:** SlotsHandler interprets the top-level keys of a dict, or the values of a list, to be categories. Categories are important as they factor into the storage name of the slot list, so if you don't want to separate your slots, give them a very general category such as `"slots"`.
+
+**Numbered vs named slots:** Slot keys must be integers or strings. Numbered slots are anonymous and mostly will not be referred to directly. Named slots are for cases where you want to refer to them directly.
+
+**Nesting:** The system is designed with simplicity in mind. You should not try to force individual slot attributes to be more than a basic `dict`. However, since we're storing objects in these slots, it's perfectly possible to store one slot attribute inside a slot in a different attribute. Your code just has to know to look for that, because `SlotsHandler` does not. The individual building blocks are simple, but they may be assembled in more complicated ways than they can achieve on their own.
+
+## SlottedObject.slots.add
+```
+add(self, slots)
+```
+> Create an array of slots, or add additional slots to an existing array.
+**Args:**
+* **slots (dict):** A dict of slots to add. Since you can't add empty categories, it would be pointless to pass a list to this function, and so it doesn't accept lists for input.
+
+## SlottedObject.slots.delete
+```
+# I feel like I need to change this syntax to conform with the dict model.
+delete(self, slots)
+```
+> Delete the named category or slots.
+**Args:**
+* **slots (list or dict):** Slot categories or individual slots to delete.
+
+## SlottedObject.slots.attach
+```
+# I feel like I need to change this syntax to conform with the dict model.
+attach(self, target, slots=None)
+```
+> Attempt to attach the target in all slots it consumes. Optionally, the target's slots may be overridden.
+**Args:**
+* **target (object):** The object being attached.
+* **slots (list or dict, optional):** Slot categories or individual slots to drop from.
+
+## SlottedObject.slots.drop
+```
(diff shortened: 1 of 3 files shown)


TehomCD:
One thing I noticed when running tests that might be worth discussing going forward is possibly making any typeclass in contribs have as specific a name as possible to prevent collisions in games when using the test runner: any typed object is treated as being from the same application no matter where they come from. The problem is that any name clashes will stop tests from running, which is far more likely with something like `SlottedObject` or `Readable` in the tutorial world than, say `ContribSlottedObject` or `TutorialReadable`. I worry that with contribs with more general names, like, say, `Wearable` or `Readable` or whatever, collisions are actually fairly likely unless they're disambiguated in some way. In this case, `SlottedObject` is fairly unlikely unless someone was making a game that had slot machines and just decided to call a class that, but still.

DamnedScholar:
Fair point. And since the names only have to be highly specific in test files, we can enforce strict namespace rules without negatively impacting the experience of developers using the contribs.

`ContribSlottedObject` would be very safe from interference, but a name like `ContribWearable` could still have a clash from a different contrib that implements armor in a different way. Let's say two community members want very different gear implementations, and both ideas are developed enough for contribs. One of them gets named `Easy Gear` and the other is called `Uberarmor`. Names like `ContribEasyWearable` and `ContribUberWearable` would be extremely safe against the chance of accidental conflict, without hampering readability.

evennia/contrib/slots_handler/readme.md
```diff
@@ -0,0 +1,110 @@
...
+```
+> Delete the named category or slots.
+> WARNING: If you have anything attached in slots when they are removed, the slots' contents will also be removed. This function will return a dict of any removed slots and their contents, so it can act as a pop(), but if you don't catch that data, it WILL be lost.
+
+**Args:**
+* **slots (list or dict):** Slot categories or individual slots to delete.
+
+## SlottedObject.slots.attach
+```
+attach(self, target, slots=None)
```
Griatch: Reading this as a new potential user I don't understand what "attach the target in all slots" means. Why would you want to regularly add the target object to *all* slots? Could you elaborate?

evennia/contrib/slots_handler/readme.md
```diff
@@ -0,0 +1,110 @@
+# Slots Handler
```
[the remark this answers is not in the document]
Griatch: Also, look at the headers/readme's of all other contribs. They follow a special format where the author's name (you) is included.

evennia/contrib/slots_handler/readme.md
Griatch: I also think most of these could probably do with a brief example (not strictly necessary but helpful)

evennia/contrib/slots_handler/slots.py
```diff
@@ -0,0 +1,385 @@
...
+
+    def __init__(self, obj):
+        self.obj = obj
+        self._objid = obj.id
+
+    # Internal utility methods
+    def is_name_valid(self, test):
+        "Internal function to check if the name is right."
+        try:
+            exec(test + " = 1")
```
Griatch: I suggest this entire check should be removed. If this is code only intended for developers, this check is pointless - if a wrong variable name is used a traceback will catch it. 
If this contrib is used by some overlying build command it's much worse - using `exec()` - and completely without checking the input too - you are now opening up yourself to a huge security exploit - if the input `test` (for example passed from some build command to `slots.add or slots.delete`) is in any way in the hands of an untrusted user they can now execute Python code unchecked on your server - the same power as `@py`. Even if you don't intend for more than internal use, the use of `exec()` is not advertised anywhere and may not be caught and the input properly verified by an user of this contrib. So, please remove this.

evennia/contrib/slots_handler/slots.py
```diff
@@ -0,0 +1,385 @@
...
+        except SyntaxError:
+            raise ValueError("You have to limit slot category names to "
+                             "characters that are valid in variable names.")
+
+    def defrag_nums(self, cats):
```
Griatch: Since this is not a private function, it should have a *Google style docstring* as per the [contribution guidelines](https://github.com/evennia/evennia/wiki/Contributing#contributing-with-contribs). This is true for all methods and functions in the contrib (I won't mark them each explicitly).

evennia/contrib/slots_handler/slots.py
```diff
@@ -0,0 +1,385 @@
...
+            return r
+
+    def add(self, slots):
+        """
+        Create arrays of slots, or add additional slots to existing arrays.
+
+        Args:
+            slots: A dict of slots to add. Since you can't add empty
+                categories, it would be pointless to pass a list to this
+                function, and so it doesn't accept lists for input.
```
Griatch: This misses a `Returns` docstring section.

evennia/contrib/slots_handler/slots.py
```diff
@@ -0,0 +1,385 @@
...
+            # Return a dict detached from the database.
+            r = {s.key: s.value for s in d}
+            return r
+
+    def add(self, slots):
+        """
+        Create arrays of slots, or add additional slots to existing arrays.
+
+        Args:
+            slots: A dict of slots to add. Since you can't add empty
```
Griatch: See the style guide for how to write the type of arguments in docstrings.

evennia/contrib/slots_handler/slots.py
```diff
@@ -0,0 +1,385 @@
...
+                categories, it would be pointless to pass a list to this
+                function, and so it doesn't accept lists for input.
+        """
+
+        if not isinstance(slots, (dict, _SaverDict)):
+            return StatMsg(False, "You have to declare slots in the form "
```
Griatch: What is `StatMsg` and where is it defined/imported?

evennia/contrib/slots_handler/slots.py
```diff
@@ -0,0 +1,385 @@
...
+        return deleted
+
+    def attach(self, target, slots=None):
+        """
+        Attempt to attach the target in all slots it consumes. Optionally, the target's slots may be overridden.
+
+        Args:
+            target (object): The object to be attached.
+            slots (list or dict, optional): If slot instructions are given,
+                this will completely override any slots on the object.
```
Griatch: A `Returns:` section is missing (it's missing all over, so won't point those out henceforth). An `Examples:` section is optional but may be useful too here.

evennia/contrib/slots_handler/slots.py
DamnedScholar: Done.

evennia/contrib/slots_handler/slots.py
DamnedScholar: Apologies. I've been playing around with a status message object and forgot to remove the references to it. I've changed all of those to `raise Exception()`.

evennia/contrib/slots_handler/slots.py
DamnedScholar: I'm not going to include `Examples:` in the docstrings because I feel like they would take up too much space and I can include a more thorough tutorial in the `README`.

evennia/contrib/slots_handler/readme.md
DamnedScholar: > Reading this as a new potential user I don't understand what "attach the target in all slots" means.

The full clause is "attach the target in all slots it consumes", but that's probably not the best way to word it. I'll make sure to talk about it more thoroughly in the walkthrough. The reason for the "all slots" wording is the fact that the `attach()` method will only work if every one of the slots requested is available.

Griatch:
@DamnedScholar Please let me know when you feel the issues have been fixed and this is ready to review again. :)

DamnedScholar:
Alright. I've added a walkthrough to the readme that should make it a breeze for new people to get into working with the contrib. Everything has better docstrings and in replicating the real-world use I ran into an edge case bug that would have been pretty embarrassing. So that's good.

Griatch:
Ah, didn't intend to close this. Please re-open against the `develop` branch.

PR #1436, opened 2017-09-19
Added basic Typeclass support to CmdTunnel

### LINKED ISSUE #1435: @tunnel doesn't support setting the exit typeclass
I was creating tunnel locations and would go and change the typeclass for the exit after the object was already created, this would break all the connetions from room to room. After looking further it appears the tunnel class doesn't have the ability to accept a typeclass for the exit

digstring = "@dig%s %s = %s;%s%s" % (telswitch, roomname,
                                             exitname, exitshort, backstring)
the dig string is passed from tunnel class into @dig this doesn't allow one to pass exit to and here typeclass on creation

@dig[/switches] <roomname>[;alias;alias...][:typeclass]
            [= <exit_to_there>[;alias][:typeclass]]
               [, <exit_to_here>[;alias][:typeclass]]

1. @tunnel e = supply shop
2.  @typeclass e = typeclasses.exit.door.Door
3.  All room connections break

cant move from room to room

chainsol:
Fixes #1435 by adding typeclass support for the exit to CmdTunnel

@Griatch: I couldn't think of or find a way to double check the typeclass is set correctly in my test - I did test my changes on my own game and it works as intended.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -866,7 +866,7 @@ class CmdTunnel(COMMAND_DEFAULT_CLASS):
     create new rooms in cardinal directions only
 
     Usage:
-      @tunnel[/switch] <direction> [= <roomname>[;alias;alias;...][:typeclass]]
+      @tunnel[/switch] <direction>[:typeclass] [= <roomname>[;alias;alias;...][:typeclass]]
 
     Switches:
       oneway - do not create an exit back to the current location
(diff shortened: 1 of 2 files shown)


chainsol:
I don't know if this should be merged to master instead; my instinct is to build off devel instead since it's merging so soon...

evennia/commands/default/building.py
```diff
@@ -911,21 +911,29 @@ def func(self):
...
             self.caller.msg(string)
             return
+
         # retrieve all input and parse it
-        exitshort = self.lhs
         exitname, backshort = self.directions[exitshort]
         backname = self.directions[backshort][0]
 
+        if ":" in self.lhs:
+            exit_typeclass = ":" + self.lhs.split(":")[-1]
```
Griatch: You need to make sure to do `self.lhs.split(":", 1)[-1]` here or extra colons will confuse the parsing.

PR #1437, opened 2017-09-21
Equipment expansion for "turnbattle" contrib

FlutterSprite:
Adds a new contrib module: turnbattle_equip, an implementation of the turnbattle contrib that includes a basic, expandable system for handling weapons and armor.

Adds a new version of the 'turnbattle' contrib module that handles weapons (including variable accuracy and damage) and armor (with variable damage reduction and hit chance). The module also includes a few prototypes for easy testing.

Equipment is a near-universal feature of MUDs featuring combat systems, so a good framework for using weapons and armor can see a lot of use. It also serves as a good example of how the 'turnbattle' module can be expanded, and therefore act as a resource for newer developers.

The unit tests aren't in yet, but it's otherwise ready to test and mess around with!

### DIFF
--- a/evennia/contrib/turnbattle_equip.py
+++ b/evennia/contrib/turnbattle_equip.py
@@ -1,0 +1,1051 @@
+"""
+Simple turn-based combat system with equipment
+
+Contrib - Tim Ashley Jenkins 2017
+
+This is a version of the 'turnbattle' contrib with a basic system for
+weapons and armor implemented. Weapons can have unique damage ranges
+and accuracy modifiers, while armor can reduce incoming damage and
+change one's chance of getting hit. The 'wield' command is used to
+equip weapons and the 'don' command is used to equip armor.
+
+Some prototypes are included at the end of this module - feel free to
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks for the contrib! Haven't looked at the code yet, some initial administrative points: 

- Given that this is a module specifically aimed for use with another contrib, I wonder if it would not be better to add a new subfolder `contrbs/turnbattle/` and put the the existing `turnbattle.py` as this new module in there. Then all modules related to turbattle (I saw in chat you might add more) could go together and be imported used from a single place. Look at the other sub-packages in contrib for hints on how to set it up (all must have a `README.md` for example). 
- Since this is a new feature, please re-point this PR against the `develop` branch (see the announcement in the forum). [Here](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request) is how you do so easily from github.

FlutterSprite:
I'm going to go ahead and start this PR over from scratch, since I moved the Turnbattle modules into their own subfolder and such!

PR #1446, opened 2017-09-26
New hooks: at_before_get, at_before_give, at_before_drop

FlutterSprite:
This PR adds three new hooks to the default object: at_before_get, at_before_give, and at_before_drop. By default, they simply return True - however, they can be made to execute additional code before an object is gotten, given, or dropped, and can even prevent the action from occurring if they return False or None.

I've often run into the problem of wanting to more easily be able to prevent getting/dropping/giving objects based on circumstance - especially for contribs, where I want to override as little default code as possible. Several in the IRC chat expressed that these hooks would be helpful as well when I brought up the issue.

This change should be backwards compatible - developers who've overloaded their get / give / drop commands with their own custom versions won't run into any problems, but will have to update their code to reference the new hooks if they want to make use of them.


FlutterSprite:
There! All ready to go!

Griatch:
Thanks for the contrib! I'll need to ponder this a little bit, especially together with other hook API fixes in the pipeline. So I'll get back to you soon!

chainsol:
I know you'll still have to ponder this - but I would already find these hooks very useful, and they're pretty simple, at least!

Griatch:
With the `at_say` issues resolved I have looked more into these. And while I think we should generally be careful with the amount of hooks we have on the default objects specifically handling Commands, these do make sense for a vast number of game types; get/give/drop is a very basic set of actions. 
So I merged this into the `develop` branch (using git cherry-pick to avoid the spurious commits). Closing as merged. Thanks!

PR #1449, opened 2017-09-30
Corrects lock on @SetDesc

CloudKeeper1:
Typo made lock check to fail for all users.


CloudKeeper1:
Disregard. Shot the gun early.

PR #1452, opened 2017-09-30
Fix CmdSay in RPSystem contrib

### LINKED ISSUE #1451: RPSystem Contrib CmdSay Doesn't use at_say Hook.
The RPSystem Contrib overrides the CmdSay command - but hasn't been updated to use the new at_say hook, it seems.

1. Install the RPSystem contrib
2. Add the CmdSay to your character
3. Attempt to speak

Instead of your message, Character says: "None"

It seems CmdSay from RPSystem doesn't know about our new at_say hook, which isn't surprising.

chainsol:
Fixes #1451 by removing the call to room.at_say

CmdSay from the RPSystem contrib doesn't work under 0.7 - because of the change to at_say. This breaks CmdSay for anything using the RPSystem, including Ainneve.

### DIFF
--- a/evennia/contrib/rpsystem.py
+++ b/evennia/contrib/rpsystem.py
@@ -858,7 +858,7 @@ class CmdSay(RPCommand):  # replaces standard say
             return
 
         # calling the speech hook on the location
-        speech = caller.location.at_say(caller, self.args)
+        speech = self.args
         # preparing the speech with sdesc/speech parsing.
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks for the PR! You are correct that `at_say` should not be called since that is these days also responsible for the actual send. The pre-say speech modification should still happen though, it's just that this has now moved to the hook `at_before_say`. I fixed that in a small commit so closing this.

chainsol:
Thanks, Griatch!

PR #1464, opened 2017-10-03
RPsystem contrib: Pass at_before_say the corrent number of arguments

amfl:

The `at_before_say` method does not accept `caller`. At present, it throws an error when you try to speak:

```
Exits: riverport(#7)
say hi
  File "/home/amfl/evennia/commands/cmdhandler.py",
line 591, in _run_command
    ret = cmd.func()
  File "/home/amfl/evennia/contrib/rpsystem.py",
line 861, in func
    speech = caller.location.at_before_say(caller, self.args)
TypeError: at_before_say() takes exactly 2 arguments (3 given)

An untrapped error occurred.
```

Introduced in 291fd6fa34080540a57524b619eb8523e283d441

### DIFF
--- a/evennia/contrib/rpsystem.py
+++ b/evennia/contrib/rpsystem.py
@@ -858,7 +858,7 @@ class CmdSay(RPCommand):  # replaces standard say
             return
 
         # calling the speech hook on the location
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks!

PR #1465, opened 2017-10-03
add `@tel/loc`

### LINKED ISSUE #1462: Feature Request: Add a switch to @tel to allow teleporting to an object's location
#### Description of requested feature:
Evennia's `@tel` command defaults to teleporting your character into another object, for instance:
`@tel Griatch`
would teleport my puppeted character into Griatch's inventory.
This is a little odd, but expected, since `@tel` is meant to be used to teleport objects into other objects, e.g. `@tel an apple = chainsol` will teleport the apple to your inventory. For teleporting to other characters, this gets a little odd.

#### Reasons for adding feature:
Most people would expect `@tel chainsol` to teleport their character to the location. For a room or a container, Evennia's current behaviour makes sense. However, for teleporting to another character, it's a little odd - you end up part of their "inventory".

This is because Evennia treats objects differently than most systems - characters don't actually have an inventory, they have "contents". However, usually if you're teleporting to a character you expect to be in their location, not their inventory.

#### Expected result of feature/ideas on how to implement
One easy way to solve this is to add a switch to `@tel`, something like `@tel/loc` that teleports to the target's location.

RyanStein:
How does this look?

Adds `@tel/loc` and also fixes an unescaped `|` in the error message.

Resolves #1462.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -2324,6 +2324,7 @@ class CmdTeleport(COMMAND_DEFAULT_CLASS):
                Note that the only way to retrieve
                an object from a None location is by direct #dbref
                reference. A puppeted object cannot be moved to None.
+      loc - teleport object to the target's location instead of its contents
 
     Teleports an object somewhere. If no object is given, you yourself
     is teleported to the target location.     """
@@ -2343,6 +2344,7 @@ class CmdTeleport(COMMAND_DEFAULT_CLASS):
         # setting switches
         tel_quietly = "quiet" in switches
         to_none = "tonone" in switches
+        to_loc = "loc" in switches
 
         if to_none:
             # teleporting to None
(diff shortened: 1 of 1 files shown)


CloudKeeper1:
Hey Ryan, 
Congrats on your first contribution to Evennia!
I threw something together when Chainsol initially raised the issue and decided to sit on it to see if someone else wanted to pick it up. If you were curious you could [have a look.](https://pastebin.com/Xnd1ePf3) Good find on the unescaped |, I missed that at first! Other things I thought was worth adding was a few more comments for easier reading and and changing strings to use format(). Just a few ideas that might be worth adding.

Griatch:
@RyanStein I think this looks fine! I ran it and things seems to work as specified in #1462. As @CloudKeeper1 points out, one could always consider more comments, but the `@tel` command is pretty sparse in its feedback as it is and expanding that was not in the specification. So thanks again. Merging.

PR #1467, opened 2017-10-04
Updated dockerfile to clean alpine image and add python + required li…

gstoel:
Dockerfile did not work for me, so fixed:

- errors with apk installing libraries
- changed dockerfile to base alpine image
- added required libraries to apk command
- added RUN commands to init and migrate evennia
- now docker container can be started by: 
- docker run -it -p 4000:4000 -p 4001:4001 -p 4005:4005 --rm evennia sh
- (user gets inital prompt for admin user, etc.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,17 +10,16 @@
 #
 # https://github.com/evennia/evennia/wiki/Running%20Evennia%20in%20Docker
 #
-FROM python:2.7-alpine
-MAINTAINER Dan Feeney "feend78@gmail.com"
+FROM alpine
 
 # install compilation environment
-RUN apk update && apk add gcc musl-dev
+RUN apk update && apk add python py-pip python-dev py-setuptools gcc musl-dev jpeg-dev zlib-dev
 
 # add the project source
 ADD . /usr/src/evennia
 
 # install dependencies
-RUN pip install -e /usr/src/evennia
+RUN pip install -e /usr/src/evennia --index-url=http://pypi.python.org/simple/ --trusted-host pypi.python.org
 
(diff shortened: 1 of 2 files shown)


gstoel:
long time since I contributed to github repo - so please let me know when i have to do it differently (fork, branch, update code, commit, create PR)

Griatch:
@gstoel The PR itself is correct, thanks! It's a toss-up if this goes into master- or develop branch but if the old docker causes issues it's a bug fix. I however recall this docker image working fine for the instructions given in the wiki, this seems to operate slightly different though. What was "not working" for you before, could you elaborate? 

As for the diff, it looks ok as far as I can tell, but I'm no expert on docker. I would like to see an expanded doc string at the top documenting how to set up and run it though; it would be nice to have this all in one place. 

I'm pinging @feend78, who is the maintainer of the existing docker image. @gstoel, assuming this gets merged, would you be willing to update https://github.com/evennia/evennia/wiki/Running%20Evennia%20in%20Docker with eventual needed changes to match?

fuzzyweapon:
Is it that the upstream one (python:2.7-alpine) was causing problems?  I was about to test using the project dockerfile but I've been busy setting up dev env.  If the upstream was having problems, it's better practice to submit changes to that one and let it trickle down to everyone rather than just fix the issues here.

Griatch:
@gstoel, @feend78: bump?

Griatch:
With no further replies from anyone on this, I dug into the docker image and reworked how it works a bit, based on this PR. I also rewrote the documentation for docker since it can now be used as a way to install as well.

PR #1469, opened 2017-10-05
Fix #1468 by checking if Notification is available

### LINKED ISSUE #1468: Webclient crashes if Notification API isn't available.
Viktyr on the IRC reported this issue, and posted the output of the Console on their browser.
If a browser doesn't support the Notification API for some reason, whether it be an old browser or a plugin/feature disabling it, Evennia's webclient_gui.js crashes without displaying anything.

1. Use a browser that doesn't support the Notification API
2. Attempt to use Evennia's webclient
3. You'll get a black screen, with an error in the console

Error: `ReferenceError: Notification is not defined:  webclient_gui.js:430:5`

This is easily fixable by checking if "Notification" is in window; if not, we don't have the Notification object in JavaScript.

chainsol:
In `webclient_gui.js`, we attempt to call the Notification API without checking if it's available. If it's not available, we get a crash as described in #1468.

Fixes #1468, also may add support for browsers/plugins that don't have the Notification object available for whatever reason.

Closes #1468

### DIFF
--- a/evennia/web/webclient/static/webclient/js/webclient_gui.js
+++ b/evennia/web/webclient/static/webclient/js/webclient_gui.js
@@ -369,28 +369,29 @@ function onNewLine(text, originator) {
     unread++;
     favico.badge(unread);
     document.title = "(" + unread + ") " + originalTitle;
+    if ("Notification" in window){
+      if (("notification_popup" in options) && (options["notification_popup"])) {
+          Notification.requestPermission().then(function(result) {
+              if(result === "granted") {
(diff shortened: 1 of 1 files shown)


chainsol:
If anyone wishes to check this out, it's available at [my development server](https://dev.perfectdisorder.net/webclient).

Griatch:
Whereas I cannot test the particular browser weirdness that prompted this I see no negative effects with having the check in. So merging this, should make it easier for the user to test if it works out.

PR #1470, opened 2017-10-05
make help command's matching respect CMD_IGNORE_PREFIXES

### LINKED ISSUE #1466: Make Evennia's default help function ignore prefixes
#### Brief summary of issue

> *For hacktoberfest folks*: This is a one-module fix that should be quite doable once getting a grasp of how things hang together. 

Evennia treats e.g. `@look` the same as `look` since the prefix is ignored by default. However, when a user, used to entering `@look` tries to get help with `help @look` the `help` command will report that it doesn't recognize `@look` as a valid command. Only `help look` works. 

1. `@look` works fine to look at the current location.
2. `help @look` gives a not-found error.

#### Expected result of feature

The `help` command must honor the prefix-ignore defined in settings, and in the same way. That is, when a user enters `help @cmdname`, the following must happen: 
 -  If a command with a key or alias `@cmdname`, `+cmdname` etc (that is, including the prefix) *actually exists* then the help entry of this should be displayed. 
 - If no exact match is found, and a prefix exists, then strip it and try to show the help entry for the stripped name.
- If no match is found neither with nor without the prerix, show not-found error. 

Since Evennia 0.7, Evennia defaults to ignoring prefixes given by [settings.CMD_IGNORE_PREFIXES](https://github.com/evennia/evennia/blob/master/evennia/settings_default.py#L305) . This means that by default a user entering `@look`, `+look`, `&look` and `/look` will get the same result as if they just entered `look`. This stripping happens in Evennia's [cmdparser](https://github.com/evennia/evennia/blob/master/evennia/commands/cmdparser.py#L91) and thus works automatically for all commands. 
However, in the case of the `help` command, what is failing is the parsing of the argument to the `help` command, which is the name of the command to get help for. The default [help command is defined here](https://github.com/evennia/evennia/blob/master/evennia/commands/default/help.py#L27) and all changes for this issue should be able to be isolated to this module. The argument to the command is available as `.args` on the Command class when `.parse()` and then `.func()` runs.

RyanStein:
Changes `CmdHelp` to respect `CMD_IGNORE_PREFIXES` by stripping prefixes from both queries and command matchsets, allowing such lookups as `help +locks` and `help locks` to redirect to the `@locks` docstring and et cetera.

Resolves #1466.

### DIFF
--- a/evennia/commands/default/help.py
+++ b/evennia/commands/default/help.py
@@ -17,6 +17,7 @@ from evennia.utils.utils import string_suggestions, class_from_module
 
 COMMAND_DEFAULT_CLASS = class_from_module(settings.COMMAND_DEFAULT_CLASS)
 HELP_MORE = settings.HELP_MORE
+CMD_IGNORE_PREFIXES = settings.CMD_IGNORE_PREFIXES
 
 # limit symbol import for API
 __all__ = ("CmdHelp", "CmdSetHelp")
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks, this looks good and works as requested! I merged with a rebase. Closing.

Issue #1458, opened 2017-10-01
Create new unit tests for Evennia - hacktoberfest


Griatch:
#### Brief summary of issue / Description of requested feature:

This is a [hacktoberfest](https://hacktoberfest.digitalocean.com/) issue for expanding Evennia's unit test coverage. 

#### Steps to reproduce the issue / Reasons for adding feature:

1. Go to [the coveralls page for Evennia](https://coveralls.io/github/evennia/evennia) 
2. Find an Evennia module under `evennia/` that lacks unit test coverage 
3. If no `tests.py`file exists in the sub-package in question, add it. See other `tests.py` in the Evennia repo for help.

#### Expected result of feature

We are currently at 53% unit test coverage, why not expand that? 

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:

This is against `master` branch.

vaibhavsingh97:
i can give it a try 😄

Keda87:
Hi, just find out this repo for hacktoberfest.
I'll try to add some test tonight, do you have any convention/documentation to create test and PR?

Griatch:
@Keda87 That's great! I updated the issue with some more info. :)

Griatch:
I also added some exceptions. No need to dig into making unit tests for the game_template since it's basically well, empty template files.

Keda87:
@Griatch how can I see coverage report for testing on my local machine?

Griatch:
@keda87 It's not necessary to run `coverage` locally for this issue. But if you want to try it you need to follow these steps (only tested on Linux): 

1. In your virtualenv, do  `pip install coverage`
2. Assuming you followed Evennia according to the `Getting Started` instructions you should have your `evennia/` folder sitting next to your game dir (called e.g. `mygame/`). `cd` into your game dir.
3. Run one of the following:        
```bash
# linux 
coverage run --source=../evennia --omit=*/migrations/*,*/urls.py,*/test*.py,*.sh,*.txt,*.md,*.pyc,*.service ../evennia/bin/unix/evennia test evennia

# windows (untested)
coverage run --source=../evennia --omit=*/migrations/*,*/urls.py,*/test*.py,*.sh,*.txt,*.md,*.pyc,*.service ../evennia/bin/windows/evennia test evennia
```
[...]

Issue #1455, opened 2017-10-01
@spawn objects have no location or home after ev 0.7


scyfris:
#### Brief summary of issue / Description of requested feature:

I recently switched to Evennia 0.7.  Spawning objects seems to have regressed, where objects without an explicit location are given locations of "None".  Previously, the behavior was to give a location of the caller if location was not specified (unless using the "noloc" switch).

#### Steps to reproduce the issue / Reasons for adding feature:

1. Add a prototype to prototypes.py, such as  
`TREE = {
    "key" : "Tree", 
    "typeclass" : "typeclasses.object", 
    "desc" : "A simple tree"
}
`
2. evennia stop/start to reload server.
3. In game, do @spawn TREE

#### Error output / Expected result of feature
After spawning tree, observe that there is no object spawned in current location.
@examine <tree's dbref> shows location to be "None.

The result previously was the location was caller's location.

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:

This is Evennia master branch , at commit 080c2a3b9f71a880811526b9884095013f24828b

As FYI - I think I found the root cause of the issue, it looks like potentially two issues:

1) in "evennia/commands/default/building.py", in CmdSpawn towards end of function:

`if "noloc" in self.switches and not "location" not in prototype:`

I think should be:

`if "noloc" not in self.switches and "location" not in prototype:`

2) In evennia/utils/spawner.py

```
def _handle_dbref(inp):
    dbid_to_obj(inp, ObjectDB)
```

should be 

```
def _handle_dbref(inp):
    return dbid_to_obj(inp, ObjectDB)
```

scyfris:
I made a fix patch on a forked copy, I'm not sure if it's correct, but for reference I put it here: https://github.com/scyfris/evennia/commit/635200ca176caf28981a49e391beb1ef899c1979

It looks like both of these spots were changed recently, so perhaps it is a regression..but not sure.

Griatch:
Fixed with the merger of the PR. Thanks!

PR #1471, opened 2017-10-06
Add a @spawn unit test

scyfris:
Tests the @spawn command and its variations.  The test is designed to
aslo check the properties of the spawned objects for some common/obvious
settings.

Test also adds a new prototypes file to be used with the Evennia unit
tester for any tests that require named prototypes.  The spawner test
requires testing named prototypes, so it is added in this change.

Mainly tests the @spawn command and its side-effects.

Test should cover my recent fix for issue #1455

I've also added a change to add new prototype testing files.  I'm not sure if this should go along with this change, or separately.  I could definitely split it up into two separate changes - but since the @spawn testing required this named prototypes, I was hoping to check it in together.

This is for issue #1458

### DIFF
--- a/evennia/commands/default/tests.py
+++ b/evennia/commands/default/tests.py
@@ -24,6 +24,8 @@ from evennia.commands.default import help, general, system, admin, account, buil
 from evennia.commands.command import Command, InterruptCommand
 from evennia.utils import ansi, utils
 from evennia.server.sessionhandler import SESSIONS
+from evennia import search_object
+from evennia import DefaultObject, DefaultCharacter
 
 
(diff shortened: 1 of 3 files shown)


Griatch:
@scyfris This looks good, I think. I was surprised we didn't actually have a spawner unit test already - an omission you have now corrected! 

(Coincidentally I think it's time to move the tests of utils into its own sub folder so as to not have too many separate test modules in `utils`. This means a bit of refactoring though, so I'll handle that one after having merged this. :)  )

scyfris:
Thanks Griatch!

PR #1474, opened 2017-10-09
Expansion of 'turnbattle' system contrib: tb_equip

FlutterSprite:
Adds a new contrib module: tb_equip, an implementation of the turnbattle contrib that includes a basic, expandable system for handling weapons and armor.

Adds a new version of the 'turnbattle' contrib module that handles weapons (including variable accuracy and damage) and armor (with variable damage reduction and hit chance). The module also includes a few prototypes for easy testing. In addition, the old 'turnbattle' module has been moved to a new subfolder, and renamed 'tb_basic'.

Equipment is a near-universal feature of MUDs featuring combat systems, so a good framework for using weapons and armor can see a lot of use. It also serves as a good example of how the 'turnbattle' module can be expanded, and therefore act as a resource for newer developers.

I'm having some trouble getting new unit tests implemented, as it seems to really dislike tb_basic and tb_equip sharing some typeclass names. Any tips on how to resolve this would be appreciated!


Griatch:
If you look at the Typeclass wiki, you'll dibs that this is one of the few differences between typeclasses and normal Python classes: they cannot share the same name with any other loaded typeclass. This is for example why we call the vanilla character in the Evennia repo `DefaultCharacter` and not just `Character` - it's so you can have the `Character` in your game dir.

FlutterSprite:
So is the solution to differentiate the names of the typeclasses in each version of the module? That shouldn't be too much trouble, as they're not super typeclass-heavy - I could just turn 'BattleCharacter' to 'TBBasicCharacter' & 'TBEquipCharacter' or something of the like

Griatch:
Yes, that sounds like a good way to do it. A typeclass' name must be unique since it represents a possible database connection - if the same name could represent the same thing in an inheritance tree there would be potential problems with referencing, which is why django forbids this.

FlutterSprite:
There we go! I might have to cover all the combat functions in tb_equip too, just for the sake of coverage...

Griatch:
@FlutterSprite Very nice! Well documented and with plenty of unit tests! What we like to see!

PR #1476, opened 2017-10-10
resolve #1448, fix 'no help entry' not popping up in webclient

### LINKED ISSUE #1448: Webclient Options popup use causes traceback pre-login
Attempting to use the webclient options pre-login causes a traceback.
Settings made pre-login are not preserved after login.
 
 
1.  Open browser to web client
2.  Attemp to use the webclient options before you log in
3.  Observe traceback in console
 
```
   File "/home/evennia/evennia/server/sessionhandler.py", line 762, in call_inputfuncs
     _INPUT_FUNCS[cname](session, *cmdargs, **cmdkwargs)
   File "/home/evennia/evennia/server/inputfuncs.py", line 440, in webclient_options
     storedoptions = account.db._saved_webclient_options or {}
 AttributeError: 'NoneType' object has no attribute 'db'
```
 
### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
 
0.7.0 (rev b3c6e9d) master
 
Pop up the "You need to be connected" modal with message changed to "You need to be logged in", or hide settings cog until after login, or do not try to save settings to account until after login.

RyanStein:
This change resolves #1448 by hiding the webclient options popup before log-in and after disconnection, in addition it will close the options dialog in the event of a disconnect, to keep things consistent. For optimal UI design, there is no reason to show the button to the user when it provides no functionality.

Included in this commit are two related bug fixes. The first fixes new accounts not having any saved webclient options despite the default settings. This is because nothing is saved to `account.db._saved_webclient_options` until the user makes a change on the other end. This explains why a server might have `helppopup` be a default setting, but it has no effect on the user's end until they toggle the checkmark off, then back on.

The second fix tags "no help entry found" messages from the built-in `help` command as help messages, which gets them popped up into the webclient as any other help file would.

### DIFF
--- a/evennia/commands/default/help.py
+++ b/evennia/commands/default/help.py
@@ -267,7 +267,7 @@ class CmdHelp(Command):
             return
 
         # no exact matches found. Just give suggestions.
-        self.msg(self.format_help_entry("", "No help entry found for '%s'" % query, None, suggested=suggestions))
+        self.msg((self.format_help_entry("", "No help entry found for '%s'" % query, None, suggested=suggestions), {"type": "help"}))
 
 
 def _loadhelp(caller):
--- a/evennia/server/inputfuncs.py
+++ b/evennia/server/inputfuncs.py
@@ -436,9 +436,12 @@ def webclient_options(session, *args, **kwargs):
     """
     account = session.account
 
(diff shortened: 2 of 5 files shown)


Griatch:
Looking good to me, straightforward and clean. Merging!

PR #1478, opened 2017-10-15
Expansion of 'turnbattle' system contrib: tb_range

FlutterSprite:
Adds a new contrib module: tb_range, an expansion for the turnbattle contrib which implements a system of movement and positioning.

The tb_range module includes an abstract system of movement and position in combat, which works by tracking different fighters and objects' relative distance to each other without tracking their absolute position. This allows for a simple and intuitive (if imprecise) system for movement in a text-based game that doesn't disadvantage players on screen readers or require navigating around a grid.

I plan to add a few more expansions to the 'turnbattle' contrib folder demonstrating various expansions to the basic combat systems, and then a 'complete' version which implements all of the subsystems of the various modules and implements interactions between them (like melee and ranged weapons for the equipment system in tb_equip, for example). Ultimately, the goal is to have a fully-featured combat system ready for aspiring MUD creators who want to get their game off the ground right away, and to give many options to let them use as much or as little of it as they like.

This system, like the 'turnbattle' contrib itself, is based around its implementation in my learning project "The World of Cool Battles", though it has been streamlined in some places (fewer possible range values and lack of multi-step movement) and expanded in others (the tracking of distances to objects and exits).


FlutterSprite:
The unit tests aren't in yet, but I'd like to know what everyone thinks! This contrib's a little more complex than my others.

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+    else:
+        damage_value = get_damage(attacker, defender)  # Calculate damage value.
+        # Announce damage dealt and apply damage.
+        attacker.location.msg_contents("%s hits %s with a %s attack for %i damage!" % (attacker, defender, attack_type, damage_value))
+        apply_damage(defender, damage_value)
+        # If defender HP is reduced to 0 or less, announce defeat.
+        if defender.db.hp <= 0:
+            attacker.location.msg_contents("%s has been defeated!" % defender)
+            
+def distance_dec(mover, target):
```
Griatch: From how this is used in-code, `distance_dec` and `distance_inc` are helper functions to `approach` and `withdraw`, the latter being the functions intended to be used externally. If this is the case you should clarify which functions are part of your API and which are internal-use only. This can be done either by marking these helper-functions as private (`_distance_dec/inc`) or by embedding them inside their main functions.

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+    """
+    mover.db.Combat_Range[target] -= 1
+    target.db.Combat_Range[mover] = mover.db.Combat_Range[target]
+    # If this brings mover to range 0 (Engaged):
+    if mover.db.Combat_Range[target] <= 0:
+        # Reset range to each other to 0 and copy target's ranges to mover.
+        target.db.Combat_Range[mover] = 0
+        mover.db.Combat_Range = target.db.Combat_Range
+        # Assure everything else has the same distance from the mover and target, now that they're together
+        for object in mover.location.contents:
```
Griatch: `object` is a reserved Python word. Never use a reserved word as a variable name even if you get no error from it here  - not only is it a potential source of insidious bugs, it's also against PEP8 and Python praxis and will also trigger all sorts of warnings in external style checkers down the line. Use something like `obj` instead.

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+        if thing != mover and thing != target:
+            # Move closer to each object closer to the target than you.
+            if mover.db.Combat_Range[thing] > target.db.Combat_Range[thing]:
+                distance_dec(mover, thing)
+            # Move further from each object that's further from you than from the target.
+            if mover.db.Combat_Range[thing] < target.db.Combat_Range[thing]:
+                distance_inc(mover, thing)
+    # Lastly, move closer to your target.
+    distance_dec(mover, target)
+    return
```
Griatch: No need for the empty `return` here; it's implicit.

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+                distance_inc(mover, thing)
+            # Move away from anything your target is engaged with
+            if target.db.Combat_Range[thing] == 0:
+                distance_inc(mover, thing)
+            # Move away from anything you're engaged with.
+            if mover.db.Combat_Range[thing] == 0:
+                distance_inc(mover, thing)
+    # Then, move away from your target.
+    distance_inc(mover, target)
+    return
```
Griatch: Empty `return` has no function here.

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+    
+    Args:
+        obj1 (obj): First object
+        obj2 (obj): Second object
+        
+    Returns:
+        range (int or None): Distance between two objects or None if not applicable
+    """
+    # Return None if not applicable.
+    if not obj1.db.Combat_Range:
```
Griatch: Out of curiosity, there is no validation of `db.Combat_Range` in `distance_inc/dec` nor in `approach/withdraw`. If there is a risk of using objects without such an Attribute the check should probably either be extended to those too, or removed here for consistency (that is, if you use objects not valid for combat you should always expect a traceback).

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+    """
+    Cleans up all the temporary combat-related attributes on a character.
+
+    Args:
+        character (obj): Character to have their combat attributes removed
+
+    Notes:
+        Any attribute whose key begins with 'combat_' is temporary and no
+        longer needed once a fight ends.
+    """
```
Griatch: It's an interesting take to use the combattants themselves to store range combat data rather than to track things on the combat handler. :)

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+    Returns true if the given character is in combat.
+
+    Args:
+        character (obj): Character to determine if is in combat or not
+
+    Returns:
+        (bool): True if in combat or False if not in combat
+    """
+    if character.db.Combat_TurnHandler:
```
Griatch: This can simply and more succinctly be expressed as `return bool(character.db.Combat_TurnHandler)`.

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+    Args:
+        character (obj): Character to determine if it is their turn or not
+
+    Returns:
+        (bool): True if it is their turn or False otherwise
+    """
+    turnhandler = character.db.Combat_TurnHandler
+    currentchar = turnhandler.db.fighters[turnhandler.db.turn]
+    if character == currentchar:
+        return True
```
Griatch: See above.

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+        Notes:
+            If this method returns False/None, the dropping is cancelled
+            before it is even started.
+
+        """
+        # Can't drop something if in combat and it's not your turn
+        if is_in_combat(dropper) and not is_turn(dropper):
+            dropper.msg("You can only drop things on your turn!")
+            return False
+        return True
```
Griatch: Minor PEP8 nag: Put one empty line between all methods in a class.

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+
+
+"""
+----------------------------------------------------------------------------
+SCRIPTS START HERE
+----------------------------------------------------------------------------
+"""
+
+
+class TBRangeTurnHandler(DefaultScript):
```
Griatch: I am a little divided on placing this important component at the end of the module. The handler is used and referenced  throughout the other components, so when reading the code one wonders what it is and where it comes from. For this reason I always push for a module to be ordered so that things are defined above the code that use them. Hence I feel this should be defined higher up, or even at the top.

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+            self.stop()  # Stop this script and end combat.
+            return
+
+        # Cycle to the next turn.
+        currentchar = self.db.fighters[self.db.turn]
+        self.db.turn += 1  # Go to the next in the turn order.
+        if self.db.turn > len(self.db.fighters) - 1:
+            self.db.turn = 0  # Go back to the first in the turn order once you reach the end.
+        newchar = self.db.fighters[self.db.turn]  # Note the new character
+        self.db.timer = 30 + self.time_until_next_repeat()  # Reset the timer.
```
Griatch: Timeouts like these seem very game-specific to hide away inside the Script body like this. Might it not be an idea to have a block of global constants up top, like `TURN_TIMEOUT = 30` for things like this? Would also make it easier for the dev to see which knobs are easy to tweak ...

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+    # If the attack value is lower than the defense value, miss. Otherwise, hit.
+    if attack_value < defense_value:
+        attacker.location.msg_contents("%s's %s attack misses %s!" % (attacker, attack_type, defender))
+    else:
+        damage_value = get_damage(attacker, defender)  # Calculate damage value.
+        # Announce damage dealt and apply damage.
+        attacker.location.msg_contents("%s hits %s with a %s attack for %i damage!" % (attacker, defender, attack_type, damage_value))
+        apply_damage(defender, damage_value)
+        # If defender HP is reduced to 0 or less, announce defeat.
+        if defender.db.hp <= 0:
```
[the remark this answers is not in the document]
FlutterSprite: This would be a good thing to separate out - I'll have to apply this change across all the 'turnbattle' modules.

evennia/contrib/turnbattle/tb_range.py
FlutterSprite: Ah! This was something I knew nothing about - I'll definitely address this. There's not really anywhere else I would think to use distance_dec and distance_inc by themselves.

evennia/contrib/turnbattle/tb_range.py
```diff
@@ -0,0 +1,1356 @@
...
+            
+def distance_dec(mover, target):
+    """
+    Decreases distance in range field between mover and target.
+    
+    Args:
+        mover (obj): The object moving
+        target (obj): The object to be moved toward
+    """
+    mover.db.Combat_Range[target] -= 1
```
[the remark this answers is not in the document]
FlutterSprite: This would be an easy change! I'll apply this across the rest of the modules too.

evennia/contrib/turnbattle/tb_range.py
FlutterSprite: Whoops!! I'll definitely fix this.

evennia/contrib/turnbattle/tb_range.py
FlutterSprite: I do think this is a good function to have, though it was intuition more than anything rational that lead me to make it instead of referencing db.combat_range directly - I'll add it into those functions too.

evennia/contrib/turnbattle/tb_range.py
FlutterSprite: Since combatants don't have absolute positions, only relative distances to each other, this seemed the most intuitive way to do it!

evennia/contrib/turnbattle/tb_range.py
FlutterSprite: I think it was my thought originally that the turn handler used lots of the other functions, it should come after them, but at some point I moved many of those functions into the turn handler itself. I'll put it higher up here and in the rest of the modules.

evennia/contrib/turnbattle/tb_range.py
FlutterSprite: This is definitely a good idea - actions per turn would be another good variable to be able to easily mess with. I think it was my original thought that adding extra flexibility with variables would complicate things too much for a system that's mostly meant to be completely overhauled, but some things like this would be universal enough that I can't imagine they'd be taken out completely!

FlutterSprite:
There! Still no unit tests, but I addressed all the concerns you raised. This involved going back and making changes in tb_basic and tb_equip as well.

Griatch:
Cool! I'll review again when the unit tests are in then. :)

FlutterSprite:
The unit tests are in! This one was quite a pain.

Griatch:
@FlutterSprite This looks good I think. There is one thing I'm missing here though, when reading the contrib's documentation, and that is how decoupled these various modules are. That is, when I look at the `README.md` I want to know if I can combine equiplent + range, if I need basic to use range ... and so on. So while the actual install instructions are in each file, I think it would be enlightening with some more info about these things, as well as pointing out where the install instructions are (since those are normally in the `README.md` file for other contribs).

FlutterSprite:
Gotcha! I'll clarify things in the readme real quick.

FlutterSprite:
How's that?

Griatch:
Thanks, merged now! One thing to remember for the next time - for your own sanity of mind, don't develop things like this in your own copy of the upstream develop (or master) branch. It's better that you keep your version of develop in sync with upstream and do your development in a new branch off that (this you can then just delete once your contributions is available upstream). Otherwise, if I should for example rebase or squash your PR's commits, your version of devel/master branch will be different from the upstream one and you'll face merge conflicts.

FlutterSprite:
Oh, that makes a lot of sense! Slowly but surely, I'm learning how to use GitHub properly. This is the first time I've actually committed my changes from the command line instead of pasting them into the website like a scrub.

Issue #1479, opened 2017-10-15
Issue with adding empty columns to EvTable and subsequent add_row


scyfris:
#### Brief summary of issue / Description of requested feature:

Calling "add_column" with a column without any specified data, only a header, to an already-created EvTable seems to cause an issue with the newly added column where it is sized differently than the rest of the table.  This leads to issues with subsequent "add_row()" commands.  See below for example.

#### Steps to reproduce the issue / Reasons for adding feature:

1. Create a new EvTable.  Example:

`        table2 = EvTable("col1", "col2", "col3", border = "cells")
`

This creates a table with only headers:

![image](https://user-images.githubusercontent.com/441039/31589030-4c299cb2-b1af-11e7-9f78-86f6e6210c84.png)

2. Call "add_column()" on the table, without specifying any data.  The expectation is that it will add an empty column...but it seems to instead append a row to the column:

`        table2.add_column(header="col4")
`

![image](https://user-images.githubusercontent.com/441039/31589032-569fab46-b1af-11e7-9d2c-4d7bca078aaa.png)

3. Next call "add_row" on the EvTable to add a row, filling in data for all columns.  This adds the first 3 columns correctly (the columns that were added with the EvTable construction), but the new column is offset by one:

`        table2.add_row("I", "am", "a", "test")
`

![image](https://user-images.githubusercontent.com/441039/31589034-5afc6602-b1af-11e7-90a0-3ed43c88a687.png)

Obviously that looks like an issue.

#### Error output / Expected result of feature

Error results:  The results of this sequence of commands:

```
        table2 = EvTable("col1", "col2", "col3", border = "cells")
        table2.add_column(header="col4")
        table2.add_row("I", "am", "a", "test")
```
leads to this output:
![image](https://user-images.githubusercontent.com/441039/31589073-e548c65c-b1af-11e7-9708-75acfd07790d.png)

Expected results:  I would expect the above sequence of commands to lead to this output:

![image](https://user-images.githubusercontent.com/441039/31589081-fb22cfae-b1af-11e7-8779-641d55f52073.png)

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:

Note this is from "master" branch at ToT as of commit 1f54d464a9293a067ffb3dc6bd1ba967c5d4ab8d

I have root-caused the issue and should have a pending fix shortly.  In summary, it looks like the code in evennia/utils/evtable.py, for the function "add_column" might be incorrect:

```
[...]

PR #1480, opened 2017-10-15
Fix an issue with adding an empty column to an EvTable

scyfris:
There seems to be an issue with adding an empty column (i.e. a column
with empty data) to an EvTable that has already been set up.  It seems
that the column with empty data is added with one extra row than the
rest of the table, and when a new call to add_rows() is made to EvTable,
the data in the column that was added previously is offset by 1.

This change fixes that by calculating the size of the new column to take
into account the headers BEFORE making the calculation on whether to
expand the column to match the table's size or not.

This change just moves calculation of new column size until after the header has been taken into account.

Fixes issue #1479

### DIFF
--- a/evennia/utils/evtable.py
+++ b/evennia/utils/evtable.py
@@ -1421,8 +1421,21 @@ class EvTable(object):
         column = EvColumn(*args, **options)
         wtable = self.ncols
         htable = self.nrows
-        excess = len(column) - htable
 
+        header = kwargs.get("header", None)
+        if header:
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks for the fix, I'll run some tests with this when I get a spare moment.  :)

scyfris:
Thanks!  Let me know if any issues pop up

Griatch:
@scyfris Looking good to me! Merged. :)

PR #1485, opened 2017-10-21
Add newline after multimatch string error message

scyfris:
Other utility functions, such as caller.msg(), don't require adding
newline because newlines are added implicitly.  This change modifies the
multimatch string parameter to also not require newline, but add one
implicitly.

Modifies the utils "at_search_result()" function to append a newline to "multimatch_string" if it is part of the kwargs.

Performing a command on object when there is ambiguity on which object to select, outputs a message  indicating there are multiple such objects. The problem is the formatting is a bit off, where the first object is on the same line as the error message.

For example, if I have two "swords" in my inventory, and I do "drop sword", I get this error message:

image

It seems like the "multimatch_error" string doesn't append a newline.  I don't know if this is on purpose, but I found this odd since other functions such as "msg()" automatically append newlines to the string.

After my fix, this is the expected output

image

If intention is not to have a newline in multimatch_string, then please ignore this commit.

### DIFF
--- a/evennia/utils/utils.py
+++ b/evennia/utils/utils.py
@@ -1786,8 +1786,12 @@ def at_search_result(matches, caller, query="", quiet=False, **kwargs):
         error = kwargs.get("nofound_string") or _("Could not find '%s'." % query)
         matches = None
     elif len(matches) > 1:
-        error = kwargs.get("multimatch_string") or \
-            _("More than one match for '%s' (please narrow target):\n" % query)
+        multimatch_string = kwargs.get("multimatch_string")
+        if multimatch_string:
+            error = "%s\n" % multimatch_string
+        else:
+            error = _("More than one match for '%s' (please narrow target):\n" % query)
+
         for num, result in enumerate(matches):
             # we need to consider Commands, where .aliases is a list
             aliases = result.aliases.all() if hasattr(result.aliases, "all") else result.aliases


Griatch:
The lack of a line break is intentional. The idea is that for a pluggable string like this, we want to impose as little hard-coded formatting on the user as we can. So if you want a line-break you add that; Maybe you want to attach some unicode icon on the same line - you can do that too. 

... That said, the multimatch handler is completely replaceable these days. So you could modify the entirety of the string output if you wanted to. So formatting is less of a concern and I can agree that one might expect a line break here. I will merge this into develop branch for now.

Griatch:
Merged into develop!  Closing.

scyfris:
ahh okay, gotcha.  Having the fully replaceable multimatch string is pretty handy.  I wasn't aware of that, but was using the default evennia-provided one, which was throwing first item on same line.

Thanks for the merge.  If it ends up causing issues down the road or you feel old behavior is in fact more desirable, then we can look at reverting if needed.  The only place I could think it would cause issues is if existing unit tests in projects are already written to check output expected first item to be on same line (i.e. testing some command which uses the default Evennia-provided multimatch string, then parsing output message).  But maybe not a big deal.  Doesn't look like Evennia itself has this issue, but maybe some user projects.

PR #1490, opened 2017-10-25
Unit test for server/deprecations

TehomCD:
Yay Hacktoberfest! I'm trying to get more in the habit of writing unit tests, so I'll just be trying to write some for modules that need coverage.

Provide test coverage for the deprecation warnings/errors module.

Part of #1458

### DIFF
--- a/evennia/server/tests.py
+++ b/evennia/server/tests.py
@@ -45,6 +45,15 @@ class EvenniaTestSuiteRunner(DiscoverRunner):
         return super(EvenniaTestSuiteRunner, self).build_suite(test_labels, extra_tests=extra_tests, **kwargs)
 
 
+class MockSettings(object):
+    def __init__(self, setting, value=True):
+        self.setting = value
+        if setting == "WEBSERVER_PORTS":
+            self.WEBSERVER_ENABLED = True
+        else:
+            self.WEBSERVER_ENABLED = False
+
+
 class TestDeprecations(TestCase):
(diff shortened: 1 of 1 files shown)


evennia/server/tests.py
```diff
@@ -46,3 +43,37 @@ def build_suite(self, test_labels, extra_tests=None, **kwargs):
         import evennia
         evennia._init()
         return super(EvenniaTestSuiteRunner, self).build_suite(test_labels, extra_tests=extra_tests, **kwargs)
+
+
+class MockSettings(object):
```
Griatch: This works, but I've recently learned there are better ways to mock the django settings. Actually the Django TestCase class contains a context manager for this. So you can do in your test:

```python
class TestIdle(django.tests.TestCase):
    def test_idle(self):
        with self.modify_settings(IDLE_TIMOUT=1000):
            # stuff testing a different idle timout
       # here the original value is back again
```

Good stuff to remember for the future. :)

Griatch:
@TehomCD Looks good, but unit tests should always be done against Master Branch. So I'll have to cherry pick these, meaning this will never look like it merges.

Griatch:
@TehomCD I squashed your commits and cherry-picked them into master. Thanks for the contribution!

evennia/server/tests.py
TehomCD: Ooh, good to know, thanks!

PR #1492, opened 2017-10-26
Add tests for general_context

TehomCD:
Another small bit of unit testing, this one for the general_context for the webpage. I struggled a bit with the testing of the global variables that were assigned - while I could mock them, this wouldn't cover testing their actual assignment. So I decided to move them into functions that would be immediately called in the module so that those could have their own separate test methods. If you like the approach, it might be something I'd do when writing unit tests for other modules that use global variables.

Add more test coverage.

Yet another #1458

### DIFF
--- a/evennia/web/utils/general_context.py
+++ b/evennia/web/utils/general_context.py
@@ -10,17 +10,18 @@ from django.conf import settings
 from evennia.utils.utils import get_evennia_version
 
 # Determine the site name and server version
-
-try:
-    GAME_NAME = settings.SERVERNAME.strip()
-except AttributeError:
-    GAME_NAME = "Evennia"
-SERVER_VERSION = get_evennia_version()
-try:
-    GAME_SLOGAN = settings.GAME_SLOGAN.strip()
-except AttributeError:
-    GAME_SLOGAN = SERVER_VERSION
-
+def set_game_name_and_slogan():
+    global GAME_NAME, GAME_SLOGAN, SERVER_VERSION
+    try:
+        GAME_NAME = settings.SERVERNAME.strip()
+    except AttributeError:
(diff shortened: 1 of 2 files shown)


evennia/web/utils/general_context.py
```diff
@@ -10,17 +10,18 @@
...
-    GAME_NAME = settings.SERVERNAME.strip()
-except AttributeError:
-    GAME_NAME = "Evennia"
-SERVER_VERSION = get_evennia_version()
-try:
-    GAME_SLOGAN = settings.GAME_SLOGAN.strip()
-except AttributeError:
-    GAME_SLOGAN = SERVER_VERSION
-
+def set_game_name_and_slogan():
```
Griatch: I'm fine with wrapping things like this, but this is now a function (outside of the test suite) that needs a proper (albeit short) doc string as per the Evennia style guide. It might also be a good idea to add a `Notes:` section to this to point out that using a function here has a purpose for testing.

evennia/web/utils/general_context.py
```diff
@@ -33,10 +34,13 @@
 
 
 # The main context processor function
-WEBCLIENT_ENABLED = settings.WEBCLIENT_ENABLED
-WEBSOCKET_CLIENT_ENABLED = settings.WEBSOCKET_CLIENT_ENABLED
-WEBSOCKET_PORT = settings.WEBSOCKET_CLIENT_PORT
-WEBSOCKET_URL = settings.WEBSOCKET_CLIENT_URL
+def set_webclient_settings():
```
Griatch: A doc string is needed here too, albeit brief.

Griatch:
Rebased and squashed, then merged into master branch. Thanks!

PR #1494, opened 2017-10-27
Add unit tests for bodyfunctions

TehomCD:
Adds unit test for the tutorial_examples bodyfunction script. I thought it'd be interesting to try to do a unit test for a script. I discovered that if you don't remember to stop the script in the `tearDown` method, `DirtyReactorAggregateError` will be raised - maybe worth adding that to `EvenniaTest.tearDown`?

Moar unit tests.

Another #1458 test.

### DIFF
--- a/evennia/contrib/tutorial_examples/tests.py
+++ b/evennia/contrib/tutorial_examples/tests.py
@@ -18,14 +18,20 @@ class TestBodyFunctions(EvenniaTest):
         self.script.stop()
 
     def test_at_repeat(self, mock_random):
+        """test that no message will be sent when below the 66% threshold"""
         mock_random.random = Mock(return_value=0.5)
         old_func = self.script.send_random_message
         self.script.send_random_message = Mock()
         self.script.at_repeat()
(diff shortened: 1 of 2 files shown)


evennia/contrib/tutorial_examples/tests.py
```diff
@@ -0,0 +1,69 @@
...
+        self.script.send_random_message = Mock()
+        self.script.at_repeat()
+        self.script.send_random_message.assert_not_called()
+        # test that random message will be sent
+        mock_random.random = Mock(return_value=0.7)
+        self.script.at_repeat()
+        self.script.send_random_message.assert_called()
+        self.script.send_random_message = old_func
+
+    def test_send_random_message(self, mock_random):
```
Griatch: Wow, this is a highly detailed test of this return ... I would say (for future reference) that there is no need to test every individual string like this. It's thorough, but no real functionality will be safeguarded if the test fails because one of these strings are rephrased.

Griatch:
Squashed commits and cherry-picked to master. Thanks!

PR #1504, opened 2017-10-30
Upgrade to Python 3

### LINKED ISSUE #1481: Task: Explore Evennia to Python3 conversion
#### Description of requested feature:

This task is for exploring just how far we can get with Evennia + Python3. 

> I'm not sure how suitable this actually is for hacktoberfest (it's unclear how big of an effort it is). It's not suitable for a Python-beginner but most likely you don't need that much Evennia-specific knowledge so it might be interesting to explore.

Currently, Evennia is solely a Python2 program. It's clear that Python3 is the future. What has held us up on this have historically been our dependencies. Django has supported py3 since several versions at this point. The remaining big dependency-block for us is Twisted, and their Py3 conversion is, by their own account, [96% complete](https://twistedmatrix.com/trac/milestone/Python-3.x). And [here is the Evennia project page tracking py3](https://github.com/evennia/evennia/projects/5). It is currently unclear if those remaining issues are blockers for us or not. 

With Django un-supporting Python2 completely in version 2.0 and Python2 itself reaching end-of-upstream-support in 2020, I don't think there is a good reason to spend effort on a Python2 + Python3 hybrid of Evennia. Only reason for going for a hybrid would be if it turns out this is somehow easier or helps the process. 

#### Process:

In a separate branch, try to run Evennia on python3, to see just how far one can get with the current state of third-party libraries.

1. Install python3
2. Set up virtualenv with py3: `virtualenv -p /path/to/python3/executable pyenv`
3. Install Evennia into this virtualenv - py3 versions of third-party packages should be used automatically. 
4. See what works and what not, and fix if possible.

#### Expected result of task:

Ideally, an Evennia running under Python3. Otherwise a report on what is missing in the third-party libraries for that to be possible (we won't patch third-party packages). 

This should be done against `develop` branch. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/50346884-task-explore-evennia-to-python3-conversion?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github).
</bountysource-plugin>

RyanStein:
It's difficult to summarize considering the scope, and I tried to make the commits as sparse and documented as possible, but a quick overview:

- ran 2to3
- converted various strings to bytes
- handled several checks for `__iter__` with the replaced `is_iter`
- deprecated `to_str` and `to_unicode`
- updated necessary dependecies
- converted WebSocket protocol to use [autobahn-python](https://github.com/crossbario/autobahn-python), had issue with both the modified txWS package and the un-modified one. autobahn-python appeared better maintained.

There's still work to be done on this, but a majority of the difficult stuff has been handled. We're going to need some manner of upgrading old databases to be compatible with Python 3. It would be worth looking into if removing all calls to `to_str` and `to_unicode` can be done easily enough, too, as well as double-checking if any remaining `hasattr(..., '__iter__')` calls will be problematic.

Lots of testing is due. I did a few things in addition to what's available in the test suite, which is failing on at most 3 tests, out of a total of 235. Specifically, the `EvForm` test is failing due to an issue outlined in `ANSITextWrapper` but is otherwise partially functional. `TestRPSystem` has two failures, which I'll leave to @Griatch to offer feedback. The tests expect raw regex pattern output, which was a little tough for me to parse on my own. 😄 

Regarding idioms, there's probably a good amount of space to make it more idiomatic. 2to3 converts the calls to `.items()`, `.keys()`, and `.values()` to list-enwrapped versions, most of which can probably be reverted back to the plain function calls. `super(..., self)` for a method calling its only parent class can be safely converted to `super()`, if I'm not mistaken. Python 3 f-strings are pretty cool, too. We could possibly replace usage of `%` and `string.format` with them, and it might result in cleaner code. I'm sure there's much room for improvement.

Somewhat related: We'll need to keep an eye out for what's necessary for Django 2 as well once it's released, as this doesn't do anything for supporting that specifically just yet other than running on Python 3.

I also had to upgrade Twisted to latest in order to get access to `twisted.conch.telnet`, which wasn't available under `== 16.0.0` in Python 3 for some reason, which I understand was changed in order to fix Docker support. I'm not familiar with Docker or the impacts thereof, but I'm noting it just in case.

Python 3 is the future.

Resolves #1481 

Feedback is appreciated, and questions are welcome.


Griatch:
Thanks a lot for your work on this! This is a substantial body of work and I appreciate your efforts in cleaning it up the commits as much as you have. We'll need to do some more polishing here, as you point out. Some points I note, from glancing at the code and your comments:

- most of the cases where `2to3` converted iterators to lists the `list()` wrapper can be removed; there may be old code or contrib code where a list is actually expected, but those should be rare. Definitely the last many years I've exclusively used things like `dct.values()` only for iterating over. 
- `super()` can be used exclusively throughout. We don't use multiple inheritance super anywhere that I know of. 
- Not sure that `to_str` and `to_unicode` should go away, but they are definitely not as needed anymore. Those are conversion helpers: They will try to convert unicode<->binary according to a user-set encoding and otherwise step through a range of encodings defined in settings (inclusive UTF-8). I suppose if correctly used they should only be needed at the protocol send/receive level now? Would need to look more at that.
- I saw a few cases where the original `input` was turned into `eval(input(...)` which is a correct conversion but more likely the original was wrong and should have been `raw_input` from the onset. :) 
- The use of f-format is something I've so far never used, it's something to consider but probably not as a first exercise. :) 

I'll look this over more over time, good job!

Griatch:
@RyanStein Again, thanks for your hard work on this! Your further modifications have adressed most of the immediate concerns. To ease collaboration I have now created a `develop-py3` branch based on this PR. I handled the immediate merge issue and can see the server starts but have not done more with the tests at the moment. So closing this PR; please direct further updates/PRs towards the `develop-py3` branch. :)

PR #1505, opened 2017-10-30
adding default commands tests

moonpatroller:
#### Adding more tests for commands/default, for https://github.com/evennia/evennia/issues/1458

### DIFF
--- a/evennia/commands/default/tests.py
+++ b/evennia/commands/default/tests.py
@@ -239,9 +239,113 @@ class TestBuilding(CommandTest):
 
     def test_set_obj_alias(self):
         self.call(building.CmdSetObjAlias(), "Obj = TestObj1b", "Alias(es) for 'Obj(#4)' set to testobj1b.")
+        self.call(building.CmdSetObjAlias(), "", "Usage: @alias <obj> [= [alias[,alias ...]]]")
+
+    def test_set_obj_alias_caller_finds_nothing(self):
+        alias = building.CmdSetObjAlias()
+        alias.caller = Mock()
+        alias.caller.search = Mock(return_value=None)
+        alias.caller.msg = Mock()
+        alias.lhs = 'dummy lhs'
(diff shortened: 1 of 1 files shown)


evennia/commands/default/tests.py
```diff
@@ -239,10 +239,114 @@ def test_examine(self):
...
+        from_obj.scripts.all = Mock(return_value=[])
+
+        cmd_copy.caller = Mock()
+        cmd_copy.caller.search = Mock(return_value=from_obj)
+
+        cmd_copy.args = 'dummy args'
+        cmd_copy.caller.msg = Mock(return_value=None)
+
+        cmd_copy.func()
+        print cmd_copy.caller.msg.call_args[0][0]
```
Griatch: Remnant debug statement?  Also, there doesn't appear to be any asserts in this method.

evennia/commands/default/tests.py
```diff
@@ -239,10 +239,114 @@ def test_examine(self):
 
     def test_set_obj_alias(self):
         self.call(building.CmdSetObjAlias(), "Obj = TestObj1b", "Alias(es) for 'Obj(#4)' set to testobj1b.")
+        self.call(building.CmdSetObjAlias(), "", "Usage: @alias <obj> [= [alias[,alias ...]]]")
+
+    def test_set_obj_alias_caller_finds_nothing(self):
```
Griatch: Why not use the existing CmdTest mechanisms for these? It already mocks the msg method and allows you to compare if what is returned, so adding mocks to check for additional changes would be easier? At any rate, it appears to be a lot of boiler plate here - these tests could go into a new test class settings things up in its `setUp` method.

Griatch:
@moonpatroller Bump?

Griatch:
I'm closing this as invalid, since it doesn't appear to be an interest to correct it.

PR #1508, opened 2017-10-30
Add simple menu tree selection contrib

FlutterSprite:
This contrib module allows developers to generate an EvMenu instance with options sourced from a multi-line string, which supports categories, back and forth menu navigation, option descriptions, and passing selections to custom callbacks. This allows for easier dynamic menus and much faster deployment of simple menu trees which does not require the manual definition of menu nodes and option dictionary-lists.

Adds a new contrib system for easily generating branching menus sourced from strings.

EvMenu is great - it's powerful and flexible - but for simple menu selection tasks, it can be sort of like using a VTOL jet to go to the grocery store. This contrib allows for a much simpler way of using EvMenus for applications where its more powerful features aren't needed.

I workshopped this one in the IRC a lot - it takes advantage of a lot of recent changes to EvMenu in the develop branch to dynamically generate its nodes on the fly.

### DIFF
--- a/evennia/contrib/tree_select.py
+++ b/evennia/contrib/tree_select.py
@@ -52,7 +52,7 @@ Categories can be nested in other categories as well - just go another '-' deepe
 can do this as many times as you like. There's no hard limit to the number of
 categories you can go down.
 
-For example, let's add some more options to our menu, turning 'Foo' into a category.
+For example, let's add some more options to our menu, turning 'Bar' into a category.
 
     TEST_MENU = '''Foo
     Bar
@@ -63,7 +63,7 @@ For example, let's add some more options to our menu, turning 'Foo' into a categ
     Baz
     Qux'''
     
-Now when we call the menu, we can see that 'Foo' has become a category instead of a
+Now when we call the menu, we can see that 'Bar' has become a category instead of a
 selectable option.
 
     _______________________________
--- a/evennia/contrib/tree_select.py
+++ b/evennia/contrib/tree_select.py
@@ -131,7 +131,7 @@ Now we see that the Baz option has a description attached that's separate from i
 Once the player makes a selection - let's say, 'Foo' - the menu will terminate and call
 your specified callback with the selection, like so:
 
-    callback(TEST_MENU, caller, 0, "Foo")
+    callback(caller, TEST_MENU, 0, "Foo")
     
 The index of the selection is given along with a string containing the selection's key.
 That way, if you have two selections in the menu with the same key, you can still
@@ -171,8 +171,8 @@ def init_tree_selection(treestr, caller, callback,
         treestr (str): Multi-lne string representing menu options
         caller (obj): Player to initialize the menu for
         callback (callable): Function to run when a selection is made. Must take 4 args:
-            treestr (str): Menu tree string given above
             caller (obj): Caller given above
+            treestr (str): Menu tree string given above
             index (int): Index of final selection
             selection (str): Key of final selection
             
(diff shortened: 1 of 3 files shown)


FlutterSprite:
Unit tests aren't in yet, but here's my simple menu tree system, all cleaned up and very thoroughly documented!

FlutterSprite:
There, that's that! Unit tests are in!

FlutterSprite:
Hi! Don't mean to pester, but this has been ready to go for a little while now!

Griatch:
@FlutterSprite Sorry, been busy. Will try to review this soon.

FlutterSprite:
No worries! Take your time!

evennia/contrib/tree_select.py
```diff
@@ -0,0 +1,522 @@
...
+
+    Make your selection:
+    _______________________________________________________________
+    
+    Foo                                      
+    Bar [+]                                  
+    Baz: Look at this one: the best option.  
+    Qux                                      
+    
+And that's all there is to it! For simple branching-tree selections, using this system is
```
Griatch: There is one thing missing from this (otherwise excellent) example/docstring - and that is an example of the callback and how it is called with the result of the menu. It's mentioned at the top, but it might be useful to put a few lines about that here when everything else is described so explicitly.

FlutterSprite: Done!

evennia/contrib/tree_select.py
```diff
@@ -0,0 +1,531 @@
...
+    
+    Foo                                      
+    Bar [+]                                  
+    Baz: Look at this one: the best option.  
+    Qux                                      
+
+Once the player makes a selection - let's say, 'Foo' - the menu will terminate and call
+your specified callback with the selection, like so:
+
+    callback(TEST_MENU, caller, 0, "Foo")
```
Griatch: This appears incorrect. The call seems to be defined as `callback(caller, treestr, index, selection)`

evennia/contrib/tree_select.py
```diff
@@ -0,0 +1,531 @@
...
+    callback = caller.ndb._menutree.callback
+    start_text = caller.ndb._menutree.start_text
+    
+    # List of options if index is 'None' or category, or 'True' if a selection
+    optlist = parse_opts(treestr, category_index=index)
+    
+    # If given index returns optlist as 'True', it's a selection. Pass to callback and end the menu.
+    if optlist == True:
+        selection = index_to_selection(treestr, index)
+        callback(caller, treestr, index, selection)
```
Griatch: It might be worth to catch errors from this callback and use `log_trace` from the `utils.logger` module to log the traceback while giving a sanitized error to the caller.

FlutterSprite: I've never done anything with logging before - is there a good example of something like this implemented elsewhere that I can use to reference?

Griatch: Yes, see for example https://github.com/evennia/evennia/search?p=2&q=log_trace&type=&utf8=%E2%9C%93

Griatch: See also https://github.com/evennia/evennia/blob/master/evennia/utils/logger.py#L62

FlutterSprite:
There! I added in the logger, and verified that it worked by changing the callback to take an extra argument so that it would cause an error - and it caught the error just fine!

```2017-11-12T11:57:49-0800 [AMPProtocol,0,127.0.0.1] [::] TypeError: change_name_color() takes exactly 5 arguments (4 give
n)
2017-11-12T11:57:49-0800 [AMPProtocol,0,127.0.0.1] [EE] Error in tree selection callback.```

PR #1520, opened 2017-11-18
Merge tb_items with tb_magic

FlutterSprite:
Since tb_items is almost ready to go as well, I figure I could combine these into the same PR! More detail to come.


FlutterSprite:
Whoops, messed up! Ignore this!

PR #1528, opened 2017-12-12
Add a setting to change telnet default encoding

vlegoff:

This adds a new setting in the default settings.  One can change Telnet default encoding to something other than utf-8.

There's a little issue though (see below).

Some countries are used to a certain encoding, and that's not necessarily utf-8.  Here in France Windows often comes with a latin-1 encoded system and natively supports it, while utf-8 may not be as widely supported.  Changing the default encoding, only for telnet, while allowing full customization per user, would be good in principle.

#### Current issue

As it is, the code is pretty straightforward.  However, if you use `@options/save` to change your encoding and save it in the account, then the encoding is dropped the next time you connect or reload the server.  I am not positive why this happens.  The overridden flag still appears in the account attributes, but it's not applied for some reason.

* Evennia develop

### DIFF
--- a/evennia/server/portal/telnet.py
+++ b/evennia/server/portal/telnet.py
@@ -50,7 +50,7 @@ class TelnetProtocol(Telnet, StatefulTelnetProtocol, Session):
         self.handshakes = 8  # suppress-go-ahead, naws, ttype, mccp, mssp, msdp, gmcp, mxp
         self.init_session(self.protocol_name, client_address, self.factory.sessionhandler)
         # change encoding to ENCODINGS[0] which reflects Telnet default encoding
-        self.protocol_flags["ENCODING"] = settings.ENCODINGS[0]
+        self.protocol_flags["ENCODING"] = settings.ENCODINGS[0] if settings.ENCODINGS else 'utf-8'
 
         # suppress go-ahead
         self.sga = suppress_ga.SuppressGA(self)
--- a/evennia/server/portal/telnet.py
+++ b/evennia/server/portal/telnet.py
@@ -24,7 +24,6 @@ _RE_LEND = re.compile(r"\n$|\r$|\r\n$|\r\x00$|", re.MULTILINE)
 _RE_LINEBREAK = re.compile(r"\n\r|\r\n|\n|\r", re.DOTALL + re.MULTILINE)
 _RE_SCREENREADER_REGEX = re.compile(r"%s" % settings.SCREENREADER_REGEX_STRIP, re.DOTALL + re.MULTILINE)
 _IDLE_COMMAND = settings.IDLE_COMMAND + "\n"
-_TELNET_ENCODING = settings.ENCODINGS[0]
 
 class TelnetProtocol(Telnet, StatefulTelnetProtocol, Session):
     """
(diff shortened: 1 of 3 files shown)


Griatch:
@v-legoff Hm, I don't think introducing a new setting only for one protocol is the way to do this. The `ENCODINGS` setting is already there and should be what's used. `ENCODINGS` is also used by the `to_str`/`to_unicode` utility functions to correctly fall back to another encoding. I think the first value in the `ENCODINGS` list should be the default one in use. That is currently `UTF-8` but should be able to be anything you want, like the encoding you want your server to prefer. I thought it would work like that already but most likely `UTF-8` is hard-coded as a default somewhere it shouldn't be (I've not had time to look into it). So won't accept this PR as it stands. If you want to rework this in the other way we can keep this PR open, otherwise I'll close it and we'll open a new Issue.

vlegoff:
I understand.  I was reluctant to using this strategy:  changing `ENCODINGS` seemed to cause lots of crashes if the first encoding wasn't utf-8 (and in that case, I don't want it to be).  Furthermore, changing the default encoding in such a way might affect other connections, the IRC, the webclient, the website, the database... while I would like for it to remain in utf-8.  That's why I thought adding a setting would be the smartest choice.  I see your point, though, I'm not just sure how to offer as much flexibility.

Griatch:
@vlegoff Internally everything will remain unicode no matter what you do. This will be even more pronounced as we move to py3 - the encoding will only come into play just before sending over the wires. The `ENCODINGS` setting is only really relevant for communicating with terminals-based clients anyway (all browsers support utf-8 so there is no point in using anything else for the webclient). The only thing changing the first value in `ENCODINGS` *should* do is to change the first tested value. You should definitely keep `UTF-8` somehwere in the list.
But if, as you say, things crash if you change the first value, then that's indeed a bug.

vlegoff:
Okay, I've removed `TELNET_ENCODING` and use `ENCODINGS[]0` instead.  I still get a strange bug:

1. Set `ENCODINGS` to a list in your settings, keeping `"utf-8"` but having another (like `"latin-"`) as first choice.
2. Log into an account.
2. Use `@options/save ENCODING = utf-8`.
3. Check that you are in utf-8 (you should be) despite the fact that you have a different default encoding in `ENCODINGS[0]`.
4. Reconnect.  The client encoding is back on `ENCODINGS[0]` even after you have logged into your account with the saved options.

evennia/server/portal/telnet.py
```diff
@@ -49,7 +49,8 @@ def connectionMade(self):
         # this number is counted down for every handshake that completes.
         # when it reaches 0 the portal/server syncs their data
         self.handshakes = 8  # suppress-go-ahead, naws, ttype, mccp, mssp, msdp, gmcp, mxp
-        self.init_session(self.protocol_name, client_address, self.factory.sessionhandler)
+        self.init_session(self.protocol_name, client_address, self.factory.sessionhandler,
+                override_flags={"ENCODING": _TELNET_ENCODING})
```
Griatch: I don't think this `override_flags` structure is really needed? It's added only for the purpose of this one setting and won't be called by any other structure. I think it's fine to use the setting directly where it's used rather than to add this faux call.

evennia/server/session.py
```diff
@@ -87,6 +88,10 @@ def init_session(self, protocol_key, address, sessionhandler):
                                "INPUTDEBUG": False,
                                "RAW": False,
                                "NOCOLOR": False}
+
+        if override_flags:
+            self.protocol_flags.update(override_flags)
```
Griatch: I think it's fine to simply do `self.protocol_flags["ENCODING"] =  _TELNET_ENCODING` here. Maybe add a comment to explain it.

vlegoff:
Done!

evennia/server/portal/telnet.py
```diff
@@ -50,6 +49,8 @@ def connectionMade(self):
         # when it reaches 0 the portal/server syncs their data
         self.handshakes = 8  # suppress-go-ahead, naws, ttype, mccp, mssp, msdp, gmcp, mxp
         self.init_session(self.protocol_name, client_address, self.factory.sessionhandler)
+        # change encoding to ENCODINGS[0] which reflects Telnet default encoding
+        self.protocol_flags["ENCODING"] = settings.ENCODINGS[0]
```
Griatch: Sorry for being a pest about it, but thinking about it, since `ENCODINGS` is controlled by the user it may be empty, maybe this would better be 

`self.protocol_flags["ENCODING"] = settings.ENCODINGS[0] if settings.ENCODINGS else 'utf-8'`

vlegoff:
Sensible.  That's done.

PR #1535, opened 2017-12-16
User input auditing, revisited

strikaco:
Adds a modified ServerSession that passes dicts containing the raw string and metadata for all strings inbound and/or outbound to a user-written method specified in settings. Object references to the account, character, room and session are appended for the sake of those who wish to manipulate any of them directly.

More enhanced logging/auditing is a common requirement for server administrators, though how it should be implemented has been a controversial issue. This implementation is designed to avoid dictation of format, storage method or content and be more in line with project principles.

Supercedes #1390. I don't understand the git ecosystem so I had to create a new PR.

Tested on Linux but not on Windows; I'm concerned some of the metadata may not generate properly (PIDs, hostname?).

### DIFF
--- a/evennia/contrib/auditing.py
+++ b/evennia/contrib/auditing.py
@@ -36,20 +36,22 @@ All strings are recorded and stored in plaintext. This includes those strings
 which might contain sensitive data (create, connect, password).
 
 You are expected to implement proper masking/scrubbing of these and any
-other commands upon receipt and storing the recorded data securely.
+other commands upon receipt and storing the recorded data securely. Please do
+not store user passwords in plaintext, and be ethical-- this records even
+private communications.
 ***************
 
 Installation:
 
-Import this class to `settings.py` and designate it as the SERVER_SESSION_CLASS.
+Designate this class as the SERVER_SESSION_CLASS in `settings.py`.
 
 i.e. SERVER_SESSION_CLASS = "evennia.contrib.auditing.AuditedServerSession"
 
 There are three additional settings that govern operation-- whether you
 want to record inbound strings and/or outbound strings, and what method
-you want the parsed dicts sent to.
+you want the data sent to.
 
-AUDIT_METHOD = just_write_it (should accept (dict, **kwargs))
+AUDIT_METHOD = just_write_it
 AUDIT_IN = True or False
 AUDIT_OUT = True or False
 


Griatch:
Interesting, to begin with please change this PR to be against the Evennia `devel` branch. You can see how to do this here: https://github.com/blog/2224-change-the-base-branch-of-a-pull-request

strikaco:
@Griatch Done.

evennia/web/webclient/static/webclient/js/webclient_gui.js
```diff
@@ -175,8 +175,11 @@ function onKeydown (event) {
     }
 
     if (code === 27) { // Escape key
```
Griatch: This seems unrelated to this PR?

evennia/contrib/auditing.py
```diff
@@ -0,0 +1,167 @@
...
+Somewhere in your game (or even right there in settings.py) should be a 
+method that receives dict objects and does something with them, whether 
+writing to a database, a file, or some external queue. A simplistic example 
+might be something as brief as:
+
+from django.utils import timezone
+from evennia.utils.logger import log_file
+import json
+
+def just_write_it(data, **kwargs):
```
Griatch: The settings generally holds strings. Adding functional code directly to settings is not a good way to do it and shouldn't be recommended. The way Evennia usually handles this is to supply the path (as a string) to the method and then use one of the utilities in `evennia.utils.utils` to load it. To this end this minimal output method (`just_write_it`) should be included in this contrib as a default if you don't supply your own method - not only does it make it easier to test, it also makes for a more complete contrib. Ideally the "scrubbing" mentioned later in this docstring should be handled with the assumption of the default cmdset (alternatively log a warning about it to console).

evennia/contrib/auditing.py
```diff
@@ -0,0 +1,167 @@
...
+
+import os
+import socket
+
+from django.utils import timezone
+from evennia import settings as ev_settings
+from evennia.utils import logger
+from evennia.server.serversession import ServerSession
+
+class AuditedServerSession(ServerSession):
```
Griatch: This requires a docstring. See the Evennia Style guide.

evennia/contrib/auditing.py
```diff
@@ -0,0 +1,167 @@
...
+    def __init__(self, *args, **kwargs):
+        super(AuditedServerSession, self).__init__(*args, **kwargs)
+        
+        # Attributes governing auditing of commands and where to send logs
+        self.audit_method = getattr(ev_settings, 'AUDIT_METHOD', None)
+        self.audit_in = getattr(ev_settings, 'AUDIT_IN', False)
+        self.audit_out = getattr(ev_settings, 'AUDIT_OUT', False)
+    
+    def audit(self, **kwargs):
```
Griatch: See the style guide for how to structure the docstring with `Args` and `Returns` sections.

evennia/contrib/auditing.py
```diff
@@ -0,0 +1,167 @@
...
+                'session': self,
+                'account': account,
+                'character': char,
+                'room': room,
+            }
+        }
+        
+        return parsed
+    
+    def data_out(self, **kwargs):
```
Griatch: See style guide for the docstring formatting.

evennia/contrib/auditing.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        Sending data from Evennia->Client
+        """
+        if self.audit_method and self.audit_out:
+            self.audit_method(self.audit(src='server', **kwargs))
+        
+        super(AuditedServerSession, self).data_out(**kwargs)
+        
+    def data_in(self, **kwargs):
```
Griatch: See the docstring formatting in the style guide.

evennia/contrib/auditing.py
```diff
@@ -0,0 +1,167 @@
...
+"""
+
+import os
+import socket
+
+from django.utils import timezone
+from evennia import settings as ev_settings
+from evennia.utils import logger
+from evennia.server.serversession import ServerSession
+
```
Griatch: This would be the place to  retrieve `AUDIT_METHOD` from settings and load it into a the real python callable using one of the Evennia utilities.  Also `AUDIT_IN/OUT` would go here.

evennia/contrib/auditing.py
```diff
@@ -0,0 +1,167 @@
...
+class AuditedServerSession(ServerSession):
+    def __init__(self, *args, **kwargs):
+        super(AuditedServerSession, self).__init__(*args, **kwargs)
+        
+        # Attributes governing auditing of commands and where to send logs
+        self.audit_method = getattr(ev_settings, 'AUDIT_METHOD', None)
```
Griatch: This could be loaded as global variables up top, no need to have it in the `__init__`.

evennia/contrib/auditing.py
```diff
@@ -0,0 +1,167 @@
...
+        room = None
+        room_token = 'null'
+        if char:
+            room = char.location
+            room_token = '%s%s' % (room.key, room.dbref)
+            
+        # Compile the IP, Account, Character, Room, and the message.
+        parsed = {
+            'time': str(timezone.now()),
+            'hostname': socket.getfqdn(),
```
Griatch: This info, plus application and pid appears identical for all sends. Is it necessary to store with every log?

evennia/contrib/auditing.py
strikaco: Agreed; I will break the example out of settings and remove the suggestion to write functions in settings. I had the example in the contrib initially but assumed since it wasn't directly related to the serversession object I'd be asked to remove it.

I like the idea of making the example scrub the default cmdset. I think it's an excellent compromise given the concerns you also brought up in the previous PR.

evennia/contrib/auditing.py
strikaco: Easy enough. What are the implications of this though? I don't understand the benefit of making them global as opposed to object-level.

evennia/web/webclient/static/webclient/js/webclient_gui.js
strikaco: Yeah when I changed the branch from master to dev, it pulled in a few subsequent unrelated PRs? I have no idea what this is about.

evennia/contrib/auditing.py
strikaco: Nope.

Time should not be identical in more than a few sends at a time-- it should resolve to millisecond accuracy. I'm not sure if that was meant to be in the scope of your question.

Hostname is identical most of the time in most deployments but it *will* change under certain containerized architectures. It can be useful to know on what node the event was recorded (especially in cases where logs should be indexed/bucketed by node name instead of date, or stored locally before forwarding).

A lot of the other metadata (appame, pid) are provided for the sake of facilitating output in syslog format, interest in which was brought up by whitenoise in previous discussion. Same with object references as per tekmunkey.

Compressed, text storage is cheap and having too much data is better than not having enough when the need arises to conduct a review. There's no point in logging anything if the logs are too incomplete to compose a narrative. These considerations are why I ended up just using dicts as a medium instead of passing strings-- if deduplication or shaving every last byte is a concern, the op can easily cherry-pick what they want from the data buffet.

strikaco:
@Griatch Thanks for the feedback, I'll check out the style guide and docstring examples and send an update when I get some time during/after holiday madness.

evennia/web/webclient/static/webclient/js/webclient_gui.js
Griatch: I think this might have been because you merged into a local `develop` branch that was not in sync with the upstream version, maybe?

evennia/contrib/auditing.py
Griatch: Time was clearly different, but if you say the host name can change in containerization, that's fair enough. I know syslog format was discussed before; I was not aware PID was necessary for this. That's ok then. :)

evennia/contrib/auditing.py
Griatch: The global have a few purposes: 

- Catch settings/load errors already at the startup stage rather than when the first session is created
- The lookup will happen once whereas if it's in the `__init__` the lookup will happen every time a session is intitated (the difference in performance is probably miniscule though)
- This pattern makes it somewhat clearer an external resource is used (IMO at least)
- Maybe most importantly, it's consistent with the pattern used for loading external resources by path in the rest of Evennia.

Griatch:
@strikaco Any updates to this?

Griatch:
@strikaco Bump.

strikaco:
Superceded by #1651.

Issue #1104, opened 2016-10-19
Feature request: @set for other typeclasses


TehomCD:
#### Brief summary of issue / Description of requested feature:
Attributes and Tags can exist in other typeclasses beyond PlayerDB and ObjectDB, but can currently only be set with @set on those two typeclasses, with @tag only working on ObjectDB. It would be a nice feature if they supported specifying typeclasses with a switch or some other mechanism.

#### Steps to reproduce the issue / Reasons for adding feature:

1. Allow @set/@tag to be more general purpose tools

#### Error output / Expected result of feature
Extending functionality for @set/@tag

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
Maybe something like checking if any of a command's switches are in a command's valid_typeclasses,  where the 'valid_typeclasses' property is a set of the allowed typeclasses for a given command, and if so, calling that typeclass's search method rather than object_search. Since they all have tag/attribute handlers, all that really needs changed is fetching them, everything else should work identically.

Griatch:
It might be good to be able to close this by adding the same functionality to the `tag` command as well. A PR anyone?

PR #1536, opened 2017-12-17
Allow other typeclasses to have their Attributes set via command.

TehomCD:
Adds support for Channels and Scripts to `@set`, as well as allowing for global-searching via typeclass on non-dbrefs for regular objects.

So I want to gradually do a pass over the existing building commands and extend them to allow users to interact with the AttributeHandlers and TagHandlers of the TypedObjects that currently aren't supported - Channels, Scripts for both, and for Tags that'll include Msg and HelpEntry objects. The latter is more complicated because they don't have all the same fields as TypedObjects, so I started with `@set` first.

Resolves the `@set` part of #1104 

Other thoughts - I'm not sure the most DRY way of implementing the searches for other stuff. It could be added as a mixin on a per-command basis to different ObjManipCommands, but some of those wouldn't make sense for different typeclasses, or would need special handling. `@ex` is one of the more interesting and tricky ones - it clearly should be supported for every typeclass, since you want an easy way of viewing tags/Attributes on everything, but it'd probably make the most sense as a method call so that it could be customized for each typeclass.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -1458,8 +1458,10 @@ class CmdSetAttribute(ObjManipCommand):
         script: If we're trying to set an attribute on a script
         channel: If we're trying to set an attribute on a channel
         account: If we're trying to set an attribute on an account
-        (room, exit, char/character may all be used as well for global
-         searches)
+        room: Setting an attribute on a room (global search)
+        exit: Setting an attribute on an exit (global search)
+        char: Setting an attribute on a character (global search)
+        character: Alias for char, as above.
 
     Sets attributes on objects. The second form clears
     a previously set attribute while the last form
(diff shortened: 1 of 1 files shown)


evennia/commands/default/building.py
```diff
@@ -1455,6 +1455,11 @@ class CmdSetAttribute(ObjManipCommand):
 
     Switch:
         edit: Open the line editor (string values only)
+        script: If we're trying to set an attribute on a script
+        channel: If we're trying to set an attribute on a channel
+        account: If we're trying to set an attribute on an account
+        (room, exit, char/character may all be used as well for global
```
Griatch: This should probably be phrased differently - rom, exit, char are valid switches but putting them in parentheses makes that unclear.

TehomCD:
@Griatch Done, let me know if you think it could still be made more clear.

PR #1550, opened 2018-01-08
Refactor ExtendedRoom return_appearance to extract desc update into its own method.

TehomCD:
This just extracts the part of return_appearance in ExtendedRoom that would update its current desc if the time/season has change into its own method.

To allow us to update the desc without either copy/pasting the code or discarding the unused result of the return_appearance `super()` call if the string has to be modified, which is usually the case when overriding return_appearance. I didn't really like doing `ExtendedRoom.return_appearance(self, looker)` just to update the seasonal desc.

Also fixed the signature to include `**kwargs` in return_appearance.

### DIFF
--- a/evennia/contrib/extended_room.py
+++ b/evennia/contrib/extended_room.py
@@ -213,37 +213,39 @@ class ExtendedRoom(DefaultRoom):
             return detail
         return None
 
-    def return_appearance(self, looker):
+    def return_appearance(self, looker, **kwargs):
         """
         This is called when e.g. the look command wants to retrieve
         the description of this object.
 
         Args:
             looker (Object): The object looking at us.
+            **kwargs (dict): Arbitrary, optional arguments for users
+                overriding the call (unused by default).
 
         Returns:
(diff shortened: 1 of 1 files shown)


evennia/contrib/extended_room.py
```diff
@@ -213,37 +213,39 @@ def return_detail(self, key):
...
             # season changed. Load new desc, or a fallback.
-            if curr_season == 'spring':
-                new_raw_desc = self.db.spring_desc
-            elif curr_season == 'summer':
-                new_raw_desc = self.db.summer_desc
-            elif curr_season == 'autumn':
-                new_raw_desc = self.db.autumn_desc
-            else:
-                new_raw_desc = self.db.winter_desc
+            new_raw_desc = self.attributes.get("%s_desc" % curr_season)
```
Griatch: nice refactor :)

Griatch:
Looks good to me, the contrib has not kept up with the latest core changes so this was needed. Merging.

PR #1552, opened 2018-01-15
`@locate` alias of CmdFind shows location of find

BlauFeuer:
If using `@locate` alias and only one object is found and the found object has a location, display that information.

Also, typo fix in CmdWhisper help docstring and access method in Command class

## Motivation for adding to Evennia
If one wants to know the location of an object, instead of doing a `@find` then an `@examine`, one can just `@locate` the object, provided the search identifier locates a unique object that has a non-None location.

## Other info (issues closed, discussion etc)

For develop branch, though the feature (and typo fix) can easily be back-ported.
Should a unit test be written for this minor addition?

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -2315,7 +2315,7 @@ class CmdFind(COMMAND_DEFAULT_CLASS):
             else:
                 result = result[0]
                 string += "\n|g   %s - %s|n" % (result.get_display_name(caller), result.path)
-                if self.cmdstring == "@locate" and not is_account and result.location:
+                if "locate" in self.cmdstring and not is_account and result.location:
                     string += "  Location: {}".format(result.location.get_display_name(caller))
         else:
             # Not an account/dbref search but a wider search; build a queryset.
@@ -2352,7 +2352,7 @@ class CmdFind(COMMAND_DEFAULT_CLASS):
                 else:
                     string = "|wOne Match|n(#%i-#%i%s):" % (low, high, restrictions)
                     string += "\n   |g%s - %s|n" % (results[0].get_display_name(caller), results[0].path)
-                    if self.cmdstring == "@locate" and nresults == 1 and results[0].location:
+                    if "locate" in self.cmdstring and nresults == 1 and results[0].location:
                         string += "  Location: {}".format(results[0].location.get_display_name(caller))
             else:
                 string = "|wMatch|n(#%i-#%i%s):" % (low, high, restrictions)
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -2238,12 +2238,14 @@ class CmdFind(COMMAND_DEFAULT_CLASS):
 
     Usage:
       @find[/switches] <name or dbref or *account> [= dbrefmin[-dbrefmax]]
+      @locate - this is a shorthand for using the /loc switch.
 
     Switches:
       room - only look for rooms (location=None)
       exit - only look for exits (destination!=None)
       char - only look for characters (BASE_CHARACTER_TYPECLASS)
       exact- only exact matches are returned.
+      loc  - display object location if exists and match has one result
 
     Searches the database for an object of a particular name or exact #dbref.
     Use *accountname to search for an account. The switches allows for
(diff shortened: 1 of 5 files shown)


BlauFeuer:
Someone screenshot the typo, so we couldn't forget where it was:
guvretm

evennia/commands/default/building.py
```diff
@@ -2315,6 +2315,8 @@ def func(self):
             else:
                 result = result[0]
                 string += "\n|g   %s - %s|n" % (result.get_display_name(caller), result.path)
+                if "locate" in self.cmdstring and not is_account and result.location:
```
Griatch: Previously 'locate' was just an alias. Now that it has an extra function, this needs to be indicated in the docstring of the Command..

evennia/commands/default/building.py
```diff
@@ -2350,6 +2352,8 @@ def func(self):
                 else:
                     string = "|wOne Match|n(#%i-#%i%s):" % (low, high, restrictions)
                     string += "\n   |g%s - %s|n" % (results[0].get_display_name(caller), results[0].path)
+                    if "locate" in self.cmdstring and nresults == 1 and results[0].location:
+                        string += "  Location: {}".format(results[0].location.get_display_name(caller))
```
Griatch: Looking at this, having just a few extra blank spaces looks a little inconsistent with other code. I think it's better to put this in a parenthesis like so: `" (location: {}"`. Also, we use `|w` color for the instruction and green for the match, so we should use that here for the location too.

BlauFeuer: `string += "  (|wlocation|n: |g{})|n".format(results[0].location.get_display_name(caller))`

Griatch: Yeah, looks good. :)

evennia/commands/default/building.py
BlauFeuer: I see the point.  With some custom commands and aliases, I've documented the switch then mention which alias performs the switch by default, which makes the docstring much cleaner and adds functionality to the entire command by way of a switch instead.

```
    Usage:
      @find[/switches] <name or dbref or *account> [= dbrefmin[-dbrefmax]]
    Switches:
      room - only look for rooms (location=None)
      exit - only look for exits (destination!=None)
      char - only look for characters (BASE_CHARACTER_TYPECLASS)
      exact- only exact matches are returned.
      loc - also displays location of object if unique (same as @locate)
```

I'll make the doc change and code for this alteration if you agree.

Griatch: Actually, while adding a switch is fine, i think it makes sense to also include the alias in the Usage line, so
```
@find[/switches] <name or dbref or *account> [= dbrefmin[-dbrefmax]]
@locate [/switches] <name or dbref or *account> [= dbrefmin[-dbrefmax]]
```
And then add a note in the main description (and in the switches) that `@locate` is the same as giving `/loc` as a switch

Griatch:
Any progress on this?

BlauFeuer:
Wow, 25 days ago.... Sorry, I haven't forgotten about it.  Looking at other similar commands' help entries, I think there may not be a standard way of indicating "this alias is the same as using these switches,"

The help entry for `@type`:

```
set or change an object's typeclass

Usage:
  @typeclass[/switch]  [= typeclass.path]
  @type                     ''
  @parent                   ''
  @swap - this is a shorthand for using /force/reset flags.
  @update - this is a shorthand for using the /force/reload flag.

Switch:
  show - display the current typeclass of object (default)
  update - *only* re-run at_object_creation on this object
          meaning locks or other properties set later may remain.
  reset - clean out *all* the attributes and properties on the
          object - basically making this a new clean object.
  force - change to the typeclass also if the object
          already has a typeclass of the same name.
```

BlauFeuer:
@griatch Not intending to tangent too far, but trying to use existing help docstrings as a guide, I wonder why "flag" is not "switch" and why there's no /reload switch documented.  I think the line that reads:
> `@update` - this is a shorthand for using the /force/reload flag.
should read:
> `@update` - this is a shorthand for using the /force/update switch.

Meanwhile, I'll work on getting the code right before nitpicking on the docstrings.

Issue #1483, opened 2017-10-19
Starting Evennia as Described Runs the Server Interactively


chainsol:
#### Brief summary of issue / Description of requested feature:
Evennia's source code, as well as the help for the Evennia command, seems to suggest that the default when calling Evennia from the command line is daemon mode - that is, a mode where input from the command line is not expected.

This would make sense, since Evennia could then be started from the command line simply and easily, without the need for GNUScreen/Tmux/etc, but instead, Evennia seems to start the *server* interactively no matter what command line options it's given.

Edit: To clear this up for anyone reading this in the future, as Griatch comments below, the server isn't starting interactively - there's a third process, the `runner` that is outputting to the terminal.

#### Steps to reproduce the issue / Reasons for adding feature:

1. Install Evennia 0.7 on Linux ( Haven't tried with Windows )
2. Start the server with `evennia start`, which is described as defaulting to daemon mode.
3. Exit your terminal session
4. Reload the server from in-game, or take any action that would print information to the terminal

#### Error output / Expected result of feature
Evennia's server process crashes, the portal stays running. The source and the help for `evennia start` reads as if it should start both the server and the portal in daemon mode, but what actually happens is the server process expects to be able to output to STDOUT, and crashes when it can't.

Edit: Again, to clear this up for anyone in the future, running `@py print('whatever')` doesn't crash the server - as Griatch explains below, the reason the server crashes when running `reload` is due to the `runner`, not the `server` or `portal`.

Overall, I think this is a two part fix - fix the bug that causes daemon mode to be ignored for the server, and also change the default for `evennia start` to be interactive mode - it makes more sense that way for new users. Then daemon mode could be a flag instead, for those of us running a production server.

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
I've only tested this on Linux but it's been consistent. Screen/Tmux is the current solution for this, but overall for me, logging to a file is sufficient for production, and interactive mode is nice for development - the current thing we have is neither of those modes.

Griatch:
I actually don't see any crashing when trying to print to stdout after having closed the terminal in which the server started. STDOUT should just be redirected nowhere, and that is what I see. Can you give an example, not related to reloading, where the Server crashes for not being able to write to STDOUT?

chainsol:
[...]

PR #1557, opened 2018-01-19
Server portal reload

Griatch:

This constitutes a rework of the Server/Portal infrastructure. Here are some of the changes:

- Only two processes. The Portal handles the reload of the Server now. No more `evennia_runner.py`.
- No interactive mode - both Server/Portal run as daemons. 
- The launcher is a proper AMP client now, meaning it can send and communicate to the Portal and also keep the connection open to wait for status etc.
- Log files can be piped to the client (this replaces the interactive mode)
- The server and portal logs are merged by default. The log format was shortened and made less verbose.
- Proper status-reporting from the server/portal to the launcher
- You can start/operate Evennia from outside the game directory by specifying the --gamedir option to the client.
- No more use of .pid and .reload files in strange ways. There are still .pid files but those are now used transparently by Twisted and not managed by Evennia. So no more "Stale PID file"-warnings.
- You can no longer reload the Portal; this seems a good compromise, most just shut it down and restarted it before (since you lose connections it doesn't matter anyway).
- Since we no longer rely on particular uses of SIGTERM signals there are less special cases and Windows users should reload just fine from the console now.

This closes https://github.com/evennia/evennia/issues/1483


BlauFeuer:
### In the Evennia Launcher menu:
```
    +----Evennia Launcher-------------------------------------------+
    |                                                               |
    +--- Common operations -----------------------------------------+
    |  1) Start Portal and Server    (also restart downed Server)   |
...
```
there's room to display the game path and name so the user knows which game folder the menu will act on, or some status information there.  Did you have plans for something to display on that line?

#### Option 4 - Reset Server
This stops and restarts both the server+portal?    If not, then what's the difference between Reload and Reset?  If so, then "(Full shutdown with restart)" would be a better wording and less ambiguous to me than (Server shutdown with restart) if I take "Server" to not include the Portal.

Griatch:
@BlauFeuer Putting the game folder path or at least the game folder name there might be a good idea. 

The reset mimics the in-game `@reset` command. Just like `@reload` this does stop- and start the Server only. The difference is that non-persistent Scripts are killed and the shutdown-hooks are called rather than the reload-hooks. So it's a Server shutdown - but with an automatic restart. It's a little hard to describe on one line though.

BlauFeuer:
Ah I understand the `@reset` versus `@reload` distinction now.

I've been using the menu via `evennia menu --gamedir MYGAME` from within pyenv but outside evennia and MYGAME, but once in the menu, I have to remember that I started with MYGAME (sometimes I have other test game code folders around), so displaying it there in the title would be perfect for me.

The full restart would be something I'd use after pulling a different branch or having updated something in the portal.   It just saves a step of stopping all and starting it again for that game folder.

BlauFeuer:
Another thing to note is `evennia start` won't show errors when it fails to start, so when in doubt about it, start with `evennia start -l`  (lowercase l for log view)

BlauFeuer:
A tiny detail, but for some reason the listing of a second telnet port is not indented as far as the first, and rest of the list in the ports listing from the menu:
```
----------------------- Evennia ---
MYGAME Portal 0.8.0-dev (rev 0b6e94a)
    external ports:
        telnet: 4000
       telnet: 8888
        webserver-proxy: 8000
        webclient-websocket: 4002
    internal_ports (to Server):
        webserver: 8001
        amp: 4006

```
Maybe when my eyes aren't as crossed, I'll see if I can find it in the code.

Griatch:
@BlauFeuer Nice catch, that's been fixed.

Griatch:
Alas, on a quick test, this branch does not work on Windows as-is. The portal/server starts up according to logs but the launcher is never notified of the started state and the server can't be connected to :(

Griatch:
It's very strange since this solution should be *more* OS-agnostic than the old solution built around sending signals ...

Griatch:
What I've found: 
- Windows maps the external Portal Websocket port to 4005, Linux to 4002. Configured value of `WEBSOCKET_CLIENT_PORT` is 4002. 
- Windows maps the internal Server Webserver port to 4002, Linux to 4005. Configured value is the second value in `WEBSERVER_PORT` tuple, which is 4005.

For some reason they are swapped on Windows, not sure why that is, there is no OS-specific differences in the port assignments of https://github.com/evennia/evennia/blob/server-portal-reload/evennia/server/portal/portal.py

Griatch:
Actually, the port-discrepancies may be a presentation bug. On Linux, the `evennia info` reports different values than what is shown actually starting in the log ... We don't get the `evennia info` on Windows (for unknown reasons) so we don't see the comparison there.

chainsol:
Hmmm, my info seems to match, unless I've missed something:
`evennia info`
```
-------------------------- Evennia ---
mygame Portal 0.8.0-dev (rev d49a005)
    external ports:
        telnet: 4010
        webserver-proxy: 4011
        webclient-websocket: 4015
    internal_ports (to Server):
        webserver: 4012
        amp: 4016

mygame Server 0.8.0-dev (rev d49a005)
    internal ports (to Portal):
        webserver: 4012
        amp : 4016
--------------------------------------
```
`cat server/logs/server.log`
```
18-01-21 18:19:57  |Portal| twistd 16.0.0 (/home/user/.virtualenvs/biomud_dev/bin/python 2.7.12) starting up.
18-01-21 18:19:57  |Portal| reactor class: twisted.internet.epollreactor.EPollReactor.
18-01-21 18:19:57  |Portal| AMP starting on 4016
18-01-21 18:19:57  |Portal| Telnet starting on 4010
18-01-21 18:19:57  |Portal| Websocket (WebSocketFactory) starting on 4015
18-01-21 18:19:57  |Portal| Webserver-proxy starting on 4011
```

Griatch:
@chainsol Yes, I had to write it on a table on-paper to get it right: It's actually the Linux ports which don't match settings and not the Windows ports. The Windows ports actually match settings (which have changed a little since `develop` branch). And the reason appears to be that I'm using and old `mygame` on Linux. If I `--init` a fresh game dir, the ports on Linux match settings (and Windows). I have no idea why this is, but it's probably unrelated to Windows not starting the Server ...

Griatch:
So the discrepancy between info and log was not there, that was just me confusing myself.

chainsol:
@Griatch OK, so that's why my version matches the settings then? I thought we had the issue nailed down, too. :(

Griatch:
By putting logging statements on the server/portal sides, it seems we die at this point: https://github.com/evennia/evennia/blob/server-portal-reload/evennia/server/amp_client.py#L108 So the AMP connection is established correctly, but the actual sending of the AMP command across the wire fails. Putting callbacks/errbacks doesn't reveal anything either...

Griatch:
I wonder if the Windows error is due to something I don't understand about how Windows spawns new processes. Cases tested:

Case 1:
- `evennia start` - Portal starts, and Portal in turn starts Server (both with Popen calls). 
- But the Server can't call back to the Portal over AMP, so it pauses in its initialization. 
- The launcher does not time out. Instead it has to be exited with Ctrl-C. At that point you get the `Terminate batch job (Y/N)` question. 
- However, when looking at the server log, it looks like doing Ctrl-C for the launcher actually affects the portal/server processes. Both shut down. There is also a very strange behavior of a second portal starting and immediately shutting down (it even overrides the log). Very strange.

Case 2:
- `evennia start` from one console. It freezes as above. 
- Open another console and try `evennia start`.
- Output freezes on the "Portal starting ..." message, but when exiting with Ctrl-C, what gets written to the log is a traceback about how port 4006 is already in use. This means the launcher tried to start a completely new Portal instead of (like on Linux it finding the old one and just saying that it's already running. This suggests that the ports from the first console are completely inaccessible/blocked. So the runner couldn't connect, thought nothing was running and tried to start (and now the blocked port was a problem). 

It *appears* python's `Popen` works differently on Windows, it appears to not detach from the process properly but run in the foreground ...

Griatch:
So, combining 
https://msdn.microsoft.com/en-us/library/windows/desktop/ms684863(v=vs.85).aspx
https://stackoverflow.com/questions/14797236/python-howto-launch-a-full-process-not-a-child-process-and-retrieve-the-pid
https://stackoverflow.com/questions/25122823/hide-console-with-subprocess-popen

It appears that a combination of these makes it work on Windows at the moment: 

- Remove `stdout=PIPE` argument from Popen call
- Set the `creationflags=create_new_process_group | create_no_window` in the call to Popen. This will detach the call (so you can close the console safely) while not opening a rogue console window at the same time. 

Alas, the missing PIPE is annoying, it means you won't get eventual errors from the Server before it starts the logger.

Griatch:
Also, as a reminder - Windows doesn't do PID files so the launcher will need to display the PIDs using some other mechanism.

Griatch:
Further looking at this, the `CREATE_NEW_PROCESS_GROUP` is not necessary to create a detached program that survives closing the console. Only the `CREATE_NO_WINDOW` is enough. Removing this flag may be beneficial since Microsoft docs suggest that processes in the new group can no longer receive SIGTERM signals.

Griatch:
One suggestion I've seen for managing the PIPE issue on Windows would be to redirect stdout to a file handle. So maybe one could append it to the log file(s). Not sure if that would mess with the normal logging somehow though.

Griatch:
Unfortunately, while replacing `stdout=PIPE` with a filehandle to the server log file works just fine in Linux, on Windows this solution appears to work at a first glance. But the moment there is something in the buffer (test by putting `1/0` on the main evennia application object in server.py), it deadlocks. Trying to flush the buffer doesn't appear to have any effect on this. Interestingly when putting some debug statements in `start_server` (`evennia/server/amp_server.py:144`) it's clear that `Popen` itself is not blocking, the method even exits normally. The Server can't start (due to the added error) but it also is not echoing anything about the error to the logfile as it should. 
If putting a print statement (not causing a traceback) on the server object, the Server starts perfectly ... but that print statement is nowhere to be seen in the logs.

Issue #1555, opened 2018-01-19
session.at_sync has typo, but is never called


TehomCD:
#### Brief summary of issue / Description of requested feature:
`session.py` has a typo at https://github.com/evennia/evennia/blob/master/evennia/server/session.py#L136

`account.attributs` instead of `account.attributes`. But this doesn't seem to be called anywhere and generate an error, since it looks like serversession just overrides the method entirely and doesn't call super.

#### Steps to reproduce the issue / Reasons for adding feature:

1. Check link

#### Error output / Expected result of feature
No AttributeError is raised because code doesn't seem to be called.

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
I'm not sure if the code is a relic of something that isn't used anymore and should be replaced by a simple pass, or if a super() call is missing and it has to be fixed.

Edit: Possibly related to #1353 ?

Griatch:
@TehomCD As noted, this was overridden in the serversession child class, but that one never called `super()`. Fixed now. We'll see if this has an effect on those having trouble with #1553

PR #1558, opened 2018-01-21
Fix at_sync update call to have right number of arguments.

TehomCD:
Whoops! There was more than one typo in #1555 . The `.attributes.get()` had the default empty dict outside it as a second argument to update, which only accepts one argument. So upon reload every at_sync would fail.

Quick fix to at_sync.

Additional fix to #1555

### DIFF
--- a/evennia/server/session.py
+++ b/evennia/server/session.py
@@ -134,7 +134,7 @@ class Session(object):
 
         """
         if self.account:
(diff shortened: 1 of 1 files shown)


Griatch:
Ah sorry. I'm merging this, but the empty dict is meant to be inside the `attributes.get` call since it's the default value if the attribute doesn't exist. I'll change that.

PR #1561, opened 2018-01-29
Webclient splits (WIP)

chainsol:
Add webclient splitting capability to Evennia.

We've had a lot of discussion about this - nice level of customization, better message handling, etc.

This is **in progress**, it's not ready to be merged, but discussion on it would be excellent. Expect many many more commits to this PR! The commit history is quite...messy, so sorry about that!

If one would like to see it in action, https://dev.perfectdisorder.net

### To-Do

- Allow server maintainers to create a custom layout
- Allow end-users to split their own panes, if the server maintainer deems it allowable
- Tweak Evennia commands to actually tag themselves in a useful way
- Allow different types of handling per-tag - for instance, in-window popup, named window, named pane, etc.

### DIFF
--- a/evennia/web/webclient/static/webclient/css/webclient.css
+++ b/evennia/web/webclient/static/webclient/css/webclient.css
@@ -93,7 +93,7 @@ div {margin:0px;}
     top: 0;
     left: 0;
     right: 0;
-    bottom: 70px;
+    /*bottom: 70px;*/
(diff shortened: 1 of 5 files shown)


chainsol:
Since #1592 solves the same feature request this PR was addressing, I'll close this in favor of that PR.

PR #1571, opened 2018-02-13
Change if statement in comms/models.py to fix #1570

### LINKED ISSUE #1570: at_msg_receive hook not firing when channel message received.
I have an object that needs to do work when a message is received from a channel.

1. Create test object with at_msg_receive hook.
2. Add object to a channel
3. send a message on that channel.

Nothing happens when the channel sends a message to the object.  verified that the object is listed and highlighted for that specific channel in allcom who

chainsol:
Applies fix described in #1570 as well as adds/fixes a couple other conveniences for people attempting to subscribe objects to channels.

Fixes #1570, better supports inter-object messages through channels

1. Applies fix described in #1570 in `comms/models.py`, changing an if statement so that objects without accounts can still be subscribed to a channel
2. Move options dict being added to kwargs in the DefaultObject `.msg` function to _prior_ to sending over to `at_msg_receive` hook - this allows us to see what channel sent the message in the `at_msg_receive` hook, as seems to have been intended.
3. Add `is_connected` property to DefaultObject, defaulting to `False` or the account's `is_connected` property, if available. This avoids an error message we would get otherwise.

### DIFF
--- a/evennia/comms/models.py
+++ b/evennia/comms/models.py
@@ -584,9 +584,7 @@ class SubscriptionHandler(object):
         for obj in self.all():
             from django.core.exceptions import ObjectDoesNotExist
             try:
-                if hasattr(obj, 'account'):
-                    if not obj.account:
-                        continue
+                if hasattr(obj, 'account') and obj.account:
                     obj = obj.account
                 if not obj.is_connected:
                     continue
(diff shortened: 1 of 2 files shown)


Griatch:
Looks good to me, thanks for the PR!

chainsol:
Thanks! I think we'll just want to edit the info in the wiki to reflect this - objects subscribing to channels will work now, but for them to receive messages they need to implement `is_connected` properly, so that it returns `True` if they should appear as online. I'll edit the wiki appropriately!

Griatch:
@chainsol Good, that sounds like a useful wiki addition.

arumford:
Actually, we should also change the docstring on objects.py to have at_msg_receive have the right signature.

PR #1580, opened 2018-03-02
Optional features for MuxCommand, added to default Evennia commands

BlauFeuer:
#### MUX command switch abbreviations

Allowing the shortest unique  text to be used as a command switch.
#### Use an alternate delimiter
https://github.com/evennia/evennia/wiki/Coding-FAQ#add-parsing-with-the-to-delimiter

This is easy to add as an optional feature, just like switch abbreviations, so adding it to `CmdTeleport` and `CmdGet` demonstrates how they work to developers creating new commands.

MUX command switch abbreviations is a useful time-saver for users familiar with all the switches of a command. 

An example is typing `@tel/l Griatch` as a shortcut for the `/loc` switch.  

A side effect of this feature is that it relies on the command class knowing which options are valid, which helps organize them, but is purely optional.

The split delimiter is something that several developers have asked about.

Although unit tests were added, to be sure, this should be user tested in develop branch.

### DIFF
--- a/evennia/commands/default/muxcommand.py
+++ b/evennia/commands/default/muxcommand.py
@@ -80,9 +80,10 @@ class MuxCommand(Command):
         start with the switch indicator /.
 
         Optional variables to aid in parsing, if set:
-          self.options = (tuple of valid /switches expected
-                          by this command (without the /))
-          self.split = Alternate string delimiter to separate left/right hand side.
+          self.switch_options  - (tuple of valid /switches expected by this
+                                  command (without the /))
+          self.rhs_split       - Alternate string delimiter to separate
+                                 left/right hand sides.
 
         This parser breaks self.args into its constituents and stores them in the
         following variables:
@@ -102,6 +103,18 @@ class MuxCommand(Command):
         """
         raw = self.args
         args = raw.strip()
+        # Temporary code to use the old settings before renaming     # #
+        if hasattr(self, "options"):                                   #
+            self.switch_options = self.options                         #
+        if hasattr(self, "split") and not hasattr(self, "rhs_split"):  #
+            self.rhs_split = self.split                            # # #
+        # Without explicitly setting these attributes, they assume default values:
+        if not hasattr(self, "switch_options"):
+            self.switch_options = None
+        if not hasattr(self, "rhs_split"):
+            self.rhs_split = "="
+        if not hasattr(self, "account_caller"):
+            self.account_caller = False
 
         # split out switches
         switches = []
@@ -114,17 +127,19 @@ class MuxCommand(Command):
             else:
                 args = ""
                 switches = switches[0].split('/')
-            # Parse mux options, comparing them against user-provided switches, expanding abbreviations.
-            if hasattr(self, "options") and self.options and switches:
-                # If specific options are known, test them against given switches.
+            # If user-provides switches, parse them with parser switch options.
+            if switches and self.switch_options:
                 valid_switches, unused_switches, extra_switches = [], [], []
                 for element in switches:
-                    option_check = [each for each in self.options if each.lower().startswith(element.lower())]
-                    if len(option_check) > 1:
+                    option_check = [each for each in self.switch_options
+                                    if each.lower() == element.lower() or
+                                    each.lower().startswith(element.lower())]
+                    match_count = len(option_check)
+                    if match_count > 1:
                         extra_switches += option_check  # Either the option provided is ambiguous,
-                    elif len(option_check) == 1:
+                    elif match_count == 1:
                         valid_switches += option_check  # or it is a valid option abbreviation,
-                    elif len(option_check) == 0:
+                    elif match_count == 0:
                         unused_switches += [element]  # or an extraneous option to be ignored.
                 if extra_switches:  # User provided switches
                     self.msg('|g%s|n: |wAmbiguous switch supplied: Did you mean /|C%s|w?' %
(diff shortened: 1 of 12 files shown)


BlauFeuer:
Travis CI completed with success, but says it's stalled. Looking for comments and feedback on the features and code.

evennia/commands/default/muxcommand.py
```diff
@@ -79,6 +79,11 @@ def parse(self):
         it here). The rest of the command is stored in self.args, which can
         start with the switch indicator /.
 
+        Optional variables to aid in parsing, if set:
```
Griatch: I feel `options` and `split` are a little too generic as names. `split` is what is usually done in Python when splitting by spaces; and `options` suggests (at least to me) some sort of internal option mechanic for the Command class. These are also options which will go with the other non-parser related arguments like `locks`, `key` etc. I suggest more explicit names: `parser_switch_options` and `parser_rhs_split`. 
(In comments below I'll keep referring to `options` and `split` for now)

evennia/commands/default/muxcommand.py
```diff
@@ -109,15 +114,44 @@ def parse(self):
             else:
                 args = ""
                 switches = switches[0].split('/')
+            # Parse mux options, comparing them against user-provided switches, expanding abbreviations.
+            if hasattr(self, "options") and self.options and switches:
+                # If specific options are known, test them against given switches.
+                valid_switches, unused_switches, extra_switches = [], [], []
+                for element in switches:
+                    option_check = [each for each in self.options if each.lower().startswith(element.lower())]
```
Griatch: I think an exact match should be given priority here. That is, consider two switches `test` and `testingthings`. If the user enters `/test`, this system will find two `.startswith` matches and give an unambiguous error. So the exact match should be tested first, and only if that does not return any match should a fuzzy match be attempted. 
(In the command parser we also consider the length of the fuzzy-matches to score the results but that's a bit overkill for this case).

evennia/commands/default/muxcommand.py
```diff
@@ -109,15 +114,44 @@ def parse(self):
...
+                    if len(option_check) > 1:
```
Griatch: Minor style thing would be to put `len(option_check)` into a variable and use that for comparison rather than having to do `len()` over and over. No big deal but it's good practice to re-use a result when possible.

evennia/commands/default/muxcommand.py
```diff
@@ -79,6 +79,11 @@ def parse(self):
         it here). The rest of the command is stored in self.args, which can
         start with the switch indicator /.
 
+        Optional variables to aid in parsing, if set:
+          self.options = (tuple of valid /switches expected
+                          by this command (without the /))
+          self.split = Alternate string delimiter to separate left/right hand side.
```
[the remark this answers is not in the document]
BlauFeuer: Is there no better name for the new optional `MuxCommand` class variables than what you've requested?
```
        Optional variables to aid in parsing, if set:
          self.parser_switch_options  - (tuple of valid /switches expected by this
                                         command (without the /))
          self.parser_rhs_split       - Alternate string delimiter to separate
                                        left/right hand side.
```
This requires updating all explicit settings of them when creating a command based on this command class:
```
    key = "@tel"
    aliases = "@teleport"
    parser_switch_options  = ("quiet", "intoexit", "tonone", "loc")
    parser_rhs_split = "to"
    locks = "cmd:perm(teleport) or perm(Builder)"
    help_category = "Building"
```
which appears out of place among the others that are not appended with `func_`.

Would `switch_options` and `rhs_split` be a good compromise, or perhaps something else?

evennia/commands/default/muxcommand.py
```diff
@@ -109,15 +114,44 @@ def parse(self):
             else:
                 args = ""
                 switches = switches[0].split('/')
+            # Parse mux options, comparing them against user-provided switches, expanding abbreviations.
```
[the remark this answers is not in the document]
BlauFeuer: I originally had done that, but I took the example of `account_caller`, which is also optional, but not set to any default.

I agree, it is cleaner to default these values instead of testing to see if they were assigned.

evennia/commands/default/muxcommand.py
BlauFeuer: This was easy to add.  I'll commit it soon.

evennia/commands/default/muxcommand.py
BlauFeuer: Done, committing an update soon.

evennia/commands/default/muxcommand.py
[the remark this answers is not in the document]
Griatch: The others wouldn't be prepended with `func` since they apply to the entire Command class, not just to `func`. These switches are the first to apply only to a single sub-method. But for the sake of  brevity I think `switch_options` and `rhs_split` are ok too.

evennia/commands/default/muxcommand.py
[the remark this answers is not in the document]
Griatch: `account_caller`, as I recall, is used by a special group of Commands, those defined in the `AccountCmdSet`. At the time of its implementation the whole concept of Character/Account was new. One could certainly argue that it could be useful existing as a default too.

BlauFeuer:
The parsing all works, but one command test fails silently, so I am still tracking that down.
> Ran 288 tests in 155.644s
> FAILED (failures=1)

```
FAIL: test_teleport (evennia.commands.default.tests.TestBuilding)
  File "/home/evennia/evennia/evennia/commands/default/tests.py", line 317, in test_teleport
    self.call(building.CmdTeleport(), "Room2", "Room2(#2)\n|Teleported to Room2.")
  File "/home/evennia/evennia/evennia/commands/default/tests.py", line 101, in call
    raise AssertionError(retval)
AssertionError:
==============================Wanted message==================================
Room2(#2)
|Teleported to Room2.
==============================Returned message================================

==============================================================================
```

BlauFeuer:
Odd: I could not get one test to pass until merging from upstream. Sorry for the delay.
Development continues.

evennia/commands/default/muxcommand.py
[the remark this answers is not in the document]
BlauFeuer: I addressed this with `account_caller`, too, and I think it turned out well.

BlauFeuer:
Ready for another review.  I can squeak in more test coverage if it seems lacking.

Griatch:
@BlauFeuer Nice! Alas I've lots of RL stuff happening this week so it may take a few days before I can review this -  just so you know I've not forgotten. :)

evennia/commands/default/muxcommand.py
```diff
@@ -109,16 +123,50 @@ def parse(self):
...
+            # If user-provides switches, parse them with parser switch options.
+            if switches and self.switch_options:
+                valid_switches, unused_switches, extra_switches = [], [], []
+                for element in switches:
+                    option_check = [each for each in self.switch_options
+                                    if each.lower() == element.lower() or
+                                    each.lower().startswith(element.lower())]
+                    match_count = len(option_check)
+                    if match_count > 1:
+                        extra_switches += option_check  # Either the option provided is ambiguous,
```
Griatch: As a hint, `extra_switches.extend(option_check)` is faster and cleaner - `+=` is pretty inefficient overall (but this is surely a minor thing so not critical at all).

evennia/commands/default/muxcommand.py
```diff
@@ -109,16 +123,50 @@ def parse(self):
             else:
                 args = ""
                 switches = switches[0].split('/')
+            # If user-provides switches, parse them with parser switch options.
+            if switches and self.switch_options:
+                valid_switches, unused_switches, extra_switches = [], [], []
+                for element in switches:
+                    option_check = [each for each in self.switch_options
```
Griatch: For eventual use by the end-user, it would be an idea to do 

```python
self.switch_options = [opt.lower() for opt in self.switch_options]
```
already above - this means you don't need to run `lower()` on every element again below but also that users wanting to use `switch_options` for their own ends can be sure it has already been parsed and is in lowercase.

BlauFeuer:
@Griatch Thanks for the additional feedback. There was a check for the exact match before considering fuzzy matching, but it wasn't a good one, and because it wasn't under test, I didn't realize it was bad.

Griatch:
Thanks for the PR, looks good now. Merged.

PR #1586, opened 2018-03-11
Remove extra words

maiki:
A plain language example had a few extra words.

### DIFF
--- a/evennia/contrib/rpsystem.py
+++ b/evennia/contrib/rpsystem.py
@@ -708,12 +708,15 @@ class RecogHandler(object):
                 than `max_length`.
 
         """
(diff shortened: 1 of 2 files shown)


Griatch:
Looks good but in the future, please make small fixes like this against the master branch, not against the develop branch. Bug fixes (which this is, in a way) should go into `master`.

maiki:
> Bug fixes should generally be done against the master branch of Evennia, while new features/contribs should go into the develop branch. **If you are unsure, just pick one and we'll figure it out.**

Cool, we used the method from the [docs](https://github.com/evennia/evennia/wiki/Contributing#contributing-through-a-forked-repository)!

I'll keep that in mind for the future. ^_^

PR #1592, opened 2018-03-17
Alpha webclient split interface support

friarzen:
Alpha webclient split interface support

To allow multiple output windows to persist information for different amounts of time.

To make a "default layout" of splits, a particular game can override the evennia webclient_ui.js as:
mygame/web/static_overrides/webclient/js/webclient_ui.js and add after

      SplitHandler.init();

something like:

      SplitHandler.dynamic_split( 'main', 'horizontal', 'left', 'right', 'append', 'append', [50,50] );
      SplitHandler.dynamic_split( 'right', 'vertical', 'look', 'sheet', 'replace', 'replace', [60,40] );
      SplitHandler.set_pane_types( 'look', ['look'] );  // this maps all text typed 'look' to the pane named 'look'.
      SplitHandler.set_pane_types( 'sheet', ['sheet'] );  // this maps all text typed 'sheet' to this pane.


Griatch:
@friarzen Haven't had time to look at this yet (also pinging @chainsol), but meanwhile, could you re-point this to the `develop` branch rather than to the `master` branch? [here](https://blog.github.com/2016-08-15-change-the-base-branch-of-a-pull-request/) you can see how to do so.

chainsol:
@Griatch I've got this running locally and it's very nice! I'm not sure what your thoughts are on this, and I'm not sure what @friarzen has in mind, but I have some things I'd like to see happen in the webclient eventually - whether that means @friarzen adds them in this PR, or someone else adds them once this is merged.

1. The current pop-up for help messages is coded into our javascript library. This could better be handled by [Bootstrap's Modals](https://getbootstrap.com/docs/4.0/components/modal/) which are much more accessible to those using screen readers. To me, this is very important - all of our custom pop-ups, including the options dialogues, could be better handled this way. The only thing we lose is the ability to drag the popup around.

2. I'd like to see a way to rename, or at least know the name, of a split, and especially I'd like a way for the developer to name each split themselves. This is doable in my original patch, but I'm not quite sure of what would have to change for this to work with @friarzen's patch.

3. Around the edge of each split there should probably be at least minimal padding - right now text goes right to the edge of the screen.

4. A way to handle mobile devices separately from desktop or tablet layouts would be much appreciated.

Overall, I'm blown away by how much @friarzen got done - everything looks and works just about perfectly!

friarzen:
I tend to believe in PR early, PR often, so that more people can be involved and help evolve the results.
Probably the simplest way to provide Item 3 would be to add a pair of text box input fields to the Split dialog box, then you'd just type in the two new pane names.   then pass those to dynamic_split() instead of split_N.

friarzen:
I have reworked the code so that the dialog box includes two text boxes to name each new pane and only use auto-generated pane names if the user leaves them blank.  I have also removed the special backwards-compatibility handling of the 'help' onText() type.  This breaks the old Help Popup (that still needs to be cleaned out).

friarzen:
Backing out a split is, I think, technically feasible, but it has a lot of edge cases.

What do we do with existing content? Is that right for all games? What about future content to the new merged window? Especially if one existing pane is appending and the other replacing.... Do we just give up and route new messages back to the default message window?  What about pane naming so that we don't break future re-splits... spawn another dialog to ask the player for a new name?

chainsol:
Yes, I had the same problem. It was one of the things I kept struggling with. In my opinion the best route probably is to either merge or clear the content, and same for the tags.

friarzen:
I haven't touched this since my last commit, everything on your "Notes for this PR" list other than merging splits was addressed.  I'll see if I can throw together a quick-n-dirty "back-out-the last-split" button, since I think that should be simple enough to do in a couple hours.

Merging arbitrary split panes is where it gets really hard, due to having to figure out what the resulting div stack should be.  I dont think I have the time/energy to implement that, at this point.

Griatch:
@friarzen Cool. Backing out of the last split is a good start. We can examine how to do merges going forward.

Griatch:
@friarzen I tested this again now. I like the cleaner look of the menus now. I still think the append/replace is confusing (and one should really see the names of the panes overlaid when trying to operate on them), but those are things we can iterate on, going forward. I had some issues with the current code though: 

- I don't see any undo button in the UI. I see it in the code/css. I fetched this latest branch anew and ran `evennia collectstatic` but nothing (tried on Firefox/Chrome). I also tried splitting a few times without seeing it. 
- Assigning the help content to anything does not appear to work. No matter which split I do and which pane I try to set the `help` content to, help text will appear in the original pane only (this functionality will need updating in the future I think; how do you un-assign content from a pane if you make a mistake?)

Griatch:
@friarzen Interesting. I fetched the branch anew and this time I *do* see the back-button. Curious, not sure what the difference was between now and this morning (pulled a new branch both times, ran `evennia collectstatic`). The undo seems to work at first glance. 

But the assignment is still not working for me. I tried:
- Create a new split (vertical or horizontal doesn't matter) off the 'main' window
- I left everything else to default to make the split
- I click the gear and Select one of the new splits and I check the 'help' content box. 
- Enter `help` on the command line
- No matter which split I try to assign to, the 'help' text only appear in the uppermost/leftmost split.

PR #1603, opened 2018-03-30
Add support for MUDINFO 1.1 query at login screen

Packetdancer:
Added a simple 'INFO' command which outputs a handful of statistics in MUDINFO 1.1 format. 

This will allow Evennia games (like Arx) to be added to mudconnector and mudstats, who insist on polling a game with MUDINFO to ensure the game is up/running/active.

### DIFF
--- a/evennia/commands/default/cmdset_unloggedin.py
+++ b/evennia/commands/default/cmdset_unloggedin.py
@@ -23,3 +23,4 @@ class UnloggedinCmdSet(CmdSet):
         self.add(unloggedin.CmdUnconnectedHelp())
         self.add(unloggedin.CmdUnconnectedEncoding())
         self.add(unloggedin.CmdUnconnectedScreenreader())
+        self.add(unloggedin.CmdUnconnectedInfo())
--- a/evennia/commands/default/tests.py
+++ b/evennia/commands/default/tests.py
@@ -14,15 +14,16 @@ main test suite started with
 
 import re
 import types
+import datetime
 
 from django.conf import settings
 from mock import Mock, mock
 
 from evennia.commands.default.cmdset_character import CharacterCmdSet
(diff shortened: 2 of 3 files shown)


Griatch:
@Packetdancer Hm, haven't heard of this format before, I thought these listing sites used MSSP. I have no issue adding it, but could you point me to a specification of the protocol? I couldn't find it on Mudstats after an (admittedly quick) search. 
Also, must each line be sent as one `msg()` or could we just send it as one line-break delimited string?

Packetdancer:
Sadly, I can no longer _find_ a copy of the MUDINFO specification; I copied it from the MUSH/MUX implementations thereof.  I am not 100% certain that mudconnector uses it, but last I checked, mudstats does use it (which is why it can get MUSH/MUX information to graph).  Either way, I know that despite repeated submissions, Arx doesn't show up on either; adding another method of getting the data in the INFO command seemed like a good way of hedging bets.

Connecting to a PennMUSH, MUX2, or TinyMUSH server and issuing the 'INFO' command at the login screen is a very quick way to test the standard output of the command, since I know for a fact all of those support it.  

(As for the output, it probably could be one single blob msg, in fairness, rather than multiple msg calls.  I can make that change easily enough.)

Griatch:
Ok, fair enough. Make it one msg-call and expand the docstring with the info you have there (including not finding the specification anymore but that it's necessary for using those sits and convenient to test con connection) and we can merge it.

Griatch:
Thanks!

BlauFeuer:
I think the more common and thorough protocol is:
Mud Server Status Protocol
http://tintin.sourceforge.net/mssp/
It looks like that's what mudstats.com uses

Packetdancer:
Agreed, but Evennia already has full MSSP support. This is just giving it support for an additional method beyond that.

PR #1604, opened 2018-04-01
Building menu

vlegoff:

This PR adds the `building_menu` contrib, to add building menus, a kind of `EvMenu` with building editing made through a menu tree.

In-line editing (through a command) is not always that simple for new builders.  Although I'd like to provide both editing modes in my own game, I think building menus could add an easier editing for these most unfamiliar with Evennia.


Griatch:
@vlegoff This looks really interesting! It won't collide (that much) with the OLC I'm making either since that is concerned solely with modifying prototypes (i.e. individual spawnable objects) whereas this 'building menu' seems to also concern itself with combining multiple objects (such as putting exits in a room). Very cool - I'll have to examine it (and run it) but it looks very promising at first glance.

evennia/contrib/building_menu.py
```diff
@@ -0,0 +1,980 @@
...
+
+    Note:
+        This function is supposed to be used as a default to
+        `BuildingMenu.add_choice`, when an attribute name is specified
+        (in the `attr` argument) but no function `on_nomatch` is defined.
+
+    """
+    attr = getattr(choice, "attr", None) if choice else None
+    if choice is None or string is None or attr is None or menu is None:
+        log_err("The `menu_setattr` function was called to set the attribute {} of object {} to {}, but the choice {} of menu {} or another information is missing.".format(attr, obj, repr(string), choice, menu))
```
Griatch: This is a _very_ long line, better to split it over multiple pages lines.

evennia/contrib/building_menu.py
```diff
@@ -0,0 +1,980 @@
...
+
+When you wish to create a building menu, you just need to import your
+class, create it specifying your intended caller and object to edit,
+then call `open`:
+
+```python
+from <wherever> import RoomBuildingMenu
+
+class CmdEdit(Command):
+
```
Griatch: This example needs a `key` for the Command. People will copy & paste this, you know! ;)

evennia/contrib/building_menu.py
```diff
@@ -0,0 +1,980 @@
...
+        if self.on_leave:
+            _call_or_get(self.on_leave, menu=self.menu, choice=self, string=string, caller=self.caller, obj=self.obj)
+
+
+class BuildingMenu(object):
+
+    """
+    Class allowing to create and set building menus to edit specific objects.
+
+    A building menu is a kind of `EvMenu` designed to edit objects by
```
Griatch: Should change this - it's not a 'kind of `EvMenu`' since it doesn't use `EvMenu` at all. It doesn't work similarly to `EvMenu` either I don't think.

vlegoff:
@griatch Thanks for the review.  I'm going to fix some issues you have pointed out in your feedback.  I have a few other questions and pointers for the others though, I would like to have your opinion about them:

1. The fact that you didn't get any text when entering "t" or "d" in your test is because the "text" argument was set to `None`, so that no text was displayed.  I wonder if it wouldn't be a good idea to provide a default text to, at least, know what you're editing.  It can be useful to move "quietly" in a menu tree however, but that doesn't prevent default text.  It could also give the `@` command to go back to the main menu.
2. Because changing the description is usually not as great in a simple menu, there's another method, `add_choice_edit` which is used to redirect to an `EvEditor` to edit the specified attribute.  Internally it still calls `add_choice` but does a couple of things just to open the editor and make sure the building menu is re-created when the editor closes.
3. Same thing, there's an `add_choice_quit` just to quit the menu.  It's not added by default, since, if you create a sub-menu, you might not want this option.  However it's worth considering, to have it by default (and allow to remove it if necessary).
4. In your example, to edit the description, you specify "desc" as attribute.  Actually, it would have worked with `attr="db.desc"`.  The building menu knows how to handle dots.  However, you are right, this is not aligned with the rest of Evennia and should perhaps be shifted.
5. I quite agree.  You can add the `CmdSet` several times which is not great.  Should I lock this in the `CmdSet` itself or in the command responsible for calling it?
6. Very good idea for demos and examples.  I should add that.  It will provide for added help in a way too.

Regarding the rest, I will definitely fix it step by step.  But I wanted to ask these questions first.

Thanks again,

Griatch:
1. I see. This was an easy mistake to make. Yes, I think a default error should be supplied, even if the error is "YOU MUST ADD YOUR TEXT". Getting no feedback is never a good thing and at least confused me. 
2. There is a reason EvMenu defaults to having useful helper functions like quit etc by default and instead allow the dev to override them. Same as here, we quickly found that devs saw missing defaults like this as bugs rather than things they forgot to add. 
3. I think I _did_ specify "desc" as the attribute? Nothing happened to the description when I did though.
4. You should handle this in the command that adds the cmdset - remove the same-keyed cmdset from the cmdset stack before adding it anew. Or make sure to clean it up as the menu context changes.

vlegoff:
Finally here!  I hope.

1. Building menus now have a default glance and default text, which is much more convenient.
2. Similarly, notifications of modifications in the building menu are announced as they occur.
3. I added `GenericBuildingMenu` and `GenericBuildingCmd` to edit any object, as a demonstration.  The documentation has been changed to reflect that.  The goal of the contrib can merely be to install this command, though it's just a demonstration of what building menus can do (a tutorial is to come when the contrib is approved, if it is ;) ).
4. Building menus cannot generate chains in cmdset, they protect themselves (see the `BuildingMenu.open` method).

Other modifications have been applied too, to reflect feedback.

Thanks!

Griatch:
As a side note, it's really unfortunate that this useful contrib is going the route of inventing its own menu system rather than using EvMenu; this introduces yet another system for the user to learn. But the PR looks a lot neater now and I'll get to reviewing it soon. :)

PR #1606, opened 2018-04-08
Add /contains switch to find.

Packetdancer:
Added a /contains switch to the find command, to allow searching with an icontains filter rather than an istartswith.

On a large game like Arx, it is _really_ hard to find objects sometimes, especially rooms, when you want to find "Hidden Room" and it's actually called "Arx - Ward of House Foobar - Smith Manor - Back Hallway - Hidden Room".  

Adding a contains switch makes it much easier to find such things, since you can merely "find/contains Hidden Room" and find the appropriate room that way.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -2270,11 +2270,12 @@ class CmdFind(COMMAND_DEFAULT_CLASS):
       @locate - this is a shorthand for using the /loc switch.
 
     Switches:
-      room - only look for rooms (location=None)
-      exit - only look for exits (destination!=None)
-      char - only look for characters (BASE_CHARACTER_TYPECLASS)
-      exact- only exact matches are returned.
-      loc  - display object location if exists and match has one result
+      room    - only look for rooms (location=None)
+      exit    - only look for exits (destination!=None)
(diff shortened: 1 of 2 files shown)


Packetdancer:
I would be fine with switching them; I honestly think given the find behavior on other servers, /contains makes sense as the default.

I just wanted to disrupt people's expected behavior as little as possible. :)

Packetdancer:
(I might make it /prefix rather than /startswith, though, just since it's a shorter name for a switch.)

Griatch:
I guess mileage may vary, but at least to me, `prefix` has a very different connotation than `startswith`; the former feels like one's adding something rather than searching from the beginning...

Packetdancer:
That's fair.

Griatch:
@Packetdancer Any time to wrap this up?

Packetdancer:
@Griatch — sorry about that.  Got caught up in day-job work.

Griatch:
@Packetdancer No worries, just checking up on the status :)

Griatch:
Looks good now, will run some tests then I'll merge.

Griatch:
Merged with a rebase. Thanks!

PR #1615, opened 2018-04-22
Fix regex in CommandTest breaking negative numbers

TehomCD:
I noticed that in a command I was testing the negative number was mysteriously converting to a positive number. Eventually I realized that the regex used by CommandTest was stripping out any hyphen, not just ones in a sequence to try to prevent EvTable stuff from breaking tests. I'm ridiculously bad at regex so hopefully this solution is actually correct and it doesn't break other stuff.

Allow negative numbers to appear in a CommandTest string

N/A


Griatch:
@TehomCD Hm, this fix appears to cause 10 errors in Travis ...

TehomCD:
I have dutifully added hyphens where once there were none. I feel I've made the world a better place today.

evennia/contrib/tests.py
```diff
@@ -688,8 +688,8 @@ def test_mail(self):
                   "You have received a new @mail from TestAccount2(account 2)|You sent your message.", caller=self.account2)
         self.call(mail.CmdMail(), "TestAccount=Message 1", "You sent your message.", caller=self.account2)
         self.call(mail.CmdMail(), "TestAccount=Message 2", "You sent your message.", caller=self.account2)
-        self.call(mail.CmdMail(), "", "| ID:   From:            Subject:", caller=self.account)
-        self.call(mail.CmdMail(), "2", "From: TestAccount2", caller=self.account)
+        self.call(mail.CmdMail(), "", "------------------------------------------------------------------------------| ID:   From:            Subject:", caller=self.account)
```
Griatch: It think this may be wrong. We *don't* want things like this to appear in the test suite. The way to correct this is rather to limit so that you don't allow more than two `-` in a row.

TehomCD: Hmm, good point. Any idea why `--+` isn't matching that in the regex, while `--*` did? My understanding is the former would match any string of two or more dashes, while the latter would be one or more, so I'm not quite getting why it doesn't match a strong series of them.

Griatch: It's actually surprising a string like `"------------------------------"` would not be stripped with your change, it seems from the reges this new form should catch and remove that string too.

TehomCD: Yeah, I just copied the regex into shell and was testing it against the test string, and `_RE.sub` is eating the test string, so I'm not sure what's going on there. How strange. I'll keep poking at it.

TehomCD:
Figured out why it was happening. The regex was matched before ansi is stripped out, and the regex does not match ansi characters mixed between them. So, as an example, the regex was checked against `|w-|n|w-|n|w-|n` repeatedly and failed to match the pattern, then the ansi was stripped afterwards and a pattern it should have matched was compared against the wanted string, raising the assert equals error.

Issue #1569, opened 2018-02-09
allcom displays objects from derived classes


arumford:
#### Brief summary of issue / Description of requested feature:
I created a derived Class from Channel called Frequency (a class used for object to object communication) and they are being shown in allcom.  I do not want these channels to be directly subscribable.

#### Steps to reproduce the issue / Reasons for adding feature:

1. Create Channel derived class
2. Create instance
3. type allcom

#### Error output / Expected result of feature
Shows derived channels

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
filter out anything that is not a true Channel object.

Henddher:
Hi @Griatch,

Is this another **good first issue** with **Hacktoberfest** label up for grabs?

I saw @ChrisLR's PR in closed state (never merged) and @TehomCD [comment](https://github.com/evennia/evennia/pull/1619#issuecomment-386765155). I have not seen the actual changes from @ChrisLR yet although it seems like there were 2 different philosophies.

What should be the goal here?

> 1. filter out anything that is not a true Channel object. 

IOW, subclasses would automatically be excluded?

2. attribute, property, or tag that prevents a channel from being included in `allcom`?

I personally like option 2.: using locks and/or tags to "hide" the channel.

ChrisLR:
Just want to say I never got around to do it after the first attempt. RL
happened :/

arumford:
I can shed light on this I think.  

From my perspective, just having the ability to optionally exclude derived classes or not is great.  This can allow a separation from standard Channels, which have pre-built commands.  Any mechanism is fine though.

Henddher:
Thank you @arumford!

I'll wait for @Griatch's advice on how to proceed. 

@ChrisLR sentiment is mutual (see my first **failed** pr attempt #589)

Griatch:
@Henddher Sorry for the very late reply to your query about this ticket, it fell between the chairs, so to speak. As part of looking through this, I implemented a change so that the default comm-related commands work on the default `Channel` entity. This seems to be the most common situation (child classes being used for very different things).

PR #1619, opened 2018-05-04
Hidden channels

ChrisLR:
Makes it possible to mark channels as hidden, which will hide them from commands using the get_all_channels() method
Those channels can still be accessed directly.

Can be used for system-only accessible channels

Fixes issue #1569
I know its not exactly what was suggested but I think this method has added value, such as being toggleable


ChrisLR:
I'll try that :)

PR #1621, opened 2018-05-10
Fix duplicate script calls

### LINKED ISSUE #1519: Duplicate script calls
Very rarely, I see the `at_repeat` for a script where there is only a single instance of it being called multiple times, usually with about a minute separation of it. 

1. It's happened very rarely, and only after long periods of usage without the server being reloaded.

`at_repeat` is called multiple times, say twice within a minute or two of the first call.

My guess is that a redundant `LoopingCall` is being created, but how or why is a complete mystery to me. It doesn't happen when the script is first created, because the script had run for days without interruption, and then we start to see the duplicate behavior, seemingly for no reason at all.

As a note, I did see this back in Evennia 0.6 as well, or at least I thought I did, with a different script. I noticed that a value which was supposed to be incremented at a script's `at_repeat` happened far, far more times than should have been possible. Since it was a timer, I just moved to datetime timedelta comparisons and thought no more of it. It was far more visible on the script I just saw because its `at_repeat` does a server broadcast, and I saw duplicate messages.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/51652751-duplicate-script-calls?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TehomCD:
This tries to fix duplicate script calls that can be caused when the idmapper flush is called on scripts without their `_task` being properly stopped.

Avoid `at_repeat` being called more than once on a script after a cache flush.

Fixes #1519

### DIFF
--- a/evennia/scripts/scripts.py
+++ b/evennia/scripts/scripts.py
@@ -18,6 +18,17 @@ from future.utils import with_metaclass
 __all__ = ["DefaultScript", "DoNothing", "Store"]
 
 
+FLUSHING_INSTANCES = False  # whether we're in the process of flushing scripts from the cache
+SCRIPT_FLUSH_TIMERS = {}  # stores timers for scripts that are currently being flushed
+
+
+def restart_scripts_after_flush():
+    """After instances are flushed, validate scripts so they're not dead for a long period of time"""
+    global FLUSHING_INSTANCES
+    ScriptDB.objects.validate()
+    FLUSHING_INSTANCES = False
+
+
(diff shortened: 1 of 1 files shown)


Griatch:
@TehomCD Awesome, I'll look over this some more, but this sounds like a good fix!

PR #1628, opened 2018-06-07
Contrib: Easy fillable form menu

FlutterSprite:
Adds a new contrib module, 'fieldfill.py', which presents a player with a form whose fields can be filled out in any order. Every field can have the sort of data it can accept tightly controlled, with built-in functionality to accept only text or numbers, with minimum / maximum lengths or values, and even the ability to define functions to verify and modify player input before it is stored. The form's data can be submitted to any callable, and can be used as a tool for builders, or as part of a player-facing mail or crafting system.

I've played a few MUDs that make heavy use of this sort of fillable form. Like my earlier 'tree select' contrib, this will allow for more easy and convenient use of EvMenu for this common sort of task.

If there is any functionality that seems core to this sort of system that I may be missing, I'm open to suggestions to add them in before merging!


evennia/contrib/fieldfill.py
```diff
@@ -0,0 +1,561 @@
...
+        return False
+    
+    # Returning anything besides True or False will replace the player's input with the returned value
+    # In this case, the value becomes a reference to the character object
+    # You can store data besides strings and integers in the 'formdata' dictionary this way!
+    return matched_character
+
+# Form template for the example 'delayed message' form
+SAMPLE_FORM = [
+{"fieldname":"Character", "fieldtype":"text", "max":30, "blankmsg":"(Name of an online player)",
```
Griatch: Just an idea, but for this type of form it may be a little easier to read by formatting it a little: 
```
{"fieldname": "Character", 
 "fieldtype": "text", 
 "max":30, "blankmsg": "(Name of an online player)",
 "required": True, 
 "verifyfunc": verify_online_player
},
{"fieldname": "Delay", 
 "fieldtype": "number", 
 "min": 3, 
 "max": 30, 
 "default": 10, 
 "cantclear": True
},
{"fieldname": "Message", 
 "fieldtype":"text", 
 "min":3, 
 "max":200, 
 "blankmsg":"(Message up to 200 characters)"
}]```

File: evennia/contrib/tests.py [hunk 2]
```diff
@@ -1283,6 +1283,24 @@ def test_tree_functions(self):
         {'goto': ['menunode_treeselect', {'newindex': 1}], 'key': ['<< Go Back', 'go back', 'back'], 'desc': 'Return to the previous menu.'}]
         self.assertTrue(tree_select.optlist_to_menuoptions(TREE_MENU_TESTSTR, test_optlist, 2, True, True) == optlist_to_menu_expected_result)
 
+# Test field fill
+
+from evennia.contrib import fieldfill
```
Griatch: You have a test-suite with examples already set up in your main module. It seems logical that you trigger this in a unit test and make sure it produces the result you expect. That way you should get good coverage of the components of your contrib.

evennia/contrib/fieldfill.py
```diff
@@ -0,0 +1,561 @@
...
+    "callback": callback,
+    "pretext": pretext,
+    "posttext": posttext,
+    "submitcmd": submitcmd,
+    "borderstyle": borderstyle,
+    "helptext": helptext
+    }
+    
+    # Initialize menu of selections
+    FieldEvMenu(caller, "evennia.contrib.fieldfill", startnode="menunode_fieldfill", **kwargs)
```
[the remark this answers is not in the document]
FlutterSprite: Not yet! I'll see about that when I go back to revisions. The only problem I foresee is having to specify db instead of ndb when storing and retrieving the menu data if the menu is specified as permanent.

evennia/contrib/fieldfill.py
FlutterSprite: Oh, that looks much nicer!

evennia/contrib/tests.py
FlutterSprite: I'll do my best! I may end up needing a bit of help calling and navigating the menu during a test.

evennia/contrib/fieldfill.py
```diff
@@ -0,0 +1,561 @@
...
+represents a field in the form, and contains the data for the field's name and
+behavior. For example, this basic form template will allow a player to fill out
+a brief character profile:
+
+    PROFILE_TEMPLATE = [
+    {"fieldname":"Name", "fieldtype":"text"},
+    {"fieldname":"Age", "fieldtype":"number"},
+    {"fieldname":"History", "fieldtype":"text"}
+    ]
+
```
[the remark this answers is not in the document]
FlutterSprite: The function that calls the form menu takes the data's target function as an argument - I'll make that clearer earlier.

FlutterSprite:
Something else I didn't think to add but I think is a fairly crucial feature, that I'll put in during my next round of changes - I should add the ability to provide pre-filled form data when you call the form menu for the first time. That would be really useful for situations in which you had already filled out a form before and need to make just a couple changes (like the profile example given in the documentation) or to dynamically pre-fill a form with data from the gameworld (like for an edit menu to change the properties of an object or room).

FlutterSprite:
Another thing I Just thought of - would it be worthwhile to give the option to let the player add or remove fields to the form? This would be really handy for stuff like more freeform profiles, or for making an object editor that lets builders add attributes to objects on the fly. It's a bit more niche, though, and I am a bit wary of feature creep...

Griatch:
@FlutterSprite Concerning the ability to add a field - I think it would be generally useful. But you decide where you think the boundaries of your contribs should be.

Issue #1305, opened 2017-04-14
Crash from File I/O Error on reload


TehomCD:
#### Brief summary of issue / Description of requested feature:
Earlier today, a staff member tried to `@reload` the server. Unlike the server hang seen in #1128 this didn't start the reload process at all, but immediately resulted in seeing repeated `... Portal has lost connection to Server.` messages until Evennia stop was called, with a repeated traceback in the console.

```
2017-04-14T10:08:52-0400 [AMPProtocol,0,127.0.0.1] Unhandled Error
Traceback (most recent call last):
  File "/home/tehom/arx/local/lib/python2.7/site-packages/twisted/python/log.py", line 101, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/home/tehom/arx/local/lib/python2.7/site-packages/twisted/python/log.py", line 84, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/home/tehom/arx/local/lib/python2.7/site-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/home/tehom/arx/local/lib/python2.7/site-packages/twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)
--- <exception caught here> ---
  File "/home/tehom/arx/local/lib/python2.7/site-packages/twisted/internet/posixbase.py", line 597, in _doReadOrWrite
    why = selectable.doRead()
  File "/home/tehom/arx/local/lib/python2.7/site-packages/twisted/internet/tcp.py", line 208, in doRead
    return self._dataReceived(data)
  File "/home/tehom/arx/local/lib/python2.7/site-packages/twisted/internet/tcp.py", line 214, in _dataReceived
    rval = self.protocol.dataReceived(data)
  File "/home/tehom/arx/local/lib/python2.7/site-packages/twisted/protocols/amp.py", line 2411, in dataReceived
    return Int16StringReceiver.dataReceived(self, data)
  File "/home/tehom/arx/local/lib/python2.7/site-packages/twisted/protocols/basic.py", line 765, in dataReceived
    self.stringReceived(packet)
  File "/home/tehom/arx/local/lib/python2.7/site-packages/twisted/protocols/basic.py", line 877, in stringReceived
    self.state = statehandler(string)
[...]

PR #1630, opened 2018-06-19
Move query/unpickling out of loop for mutelist.

TehomCD:
I was trying to figure out why there was so much lag in channels on peak times, and I realized that the mutelist property I created was being retrieved inside a loop. Whoops!

Reduce execution time for high-traffic channels by a large amount.

The amount of horsepower it needed for channels during peak hours could have been a contributing factor in #1305 - hopefully it'll be reduced a bit now.


Griatch:
Good catch. Merged!

PR #1633, opened 2018-06-30
initial support for webclient HotButtons

friarzen:
This adds a small extra split pane with 9 "hot buttons" on it that can send frequently used commands (like @who). This initial PR is intended more for starting the discussion around this functionality than a final form.

What Works:
- click an "unassigned" button to record the current command that has been typed into the input textarea.
- click to send that command to the server
- shift-click to clear the button back to "unassigned" status.

Many games have very-frequently-used commands that become a pain to type over-and-over on mobile devices and similar browser situations.

### DIFF
--- a/evennia/web/webclient/static/webclient/js/plugins/hotbuttons.js
+++ b/evennia/web/webclient/static/webclient/js/plugins/hotbuttons.js
@@ -1,0 +1,154 @@
+/*
+ *
+ * Assignable 'hot-buttons' Plugin
+ *
+ * This adds a bar of 9 buttons that can be shift-click assigned whatever is in the textinput buffer, so you can simply
+ * click the button again and have it execute those commands, instead of having to type it all out again and again.
+ *
+ * It stores these commands as server side options.
+ *
+ * NOTE:  This is a CONTRIB.  To use this in your game:
+ *
+ *     Stop Evennia
+ *
+ *     Copy this file to mygame/web/static_overrides/webclient/js/plugins/hotbuttons.js
+ *     Copy evennia/web/webclient/templates/webclient/base.html to mygame/web/template_overrides/webclient/base.html
+ *
+ *     Edit  mygame/web/template_overrides/webclient/base.html to add:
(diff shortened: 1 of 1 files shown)


friarzen:
Added server side persistence using webclient_options inputFunc

Griatch:
This sounds really useful, just need to set up my dev machine to test it properly.

Griatch:
@friarzen While I like this concept of this, I don't think this GUI solution is really suitable for a newcomers'  first impression of the client. At the very least, the row of initailized buttons should be optional (start out hidden, set in options). But going with the overall theme of the editor it makes even more sense to let the button-menu be a special pane-'type', that a user can select when splitting it off. 
Not sure how much work those extra features would be at this point but this can't merge as-is.

busbyra:
@friarzen I think this could be a great idea.  To meet in the middle between you an @Griatch, maybe have it optional, but make sure the user is aware of the option.  There are a lot of clients out there that have cool little options that a general user would never think to look for, so simply just hiding it would defeat the purpose.

Anyhow, I'm all for buttons or things that further pull us out of the 1980s/90s!

friarzen:
I am decidedly NOT married to this current implementation, in fact, after poking around the webclient code some more I think I have come up with a more modular, "plugin"-style that will make turning this into an 'option controlled' feature reasonable.  I'll work on it this week and get back to you.

evennia/web/webclient/static/webclient/js/plugins/hotbuttons.js
```diff
@@ -0,0 +1,141 @@
+/*
+ *
+ * Assignable 'hot-buttons' Plugin
+ *
+ * This adds a bar of 9 buttons that can be shift-click assigned whatever is in the textinput buffer, so you can simply
+ * click the button again and have it execute those commands, instead of having to type it all out again and again.
+ *
```
Griatch: Since this is in effect an inactive "contrib", I think this intro needs a mention of 

- The fact that this is disabled by default
- How to practically activate it in the webclient

friarzen:
Install procedure added.

Issue #1637, opened 2018-07-23
Docker: Additional depends. needed for Telnet TLS


robotfactory:
#### Brief summary of issue / Description of requested feature:
The Dockerfile doesn't include the py2-openssl package and a few python dependencies for running TLS on telent.

#### Steps to reproduce the issue / Reasons for adding feature:

1. Probably a good thing to have out of the box
2. Start a docker container with TLS on Telnet enabled, get errors about missing depends

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
1. Recommend adding py2-openssl package to the RUN apk add... line
2. Recommend adding additional RUN line to install depnds:
RUN pip install cryptography pyasn1 service_identity

This creates an additional layer but it shouldn't be very big.

robotfactory:
Submitted pull request with proposed changes.

robotfactory:
On second thought, it's probably much better to include the depends in requirements.txt.

bookofportals:
This can be closed now, right?

robotfactory:
I think so unless the PR caused issues for anyone else.

PR #1638, opened 2018-07-23
Python depends for telnet tls

robotfactory:

Allows for tls on telent out of the box when starting up a docker container

Addresses feature request in issue #1637

Installs py2-openssl as well as cryptography, pyasn1, and service_identity

Allows for tls on telent out of the box when starting up a docker container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,13 +24,14 @@ FROM alpine
 MAINTAINER www.evennia.com
 
 # install compilation environment
-RUN apk update && apk add python py-pip python-dev py-setuptools gcc musl-dev jpeg-dev zlib-dev bash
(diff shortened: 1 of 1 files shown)


Griatch:
The reason for not having the pycrypto dependency in the requirement file is because not everyone wants or care about Telnet encryption (it's not active by default) so this is an optional install. Also, using the crypto required/requires some additional setup that was not something we wanted to incur on new users. 

The dockerfile is a different thing in that respect - there it becomes a hindrance to not have the packages available. 

We'll need to rework the dockerfiles for the develop branch of Evennia anyway, so this has not been a priority. But it sounds reasonable to me to merge this for master for now.

PR #1640, opened 2018-07-30
Change callability check

bookofportals:
Changes an `except TypeError ` to a `callable()` check in the `_instantiate()` function of `cmdset.py`.

#### Motivation for adding to Evening
`TypeError`s can be thrown for many different reasons; expecting them to always come from calling an uncallable object should be considered a bug. Python has a built-in function, `callable()`, to test whether an object can be invoked. If a class's `__init__` method unexpectedly throws a `TypeError`, an extremely cryptic exception about unbound methods is raised from `cmdparser.py`, which is, to say the least, counterintuitive.


Griatch:
I like that, good change!

PR #1645, opened 2018-08-14
Update to fix #1644

### LINKED ISSUE #1644: copy_object copies attributes, cmdsets and scripts - but not tags
Using copy_object from [https://github.com/evennia/evennia/wiki/evennia.objects.manager](https://github.com/evennia/evennia/wiki/evennia.objects.manager)

> copy_object(self, original_object, new_key=None, new_location=None, \
>     new_home=None, new_permissions=None, new_locks=None, \
>     new_aliases=None, new_destination=None)
> 
>     Create and return a new object as a copy of the original object. All will be identical to the original except for the arguments given specifically to this method.
> 

Original object has _tags_. Copy created does _not_ .

1. Run this batchcode:

```


#
    
#HEADER
    
# This will be included in all other #CODE blocks
    
from evennia.utils import create, search
from typeclasses.objects import Object
from evennia.objects.models import ObjectDB
    
    
#CODE 

item = create.create_object('typeclasses.objects.Object', key='item')

item.tags.add("some_tag")
item.db.someattr = "someattr"

newitem = ObjectDB.objects.copy_object(item, new_key="newitem")

caller.msg("\n" + "item attribute: " + str(item.db.someattr) + "; item tags: " + str(item.tags.get()) + "\n")
caller.msg("\n" + "new item attribute: " + str(newitem.db.someattr) + " ; new item tags: " + str(newitem.tags.get()) + "\n")
```

@batchcode test
Running Batch-code processor - Automatic mode for test ...01/02:
02/02: item = create.create_object('typeclasses.objects.Object[...]

item attribute: someattr ; new item tags: some_tag

new item attribute: someattr ; new item tags: None
End of batch file.  Batchfile 'test' applied.

Latest stable, running on a Raspberry Pi 3, OS is raspbian stretch, python 2.7.13

I believe copying tags was/should be the intended behaviour.

If copy_objects is reviewed, I am less certain of what should happen to object.contents.

- Should each object in contents be duplicated, recursively?
- If no, should there be an optional parameter to switch this behaviour on?
- If neither, should there be a comment in the source to that effect?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62108751-copy_object-copies-attributes-cmdsets-and-scripts-but-not-tags?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github).
</bountysource-plugin>

FatherGrishnak:
Added code to copy tags

Fix oversight #1644

### DIFF
--- a/evennia/objects/manager.py
+++ b/evennia/objects/manager.py
@@ -503,9 +503,9 @@ class ObjectDBManager(TypedObjectManager):
             ScriptDB.objects.copy_script(script, new_obj=new_object)
 
         # copy over all tags, if any
-        for tag in original_object.tags.get():
-            new_object.tags.add(tag)
-
+        for tag in original_object.tags.get(return_tagobj=True, return_list=True):
+            new_object.tags.add(tag=tag.key, category=tag.category, data=tag.data)
+    
         return new_object
 
     def clear_all_sessids(self):
--- a/evennia/objects/manager.py
+++ b/evennia/objects/manager.py
@@ -437,7 +437,7 @@ class ObjectDBManager(TypedObjectManager):
         """
         Create and return a new object as a copy of the original object. All
         will be identical to the original except for the arguments given
-        specifically to this method.
+        specifically to this method. Object contents will not be copied.
 
         Args:
             original_object (Object): The object to make a copy from.
@@ -501,6 +501,10 @@ class ObjectDBManager(TypedObjectManager):
         # copy over all scripts, if any
         for script in original_object.scripts.all():
             ScriptDB.objects.copy_script(script, new_obj=new_object)
+
+        # copy over all tags, if any
+        for tag in original_object.tags.get():
+            new_object.tags.add(tag)
 
         return new_object
 


evennia/objects/manager.py
```diff
@@ -502,6 +502,10 @@ def copy_object(self, original_object, new_key=None,
         for script in original_object.scripts.all():
             ScriptDB.objects.copy_script(script, new_obj=new_object)
 
+        # copy over all tags, if any
+        for tag in original_object.tags.get():
```
Griatch: Now that I can comment correctly, I should note that this will be a list of tag names (`key`). That is, if the tags to copy have `category` or `data` fields, those will be lost in the copy. There is also a chance your loop fails if there is no- or one tag found. I think the way to get around this would be to get the `Tag` instances themselves instead: 

```python
for tag in original_object.tags.get(return_tagobj=True, return_list=True):
    new_object.tags.add(tag=tag.key, category=tag.category, data=tag.data)
```

Griatch: Also note that the test suite fails with this change, that needs to be fixed before this can merge.

FatherGrishnak:
Changes made, thanks

PR #1649, opened 2018-08-27
Plugin-ify the webclient

friarzen:

This is a pretty extensive re-organization of the Webclient to allow easier functional updates and extensions by compartmentalizing existing functionality into 'plugins,' self-contained units of code that can be added with a single <script> line in the base.html.

The current implementation makes it hard to swap out components like the split.js interface or provide new functionality such as my hotbuttons branch.  This provides a more flexible base.

Known issue: pulling out the splithandler exposes a CSS problem in the base gui where the scrollbars do not work and then eventually the input field scrolls off the bottom of the screen.

### DIFF
--- a/evennia/web/website/urls.py
+++ b/evennia/web/website/urls.py
@@ -18,7 +18,7 @@ urlpatterns = [
     url(r'^tbi/', website_views.to_be_implemented, name='to_be_implemented'),
 
     # User Authentication (makes login/logout url names available)
-    url(r'^authenticate', include('django.contrib.auth.urls')),
+    url(r'^authenticate/', include('django.contrib.auth.urls')),
 
     # Django original admin page. Make this URL is always available, whether
     # we've chosen to use Evennia's custom admin or not.


evennia/web/webclient/static/webclient/js/default_io.js
```diff
@@ -0,0 +1,41 @@
+/*
+ *
+ * Evennia Webclient default 'send-text-on-enter-key' IO plugin
+ *
+ */
+plugin_handler.add('defaultio', (function () {
```
Griatch: Style peeve that is probably more of a personal thing than anything: I know JS code loves to do these nested function calls, but they lead to this bunch of parentheses and curly brackets at the bottom which is just hard to parse and check. Would it not be more readable to define the function in a variable first and then do the `plugin_handler.add()` call on its own near the bottom? Not necessary to change, but maybe worth considering?

evennia/web/webclient/static/webclient/js/default_io.js
```diff
@@ -0,0 +1,41 @@
...
+            });
+
+        console.log('DefaultIO initialized');
+    }
+
+    return {
+        init: init,
+        onKeydown: onKeydown,
+    }
+})());
```
Griatch: Seriously, JS ...

evennia/web/webclient/static/webclient/js/webclient_gui.js
```diff
@@ -12,618 +12,258 @@
...
-function onKeydown (event) {
-    var code = event.which;
-    var history_entry = null;
-    var inputfield = $("#inputfield");
-    if (code === 9) {
-      return;
+    // Ask if user really wants to exit session when closing
+    // the tab or reloading the page. Note: the message is not shown
+    // in Firefox, there it's a standard error.
+    var onBeforeUnload = function () {
```
Griatch: Could this not be refactored as a (very simple) plugin as well? Like a hook?

evennia/web/webclient/static/webclient/js/webclient_gui.js
```diff
@@ -12,618 +12,258 @@
...
+    // Evennia Public Event Handlers
+    //
 
-    if (code === 27) { // Escape key
-        if ($('#helpdialog').is(':visible')) {
-          closePopup("#helpdialog");
-        } else {
-          closePopup("#optionsdialog");
-        }
+    // Handle onLoggedIn from the server
```
Griatch: I think a summary of these handled events and what they expect the plugin to provide should be in the top header of this file.

evennia/web/webclient/static/webclient/js/webclient_gui.js
```diff
@@ -12,618 +12,258 @@
...
+                if( plugins[plugin].onText(args, kwargs) ) {
+                    // True -- means this plugin claims this text.
+                    // It should not be appended to #messagewindow
+                    // Note: multiple plugins can all claim this text.
                     use_default_pane = false;
                 }
             }
+        });
+
+        // append message to default pane, then scroll so latest is at the bottom.
```
Griatch: This seems to break with the plugin separation used elsewhere in this code. Would it not make more sense to provide this handling as a default plugin-implementation rather than to hard-code it here?

evennia/web/webclient/static/webclient/js/webclient_gui.js
```diff
@@ -12,618 +12,258 @@
...
-        .addClass("out")
-        .html(args[0]);
-    doWindowResize();
-
-    // also display the prompt in the output window if gagging is disabled
-    if (("gagprompt" in options) && (!options["gagprompt"])) {
-        onText(args, kwargs);
+    // Handle prompt output from the server
+    var onPrompt = function (args, kwargs) {
+        var use_default_prompt = true;
```
Griatch: Same here, should not the default_prompt just be a default-provided plugin?

evennia/web/webclient/static/webclient/js/webclient_gui.js
```diff
@@ -12,618 +12,258 @@
...
-    onText(["The connection was closed or lost."], {'cls': 'err'});
-}
-
-// Handle unrecognized commands from server
-function onDefault(cmdname, args, kwargs) {
-    var mwin = $("#messagewindow");
-    mwin.append(
+
+
+    // Handle unrecognized commands from server
```
Griatch: One could argue that even unhandled command-handling should be changeable.

evennia/web/webclient/static/webclient/js/webclient_gui.js
```diff
@@ -12,618 +12,258 @@
  *
  */
```
Griatch: This docstring needs to be updated to indicate that this is now basically a plugin-handler. It also needs a summary of how a plugin must look to work and how to add it. Finally now that this is made to be more replaceable by users it might be a good time to make a listing of the supported Events and what each event expects from a plugin that uses it.

Griatch:
@friarzen Any updates to this?

PR #1651, opened 2018-08-30
Implements user input and server output auditing.

strikaco:

Adds a modified ServerSession that passes dicts containing the raw string and metadata for all strings inbound and/or outbound to a user-defined method specified in settings. Object references to the account, character, room and session are appended for the sake of those who wish to manipulate any of them directly.

Adds previously requested docstrings and scrubs default commands for password leakage.

## Motivation for adding to Evennia

More enhanced logging/auditing is a common requirement for server administrators, though how it should be implemented has been a controversial issue. This implementation is designed to avoid dictation of format, storage method or content and be more in line with project principles.

## Other info (issues closed, discussion etc)

Supercedes #1390 and #1535. Git is too hard for me; sorry for all the duplicate PRs.

Tested on Linux but not on Windows; some of the metadata may not generate properly (PIDs, hostname?).

### DIFF
--- a/evennia/prototypes/menus.py
+++ b/evennia/prototypes/menus.py
@@ -2027,14 +2027,14 @@ def node_apply_diff(caller, **kwargs):
                         sep=" |r->|n ", new='', change=inst))
             options.append(_keep_option(key, prototype,
                            base_obj, obj_prototype, diff, update_objects, back_node))
-        options.extend(
-            [{"key": ("|wu|Wpdate {} objects".format(len(update_objects)), "update", "u"),
-              "desc": "Update {} objects".format(len(update_objects)),
-              "goto": (_apply_diff, {"prototype": prototype, "objects": update_objects,
-                  "back_node": back_node, "diff": diff, "base_obj": base_obj})},
-             {"key": ("|wr|Weset changes", "reset", "r"),
-              "goto": ("node_apply_diff", {"prototype": prototype, "back_node": back_node,
-                                           "objects": update_objects})}])
+    options.extend(
+        [{"key": ("|wu|Wpdate {} objects".format(len(update_objects)), "update", "u"),
+          "desc": "Update {} objects".format(len(update_objects)),
+          "goto": (_apply_diff, {"prototype": prototype, "objects": update_objects,
+              "back_node": back_node, "diff": diff, "base_obj": base_obj})},
+         {"key": ("|wr|Weset changes", "reset", "r"),
+          "goto": ("node_apply_diff", {"prototype": prototype, "back_node": back_node,
+                                       "objects": update_objects})}])
 
     if ichanges < 1:
         text = ["Analyzed a random sample object (out of {}) - "
--- a/evennia/utils/evmenu.py
+++ b/evennia/utils/evmenu.py
@@ -866,19 +866,34 @@ class EvMenu(object):
         props = {prop: value for prop, value in all_props if prop not in all_methods and
                  prop not in all_builtins and not prop.endswith("__")}
 
+        local = {key: var for key, var in locals().items()
+                 if key not in all_props and not key.endswith("__")}
+
         if arg:
             if arg in props:
                 debugtxt = " |y* {}:|n\n{}".format(arg, props[arg])
+            elif arg in local:
+                debugtxt = " |y* {}:|n\n{}".format(arg, local[arg])
             elif arg == 'full':
-                debugtxt = ("|yMENU DEBUG full ... |n\n" + "\n".join(props) +
+                debugtxt = ("|yMENU DEBUG full ... |n\n" + "\n".join(
+                                "|y *|n {}: {}".format(key, val)
+                                for key, val in sorted(props.items())) +
+                            "\n |yLOCAL VARS:|n\n" + "\n".join(
+                                "|y *|n {}: {}".format(key, val)
+                                for key, val in sorted(local.items())) +
                             "\n |y... END MENU DEBUG|n")
             else:
                 debugtxt = "|yUsage: menudebug full|<name of property>|n"
         else:
-            debugtxt = "|yMENU DEBUG properties:|n\n" + "\n".join("|y *|n {}: {}".format(
-                            prop, crop(to_str(value, force_string=True), width=50))
-                                for prop, value in sorted(props.items()))
-            debugtxt += "\n|y... END MENU DEBUG (use menudebug <name> for full value)|n"
+                debugtxt = ("|yMENU DEBUG properties ... |n\n" + "\n".join(
+                                "|y *|n {}: {}".format(
+                                    key, crop(to_str(val, force_string=True), width=50))
+                                for key, val in sorted(props.items())) +
+                            "\n |yLOCAL VARS:|n\n" + "\n".join(
+                                "|y *|n {}: {}".format(
+                                    key, crop(to_str(val, force_string=True), width=50))
+                                for key, val in sorted(local.items())) +
+                            "\n |y... END MENU DEBUG|n")
         self.caller.msg(debugtxt)
 
     def parse_input(self, raw_string):


evennia/contrib/auditing/example.py
```diff
@@ -0,0 +1,21 @@
+from evennia.utils.logger import log_file
```
Griatch: this file should probably not be named `example` since it actually implements a useful logging-type that many may use out of the box. Probably better to go with `default_logger.py` or somesuch.

evennia/contrib/auditing/example.py
```diff
@@ -0,0 +1,21 @@
...
+import json
+
+def output(data, *args, **kwargs):
```
Griatch: Call this `default_logger` or maybe `daily_logger` ?

evennia/contrib/auditing/server.py
```diff
@@ -0,0 +1,266 @@
...
+        AUDIT_CALLBACK = mod_import(AUDIT_CALLBACK).output
+        logger.log_info("Auditing module online.")
+        logger.log_info("Recording user input = %s." % AUDIT_IN)
+        logger.log_info("Recording server output = %s." % AUDIT_OUT)
+    except Exception as e:
+        logger.log_err("Failed to activate Auditing module. %s" % e)
+
+class AuditedServerSession(ServerSession):
+    """
+    This class represents a player's session and is a template for
```
Griatch: Good documentation, but the bit about portal/server sessions may be better under a `Notes:` section at the end. For those that know what a Session is, it's better to start documenting the bits unique to this implementation (the sections about the auditing etc).

evennia/contrib/auditing/server.py
```diff
@@ -0,0 +1,266 @@
...
+    AUDIT_MASK_KEEP_BIGRAM = ['create', 'connect', '@userpassword']
+    """
+    def audit(self, **kwargs):
+        """
+        Extracts messages and system data from a Session object upon message 
+        send or receive.
+    
+        Kwargs:
+            src (str): Source of data; 'client' or 'server'. Indicates direction.
+            text (list): Message sent from client to server.
```
Griatch: Write this as `text (str or list): ... ` and then detail that the type depends on where the data comes from. 

It's also very important to note in the doc that this implementation only handles the `text` input/outputfunc but that **kwargs can well include other types of data.

evennia/contrib/auditing/server.py
```diff
@@ -0,0 +1,266 @@
...
+                data = str(kwargs['text'][0][0])
+            except IndexError:
+                logger.log_err('Failed to parse client-submitted string!')
+                return False
+                
+        elif src == 'server':
+            # Server outputs can be unpredictable-- sometimes tuples, sometimes
+            # plain strings. Try to parse both.
+            try: 
+                if isinstance(kwargs.get('text', ''), (tuple,list)):
```
Griatch: For future-proofing, I would suggest using `evennia.utils.utils.is_iter` here instead of looking for list/tuple types. `is_iter` will find all iterable structures but not strings. That way you'll catch other types of iterables for free (including custom ones like `_SaverDict` etc).

evennia/contrib/auditing/server.py
```diff
@@ -0,0 +1,266 @@
...
+                    data = kwargs['text'][0]
+                elif not 'text' in kwargs and len(kwargs.keys()) == 1:
+                    data = kwargs.keys()[0]
+                else:
+                    data = str(kwargs['text'])
+                    
+            except: data = str(kwargs)
+            
+        bytes = len(data.encode('utf-8'))
```
Griatch: `bytes` is a reserved keyword in Python. It's not a good idea to replace it with a same-named local variable. Especially once we start moving towards python3.

evennia/contrib/auditing/server.py
```diff
@@ -0,0 +1,266 @@
...
+
+        return log
+        
+    def mask(self, msg, **kwargs):
+        """
+        Masks potentially sensitive user information within messages before
+        writing to log. Recording cleartext password attempts is bad policy.
+    
+        Args:
+            msg (str): Raw text string sent from client <-> server
```
Griatch: The `**kwargs` is passed on from `audit` so need to mentioned here too.

evennia/contrib/auditing/server.py
```diff
@@ -0,0 +1,266 @@
...
+    
+        Returns:
+            msg (str): Text string with sensitive information masked out.
+
+        """
+        # Get command based on fuzzy match
+        command = next((x for x in re.findall('^(?:Command\s\')*[\s]*([create]{5,6}|[connect]{6,7}|[@userpassword]{6,13}).*', msg, flags=re.IGNORECASE)), None)
```
Griatch: This appears to be a hard-coded regex assuming the name of the create/connect/userpassword commands. Not only is it not sure a given gave name them such,  I thought these could be customized in the settings?

Griatch: I don't really understand the use of `next()` here. Could you elaborate a bit what the intention is? To get the first match? Note also that for creating an iterator, `re.finditer` can be used.

evennia/contrib/auditing/server.py
```diff
@@ -0,0 +1,266 @@
...
+        if not command or command in AUDIT_MASK_IGNORE:
+            return msg
+            
+        # Break msg into terms
```
Griatch: As mentioned, I feel you can probably skip the manual handling of terms in favor of applying a regex for finding and masking the parts that should be removed.

evennia/contrib/auditing/server.py
```diff
@@ -0,0 +1,266 @@
...
+        
+        # If the first term was typed correctly, grab the appropriate number
+        # of subsequent terms and mask the remainder
+        if command in AUDIT_MASK_KEEP_BIGRAM and num_terms >= 3:
+            terms = terms[:2] + ['*' * sum([len(x.zfill(8)) for x in terms[2:]])]
+        else:
+            # If the first term was not typed correctly, doesn't have the right
+            # number of terms or is clearly password-related,
+            # only grab the first term (minimizes chances of capturing passwords
+            # conjoined with username i.e. 'conect johnnypassword1234!').
```
Griatch: while a good idea, the problem here is that it appears to make big assumptions about the command syntax without allowing the user to customize it (see previous comments).

evennia/contrib/auditing/server.py
```diff
@@ -0,0 +1,266 @@
...
+        elif src == 'server':
+            # Server outputs can be unpredictable-- sometimes tuples, sometimes
+            # plain strings. Try to parse both.
+            try: 
+                if isinstance(kwargs.get('text', ''), (tuple,list)):
+                    data = kwargs['text'][0]
+                elif not 'text' in kwargs and len(kwargs.keys()) == 1:
+                    data = kwargs.keys()[0]
+                else:
+                    data = str(kwargs['text'])
```
Griatch: This will not be correct for different types of input/outputfuncs. It's perfectly possible for the system to receive (for example) `msg(text="Hello", prompt="hp=20|st=10|ma=15", pane=2)` which will translate to more keys in `**kwargs`. For OOB messages it's also possible to have no `text` key at all (since nothing should be displayed in the main window). In that case this line will lead to a traceback.

strikaco: For the sake of unpredictability, I think I will just cast kwargs to string and call it a day. Capture of server output was intended to be more of a diagnostic tool for admins anyway.

Griatch: Just casting raw kwargs will likely make things quite hard to read. It probably makes sense to identify the different input/outputfuncs when they are different to `text`. It's in fact a good candidate for a filter, to specify which input/outputfuncname to include/exclude in the logging. Things like regularly updating client-side OOB health-bar messages could quickly get very, very spammy.

evennia/contrib/auditing/server.py
strikaco: Yes, the next() call was supposed to get the first match. Nice to know about finditer but I don't think I need this block at all anymore :)

evennia/contrib/auditing/server.py
strikaco: Argh...good point. You are not making this easy on me :)

evennia/contrib/auditing/server.py
strikaco: How do I document arbitrary pass-through kwargs? This function only concerns itself with `src` and `text`, but I do want to allow anything else provided by a dev to flow through to the logger (since `text` may not be what they want to capture.)

evennia/contrib/auditing/server.py
strikaco: ...and after trying things that way, it becomes much more difficult to reliably scrub commands for credentials.

My current compromise is to pop `text` from data when it's available, scrub it, and push it to the template as `text`. All other kwargs provided simply get passed through as `data` in the final object. I think this should accommodate everybody.

PR #1652, opened 2018-08-31
Urlfix - /authenticatelogin to /authenticate/login

strikaco:
Evennia has always had a confusing typoed path for login authentication at `/authenticatelogin`. This PR corrects this to the proper `/authenticate/login` path that is more consistent with most Django projects.

Spelling counts!

### DIFF
--- a/evennia/prototypes/menus.py
+++ b/evennia/prototypes/menus.py
@@ -2027,14 +2027,14 @@ def node_apply_diff(caller, **kwargs):
                         sep=" |r->|n ", new='', change=inst))
             options.append(_keep_option(key, prototype,
                            base_obj, obj_prototype, diff, update_objects, back_node))
-        options.extend(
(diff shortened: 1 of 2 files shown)


strikaco:
Ugh I don't know why it pulled in my previous PRs; I set up a new branch specifically for this fix. :\

Griatch:
Heh, I never noticed that lack of `/` ! 

I think the problem with the commits is that you made your original PR-commits directly in your copy of the develop branch. Now you are making a new branch based on that branch, meaning the new branch includes those commits from that previous PR. You should always make PRs based of local branches, that way they become independent of each other (especially necessary if upstream changes anything before merging - your version of `develop` branch will then be different from upstream). To fix this for this PR, easiest may be to go back to your develop branch and then revert back to a commit-point before you did any changes create a new branch from that point (so it doesn't have any of your PRs). Then you `git cherry-pick` the commits you want into your new branch, Finally you can delete that commit in your `develop` branch.

PR #1661, opened 2018-09-21
Adds a logger for security-related events.

strikaco:
Adds a logger to the existing collection that classifies log entries as security events (prefix "[SS]").

#1383 and others have highlighted the need for better logging of security-related events, but we do not currently have a commonly accepted destination to send them to. This PR aims to define one.


strikaco:
A change to this is forthcoming, please await an additional commit.

strikaco:
@Griatch Nevermind; this is all it needs to be.

PR #1664, opened 2018-09-26
Pathfinder

strikaco:
Implements a Pathfinder class (no relation to the tabletop game).

* Upon initialization, the Pathfinder will build a directional graph of all of your game's Rooms and Exits. 
* It can then be queried to get explicit directions navigating you (or a mob AI) between any two Rooms using the shortest path available ("to get from Limbo to Downtown, go `['down', 'east', 'down', 'north', 'down']`"). Useful for enabling Terminator or enforcer mobs to hunt anyone down anywhere in the game.
* It can also be queried to get a list of unobstructed Room objects visible from a given location in any supported direction (line-of-sight). Useful for long-range observation/sniper rifle mechanics.
* Also comes with a script that loads and periodically refreshes the graph if deployed that way.

All games eventually need pathfinding :)

Requires third-party `networkx` library.


evennia/contrib/pathfinding/pathfinder.py
```diff
@@ -0,0 +1,212 @@
...
+        
+        # Get the edge attributes for each hop
+        edge_ids = zip(path[0:], path[1:])
+        for src, dst in edge_ids:
+            direction = self._graph[src][dst]['direction']
+            steps.append(direction)
+        
+        return steps
+        
+    def get_line_of_sight(self, source, direction, ordered=True):
```
Griatch: This appears to get all rooms in a line. Would not a `max_distance` argument be a useful and practical optimization for this method?

evennia/contrib/pathfinding/pathfinder.py
```diff
@@ -0,0 +1,212 @@
+from evennia import DefaultRoom as Room
```
Griatch: Please add a short module docstring giving and overview of  the functionality in this module.

evennia/contrib/pathfinding/scripts.py
```diff
@@ -0,0 +1,29 @@
+from evennia import DefaultScript
```
Griatch: Please add docstring.

evennia/contrib/pathfinding/pathfinder.py
strikaco: Yeah, that'll break it. Nice catch :)

A distance arg would be useful too.

strikaco:
I don't know how well it would scale, but the thought occurs to me that if I extend this to graph all objects (not just rooms/exits), the fact that we can serialize/deserialize graphs as JSON and store attributes on edges and nodes means this might be an easier way of batch-creating entire worlds.

Pulling that off is way too ambitious for this PR but I'd be interested in hearing anybody's thoughts.

FlutterSprite:
Really excited to see a pathfinder contrib!! Just a quick question - if there are multiple disconnected 'maps' of rooms that aren't traversable between each other via exits, how does the charter handle it? Say, for example, a map with two different islands, or multiple planes of existence that can't be traversed just by walking? Does the map-charter still cover all rooms that way? If so, what happens if you try to find a path between two rooms that aren't connected?

strikaco:
> 
> 
> Really excited to see a pathfinder contrib!! Just a quick question - if there are multiple disconnected 'maps' of rooms that aren't traversable between each other via exits, how does the charter handle it? Say, for example, a map with two different islands, or multiple planes of existence that can't be traversed just by walking? Does the map-charter still cover all rooms that way? If so, what happens if you try to find a path between two rooms that aren't connected?

You people and your multiple planes :P

It's tricky. As it is, each island/plane would get its own local geography graphed, but if they aren't joined by a proper Exit of any sort then there's no logical way for me to anticipate the connection.

Now, there is possibly a way around this, but I'm hesitant to go too far down the path of bespoke solutions with this contrib. What would have to be done in your case is to create a "universe" graph, then create discrete graphs for each isolated plane. Each plane would route its own traffic internally (like your home network). The Universe graph would route traffic between each plane (like your ISP and the higher ASNs). This is actually a better solution in terms of overall scalability and memory management but there is no native Evennia concept of a "Plane" or how to move between them, only "Rooms" and "Exits"-- so I don't know what I'm building against when dealing with your idea of a Plane versus someone else's.

You've made me think though-- conditional movement isn't something I anticipated; I can give you directions from A to B, but I can't predict that you won't encounter a locked door or chasm that is technically passable but only with climbing tools and a sherpa, or an exit between planes that requires a handheld teleporter. (The obvious failure mode is to try multiple paths, but this leads to mobs/NPCs running around endlessly like idiots until they starve or get beaten to death). One solution would be to store the permission string for each edge as an attribute on itself, then evaluate each generated path within the context of the calling object's permission set. That's a bit more involved than I have time for right now though.

FlutterSprite:
Gotcha! I wasn't really expecting that the pathfinder would be able to return something like "you need a teleporter to get here" or whatever - what I would expect the behavior to be would that it just returns "None" if no exit-connection appears between two rooms. I don't really plan on having multiple planes of existence per se, but I was interested in having multiple 'maps' for discreet locations - a building map being nested in a town map, being nested in a wilderness map etc., each with their own coordinates - though I suppose if maps were traversible via exits then the pathfinder will work just fine!

strikaco:
> Gotcha! I wasn't really expecting that the pathfinder would be able to return something like "you need a teleporter to get here" or whatever - what I would expect the behavior to be would that it just returns "None" if no exit-connection appears between two rooms. I don't really plan on having multiple planes of existence per se, but I was interested in having multiple 'maps' for discreet locations - a building map being nested in a town map, being nested in a wilderness map etc., each with their own coordinates - though I suppose if maps were traversible via exits then the pathfinder will work just fine!

Yep, that setup should graph just fine, as long as there is a connection between each area/subarea by means of an Exit. 

It would only cause trouble if someone implements something like a TARDIS in which someone inside a booth pushes a button and is \@tel'ed to a new room without having used an Exit. Many `at_*` hooks are bypassed doing things that way as well, so the "correct" thing to do in these cases (independent of this contrib) would be to create/enable an Exit connecting the TARDIS interior to the new location and shoving the character through that when the button is pressed, which is the behavior expected by Pathfinder.

Griatch:
@strikaco I think most MUD pathfinder algorithms does not consider stuff like locked doors. Unfortunately this is a tricky problem, as you say, because the graph assumes a general, topological solution that applies to everyone, when in practice it may well be highly character-dependent: Through locks you could easily design exits that are invisible and impassable without a given item or tag set on you (see for example the northern path in the tutorial world). The exit is still there in the room, just selectibly visible. the flexibility of this makes it nigh impossible to make a general solution for it from a pathfinding perspective (and why I've not made one in core Evennia. 
Since this is so game-dependeint, a contrib works well for it, but it's probably wise to include a mention of these caveats in the README. Maybe one could consider adding a hook for individual devs to easier plug in more complex analyis of exists/rooms when the pathfinder crawls the map ...

Griatch:
@strikaco also, since we are not adding `networkx` as a required dependency, you'll need to mock it somehow in your testing or unittests will continue to fail.

vadi2:
I'd recommend setting being able to set a limit on the amount of rooms pathfinding will scan - MUDs can get pretty big, and since this is server-side, you wouldn't want to give someone an ability to lag the game out.

strikaco:
Ok, so I've tweaked quite a bit by this point.

* The Pathfinder class itself is no longer a wrapper for a nxgraph object, it *is* a nxgraph object. So this lends itself well to being used to create any number of discrete subnetworks, though I don't think there is any inherent routing between them (it's way out of scope for this contrib anyway, but I at least wanted to enable the possibility).
* You can now specify what queryset you want to use when generating graphs. For example, if you have a bunch of niche rooms with a particular tag (like "chargen" or any sort of instancing setup) that you don't want to include, then feed Pathfinder an iterable containing all the rooms you *do* want (or a queryset that results in them being excluded when run).
* Of greater interest is the fact that if you provide a caller object to functions, its lockset will get factored into the path calculation. This depends on `traverse` locks being set properly on all exits.  I'm unclear on how to do the same with permissions and dynamic stuff like `must_possess: rope` but we'll get there.
* Providing a caller object when calculating line of sight will do the same with `view` locks on rooms. If you can `view` a room, you can see it and further. If you cannot `view` a room, you cannot see it or past it. If you absolutely want a birds-eye/god view, leave caller as None and it will skip lock checks altogether.

evennia/contrib/pathfinding/pathfinder.py
```diff
@@ -0,0 +1,394 @@
...
+                dst_key = self.get_key(dst)
+                try: self[dst_key]
+                except: self.add_node(dst_key, key=dst.db_key, id=dst.id, type='room', locks_str=str(dst.locks))
+                
+                # Create an edge representing the exit
+                self.add_edge(src_key, dst_key, direction=exit.db_key, id=exit.id, type='exit', locks_str=str(exit.locks))
+                
+        return self
+                
+    def get_path(self, source, dest):
```
Griatch: Similarly to the line-of-sight calculation, how does this handle cases where a room (say, on the way to the destination), has exits that loop back on itself or to a room earlier in the path?

evennia/contrib/pathfinding/scripts.py
```diff
@@ -0,0 +1,96 @@
...
+Get or create an instance of this script at server start/reload.
+
+To do so, modify `yourgame/server/conf/at_server_startstop.py`:
+
+Add this import to the top of the file:
+
+    from evennia.contrib.pathfinding.scripts import PathfinderScript
+    
+Add this line to the beginning or end of your `at_server_start()` function, but
+above the word 'pass' if it's there:
```
Griatch: Actually, you replace the `pass`, it has no purpose if you are actually doing anything in the function  ...

evennia/contrib/pathfinding/scripts.py
```diff
@@ -0,0 +1,96 @@
...
+            obj = cls.objects.first()
+            
+        kwargs['obj'] = None
+        if not obj: obj = create.create_script(cls, **kwargs)
+        return obj
+    
+    def at_script_creation(self):
+        self.key = "pathfinder"
+        self.desc = "Maintains a global Pathfinder instance."
+        self.interval = 60 * 30 # every 30 minutes
```
Griatch: This sounds like it would be suitable to set during creation rather than having it hard-coded. For a production game that is not into a phase of heavy building, running this every 24h seems more than enough.

evennia/contrib/pathfinding/tests.py
```diff
@@ -0,0 +1,211 @@
+from evennia import create_script, DefaultScript, DefaultRoom
+from evennia.contrib.pathfinding.pathfinder import Pathfinder
```
Griatch: Unittests are currently failing. This is because we don't include `networkx` as part of our dependencies. So while this may seem counter-intuitive, the `networkx` library needs to be mocked out for the unit tests. See [here](https://docs.python.org/3/library/unittest.mock-examples.html) for help with mocking.

evennia/contrib/pathfinding/pathfinder.py
```diff
@@ -0,0 +1,394 @@
...
+        """
+        Get all the Room objects to be included in this graph.
+        
+        Returns:
+            objs (generator): Any iterable generator of Room objects you
+                wish to be part of the graph.
+            
+        """
+        if self.queryset: 
+            return (x for x in self.queryset)
```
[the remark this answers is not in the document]
Griatch: Since you call the variable `self.queryset`, it's not clear to the user that slapped-together lists would be ok. But sure, if you want to support that then this solution would work. You should probably make that clearer in the documentation in that case.

evennia/contrib/pathfinding/pathfinder.py
strikaco: I understand your question but I'm having a hard time simulating it, so I suspect networkx is refusing to chase its own tail in this manner. An infinite loop is never a valid path, and I can't seem to make it get stuck trying to traverse one.

LOS calculation was a bit different as that was my own implementation and I hadn't thought to do such checks when you raised that issue, but in other contexts nx is smart enough to return "inf" (infinity) as the length of a path so given my inability to break it I'm assuming it's doing the same here.

Griatch: Ok, good to know! Worthy of a note in the docstring to show that it was considered in case people wonder later (and they will). :)

evennia/contrib/pathfinding/scripts.py
strikaco: That's what I tried to facilitate but I don't fully understand script construction. Any kwargs (like interval) passed to the spawn() function get forwarded to create.create_script(), and I was under the assumption constructor kwargs took precedence over anything hardcoded. I don't see any other way to pass data to at_script_creation. Advice plz? :)

Griatch: No, that sounds right, I think I just missed that spawn forwards to `create_script`, sorry.

evennia/contrib/pathfinding/tests.py
strikaco: Counterintuitive indeed...I understand where you're coming from wrt not making nx a dependency and still wanting tests to pass, but having a hard time wrapping my mind around the rest of it. 

What becomes the point of the unit tests if we're faking the entire procedure? The current suite tests integration between libraries and will fail if Evennia or networkx deviate from each other in an incompatible manner. Mocking inputs/outputs seems like it would yield false assurances in such an event.

If mocking is just for the sake of passing tests in the absence of dependencies, I could much more easily catch an ImportError and skip all the tests, rather than write a lot of additional code that pretends to make sure everything's working in favor of code that actually does.

I don't mean to come across as defensive/combative, just feeling in over my head. I hear what you're asking but am struggling with comprehension of the concept :\

Griatch: @strikaco It may feel a bit unexpectedl, but it's worth remembering that we are not testing the networkx library. The pathfinder contrib has a lot of code that needs unit testing - this is the purpose of a unit test - to test the code in isolation. Mocking out a third-party library (for example because it requires a live network connection or, in this case because it's not necessarily available at the time of testing) is not weird at all. The networkx library has its own unit tests, it is not our job to unit test it - rather we expect a certain set of inputs and outputs from it. It's the *effect* of those inputs and outputs on the pathfinder code that we are testing, not the functionality of the networkx library itself.  

If you are worried about networkx's API changing in relation to Evennia you need to establish the specific version you support and lock into that. This is not unreasonable and something we do with actual default dependencies.

Of course the live networkx library is convenient to use when designing the unit test. But for consistency, there should be no use of the original networkx library in the final unit test suite (even if it's available) - the point being that the suite should be designed to act exactly the same regardless of if the inputs are coming from the networkx lib or from the mock. 

If you are unsure about how mocking works, I'm happy to help.

strikaco: > It's the effect of those inputs and outputs on the pathfinder code that we are testing, not the functionality of the networkx library itself.

@Griatch I'm still failing to understand how to apply this. What I'm envisioning is something like:

To test something like get_path(<basement obj>, <attic obj>)-- check to make sure after calling that, there is a subsequent call to nx.shortest_path(self, source=src_key, target=dst_key).

To test get_usable_path(<basement obj>, <attic obj>)-- check to make sure that after calling that, a request for a nx.DiGraph() is made, a bunch of calls are made to out_edges() and in_edges(), *.locks.check(), and *.add_edge().

This can't yield any output by itself. How much further am I supposed to go with each test? Is the entire point just to count/see what calls are being made?

strikaco: I've made an attempt to resolve this with 575737f; let me know if I'm on the right track...

strikaco:
It does, thanks. I was struggling with how to go about mocking the library itself, or creating an entirely separate class that implements and mocks *all* of a DiGraph's methods, but I see what you did there. That makes more sense. I'll circle back to this soon.

strikaco:
@Griatch Something isn't working but I'm not sure what. Using your template and having uninstalled networkx from my venv to troubleshoot, the module appears to fail to patch-- I'm getting `ImportError: No module named networkx.readwrite` when the pathfinder class is instantiated.

I added a `self.pfinder = self.pathfinder_module.Pathfinder()` to the last line of setUp to create it.

Any ideas what's wrong here?

Griatch:
Still stuck with the unit testing of this?

strikaco:
Sort of. I now know how to use mock thanks to your sample code, but I have not had the time to revisit this. I haven't forgotten it though :)

Griatch:
@strikaco Still issues with the mocking?

Griatch:
This would be good to merge, as long as the tests are in place.

strikaco:
I don't have time to see this one through; too busy with work. I appreciate your help with the mocking though, it's come in handy in a lot of other places. Closing this.

Griatch:
@strikaco That's sad to hear, this was so close to being ready :(

PR #1666, opened 2018-09-28
Enhanced geoip and contextual connection wrappers

strikaco:
This PR addresses two issues:

Evennia `Session` objects and Django `HttpRequest` objects both handle TCP connections yet have almost no properties in common. They both share the concept of useragents, source IP address, protocol and others but the names of these properties are not consistently accessible. So for operations like banning an IP address from the terminal and webserver, you have to implement two different procedures for handling what is ultimately a simple text string. This PR provides a wrapper for both classes that handles identification and extraction of appropriate properties, so to get the IP address from either, you can do `SessionWrapper(session).ip` or `RequestWrapper(request).ip` to get the same data.

The second component to this PR extends Django's own geoip contrib to retrieve AS data (or ISP data, for those with the appropriate licensing) as well as the specific CIDR block of a given IP. With this added context, one can more accurately identify where users are coming from and what infrastructure they are using when anonymously conducting harassment campaigns. No more assumptions of /24 subnets need to be made or blocking entire class Bs using wildcards; we can extend the ban mechanisms to block by any artifact-- admins can now be empowered to ban an entire country, a particular service provider, or a specific subnet, which are functions you normally have to have a standalone WAF or pay Cloudflare to do.

The two integrate in a manner such that if you wrap a `Session` or `Request`, you can easily drill into connection details without any special magic. `SessionWrapper(session).isp` will output something along the lines of `Verizon Wireless`.

The capacity for trolls to cause damage generally outscales most applications' ability to stop them, Evennia being no exception. Being terminal-based it is susceptible to all of the same sorts of scripted problems that plague IRC. Being able to classify traffic would go a long way towards filtering out that which is bad.

### DIFF
--- a/evennia/contrib/security/geoip/__init__.py
+++ b/evennia/contrib/security/geoip/__init__.py
@@ -1,0 +1,120 @@
+from django.conf import settings
+from django.contrib.gis.geoip2 import GeoIP2 as DjangoGeoIP2
+from django.core.validators import validate_ipv46_address
+
+from pathlib import Path
+
+import geoip2.database
+import os
+import pyasn
+import socket
+
+# Creating the settings dictionary with any settings, if needed.
+GEOIP_SETTINGS = {
+    'GEOIP_PATH': getattr(settings, 'GEOIP_PATH', os.getcwd()),
+    'GEOIP_CITY': getattr(settings, 'GEOIP_CITY', 'GeoLite2-City.mmdb'),
+    'GEOIP_COUNTRY': getattr(settings, 'GEOIP_COUNTRY', 'GeoLite2-Country.mmdb'),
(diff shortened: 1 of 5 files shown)


Griatch:
Note that I moved the `auditing` PR into this same `contrib/security/` package too.

Griatch:
For mocking out the unavailable imports, see my comments to pull #1664 [here](https://github.com/evennia/evennia/pull/1664#commitcomment-30795923)

Griatch:
@strikaco This appears to still have issues with the mocking?

Griatch:
Any progress on this?

strikaco:
I don't have time to continue work on this; closing.

Griatch:
@strikaco Sorry to hear that :(

PR #1667, opened 2018-09-29
Add a systemd unit file using the latest "foreground portal" options

friarzen:

Added a sample script for running Evennia as a systemd unit

Using newer distributions, the preferred method of managing system services is through systemd.

The documentation on the wiki would need to be updated to reflect systemd
This PR supersedes  Systemd unit #1317

### DIFF
--- a/bin/unix/evennia.service
+++ b/bin/unix/evennia.service
@@ -1,0 +1,34 @@
+# Evennia systemd unit script
+#
+# Copy this to /usr/lib/systemd/system/ and Edit the paths to match your game.
+#
+# Then, register with systemd using:
+#
+#   sudo systemctl daemon-reload
+#   sudo systemctl enable evennia.service
+#
+
+[Unit]
+Description=Evennia Server
+
+[Service]
+Type=simple
+
+#
+# Change this to the user the game should run as.
+# Don't run this as root.  Please, I beg you.
+#
+User=your-user
+
+#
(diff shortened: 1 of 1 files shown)


bin/unix/evennia.service
```diff
@@ -0,0 +1,34 @@
...
+# Change this to the user the game should run as.
+# Don't run this as root.  Please, I beg you.
+#
+User=your-user
+
+#
+# The command to start Evennia as a Systemd service.  NOTE:  These must be absolute paths.
+# Replace /your/path/to with whatever is appropriate.
+#
+ExecStart=/your/path/to/pyenv/bin/python /your/path/to/evennia/bin/unix/evennia ipstart --gamedir /your/path/to/mygame
```
Griatch: I don't know offhand what syntax systemd scripts allow, but would it not be possible to allow the user to enter the path once as a variable at the top?

friarzen: Nope.  Systemd is SUPER pedantic.  It does not support any kind of variable on the ExecStart pathing.  It is often quoted as being a security "feature"....

Russell-Jones: See https://askubuntu.com/questions/1014480/how-do-i-add-bin-to-path-for-a-systemd-service I needed to add a PATH to get things going (to put twistd from my venv into the PATH), so it was handy for that, but the post also emphasises what friarzen said.

Issue #533, opened 2014-06-30
Player/Character creation via web [$35]


Griatch:
Player creation via the web admin works but it is not feasible to create a full Player-Character pairing via the admin at this point. This issue concerns the full pipeline of creating Player and Character in the admin and then log in and play normally under any `MULTISESSION_MODE`. 
- ~~Player cannot get permissions. Permissions are Tags with the `db_tagtype = "permission"`, and there is currently no way to create new Tags to go with the Player permission. One can select an existing tag (which has to be pre-created in-game at the moment), but the Tag display does not currently show the different tag types so one could accidentally pick a non-permission Tag here if it happens to be named the same as a permission level.~~ resolved in #1115
- ~~Attributes should be fully modifiable from the admin.~~ resolved in #1115
- ~~Aliases are Tags with `db_tagtype="alias"`, there should be a way to set this (nicks are not needed though).~~ resolved in #1115 
- ~~The superuser flag sits in the "web only" section. This is the only setting that does matter in-game in the same way as on the web, so it should probably be described somehow in that header (this is just a documentation issue).~~ -Fixed in develop
- ~~For `MULTISESSION_MODE=0,1`, one should be able to tie a Character to the Player by setting the Object's `db_player` foreignfield. In `MULTISESSION_MODE=2`, one might instead want to assign the character to the `_playable_characters` Attribute on the Player. A custom form is probably needed for this linking.~~ -Added with button in develop
- ~~Possibly there should be some sort of list of suggested Object cmdsets so one doesn't need to remember and write the path every time. This could be a help text or if possible some selector based on `settings.CMDSET_*`. Maybe one could consider some sort of clickable buttons that fill in some standard presets ...?~~ Added with dynamic help entry in develop
- ~~When setting a cmdset on a Character in the admin, the cmdset seems to be set to be visible outside the character. The result is that you get multi-command errors when executing commands in the same room as the new Character.~~ Fixed a long time ago 
- ~~Tags-Permissions for Characters are needed in the same way as for Players.~~ resolved in #1115 
- ~~There should probably be some help info or preset for how to set the `puppet:` lock to allow a Player to puppet a given Character, such as `puppet:id(<player id>)`.~~ Handled with button and info-text in develop
- ~~There should probably be a way to connect a Player to a Character in the admin, for `MULTISESSION_MODE=0,1` to not need to use `@ic` to connect.~~ Fixed in develop.

Griatch:
[...]

PR #1669, opened 2018-10-02
Typeclass-based object creation, validation, and authentication

strikaco:
* Implements username validation through configurable Django validators.
* Accounts now handle reflexive actions.
  * Check if a name/IP are banned with a single call to Account.is_banned()
  * Get or create a usable guest account (where enabled) with a single call to Account.authenticate_guest()
  * Authenticate regular accounts with a single call to Account.authenticate(), which also checks login throttles, banlists and emits events to the security log upon success or failure.
  * Usernames have excessive spaces stripped and visually-similar Unicode characters are treated as identical for the purposes of preventing duplicate (or duplicate-looking) usernames via Account.normalize_username().
  * Create a fully-functional default Account or Account+Character pair with Account.create()
* Cleans up a lot of redundant/migrated/consolidated logic from unloggedin cmdset.
* Adjusts throttle so that logging of continued auth failures is suppressed while throttle is active; also accepts custom activation message on update/triggering

Migration of logic out of Commands and into more contextually-agnostic locations.

Apologies for the torrent of changes in a single PR; once I started messing with one thing, the effects cascaded.

#1383:
* Authentication failures now logged (limited by throttle)
* Account creations now logged (guest and normal)
* Throttle activations now logged
* Password changes now logged
* IP address of creator (at time of creation) stored on new Account objects (makes bulk-created/spam accounts easier to identify without having to rely on timestamps)
* IP address of creator (at time of creation) stored on new Character objects during Account creation. Creator's id (as int) is also stored on Character objects.

#533: 
* Account creation and the security measures surrounding it can now be consistently applied and audited by both terminal commands and Django forms.

### DIFF
--- a/evennia/accounts/tests.py
+++ b/evennia/accounts/tests.py
@@ -84,16 +84,14 @@ class TestDefaultGuest(EvenniaTest):
         self.assertFalse(account, 'Two guest accounts were created with a single entry on the guest list!')
         
         settings.GUEST_ENABLED = False
-
-class TestDefaultAccount(EvenniaTest):
-    "Check DefaultAccount class"
-
+        
+class TestDefaultAccountAuth(EvenniaTest):
+    
     def setUp(self):
-        self.s1 = MagicMock()
-        self.s1.puppet = None
-        self.s1.sessid = 0
+        super(TestDefaultAccountAuth, self).setUp()
         
         self.password = "testpassword"
+        self.account.delete()
         self.account = create.create_account("TestAccount%s" % randint(100000, 999999), email="test@test.com", password=self.password, typeclass=DefaultAccount)
         
     def test_authentication(self):
@@ -138,10 +136,6 @@ class TestDefaultAccount(EvenniaTest):
         result, error = DefaultAccount.validate_username('xx')
         self.assertFalse(result, "2-character username passed validation.")
 
-    def tearDown(self):
-        if hasattr(self, "account"):
-            self.account.delete()
-
     def test_password_validation(self):
         "Check password validators deny bad passwords"
 
@@ -166,6 +160,14 @@ class TestDefaultAccount(EvenniaTest):
 
         # Try setting a better password (test for False; returns None on success)
         self.assertFalse(self.account.set_password('Mxyzptlk'))
+
+class TestDefaultAccount(TestCase):
+    "Check DefaultAccount class"
+
+    def setUp(self):
+        self.s1 = MagicMock()
+        self.s1.puppet = None
+        self.s1.sessid = 0
 
     def test_puppet_object_no_object(self):
         "Check puppet_object method called with no object param"
--- a/evennia/accounts/tests.py
+++ b/evennia/accounts/tests.py
@@ -75,6 +75,41 @@ class TestDefaultAccount(TestCase):
         obj, errors = DefaultAccount.authenticate(self.account.name, 'xyzzy')
         self.assertFalse(obj, 'Account authenticated using invalid credentials.')
         
+    def test_create(self):
+        "Confirm Account creation is working as expected."
+        # Create a normal account
+        account, errors = DefaultAccount.create(username='ziggy', password='stardust11')
+        self.assertTrue(account, 'New account should have been created.')
+        
+        # Try creating a duplicate account
+        account, errors = DefaultAccount.create(username='Ziggy', password='starman11')
+        self.assertFalse(account, 'Duplicate account name should not have been allowed.')
+        
+        # Guest account should not be permitted
+        account, errors = DefaultAccount.authenticate_guest()
+        self.assertFalse(account, 'Guest account was created despite being disabled.')
+        
+        settings.GUEST_ENABLED = True
+        settings.GUEST_LIST = ['bruce_wayne']
+        
+        # Create a guest account
+        account, errors = DefaultAccount.authenticate_guest()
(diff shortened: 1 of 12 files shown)


strikaco:
> We already have a create functions in the Account manager, wrapped by the utils.create.account function - the Account.create is a game-specific wrapper on top of that which confuses the matter; which one should one use?

Unfortunately it depends. Whether you use Account.objects.create() or calling utils.create_account(), both are low-level calls-- you ask for an account, you get one, nothing more. It does not perform the post-creation steps currently implemented only in CmdUnconnectedCreate.

As you point out, the benefit of implementing this as a class method is that if one decides they want additional steps taken (like subscribing to additional channels or granting starter bonuses) they can extend the functionality instead of having to reimplement all of it.

> Linking back to the first negative point - this breaks consistency with other typeclasses. You can't do Object.create or Script.create, for example - the Account class becomes a discrepancy. This inconsistency does not an acceptable API do.

We could though. I just hadn't thought of that :)

> Consolidate all typeclasses to have a .create method on themselves, in the simplest form only as a wrapper of the manager method. This is a big or small API change depending on if we then consider utils.create deprecated or not.

I think this is the best option, but we don't really need to deprecate utils.create at all. Leave that for power users who don't want any hand-holding, while implementing friendlier and more thorough object.create() methods for the sake of easier adoption and configuration. My opinion (and part of the basis for this PR) is that a naive dev should be able to call something like object.create() and receive a default feature-complete object where possible, unless overridden or extended to do otherwise.

Griatch:
@strikaco After some thinking I agree that this option is the sanest one - I think the way to go would be to add a new `.create` method on the Typeclass base (defaulting it to be a simple wrapper for the `utils.create` functions) and then override it in the subclasses where it's needed (like for `Account.create` in this case).

strikaco:
> @strikaco After some thinking I agree that this option is the sanest one - I think the way to go would be to add a new `.create` method on the Typeclass base

For additional consideration in support of this course of action, note that these objects already had `delete()` methods implemented directly on themselves. It seems logical that `create()` should also be accessible in this manner.

> (defaulting it to be a simple wrapper for the `utils.create` functions)

Looking into it, I notice that DefaultObjects are affected by the same workflow issue that led to this PR for Account objects. `CmdCreate` implements useful post-creation functionality (like setting locks and default descriptions) that is only available in-game through `CmdCreate`. I'm sure I'll find the same when I get to Character/Room/Exit objects too.

Wouldn't it make more sense to migrate the effective functionality of `CmdCreateWhatever` to `WhateverObject.create()` across the board instead of making them simple wrappers? I think you alluded to this with "`in the simplest form only as a wrapper of the manager method`" but want to make sure we're on the same page. This would be more consistent in application (not just an exception for Accounts) and much easier for newcomers to grasp.

evennia/accounts/tests.py
```diff
@@ -66,7 +68,68 @@ def setUp(self):
         self.s1 = MagicMock()
         self.s1.puppet = None
         self.s1.sessid = 0
-        self.s1.data_outj
```
strikaco: This got merged in somewhere along the lines; appears to be an error? Removed it. Tests still pass...

evennia/settings_default.py
```diff
@@ -810,6 +810,28 @@
     {'NAME': 'django.contrib.auth.password_validation.CommonPasswordValidator'},
     {'NAME': 'django.contrib.auth.password_validation.NumericPasswordValidator'},
     {'NAME': 'evennia.server.validators.EvenniaPasswordValidator'}]
+    
+# Username validation plugins
+AUTH_USERNAME_VALIDATORS = [
+    {
```
Griatch: From a pure formatting perspective, we generally don't format dicts this way (javascript-like, maybe?), but tend to use a more compact form in this settings file.

strikaco:
How many typeclassed entities are there? I see "create_object", "create_script", "create_help_entry", "create_message", "create_channel", "create_account" listed in utils.create(). So far it seems I'm overlooking help_entry, message and channel?

What should I have the typeclass base do? I can't account for custom entities or their creation path. Should I just raise a not-implemented exception?

Griatch:
The base typeclasses are `DefaultAccount`, `DefaultObject`, `DefaultScript` and `DefaultChannel`.

I mis-spoke about the `Msg`, this is not a typeclass, just an idmapped entity. Same with `HelpEntry` (although I have toyed with the idea making help entries typeclassed, they are not at this time). 

While the `TypeclassBase` *could* forward its `.create` to `cls.objects.create`, I think it's probably best to not touch it further now that I look at it - keep the `create` functions only on the parent typeclasses above.

strikaco:
I believe I've addressed the feedback to date; whenever you get a moment this is ready for review.

Griatch:
@strikaco The code changes look good, but you must test your PRs against the entire test suite - it appears to pass when you run your tests in isolation, but fail when ran in combination with the rest of the Evennia test suite (and I've compared with and without this branch - the fault is introduced here). See the results from the Travis integration. 
The likely cause of this is because you are making assumptions about the state of your test environment what is not valid when run with other tests. In principle unit tests should be isolated enough that this should not happen (so the problem likely lies in another test that runs before this), but one way or another - the test suite does not pass.

strikaco:
@Griatch All bugs should be squashed; comprehensive tests pass locally but it doesn't look like Travis picked up the last commit. What now?

strikaco:
...Looks like it picked it up after all. This one's ready.

PR #1671, opened 2018-10-04
Frontend CRUD enhancements (character/account creation and management)

strikaco:
* Updates Bootstrap to v4 stable (the beta in use had some irritating bugs that have since been resolved)
* Enables the use of Django Messages framework
* Adds a web-based account registration form
* Adds links to reset password / create new account on login pane (resets require SMTP configuration to actually work, but the functionality was already enabled by account of django.contrib.auth.urls-- I just made it prettier). Uses Django's password reset workflow.
* Modifies the series of password reset forms to use Evennia's CSS instead of Django admin's.
* Adds a dropdown for authenticated users allowing them to change their password.
* Prettifies the "change password" form to use Evennia's CSS instead of Django admin's.
* Adds placeholders to dropdown menu for players to quickpuppet/manage characters on the website.
* Fixes a bug on the login pane where error messages weren't being displayed (was calling form.has_errors instead of just checking form.errors)
* Web-based chargen based on predefined form
* Character puppeting within the context of the website

Another notch towards #533


chainsol:
I meant to update to the latest stable! It's a good idea to ensure nothing broke due to this update, so I'll test that out ASAP.

evennia/accounts/accounts.py
```diff
@@ -189,6 +189,10 @@ def nicks(self):
     @lazy_property
     def sessions(self):
         return AccountSessionHandler(self)
+        
+    @lazy_property
+    def characters(self):
+        return self.db._playable_characters
```
strikaco: Django templates cannot access attributes with underscores; this is a valuable field anyway so I'm not sure why it's private/hidden.

evennia/web/website/templates/website/registration/password_reset_email.html
```diff
@@ -0,0 +1,15 @@
+{% autoescape off %}
+To initiate the password reset process for your {{ user.get_username }} {{ site_name }} account,
+click the link below:
+
+{{ protocol }}://{{ domain }}{% url 'password_reset_confirm' uidb64=uid token=token %}
```
strikaco: Currently, domain is set to "example.com" per the default configuration of what appears to be the Django Sites framework. This should probably be resolved during `evennia --init` but is beyond me to fix-- it also impacts the reverse "view on site" link in the Django admin (if get_absolute_url is implemented on the object in question).

evennia/web/website/forms.py
```diff
@@ -0,0 +1,142 @@
...
+    
+    email = forms.EmailField(help_text="A valid email address. Optional; used for password resets.", required=False)
+    
+class CharacterCreationForm(forms.Form):
+    name = forms.CharField(help_text="The name of your intended character.")
+    age = forms.IntegerField(min_value=3, max_value=99, help_text="How old your character should be once spawned.")
+    description = forms.CharField(widget=forms.Textarea(attrs={'rows': 3}), max_length=2048, min_length=160, required=False)
+    
+    @classmethod
+    def assign_attributes(cls, attribute_list, points, min_points, max_points):
```
Griatch: This appears to be game specific and does not belong in core as far as I can tell?

evennia/web/website/forms.py
```diff
@@ -0,0 +1,142 @@
...
+            
+        """
+        num_buckets = len(attribute_list)
+        point_spread = (x for x in self.random_distribution(points, num_buckets, min_points, max_points))
+        
+        # For each field, get the point calculation for the next attribute value generated
+        return {attribute: next(point_spread) for k in attribute_list}
+    
+    @classmethod
+    def random_distribution(cls, points, num_buckets, min_points, max_points):
```
Griatch: Same here, this does not belong in core. If wanting to offer these functions, they rather belong as utilities that the user can pick and chose from.

evennia/web/website/forms.py
```diff
@@ -0,0 +1,142 @@
...
+            # Add to bucket
+            buckets[index] = buckets[index] + 1
+            points = points - 1
+            
+            # Get the indices of eligible buckets
+            indices = [i for (i, value) in enumerate(buckets) if (value < min_points) or (value < max_points)]
+            
+        return buckets
+    
+class ExtendedCharacterCreationForm(forms.Form):
```
Griatch: This definitely does not belong in core. This should rather be a contrib or - maybe better - a wiki tutorial.

evennia/web/website/forms.py
```diff
@@ -0,0 +1,142 @@
...
+    
+    strength = forms.IntegerField(min_value=1, max_value=10)
+    perception = forms.IntegerField(min_value=1, max_value=10)
+    intelligence = forms.IntegerField(min_value=1, max_value=10)
+    dexterity = forms.IntegerField(min_value=1, max_value=10)
+    charisma = forms.IntegerField(min_value=1, max_value=10)
+    vitality = forms.IntegerField(min_value=1, max_value=10)
+    magic = forms.IntegerField(min_value=1, max_value=10)
+    
+    def __init__(self, *args, **kwargs):
```
Griatch: I think this whole extended character creation form rather belongs either in a contrib or explained in a tutorial. It might also be good to note that users of telnet will be left out if this is the main way to create a character - one would need to make a separate character creation mechanism for those not using the web interface.

evennia/web/website/templates/website/_menu.html
```diff
@@ -40,8 +40,21 @@
             {% endblock %}
             {% block navbar_user %}
             {% if user.is_authenticated %}
-            <li class="nav-item">
-                <a class="nav-link">Logged in as {{user.username}}</a>
+            <li class="nav-item dropdown">
+              <a class="nav-link dropdown-toggle" data-toggle="dropdown" href="#" id="user_options" aria-expanded="false">Logged in as {{user.username}} <span class="caret"></span></a>
+              <div class="dropdown-menu" aria-labelledby="user_options">
+                <a class="dropdown-item" href="{% url 'chargen' %}">Create</a>
+                <a class="dropdown-item" href="#">Manage</a>
```
Griatch: I think this should be removed - it doesn't do anything and that feels like a bug to user. If it's meant as an example it should show some sort of information saying that it's up to the dev to expand on it.

evennia/web/website/templates/website/_menu.html
```diff
@@ -40,8 +40,21 @@
...
+                <div class="dropdown-divider"></div>
+                {% for character in user.characters|slice:"10" %}
+                  <a class="dropdown-item" href="#">{{ character }}</a>
+                {% empty %} 
+                  <a class="dropdown-item" href="#">No characters found!</a>
```
Griatch: This section didn't work for me. When I logged in with my pre-existing superuser, my current Character was shown here, but creating a new character in the web chargen did not expand the list. I also tried registering a new Account, and even after having created a new Character (and played with it in the web client), this menu consistently showed "No characters found!". I have not examined why this is.

evennia/web/website/templates/website/_menu.html
```diff
@@ -40,8 +40,21 @@
             {% endblock %}
             {% block navbar_user %}
             {% if user.is_authenticated %}
-            <li class="nav-item">
-                <a class="nav-link">Logged in as {{user.username}}</a>
+            <li class="nav-item dropdown">
+              <a class="nav-link dropdown-toggle" data-toggle="dropdown" href="#" id="user_options" aria-expanded="false">Logged in as {{user.username}} <span class="caret"></span></a>
+              <div class="dropdown-menu" aria-labelledby="user_options">
+                <a class="dropdown-item" href="{% url 'chargen' %}">Create</a>
```
Griatch: For clarity of intent, it would be better to call this menu item `New Character` or `Create Character`.

evennia/web/website/templates/website/_menu.html
```diff
@@ -40,8 +40,21 @@
             {% endblock %}
             {% block navbar_user %}
             {% if user.is_authenticated %}
-            <li class="nav-item">
-                <a class="nav-link">Logged in as {{user.username}}</a>
+            <li class="nav-item dropdown">
+              <a class="nav-link dropdown-toggle" data-toggle="dropdown" href="#" id="user_options" aria-expanded="false">Logged in as {{user.username}} <span class="caret"></span></a>
```
Griatch: I like using a dropdown for this, feels very flexible without cluttering the rest of the UI. :+1:

evennia/accounts/accounts.py
strikaco: Hmm. The locks thing makes sense when determining who can puppet what, but it doesn't lend itself well to identifying which players "own" which characters at the db level. I think the reason myself and others have just taken advantage of this attribute is because it's simple and intuitive. Some sort of wrapper would be beneficial. Not sure how to proceed with this though.

evennia/web/website/forms.py
```diff
@@ -0,0 +1,142 @@
...
+class AccountCreationForm(UserCreationForm):
+    
+    class Meta:
+        model = class_from_module(settings.BASE_ACCOUNT_TYPECLASS)
+        fields = ("username", "email")
+        field_classes = {'username': UsernameField}
+    
+    email = forms.EmailField(help_text="A valid email address. Optional; used for password resets.", required=False)
+    
+class CharacterCreationForm(forms.Form):
```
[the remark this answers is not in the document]
strikaco: Agreed, I've since distilled it into a more generic name/description form and intend to use this more complex form as an example of sorts. The hardest part is coming up with a sane way of letting users customize chargen forms without (a) creating and forcing users to learn yet another DSL and (b) making sure that in-game chargen adheres to the same template.

evennia/web/website/forms.py
strikaco: Inclusion on this form is debatable; it's more of a general utility function. It might be better suited to inclusion in utils or just a contrib. Most games involving rolling initial stats and distributing points end up having to come up with a similar algorithm for this anyway.

evennia/web/website/forms.py
strikaco: Agreed; this is a bit of a mess currently with a lot of nuances to work through. Ultimately we need a common interface for creating and validating characters and directing telnet or webserver requests for new characters through it so there are no inconsistencies.

evennia/web/website/templates/website/_menu.html
strikaco: Which multisession mode? That might be related.

evennia/web/website/templates/website/_menu.html
strikaco: That was my own placeholder; I'm creating a character management menu to go along with this as well but enabling and implementing it hinges on how #1674 gets resolved. For development I just needed Django to not yell at me about nonexistent links :)

evennia/web/website/templates/website/_menu.html
Griatch: I think this probably was with mode 0.

strikaco: Well, obviously the goal is to do it *right*, but I was asking for the sake of diagnosis :)

Griatch:
@strikaco Is this ready for re-review?

strikaco:
No, not ready yet; the dropdown bug is still outstanding.

evennia/web/website/templates/website/_menu.html
strikaco: This was due to the use of lazy-property; the values were only being refreshed on server reload.

strikaco:
Fixed the dropdown bug but still have some tidying up to do. Further updates as events warrant...

evennia/web/website/templates/website/_menu.html
strikaco: This now lets users select a puppet to use within the context of the website.

strikaco:
So we now have a form-based character creation page, a character update page, a character management page, a character deletion page, ownership and lock checks for both, character puppeting on the website, a password change workflow, and unit tests for all website views. Everything works as far as I've thought to test it.

I think that's enough changes for one PR... for all this entailed, I'm missing out on a lot of Hacktoberfest cred :\

@Griatch ready for you to tell me how much my documentation sucks :)

Griatch:
The documentation of some of these functions is too sparse :stuck_out_tongue: But this is great work, I think it's worth merging and resolve any remaining issues as they come along. Thanks for your hard work!

PR #1674, opened 2018-10-05
Object URL references, named paths and pseudo-slugs

strikaco:
This PR is designed to let one generate paths to an object's web interface(s) from within a Django template.

It adds a set of methods to retrieve HTTP paths for DefaultAccount objects and all children of DefaultObject:
* `get_admin_url()`: Returns a path that, if followed, will display the object in the Admin backend.
* `get_absolute_url()`: Django construct; returns a path that should display the object in a DetailView.
* `get_update_url()`: Returns a path that should display the object in an UpdateView.
* `get_delete_url()`: Returns a path that should display the object in a DeleteView.

Note that this does not magically expose all your objects, create DeleteViews for everything, let random users into Admin or enable any sort of new or scary functionality.

The admin function works out of the box. It lets you do things in templates like:
```
{% if request.user.is_staff %}
<a href="{{ object.get_admin_url }}">Admin</a>
{% endif %}
```
...so when looking at an object (say, on a DetailView) staff can quickly pivot to the Admin backend to change things.

For the rest to work, you need to have defined named URLs that map typeclasses to actions. It must follow a particular naming convention (`object-action`, so `account-update`, `character-delete` or `room-detail` would yield the link to an appropriate view).

```
    # Character management
    url(r'^characters/create/$', website_views.CharacterCreateView.as_view(), name="character-create"),
    url(r'^characters/manage/$', website_views.CharacterManageView.as_view(), name="character-manage"),
    url(r'^characters/update/(?P<slug>[\w\d\-]+)/(?P<pk>[0-9]+)/$', website_views.CharacterUpdateView.as_view(), name="character-update"),
    url(r'^characters/delete/(?P<slug>[\w\d\-]+)/(?P<pk>[0-9]+)/$', website_views.CharacterDeleteView.as_view(), name="character-delete"),
```

This enables template functionality like:
```
{% for character in characters %}
<li>{{ character }} 
    <a href="{{ character.get_absolute_url }}">View</a> 
    <a href="{{ character.get_update_url }}">Update</a> 
    <a href="{{ character.get_delete_url }}">Delete</a>
</li>
{% endfor %}
```
...to let developers quickly generate links to the appropriate views governing disposition of an object. Securing those views and restricting access is entirely the responsibility of the developer; this PR does not facilitate anything that isn't already possible, just makes things more convenient for frontend development.

This PR also makes it more difficult to enumerate predictable endpoints (`/accounts/1/`, `/accounts/2/`, etc.). Evennia doesn't do slugs natively, though Django heavily encourages their use. When the link is generated, it passes the object pk as well as the slugified object name as kwargs. This means that the corresponding view must also be configured to receive and check an arbitrary `slug` kwarg, which is not the case by default. The receiving view should retrieve the object by pk but confirm the slug relates to the object name. This will inhibit scrapers and crawlers from probing random object IDs by requiring knowledge of both the object ID *and* the slugified version of the object name to be supplied to the view. It also makes the URL more intelligible (`/accounts/johnny/1/`, `/accounts/frank/2/`, etc.).

### DIFF
--- a/evennia/accounts/accounts.py
+++ b/evennia/accounts/accounts.py
@@ -14,7 +14,9 @@ instead for most things).
 import time
 from django.conf import settings
 from django.contrib.auth import password_validation
+from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
+from django.urls import reverse_lazy
 from django.utils import timezone
 from evennia.typeclasses.models import TypeclassBase
 from evennia.accounts.manager import AccountManager
@@ -189,6 +191,25 @@ class DefaultAccount(with_metaclass(TypeclassBase, AccountDB)):
     @lazy_property
     def sessions(self):
         return AccountSessionHandler(self)
+        
+    def get_absolute_url(self):
+        """
+        Returns the canonical URL for an Account. 
+        
+        To callers, this method should appear to return a string that can be 
+        used to refer to the object over HTTP.
+        
+        https://docs.djangoproject.com/en/2.1/ref/models/instances/#get-absolute-url
+        """
+        try: return reverse_lazy('account-detail', kwargs={'pk': self.pk, 'slug': slugify(self.name)})
+        except: return '#'
+        
+    def get_admin_url(self):
+        """
+        Returns a link to this object's entry within the Django Admin panel.
+        """
+        content_type = ContentType.objects.get_for_model(self.__class__)
+        return reverse_lazy("admin:%s_%s_change" % (content_type.app_label, content_type.model), args=(self.id,))
 
     # session-related methods
 
--- a/evennia/accounts/accounts.py
+++ b/evennia/accounts/accounts.py
@@ -14,9 +14,7 @@ instead for most things).
 import time
 from django.conf import settings
 from django.contrib.auth import password_validation
-from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
-from django.urls import reverse
 from django.utils import timezone
 from evennia.typeclasses.models import TypeclassBase
 from evennia.accounts.manager import AccountManager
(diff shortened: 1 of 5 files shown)


evennia/accounts/accounts.py
```diff
@@ -189,6 +191,41 @@ def nicks(self):
     @lazy_property
     def sessions(self):
         return AccountSessionHandler(self)
+        
+    def get_absolute_url(self):
```
Griatch: I think these should be marked somehow to note that they are web-specific methods. Maybe prepend all such methods with `web_` or somesuch? They make no real sense for people coming at them from the pure Python side.

evennia/accounts/accounts.py
```diff
@@ -189,6 +191,41 @@ def nicks(self):
...
+        """
+        Returns the canonical URL for an Account. 
+
```
Griatch: You need a `Returns:` docstring heading here. Would be good with an `Example:` block too, giving an example of how the url actually will look like. Same for the other methods below.

evennia/accounts/accounts.py
strikaco: Fair point, but get_absolute_url is a Django construct and the Admin pane looks for this function when generating reverse links to go back to the object detail page from the Admin page. The rest of the functions I felt made more sense to follow this convention for as well. Your call...

Griatch: I still think marking web-specific methods with something like  `web_`, yes. In the interest of consistency, this method should then also be renamed such - it's easy enough to add an alias just below (with an explaining comment) aliasing it to the name Django expects.

strikaco:
> There is an API inconsistency here - if Accounts and Objects have these methods, the user will expect other typeclasses like Scripts and Messages to have them too. I think it would be an idea to explore if these methods could be moved up and generalized into the main Typeclass parent instead.

It's a good idea. I moved the methods to TypedObject and was able to eliminate some duplicate code. I initially couldn't think of any reasons why one would want to render a ScriptDetailView, but given that the docs encourage the use of scripts for global settings (like economic engines) I'm sure someone might find it useful to be able to display the current price of pork futures on their website.

strikaco:
I think I've addressed all the feedback on this one; ready for review!

evennia/typeclasses/models.py
```diff
@@ -733,3 +736,145 @@ def at_rename(self, oldname, newname):
...
+        Django construct; used by Django Sites framework and within the Admin
+        panel for reverse linking to the object detail page. 
+        
+        https://docs.djangoproject.com/en/2.1/ref/models/instances/#get-absolute-url
+        
+        Returns:
+            path (str): URI path to object detail page, if defined.
+            
+        """
+        return self.web_detail_url()
```
Griatch: Name this method `web_get_absolute_url` for consistency. Since Django requires it on a specific form, you add that as an alias below: 
```python
get_absolute_url = web_get_absolute_url   # django requires this name
```

strikaco:
@Griatch This should be good to go now...

Griatch:
@strikaco Thanks for your work with this! Looks good now! Merged.

evennia/typeclasses/models.py
```diff
@@ -733,3 +736,135 @@ def at_rename(self, oldname, newname):
...
+        
+        Returns:
+            path (str): URI path to object deletion page, if defined.
+            
+        """
+        try: return reverse('%s-delete' % self._meta.verbose_name.lower(), kwargs={'pk': self.pk, 'slug': slugify(self.name)})
+        except: return '#'
+        
+    # Used by Django Sites/Admin
+    get_absolute_url = web_get_detail_url
```
Griatch: Why is `web_get_detail_url` not named `web_absolute_url` if `get_absolute_url` is what django expects - would not that name (just adding `web_`) be more expected?

strikaco:
I get where you're coming from and agree; Django's use of the term
'absolute' is a bit nonsensical given that what's expected on the other end
of it is in fact a DetailView (or equivalent function). I'm not aware of
anywhere else they use the term 'absolute'.

One could also argue it should be get_read_* to adhere to the CRUD idea but
Django deviates there too-- their idea of a Read view is again the
DetailView.

So Detail felt like a more natural term to use. Where should the
get_update_url function go? An UpdateView. Where should get_detail_url go?
A DetailView. And so on.

Issue #1477, opened 2017-10-13
Add a /delete switch to @accounts


vincent-lg:
#### Brief summary of issue / Description of requested feature:

I suggest we remove the `@delaccount` command, instead, we could add a `/delete` switch to the `@accounts` command.

#### Steps to reproduce the issue / Reasons for adding feature:

1. Most Evennia commands respect this syntax: a master command, switches for operations.  They are few exceptions, but I believe it would be worth having a uniform system.
2. `@accounts` could have an alias on `@account` and allow additional switches in the future to perform more advanced searches or operations.

#### Error output / Expected result of feature

`@accounts/delete <name>`

If you approve of it, I would be happy to offer a PR.

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:

* Evennia : current develop.

Griatch:
I think this is a good idea. There are overall way too many commands doing very similar things. While these are all inspired by MUX, it's not like the concept of `Account` is present in MUX anyway.

Griatch:
Implemented in #1682.

PR #1682, opened 2018-10-12
Move delaccount functionality to `@accounts/delete`

whutch:

This PR implements feature request #1477.

It moves the code from CmdDelAccount into a switch check in CmdAccounts, making some minor style adjustments (mostly normalizing the quotes around strings and compressing newlines), and adding an explicit permission check to the new switch (as `@delaccount` has a higher permission lock than `@accounts`). It then completely removes all references to CmdDelAccount.

It does _not_ carry over the /delobj switch from CmdDelAccount as that switch appeared to be completely unimplemented.


Griatch:
@whutch Thankyou! I made some minor tweaks and added a yes/no question before merging this.

Issue #1663, opened 2018-09-24
error in scripts with repeats = 1


kaleara:
#### Brief summary of issue / Description of requested feature:
When creating custom scripts I cannot set repeats = 1 or it crashes.

#### Steps to reproduce the issue / Reasons for adding feature:

1. create script below
```
class TestScript(DefaultScript):
    def at_script_creation(self):
        self.key = 'testing_script'
        self.interval = 10
        self.repeats = 1
        self.persistent = False
```

2. apply in game either through create_script or @scripts command

#### Error output / Expected result of feature
```
Traceback (most recent call last):
  File "f:\repositories\evennia\evennia\commands\cmdhandler.py", line 591, in _run_command
    ret = cmd.func()
  File "f:\repositories\evennia\evennia\commands\default\building.py", line 2504, in func
    ok = obj.scripts.add(self.rhs, autostart=True)
  File "f:\repositories\evennia\evennia\scripts\scripthandler.py", line 78, in add
    autostart=autostart)
  File "f:\repositories\evennia\evennia\utils\create.py", line 217, in create_script
    new_script.save()
  File "f:\repositories\evennia\evennia\utils\idmapper\models.py", line 413, in save
    _MONITOR_HANDLER.at_update(self, fieldname)
  File "f:\repositories\evennia\evennia\scripts\monitorhandler.py", line 99, in at_update
    if obj in self.monitors and fieldname in self.monitors[obj]:
  File "f:\repositories\pyenv\Lib\site-packages\django\db\models\base.py", line 615, in __hash__
    raise TypeError("Model instances without primary key value are unhashable")
TypeError: Model instances without primary key value are unhashable
```

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
Happens when game is hosted on Windows and LInux. Same error happened when I tried to manually stop the script using at_repeat when repeats was greater than 1. Doesn't happen if start_delay is used in at_script_creation.

Henddher:
Is this as simple as it sounds?

The stacktrace looks tricky but the description seems straight forward.

Henddher:
I'm about to place a simple test in `evennia/utils/test_resources.py` as it seems like the closest place to test `create_script` function. 

I don't think that's the wrong place but I can't find where `evennia.utils.create_script` function has its tests ... I'll be happy to move it to the correct location afterwards.

kaleara:
[...]

PR #1690, opened 2018-10-14
#1663 [TRIAGE HELP NEEDED] test evennia.utils.create.create_script

Henddher:
evennia/issues/#1663


Henddher:
When running tests, the following appears on the terminal

```
../.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-
packages/django/db/models/base.py:319: RuntimeWarning: 
Model 'typeclasses.testscript' was already registered. 
Reloading models is not advised as it can lead to inconsistencies, 
most notably with related models.
```

```
(py2_evennia) $ evennia test evennia.utils.test_resources

    TESTING: Using Evennia's default settings file (evennia.settings_default).
    (use 'evennia --settings settings.py test .' to run tests on the game dir)
    
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
./Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py:319: RuntimeWarning: Model 'typeclasses.testscript' was already registered. Reloading models is not advised as it can lead to inconsistencies, most notably with related models.
  new_class._meta.apps.register_model(new_class._meta.app_label, new_class)
.
Ran 2 tests in 0.054s

OK
Destroying test database for alias 'default'...

```

Henddher:
```
---> 99         if obj in self.monitors and fieldname in self.monitors[obj]:
    100             for idstring, (callback, persistent, kwargs) in self.monitors[obj][fieldname].iteritems():
    101                 try:
    102                     callback(obj=obj, fieldname=fieldname, **kwargs)
    103                 except Exception:
    104                     to_delete.append((obj, fieldname, idstring))
    105                     logger.log_trace("Monitor callback was removed.")
    106         # we cleanup non-found monitors (has to be done after loop)
    107         for (obj, fieldname, idstring) in to_delete:
    108             del self.monitors[obj][fieldname][idstring]
    109 

ipdb> p obj.__dict__
{'__dbclass__': <class 'evennia.scripts.models.ScriptDB'>, '_state': <django.db.models.base.ModelState object at 0x10f72f510>, 'db_lock_storage': u'', '_is_deleted': True, 'db_interval': 10, '_createdict': {'account': None, 'obj': None, 'autostart': True, 'interval': None, 'persistent': None, 'start_delay': None, 'locks': None, 'key': 'test_script', 'report_to': None, 'repeats': None}, 'id': None, 'db_is_active': True, 'aliases': <evennia.typeclasses.tags.AliasHandler object at 0x10f7a3bd0>, '_db_account_cache': None, 'db_desc': u'', 'db_typeclass_path': 'evennia.utils.test_resources.TestScriptB', 'db_account_id': None, 'db_date_created': datetime.datetime(2018, 10, 15, 0, 29, 50, 299227, tzinfo=<UTC>), 'db_key': 'test_script', '_db_obj_cache': None, 'permissions': <evennia.typeclasses.tags.PermissionHandler object at 0x10fa32cd0>, 'db_repeats': 1, 'db_persistent': False, 'nattributes': <evennia.typeclasses.attributes.NAttributeHandler object at 0x10fa31e50>, '_ndb_holder': <evennia.typeclasses.models.DbHolder object at 0x10fa32c50>, '_db_holder': <evennia.typeclasses.models.DbHolder object at 0x10f81b790>, 'db_start_delay': False, 'attributes': <evennia.typeclasses.attributes.AttributeHandler object at 0x10f7a35d0>, 'delete': <bound method TestScriptB._deleted of <TestScriptB: test_script>>, 'db_obj_id': None}

ipdb> !for k,v in self.monitors.items():print(k, v)

ipdb> obj in self.monitors
*** TypeError: Model instances without primary key value are unhashable

ipdb> p obj
<TestScriptB: test_script>

```

I see `id: None` and `'db_obj_id': None` and `self.monitors` is empty?

Griatch:
@Henddher 
`db_obj_id = None` would be because the script is not stored on an Object. `self.monitors` would be empty because there are not OOB monitors assigned to the object. 

`id == None` suggests that the entity was not saved to the database yet (so `obj.save()` was not called before this code). If the script was created with `create_script`, this should be handled automatically. If started with `script = Script(...)`, then `script.save()` need to be called explicitly.

evennia/utils/test_resources.py
```diff
@@ -66,3 +66,42 @@ def tearDown(self):
...
+        super(EvenniaTest, self).tearDown()
+
+    def test_create_script(self):
+        class TestScriptA(DefaultScript):
+            def at_script_creation(self):
+                self.key = 'test_script'
+                self.interval = 10
+                self.persistent = False
+
+        script = create.create_script(TestScriptA, key='test_script')
```
Griatch: This will cause problems down the line due to the async execution of scripts (basically the callback will fire after the test method already concluded). You should probably make sure to run script.stop() at the end of all methods using scripts with an interval.

evennia/utils/test_resources.py
```diff
@@ -66,3 +66,42 @@ def tearDown(self):
...
         self.account2.delete()
         super(EvenniaTest, self).tearDown()
+
+
+import unittest
+import inspect
+
+class TestCreateScript(EvenniaTest):
+
+    def setUp(self):
```
Griatch: Placeholdsrs for future code? `setUp/tearDown` are empty and not needed.

Henddher:
Hi @Griatch. Thanks for the comments!

Sorry I wasn't clear of how I'm going with this.

My goal is to root-cause #1663; hence, I wrote 2 tests: One that succeeds (https://github.com/evennia/evennia/pull/1690/files#diff-2dc9e2642e2bd3cb1b63df0aee33ecd4R82) and the other that fails exactly as originally reported (in #1663) (https://github.com/evennia/evennia/pull/1690/files#diff-2dc9e2642e2bd3cb1b63df0aee33ecd4R95).

The first test case is "my control group": all works ok. 
The second test case is the "experimental group" showing the problem in #1663.

The stacktrace for the second is the following:

```
ipdb> bt
  /Users/henddher/.pyenv/versions/py2_evennia/bin/evennia(7)<module>()
      5 __file__ = '/Users/henddher/Documents/github/evennia/bin/unix/evennia'
      6 with open(__file__) as f:
----> 7     exec(compile(f.read(), __file__, 'exec'))

  /Users/henddher/Documents/github/evennia/bin/unix/evennia(10)<module>()
      8 
      9 from  evennia.server.evennia_launcher import main
---> 10 main()

  /Users/henddher/Documents/github/evennia/evennia/server/evennia_launcher.py(1338)main()
   1337         try:
-> 1338             django.core.management.call_command(*args, **kwargs)
   1339         except django.core.management.base.CommandError as exc:

  /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/core/management/__init__.py(131)call_command()
    130 
--> 131     return command.execute(*args, **defaults)
    132 

  /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/core/management/base.py(330)execute()
    329                 self.check_migrations()
--> 330             output = self.handle(*args, **options)
    331             if output:

  /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/core/management/commands/test.py(62)handle()
     61         test_runner = TestRunner(**options)
---> 62         failures = test_runner.run_tests(test_labels)
     63 

  /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/test/runner.py(603)run_tests()
    602         self.run_checks()
--> 603         result = self.run_suite(suite)
    604         self.teardown_databases(old_config)

  /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/test/runner.py(567)run_suite()
    566         runner = self.test_runner(**kwargs)
--> 567         return runner.run(suite)
    568 

  /Users/henddher/.pyenv/versions/2.7.14/lib/python2.7/unittest/runner.py(151)run()
    150         try:
--> 151             test(result)
    152         finally:

  /Users/henddher/.pyenv/versions/2.7.14/lib/python2.7/unittest/suite.py(70)__call__()
     69     def __call__(self, *args, **kwds):
---> 70         return self.run(*args, **kwds)
     71 

  /Users/henddher/.pyenv/versions/2.7.14/lib/python2.7/unittest/suite.py(108)run()
    107             if not debug:
--> 108                 test(result)
    109             else:

  /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/test/testcases.py(213)__call__()
    212                 return
--> 213         super(SimpleTestCase, self).__call__(result)
    214         if not skipped:

  /Users/henddher/.pyenv/versions/2.7.14/lib/python2.7/unittest/case.py(393)__call__()
    392     def __call__(self, *args, **kwds):
--> 393         return self.run(*args, **kwds)
    394 

  /Users/henddher/.pyenv/versions/2.7.14/lib/python2.7/unittest/case.py(329)run()
    328                 try:
--> 329                     testMethod()
    330                 except KeyboardInterrupt:

  /Users/henddher/Documents/github/evennia/evennia/utils/test_resources.py(105)test_create_script_w_one_repeat()
    104         ipdb.set_trace()
--> 105         script = create.create_script(TestScriptB, key='test_script')
    106         assert script is not None

  /Users/henddher/Documents/github/evennia/evennia/utils/create.py(217)create_script()
    216     # can be used.
--> 217     new_script.save()
    218     return new_script

  /Users/henddher/Documents/github/evennia/evennia/utils/idmapper/models.py(413)save()
    412             # trigger eventual monitors
1-> 413             _MONITOR_HANDLER.at_update(self, fieldname)
    414             # if a hook is defined it must be named exactly on this form

> /Users/henddher/Documents/github/evennia/evennia/scripts/monitorhandler.py(99)at_update()
     98         to_delete = []
---> 99         if obj in self.monitors and fieldname in self.monitors[obj]:
    100             for idstring, (callback, persistent, kwargs) in self.monitors[obj][fieldname].iteritems():

```

The `ipdb` console output I showed above https://github.com/evennia/evennia/pull/1690#issuecomment-429678504 depicts where the problem happens. 

It's at `.../evennia/utils/create.py(217)create_script()", line 217, in create_script
    new_script.save()`. I.e. it's when `save` operation is being executed.

The `ipdb` log shows what the contents of `self` (the script instance) and its `__dict__` contains when `_MONITOR_HANDLER.at_update(self, fieldname)` for `fieldname` with value `u'id'`.

screen shot 2018-10-15 at 6 29 21 pm

Unfortunately, I know nothing about `_MONITOR_HANDLER.at_update()` and why it expects that `fieldname` `u'id'` must be present in the `self.monitors`.  

I think that's where the **culprit of the problem is** but such is only exposed when the `repeats` attribute of the script is left uninitialized. 

WDYT?

Henddher:
Hi @Griatch ... did you get a chance to digest my previous [comment](https://github.com/evennia/evennia/pull/1690#issuecomment-430049676) ... there's a lot of info there so let me know if something's unclear and I'll explain where/why I am stuck.

It basically boils down to `_MONITOR_HANDLER.at_update()` throwing exception because `fieldname` `u'id'` is not present in `self.monitors.` Notice that the fieldname is `u'id'` (unicode key)

Henddher:
> Changing line `if obj in self.monitors and fieldname in self.monitors[obj]:` to:
> `if obj.pk and obj in self.monitors and fieldname in self.monitors[obj]:`
> 
> Should resolve this issue. As the error states, a model without a primary key is unhashable. Attempting to check if it's inside a dictionary requires it to be hashable, and raises that error. I'm uncertain why there isn't a check before that point of it not being saved (or having been deleted, which would also cause it to have the `.id` field set to `None`), which leads to that error, but that's what's causing the traceback on that line.

Thank you @TehomCD !

I'll ipdb it again and check if `obj.pk` is present and report back.

Henddher:
Ok, this is what I found out:

In the second test case (`repeats = 1`), `obj.pk` and `obj.id` are both `None`.

Here is the ipdb output showing so:

```pdb
  /Users/henddher/Documents/github/evennia/evennia/utils/test_resources.py(106)test_create_script_w_repeats_equal_1()
    105 
--> 106         script = create.create_script(TestScriptB, key='test_script')
    107         assert script is not None

  /Users/henddher/Documents/github/evennia/evennia/utils/create.py(231)create_script()
    230     # can be used.
--> 231     new_script.save()
    232     return new_script

  /Users/henddher/Documents/github/evennia/evennia/utils/idmapper/models.py(413)save()
    412             # trigger eventual monitors
--> 413             _MONITOR_HANDLER.at_update(self, fieldname)
    414             # if a hook is defined it must be named exactly on this form

> /Users/henddher/Documents/github/evennia/evennia/scripts/monitorhandler.py(99)at_update()
     98         to_delete = []
1--> 99         if obj in self.monitors and fieldname in self.monitors[obj]:
    100             for idstring, (callback, persistent, kwargs) in self.monitors[obj][fieldname].iteritems():

ipdb> p obj.id
None
ipdb> p obj.pk
None
ipdb> p obj
<TestScriptB: test_script>
ipdb> p self.monitors
defaultdict(<function <lambda> at 0x1062fa0c8>, {})
ipdb> p fieldname
u'id'

ipdb> p obj.__dict__
{'__dbclass__': <class 'evennia.scripts.models.ScriptDB'>, 
'_state': <django.db.models.base.ModelState object at 0x106d4dbd0>, 
'db_lock_storage': u'', '_is_deleted': True, 'db_interval': 10, 
'_createdict': {'account': None, 'obj': None, 'autostart': True, 'tags': None, 
'interval': None, 'persistent': None, 'start_delay': None, 'locks': None,
 'key': 'test_script', 'report_to': None, 'attributes': None, 'repeats': None, 'desc': None}, 
'id': None, 'db_is_active': True, 
'aliases': <evennia.typeclasses.tags.AliasHandler object at 0x10767a850>, 
'_db_account_cache': None, 'db_desc': u'', 
'db_typeclass_path': 'evennia.utils.test_resources.TestScriptB', 
'db_account_id': None, 
'db_date_created': datetime.datetime(2018, 10, 21, 1, 37, 43, 894189, tzinfo=<UTC>), 
'db_key': 'test_script', '_db_obj_cache': None, 
'permissions': <evennia.typeclasses.tags.PermissionHandler object at 0x107693bd0>, 
'db_repeats': 1, 'db_persistent': False, 
'nattributes': <evennia.typeclasses.attributes.NAttributeHandler object at 0x10769a610>, 
'_ndb_holder': <evennia.typeclasses.models.DbHolder object at 0x107693b90>, 
'_db_holder': <evennia.typeclasses.models.DbHolder object at 0x10746cf10>, 
'db_start_delay': False, 
'attributes': <evennia.typeclasses.attributes.AttributeHandler object at 0x10738cc50>, 
'delete': <bound method TestScriptB._deleted of <TestScriptB: test_script>>, 
'db_obj_id': None}

```

I'll start checking why we land here `obj.pk` and `obj.id` are None.

Henddher:
Here is where I am now:

Right before `post_save()`, `self.id` and `self.pk` have value `1` (valid pk). A little after, the values are `None`. See the `<<<<<<<<<` markers below:


```ipdb
> /Users/henddher/Documents/github/evennia/evennia/utils/create.py(221)create_script()
    220     # create new instance
--> 221     new_script = typeclass(**kwarg)
    222 

ipdb> p kwarg
{'db_key': 'test_script'}
ipdb> p typeclass
<class 'evennia.utils.test_resources.TestScriptB'>

> /Users/henddher/Documents/github/evennia/evennia/utils/create.py(225)create_script()
    224     new_script._createdict = dict(key=key, obj=obj, account=account, locks=locks, interval=interval,
--> 225                                   start_delay=start_delay, repeats=repeats, persistent=persistent,
    226                                   autostart=autostart, report_to=report_to, desc=desc,

ipdb> p new_script._createdict
{'account': None, 'obj': None, 'autostart': True, 'tags': None, 'interval': None, 'persistent': None, 'start_delay': None, 'locks': None, 'key': 'test_script', 'report_to': None, 'attributes': None, 'repeats': None, 'desc': None}

--> 231     new_script.save()

> /Users/henddher/Documents/github/evennia/evennia/utils/idmapper/models.py(391)save()
    390 
--> 391         if _IS_MAIN_THREAD:
    392             # in main thread - normal operation

ipdb> n
> /Users/henddher/Documents/github/evennia/evennia/utils/idmapper/models.py(393)save()
    392             # in main thread - normal operation
--> 393             super(SharedMemoryModel, self).save(*args, **kwargs)
    394         else:

ipdb> p args, kwargs
((), {})

> /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py(743)save()
    742         # nullable, allowing the save() would result in silent data loss.
--> 743         for field in self._meta.concrete_fields:
    744             if field.is_relation:

ipdb> p self._meta.concrete_fields
(<django.db.models.fields.AutoField: id>, 
<django.db.models.fields.CharField: db_key>, 
<django.db.models.fields.CharField: db_typeclass_path>, 
<django.db.models.fields.DateTimeField: db_date_created>, 
<django.db.models.fields.TextField: db_lock_storage>, 
<django.db.models.fields.CharField: db_desc>,
 <django.db.models.fields.related.ForeignKey: db_obj>, 
<django.db.models.fields.related.ForeignKey: db_account>,
 <django.db.models.fields.IntegerField: db_interval>, 
<django.db.models.fields.BooleanField: db_start_delay>, 
<django.db.models.fields.IntegerField: db_repeats>, 
<django.db.models.fields.BooleanField: db_persistent>, 
<django.db.models.fields.BooleanField: db_is_active>)

> /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py(768)save()
    767         using = using or router.db_for_write(self.__class__, instance=self)
--> 768         if force_insert and (force_update or update_fields):
    769             raise ValueError("Cannot force both insert and updating in model saving.")

ipdb> p self
<TestScriptB: test_script>
ipdb> p using
'default'
ipdb> p self.__class__
<class 'evennia.utils.test_resources.TestScriptB'>
ipdb> p force_insert
False
ipdb> p force_update
False
ipdb> p update_fields
None

> /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py(807)save()
    806 
--> 807         self.save_base(using=using, force_insert=force_insert,
    808                        force_update=force_update, update_fields=update_fields)

> /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py(828)save_base()
    827         if cls._meta.proxy:
--> 828             cls = cls._meta.concrete_model
    829         meta = cls._meta

ipdb> p cls._meta.concrete_model
<class 'evennia.scripts.models.ScriptDB'>

> /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py(831)save_base()
    830         if not meta.auto_created:
--> 831             pre_save.send(
    832                 sender=origin, instance=self, raw=raw, using=using,

ipdb> p origin
<class 'evennia.utils.test_resources.TestScriptB'>
ipdb> p self
<TestScriptB: test_script>
ipdb> p raw
False
ipdb> p using
'default'
ipdb> p update_fields
None

<<<<<<<<<<<<<<<<<<<<<<<<<<<

> /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py(846)save_base()
    845         if not meta.auto_created:
--> 846             post_save.send(
    847                 sender=origin, instance=self, created=(not updated),

ipdb> p origin
<class 'evennia.utils.test_resources.TestScriptB'>
ipdb> p self
<TestScriptB: test_script>
ipdb> p self.id
1

> /Users/henddher/Documents/github/evennia/evennia/utils/idmapper/models.py(402)save()
    401         new = False
--> 402         if "update_fields" in kwargs and kwargs["update_fields"]:
    403             # get field objects from their names

ipdb> p self.id
None
ipdb> p self
<TestScriptB: test_script>
ipdb> p self.pk
None
```

Any ideas/suggestions?

Henddher:
I'm starting to understand what's going on:

As the script is being saved, `at_script_creation()` is being **called back**. 
For each attribute being set, a `pre_save()` and `post_save()` happens.

I see that `self.id` and `self.pk` are both valid (= '1') until here:

```
  /Users/henddher/Documents/github/evennia/evennia/utils/test_resources.py(100)at_script_creation()
     99                 self.interval = 10
--> 100                 self.repeats = 1
    101                 self.persistent = False

  /Users/henddher/Documents/github/evennia/evennia/utils/idmapper/models.py(183)fset()
    182             else:
--> 183                 def fset(cls, val): return _set(cls, fieldname, val)
    184 

  /Users/henddher/Documents/github/evennia/evennia/utils/idmapper/models.py(138)_set()
    137                 update_fields = [fname] if _GA(cls, "_get_pk_val")(_GA(cls, "_meta")) is not None else None
--> 138                 _GA(cls, "save")(update_fields=update_fields)
    139 

  /Users/henddher/Documents/github/evennia/evennia/utils/idmapper/models.py(393)save()
    392             # in main thread - normal operation
--> 393             super(SharedMemoryModel, self).save(*args, **kwargs)
    394         else:

  /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py(808)save()
    807         self.save_base(using=using, force_insert=force_insert,
--> 808                        force_update=force_update, update_fields=update_fields)
    809     save.alters_data = True

> /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py(846)save_base()
    845         if not meta.auto_created:
--> 846             post_save.send(
    847                 sender=origin, instance=self, created=(not updated),

ipdb> p self, self.id, self.pk
(<TestScriptB: test_script>, 1, 1)

```

Then, as I step over `post_save()`, an exception is thrown `AttributeError` `_db_account_cache`:

```
ipdb> n
--Return--
None
> /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py(848)save_base()
    847                 sender=origin, instance=self, created=(not updated),
--> 848                 update_fields=update_fields, raw=raw, using=using,
    849             )

ipdb> import sys
ipdb> p sys.exc_info()
(<type 'exceptions.AttributeError'>, 
AttributeError("'TestScriptB' object has no attribute '_db_account_cache'",), 
<traceback object at 0x109ea0c68>)
```

Here is the stack

```
  /Users/henddher/Documents/github/evennia/evennia/utils/test_resources.py(100)at_script_creation()
     99                 self.interval = 10
--> 100                 self.repeats = 1
    101                 self.persistent = False

  /Users/henddher/Documents/github/evennia/evennia/utils/idmapper/models.py(183)fset()
    182             else:
--> 183                 def fset(cls, val): return _set(cls, fieldname, val)
    184 

  /Users/henddher/Documents/github/evennia/evennia/utils/idmapper/models.py(138)_set()
    137                 update_fields = [fname] if _GA(cls, "_get_pk_val")(_GA(cls, "_meta")) is not None else None
--> 138                 _GA(cls, "save")(update_fields=update_fields)
    139 

  /Users/henddher/Documents/github/evennia/evennia/utils/idmapper/models.py(393)save()
    392             # in main thread - normal operation
--> 393             super(SharedMemoryModel, self).save(*args, **kwargs)
    394         else:

  /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py(808)save()
    807         self.save_base(using=using, force_insert=force_insert,
--> 808                        force_update=force_update, update_fields=update_fields)
    809     save.alters_data = True

None
> /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py(848)save_base()
    847                 sender=origin, instance=self, created=(not updated),
--> 848                 update_fields=update_fields, raw=raw, using=using,
    849             )
```

Suggestions?

evennia/utils/test_resources.py
Henddher: Hm. I don't understand what you mean.

What is async in this code? the database save is async?
What `callback` are you referring to?
Where should I call `script.stop()`? After `create.create_script()` returns?

Henddher:
I added another test setting `repeats = 2` and it succeeds. 
😕 

i.e. the creation seems to fail only when `self.repeats = 1`. When `self.repeats` is left uninitialized or if the value is other than `1`, it works correctly.

Henddher:
**This is interesting!**

[`barter.py`](/evennia/evennia/blob/master/evennia/contrib/barter.py) sets `self.repeats = 1` right after `self.start_delay = True`. 

https://github.com/evennia/evennia/blob/0017547a8bfc8dcaccfae4937f24f613f50f9692/evennia/contrib/barter.py#L109-L118

I tried the same with `test_create_script_w_repeats_equal_1` and it now passes!

How can the order of attributes assignment affect the creation of the object?

(sounds like a side-effect kicking in when one attribute is set before another???? e.g. (`start_delay` before `repeats`)

evennia/utils/test_resources.py
Griatch: The entire repeating Script is async. The callback is the `at_repeat` method of the Script. Which, unless properly handled, may fire 10 seconds after this test-script was created, at which point the test suite has long since moved on to something else. Or worse, may have concluded and shut down the event loop already. So make sure to stop your script, for example in the tearDown method or yes, at the end of your test body.

Griatch:
@Henddher On typeclasses, assigning a field will automatically save that field (pre/post save signals, as you pointed out). When  `start_delay=True` is set, it means the script will not start its first iteration until after `interval` time. My guess is that for one reason or the other, the Script actually `.start` at this point, but if `start_delay` is set first, there will be an `interval`-long delay until the first iteration whereas if not set, it will fire immediately, causing some sort of race condition because everything is not ready yet.

_(Of course, you may ask why you would ever actually  *need* a one-repeat, no-start-delay script ... you could just put your `Script.at_repeat` code into a function and call that once ...)_

Henddher:
Thank you @Griatch (**now** I understand what you meant by https://github.com/evennia/evennia/pull/1690#discussion_r225140227

> When `start_delay=True` is set, it means the script will not start its first iteration until after `interval` time. My guess is that for one reason or the other, the Script actually `.start` at this point, but if `start_delay` is set first, there will be an `interval`-long delay until the first iteration whereas if not set, it will fire immediately, causing some sort of race condition because everything is not ready yet.

Aha! **that** explains it!
I'll try to verify if the script is running immediately. Will post again with whatever I observe.

> _(Of course, you may ask why you would ever actually _need_ a one-repeat, no-start-delay script ... you could just put your `Script.at_repeat` code into a function and call that once ...)_

That's very true - pragmatically speaking.

Should this behavior be documented more explicitly in the [Scripts Wiki](https://github.com/evennia/evennia/wiki/Scripts)?

I'd be happy to do so. WDYT?

* * * 

Now ... just for the sake of carrying out an interesting technical conversation. Ok, more like *philosophically speaking* ... bare with me please ...

Why would the script start executing **before** its `at_script_created()` call back completes?

I would have imagined a state-machine of sorts that each script follows:

```
[ created ]    ------>    [  running  ]     ------>    [  stop  ]
                           v          ^
                            \          \
                             \----->----\   
                   (if repeats > 0?  => repeats--; else goto stop)

```  

`[created]` state would transition to `[running]` state **after** `at_script_created()` callback completes and after `interval` expires iff `start_delay is True`.

Perhaps the state machine is not being honored? 
Should it be honored?
Without looking at that code and understanding it deeply, I would not dare to even guess what the effort of fixing this issue might be ... I just wonder about other "side effects" that could be lurking around ...

evennia/utils/test_resources.py
Henddher: Will do that!

Of course, that won't work for the *failing* test-case `repeats = 1` as the `create_script` function never gets to return.

Griatch:
@Henddher Sure, it might be worth making a brief note that if you are creating a 1-repeat, no-start-delay Script you should probably just be creating a normal function instead. 

Correct - the Script should not start its timer component before `at_script_creation` has finished. It doing so is a bug or regression and the cause of this issue in the first place most likely.

Henddher:
> ... worth making a brief note that if you are creating a 1-repeat, no-start-delay Script you should probably just be creating a normal function instead.

Will do that!

> Correct - the Script should not start its timer component before `at_script_creation` has finished. It doing so is a bug or regression and the cause of this issue in the first place most likely.

I'll take a look into that - getting into that machinery sounds interesting

Griatch:
@Henddher I added to the Script wiki page already as part another update, just so you know.

Henddher:
@Griatch sounds good - I just saw the edit in https://github.com/evennia/evennia/wiki/Scripts#properties-and-functions-defined-on-scripts

BTW (off topic), what's the workflow for wiki changes? Are there PRs for them? ... I haven't searched the wiki for howto-modify-evennia-wiki so I was wondering :P

Henddher:
Ok - 

Here is where I am in code (and ipdb): line 262.

https://github.com/evennia/evennia/blob/649a6db3a0b832d20453c967c81212a87113232e/evennia/scripts/scripts.py#L251-L316

As @Griatch confirmed earlier, setting any value within the context of `at_script_creation()` is triggering `pre_save.send()` and `post_save.send()`. 

The latter is calling  
 `django/dispatch/dispatcher.py(193)send()` **even though** the script is supposed to start at the end of `at_first_save()` line 316 above.

However, take a look at this:

```
    811     def save_base(self, raw=False, force_insert=False,
    812                   force_update=False, using=None, update_fields=None):
    813         """
    814         Handles the parts of saving which should be done only once per save,
    815         yet need to be done in raw saves, too. This includes some sanity
    816         checks and signal sending.
    817 
    818         The 'raw' argument is telling save_base not to save any parent
    819         models and not to do any changes to the values before save. This
    820         is used by fixture loading.
    821         """
    822         using = using or router.db_for_write(self.__class__, instance=self)
    823         assert not (force_insert and (force_update or update_fields))
    824         assert update_fields is None or len(update_fields) > 0
    825         cls = origin = self.__class__
    826         # Skip proxies, but keep the origin as the proxy model.
    827         if cls._meta.proxy:
    828             cls = cls._meta.concrete_model
    829         meta = cls._meta
    830         if not meta.auto_created:
    831             pre_save.send(
    832                 sender=origin, instance=self, raw=raw, using=using,
    833                 update_fields=update_fields,
    834             )
    835         with transaction.atomic(using=using, savepoint=False):
    836             if not raw:
    837                 self._save_parents(cls, using, update_fields)
    838             updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
    839         # Store the database on which the object was saved
    840         self._state.db = using
    841         # Once saved, this is no longer a to-be-added instance.
    842         self._state.adding = False
    843 
    844         # Signal that the save is complete
    845         if not meta.auto_created:
    846             post_save.send(
    847                 sender=origin, instance=self, created=(not updated),
--> 848                 update_fields=update_fields, raw=raw, using=using,
    849             )

```

Lines 840 and 842 seem to **mark** the object as **saved** although the `post_save.send()` has not completed yet. Could this be the issue?

Another interesting point is the list of receivers:

```
> /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/db/models/base.py(848)save_base()
    847                 sender=origin, instance=self, created=(not updated),
--> 848                 update_fields=update_fields, raw=raw, using=using,
    849             )

ipdb> l
    843 
    844         # Signal that the save is complete
    845         if not meta.auto_created:
    846             post_save.send(
    847                 sender=origin, instance=self, created=(not updated),
--> 848                 update_fields=update_fields, raw=raw, using=using,
    849             )
    850 
    851     save_base.alters_data = True
    852 
    853     def _save_parents(self, cls, using, update_fields):

ipdb> d
> /Users/henddher/.pyenv/versions/2.7.14/envs/py2_evennia/lib/python2.7/site-packages/django/dispatch/dispatcher.py(193)send()
    192             (receiver, receiver(signal=self, sender=sender, **named))
--> 193             for receiver in self._live_receivers(sender)
    194         ]

ipdb> p self._live_receivers
<bound method ModelSignal._live_receivers of <django.db.models.signals.ModelSignal object at 0x102eddf90>>
ipdb> p sender
<class 'evennia.utils.test_resources.TestScriptC'>
ipdb> p self._live_receivers(sender)
[<function update_cached_instance at 0x103a6bb18>, <function post_save at 0x103b2a848>]

```

Does it make sense that `post_save` is one of the `_live_receivers`?

Sounds strange ... I wonder if `post_save` was added to `_live_receivers` before it was the right time?

Any suggestion or experiment is welcome :)

Henddher:
Ok - 

So I verified that `at_script_created` actually completed before `script.start()` was called. 

So my first observation, although it was accurate (`id` and `pk` both `None`), it wasn't correct in terms of call stack.  The `id` being `None` wasn't within the context of `at_script_created` (as I first interpreted from the call stack), but much later after `script.start()` had completed but before `create_script()` had returned.

I just modified the test and changed `self.persistent = True` and the issue no longer happens. 

IOW, the issue seems to be bound to `self.persistent = False` if `self.repeats = 1`. 

Could it be possible that the script execution machinery decides to `delete` (or not commit to db) the script? disregarding the fact that the `create_script` function is being called and *expected* to return a valid script *no matter what*?

TehomCD:
What I would suspect is that somehow the script is being deleted immediately and that causes the errors we see, though I haven't had time to try to trace through the code to see if I could figure out why that's happening. If it's helpful, `script.stop()` causes script deletion, so if you can see that happening, like due to the script being invalid or having run its course due to limited intervals before the monitorhandler tries to update it, that'd be what's going on.

If it's helpful, a deleted model object will still be a valid object that can be returned from functions or otherwise interacted with, it'll just raise errors when you try to use it in certain contexts - like try to use it as a key in a dictionary since it's no longer hashable, or try to add it as the target of a foreignkey, or use it in Many-to-Many relationship. A model instance is known to be deleted if its primary key is set to None, and will then raise an error in all the above situations. A model instance that has yet to be saved will similarly have its primary key set to None, and will be assigned one when it's first saved to the database.

Henddher:
Hi @TehomCD, that's something I can try to debug and report back.

Nonetheless, kinda a philosophical question for you all though ... 

Shouldn't the script *not* start automatically when `create_script()` is being used?

I'd say the function should perhaps just create it, return it, and let the caller explicitly invoke `start()` afterwards?

I suppose there are other complications to take into account but I can't remember ever seeing anything within the realm of concurrency that auto-starts as part of the creation process. I am thinking of timers, threads, callbacks, etc. ... they are all created first, and then either started or scheduled explicitly. 

WDYT?

Griatch:
@Henddher Btw, I said above that the test suite was correctly placed but I was wrong. The `TestCreateScript` class belongs in a separate module in `evennia/utils/tests/`, for example `evennia/utils/tests/test_create_functions.py` or similar. The `test_resources.py` module is just for useful testing classes used by all other tests. Sorry for the confusion!

Griatch:
So the reason for this error is actually pretty simple. The process goes like this: 

1. In `script.save`, the script is saved correctly (so it gets an `.id`). This will fire the `at_post_save` signal
2. The `at_post_save` signal will fire `Script.at_first_start`, 
3. The beginning of `Script.at_first_start` will call `self.at_script_creation` which is benign. However at the end it will call `self.start`
4. `self.start` sets up the timer mechanism. Since `start_delay=False` it will fire `at_repeat` immediately.
5. After `at_repeat` has fired, the system will count down `repeats`. Aha - since `repeats=1` was set, we have run out of repeats! *The script is stopped*!  Herein lies the problem.
6. The script stopping means `self.delete()` is called, deleting the script (so it no longer has an `id`). 
7. Now the `at_post_save` signal is done and we continue in `script.save`.
8. In `script.save` we get to the assignment of Monitorhandler handlers but the Script *is already deleted* so we get the error. 

I'll need to ponder a bit how to best solve this.


Since `start_delay=None` and `repeats=1`, the `at_script_

Griatch:
@Henddher I added a fix for this and merged your branch. The solution I picked was to have have the `save` function check for the existence of `self.id` and return if the object (Script in this case) was already deleted as part of the very save operation. 

In the same vein, `create_script` will return `None` in this case. It's a pretty weird situation and I wonder if it would be better to have `start_delay` be the default rather than not.

Henddher:
Nice!

Considering the scope of scripts nowadays and backward compatibility, I’d vote for opt-out the auto-start. 

Iow, caller of ‘create_script()’ should pass ‘auto_start=False’ if it didn’t want the script to automatically start. That would not break backward compatibility for older scripts. 

For newer scripts, they would have nearly full control over the lifecycle (concurrency-wise speaking).

Ok but ....

I would **really** advocate to rename the function as ‘create_and_start()’. Break and then update all existing callers accordingly. Once that’s done, reintroduce ‘create_script()’ that just creates and persists the script so the caller would always control when the script is to be started.

In terms of auto deletion, I would extrapolate to “let the caller control it”. Should the caller want to auto-delete at ‘stop()’ if ‘repeats==0’, would be part of the script attributes: ‘destroy_after_stop=True’

Griatch:
@Henddher Might be an idea; that would be something for `develop` branch though.

evennia/utils/test_resources.py
```diff
@@ -66,3 +66,85 @@ def tearDown(self):
...
+                self.interval = 10
+                self.persistent = False
+
+        script = create.create_script(TestScriptA, key='test_script')
+        assert script is not None
+        assert script.interval == 10
+        assert script.key == 'test_script'
+        script.stop()
+
+    @unittest.expectedFailure
```
Henddher: You noticed I left this test as **failure expected**? 

I think it should be fixed and start passing instead.

Issue #1499, opened 2017-10-28
No way to clear object descs without resorting to @py


rnmissionrun:
#### Brief summary of issue / Description of requested feature:

Typing '@desc object=' does not clear the object description like it should.
 
#### Steps to reproduce the issue / Reasons for adding feature:

1. Typing the following command:

>> @desc here=
2. 
results in the description of the current room being set to 'here=', rather than being cleared.
3.

rnmissionrun:
I wrote a small patch that fixes this issue.

https://pastebin.com/idRRgu5J

Griatch:
Closed with #1691

PR #1691, opened 2018-10-17
Bug 1499: [READY FOR REVIEW AND MERGE] No way to clear object descs without resulting to `@py`

Henddher:
- Resolution for #1499
- Tests for `@desc obj=`

Failing tests demonstrates 2 issues (commit fde9f151889):
1. `@desc obj=` doesn't clear `obj` description.
2. `@desc obj=` sets the current room description as `'obj='`.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -589,12 +589,12 @@ class CmdDesc(COMMAND_DEFAULT_CLASS):
             self.edit_handler()
             return
 
-        if self.rhs:
+        if '=' in self.args:
             # We have an =
             obj = caller.search(self.lhs)
             if not obj:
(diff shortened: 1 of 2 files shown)


Henddher:
Hi @Griatch 

I added 2 test cases to validate the incorrect behavior originally reported in #1499. 

The patch in `buildings.py` produces the correct result without any regressions.

I opted to patch `@desc` command because it's lower risk compared to patching `muxcommand.py`. I think the root-cause is the following but of course that would affect all MUX commands.

https://github.com/evennia/evennia/blob/c69675e925f960533eb7d45768a200d8c37b8289/evennia/commands/default/muxcommand.py#L167

PR #1694, opened 2018-10-18
Implement local-first search for @link command

### LINKED ISSUE #1355: Feature Request: `@link` exits always searches globally. Searching local-first is more helpful.
When `@link` is used on a local exit to link or relink an exit to a proper destination, it returns a global list of exits when searching by name, even though there may be a single match locally.  This then requires all exits to be referenced by `#dbref`.

An initial local search, assuming the builder is in a valid location and exits are also in that location, would allow by-name searches to quickly match the intended local exit instead of returning all global exits.  Listing global exits should only occur under the condition that no local exits match the given exit name.

Example use, not reproduced error.
1.  `@link south=#2`
2. (Have 1000+ exits returned, as "south" is a common exit name.)
3. `@link #523=#2` (after resorting to using the `#dbref`

`@link south=#2` would first look locally, find the south exit, then link it.  If multiple south exits are local, then multimatch only those.  If no matches are found, then look for a match globally.

Current version, all OS.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46706524-feature-request-link-exits-always-searches-globally-searching-local-first-is-more-helpful?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TehomCD:
Add searching for local exits before trying to search globally.

Convenience change for the `@link` command to prevent always having to use dbrefs.

Resolves #1355

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -10,7 +10,7 @@ from evennia.objects.models import ObjectDB
 from evennia.locks.lockhandler import LockException
 from evennia.commands.cmdhandler import get_and_merge_cmdsets
 from evennia.utils import create, utils, search
-from evennia.utils.utils import inherits_from, class_from_module, get_all_typeclasses
+from evennia.utils.utils import inherits_from, class_from_module, get_all_typeclasses, variable_from_module
 from evennia.utils.eveditor import EvEditor
 from evennia.utils.evmore import EvMore
 from evennia.prototypes import spawner, prototypes as protlib, menus as olc_menus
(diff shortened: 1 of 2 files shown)


Griatch:
@TehomCD You are on a roll here :)

PR #1695, opened 2018-10-18
Add portal uptime to @time command.

### LINKED ISSUE #1420: `@uptime` only reports server uptime, not portal
Server uptime only tells half the story:

The dual nature of portal/server construction in Evennia isn't unique, but it's still uncommon, so it can be confusing to some to see server uptime shorter than the login times of some users (i.e. Some characters have been connected longer than the server's been up!).

I'd like to find the best way to also have the server report the portal's uptime in `@uptime` command, also.

1. Run Evennia for awhile, then `@reload`
2. `@uptime`

`@uptime` doesn't show portal uptime.

### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
All Evennia branches, all OS

I've considered the easiest way might be to look at the creation date on server/portal.pid, but another way could be to have portal tell server when it was started, when server connects to portal, so server will always have portal's most recent startup timestamp (unless portal is restarted but server is not, which isn't common, but may be possible).

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49087126-uptime-only-reports-server-uptime-not-portal?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TehomCD:
Puts in tracking for the uptime of the portal and transmits that data upon the server connecting to the portal via the amp bridge, then makes that information accessible in `gametime`, which is then displayed in the `@time` command.

Implementing an old feature request.

Resolves #1420

### DIFF
--- a/evennia/commands/default/system.py
+++ b/evennia/commands/default/system.py
@@ -710,6 +710,7 @@ class CmdTime(COMMAND_DEFAULT_CLASS):
         """Show server time data in a table."""
         table1 = EvTable("|wServer time", "", align="l", width=78)
         table1.add_row("Current uptime", utils.time_format(gametime.uptime(), 3))
+        table1.add_row("Portal uptime", utils.time_format(gametime.portal_uptime(), 3))
         table1.add_row("Total runtime", utils.time_format(gametime.runtime(), 2))
         table1.add_row("First start", datetime.datetime.fromtimestamp(gametime.server_epoch()))
         table1.add_row("Current time", datetime.datetime.now())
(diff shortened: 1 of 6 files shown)


Griatch:
I was a little sceptical at first about passing the portal uptime as its own kwarg in the AMP admin command. But I suppose it *is* a separate thing from the `sessiondata`, so it makes sense :) Looks good, merging.

PR #1696, opened 2018-10-18
Removes deleted characters from account playable characters list

strikaco:
Adds a line of code to the object deletion hook to check for and remove the object from an account's playable characters list.

For multisession modes greater than 1, when a character is deleted, it leaves a null value in the account object's playable characters list instead of shrinking the list.


Griatch:
Looking good!

PR #1697, opened 2018-10-19
Cross-platform installation script

strikaco:
Once dependencies are installed, this script finishes installation and configuration by making use of native Python libraries to download and install source and create an instance of a new game.

Only tested on Linux; I'll need some assistance testing it out on other platforms.

Virtualenv and git are confusing concepts to drop on newcomers. This PR aims to boil the entire installation process down to configuring a few variables and abstracting developers' interaction with virtualenv down to simply remembering to activate the damn thing :)

* Install dependencies.
* Run script.
* Activate virtualenv
* `evennia start`

This should resolve all issues of virtualenvs accidentally being created in bizarre places, games being installed within virtualenvs, installing Evennia using the incorrect virtualenv path, etc.

http://www.musoapbox.net/topic/2490/make-evennia-more-accessibility-ideas


Griatch:
An interesting concept! Not sure about adding it to the repo though. This feels more like the sort of third-party helper script that one would distribute in parallel with Evennia than being a part of it. It'd probably be better to have it as a third-party link we can refer to; something like this: 

```
Unsure about how to install? User strikaco has made a install script that automates 
most of the installation. You can download it [here]!
```

strikaco:
The idea was to submit it as a contrib and link to the script itself so the user could download it in advance of everything else, but my concern is that since I don't personally need this and can't test cross-platform without communal help, if I maintain it alone it won't get the attention it needs. 

Delegating initial user experience to third parties rarely ends well, so if it's simply not suitable for contrib, let's just close this one.

Griatch:
It's a good thought and a nice piece of code overall but no, I don't think it works as a contrib. So ok, let's close it.

PR #1706, opened 2018-10-24
Converts shared_login view function to a middleware class, updates middleware setting

strikaco:
* Renames `MIDDLEWARE_CLASSES` to `MIDDLEWARE` to adhere to Django 1.10+ new-style middleware expectations.
* Reimplements the `shared_login` view function as a middleware class.

I haven't tested it, but based on the current logic I believe that given the way we had this set up, `_shared_login` would only kick off if someone went to the index page. It would not run if someone tried to access a protected page, received an auth prompt, was redirected back to the destination page, and then hopped over to the webclient. The webdev would likely have to call this method on all views.

Implementing this as middleware forces the credentials to synchronize on all submitted requests.

As for the rename, we didn't have any custom middleware written so it's a matter of renaming a setting for the sake of future-proofing. 

https://docs.djangoproject.com/es/1.10/topics/http/middleware/#writing-your-own-middleware

### DIFF
--- a/evennia/web/utils/middleware.py
+++ b/evennia/web/utils/middleware.py
@@ -1,0 +1,60 @@
+from django.contrib.auth import authenticate, login
+from evennia.accounts.models import AccountDB
+from evennia.utils import logger
+
+class SharedLoginMiddleware(object):
+    """
+    Handle the shared login between website and webclient.
+
+    """
+    def __init__(self, get_response):
+        # One-time configuration and initialization.
+        self.get_response = get_response
+        
+    def __call__(self, request):
+        # Code to be executed for each request before
+        # the view (and later middleware) are called.
+        
+        # Process view
+        response = self.get_response(request)
+
+        # Code to be executed for each request/response after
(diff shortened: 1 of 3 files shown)


evennia/web/utils/middleware.py
```diff
@@ -0,0 +1,60 @@
+from django.contrib.auth import authenticate, login
```
Griatch: I would think this belongs in the `evennia/web` folder since it's explicitly dealing with web-related stuff. Maybe in `web/utils` ?

strikaco: I'm not sure what you're suggesting? It already is in `evennia/web`, specifically in `web/utils`. We already had `backends` and `general context` there so this location felt appropriate.

Griatch: Oh, I'm apparently tired and blind. :man_facepalming: For some reason I read `utils` and immediately assumed it was `evennia/utils`. Never mind then, carry on.

Issue #1707, opened 2018-10-25
Feature Request: @restart as alias to @reset


thenomain:
#### Brief summary of issue / Description of requested feature:

For us Mushers, we use "@restart" to, er, restart the game. So I think it'd be a good alias for @reset.

Thanks.

Griatch:
Thanks for the suggestion! I think having a `@restart` alias in game is fine, but it should apply to the `@reload` command instead of the `@reset` (see comments in #1708). So closing this since the PR could be changed. 

> Hint for the future: When you refer to `@restart`, try to put it within backticks. Otherwise you are pinging another github user named `restart` (I'm sure there is one). :)

Griatch:
Closing this, but you can still comment in the thread if you want :)

PR #1708, opened 2018-10-25
Adds '@restart' as alias for '@reset'

strikaco:
Adds restart as alias for reset.

Addresses #1707

### DIFF
--- a/evennia/commands/default/system.py
+++ b/evennia/commands/default/system.py
@@ -47,6 +47,7 @@ class CmdReload(COMMAND_DEFAULT_CLASS):
     @reset to purge) and at_reload() hooks will be called.
     """
     key = "@reload"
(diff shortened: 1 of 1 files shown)


Griatch:
I saw the issue and don't agree. The `@restart` command should be an alias for the `@reload` command instead - the main reason for this is consistency - if you run `evennia restart` on the command line, it acts as an alias for `evennia reload`.  But you can also argue that `@reset` is a more 'destructive' command than `@reload` and you should specify it explicitly if you want it. `@reload` is used far more often. 

@strikaco If you want to change this PR to apply this to the `@reload` command instead, feel free to do so (although it won't close that particular issue).

strikaco:
Fair enough. I don't think I've ever encountered a problem that a reload failed to fix anyway.

PR #1711, opened 2018-10-26
Implements generic CharacterDetailView and documents all existing views

strikaco:
* Documents all existing class-based views and forms
* Adds a generic DetailView for Character objects.
* Fixes a cosmetic issue where error messages were not rendering properly.

### DIFF
--- a/evennia/web/website/views.py
+++ b/evennia/web/website/views.py
@@ -139,12 +139,54 @@ def admin_wrapper(request):
 #
 
 class EvenniaIndexView(TemplateView):
-    # Display this HTML page
+    """
+    This is a basic example of a Django class-based view, which are functionally
+    very similar to Evennia Commands but differ in structure. Commands are used
+    to interface with users using a terminal client. Views are used to interface
+    with users using a web browser.
+    
+    To use a class-based view, you need to have written a template in HTML, and
+    then you write a view like this to tell Django what values to display on it.
+    
+    While there are simpler ways of writing views using plain functions (and 
+    Evennia currently contains a few examples of them), just like Commands, 
+    writing views as classes provides you with more flexibility-- you can extend
+    classes and change things to suit your needs rather than having to copy and
+    paste entire code blocks over and over. Django also comes with many default
+    views for displaying things, all of them implemented as classes.
+    
+    This particular example displays the index page.
+    
+    """
+    # Tell the view what HTML template to use for the page
     template_name = 'website/index.html'
     
-    # Display these variables on it
+    # This method tells the view what data should be displayed on the template.
     def get_context_data(self, **kwargs):
-        # Call the base implementation first to get a context object
+        """
+        This is a common Django method. Think of this as the website
+        equivalent of the Evennia Command.func() method.
+        
+        If you just want to display a static page with no customization, you
+        don't need to define this method-- just create a view, define
+        template_name and you're done.
+        
+        The only catch here is that if you extend or overwrite this method,
+        you'll always want to make sure you call the parent method to get a 
+        context object. It's just a dict, but it comes prepopulated with all 
+        sorts of background data intended for display on the page. 
+        
+        You can do whatever you want to it, but it must be returned at the end 
+        of this method.
+        
+        Kwargs:
+            any (any): Passed through.
+            
+        Returns:
+            context (dict): Dictionary of data you want to display on the page.
+        
+        """
+        # Always call the base implementation first to get a context object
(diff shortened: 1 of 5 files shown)


evennia/web/website/views.py
```diff
@@ -375,23 +705,22 @@ def form_valid(self, form):
         description = self.attributes.pop('desc')
         
         # Create a character
-        try:
-            character, errors = self.model.create(charname, account, description=description)
+        character, errors = self.model.create(charname, account, description=description)
```
Griatch: Reading this now I wonder if using the variable name `model` is not confusing. I know `model` is the term in Django but the `model` elsewhere in Evennia tends to refer to the actual real database model, such as `ObjectDB` or `AccountDB`. By contrast `self.model` here actually represents what we would elsewhere would call the `Typeclass` - that is, a Django proxy-model with some Evennia-specific changes. 

I'm not 100% sure about how much Django depends on `self.model` being called exactly that but probably it needs that name. I think it would help to create a get/setter property `typeclass` that uses `self.model` under the hood (for Django's benefit), but which allows you to use `self.typeclass.create` in situations like this. I just predict people will be confused by this nomenclature otherwise.

strikaco: I agree with you, but here's the other side-- Django is already a tricky and very intimidating beast for those unfamiliar with it. My goal was to avoid plastering abstractions over already complex Django concepts so people could leverage the extensive Django docs and community for their web development questions and not get frustrated trying to figure out where Django ends and Evennia begins.

We may call them Typeclasses because we can, but at the end of the day Typeclassed proxy models are still just models as far as Django and its CBVs are concerned. Most generic CBVs (and all tutorials involving them) expect a `model` field so changing this would break/require extension of a bunch of core classes and mixins and invalidate all existing Django tutorials by deviating from standard Django practice.

There are other inherent nomenclature conflicts that have no clear resolution either; Sessions as Evennia knows them have nothing to do with the Django Sessions framework, for example. I've tried to navigate this divide as delicately as I can with extensive comments ("this is a common Django thing" / "this is something special we have to do for Evennia" / "this Django thing has the same name as this Evennia thing but they are completely different") but in the end it might be safer to just accept that there's a way Evennia does things and a way Django does things and as long as we understand and document what those differences are, we don't need to overcomplicate an already-quite-complicated framework-upon-a-framework.

I'm not invested in it either way (I mean, *I* know what I'm doing), but in my opinion I don't think it does new developers any favors to force them to learn non-standard Django abstractions for no material benefit, especially when they're already complaining about its inaccessibility. But that's just my opinion, and I'm willing to admit I don't actually know what I'm doing after all :)

Griatch: I suppose the `EvenniaForm` could act as a base class for the form case of `self.model`. As for the views, there are already Evennia-specific mixins `CharacterMixin`, `AccountMixin` etc. Rather than have them all be their own bases, why not have them inherit from common base `EvenniaMixin`?

Griatch: Coincidentally, I really like your idea of likening views to Commands conceptually. I think it would help MU* users understand Django's role better if one were to present them next to each other - on one hand people entering MU commands that get handled by Commands and on the other hand people requesting web pages that get served by Django views ("Web Commands").

strikaco: > Coincidentally, I really like your idea of likening views to Commands conceptually. I think it would help MU* users understand Django's role better if one were to present them next to each other - on one hand people entering MU commands that get handled by Commands and on the other hand people requesting web pages that get served by Django views ("Web Commands").

Heh thanks. They really are truly alike-- everything you do in Command.func() is equally applicable within View.get(), hence my recent efforts to liberate logic from Commands and into methods accessible by both sides of the fence.

strikaco:
I've implemented a TypeclassMixin with a getter/setter that allows users to reference `self.typeclass` in favor of `self.model` within View methods. It's inherited by all the CBVs I wrote except Index, which doesn't need it.

They do still need to specify `model` at the class level though; I'm not experienced enough to know of a way around this (that doesn't involve rewriting a significant amount of core Django).

Griatch:
Hmm, not allowing `typeclass` on the class level leave things rather inconsistent from an API perspective. I'm pretty sure this could be rather easily handled transparently without changing any upstream Django code, but I would need to test a bit. Holding off merging this until I get some time to check it out.

Griatch:
I'm merging this for now, will consider API homogenization later :) Thanks for the nice work with it!

PR #1716, opened 2018-10-27
#1704: traceback when prototype key contains `#Number` [READY FOR REVIEW/MERGE]

### LINKED ISSUE #1704: `@spawn` traceback within evennia.prototypes.prototypes.protfunc_parser

Based on https://github.com/evennia/evennia/wiki/evennia.prototypes.protfuncs, it seems like all `protofunc`s are preceeded by `$` character. However, prototype values without preceding `$` are getting treated as `$protofunc`.

1. `@spawn {'key':'stone(#1234)'}`

```
18-10-24 01:46:49 [::] Traceback (most recent call last):
18-10-24 01:46:49 [::]   File "/usr/src/evennia/evennia/commands/cmdhandler.py", line 594, in _run_command
18-10-24 01:46:49 [::]     _progressive_cmd_run(cmd, ret)
18-10-24 01:46:49 [::]   File "/usr/src/evennia/evennia/commands/cmdhandler.py", line 193, in _progressive_cmd_run
18-10-24 01:46:49 [::]     value = generator.next()
18-10-24 01:46:49 [::]   File "/usr/src/evennia/evennia/commands/default/building.py", line 3098, in func
18-10-24 01:46:49 [::]     for obj in spawner.spawn(prototype):
18-10-24 01:46:49 [::]   File "/usr/src/evennia/evennia/prototypes/spawner.py", line 694, in spawn
18-10-24 01:46:49 [::]     create_kwargs["db_key"] = init_spawn_value(val, str)
18-10-24 01:46:49 [::]   File "/usr/src/evennia/evennia/prototypes/prototypes.py", line 735, in init_spawn_value
18-10-24 01:46:49 [::]     value = protfunc_parser(value)
18-10-24 01:46:49 [::]   File "/usr/src/evennia/evennia/prototypes/prototypes.py", line 583, in protfunc_parser
18-10-24 01:46:49 [::]     stacktrace=stacktrace, testing=testing, **kwargs)
18-10-24 01:46:49 [::]   File "/usr/src/evennia/evennia/utils/inlinefuncs.py", line 425, in parse_inlinefunc
18-10-24 01:46:49 [::]     retval = "".join(_run_stack(item) for item in stack)
18-10-24 01:46:49 [::]   File "/usr/src/evennia/evennia/utils/inlinefuncs.py", line 425, in <genexpr>
18-10-24 01:46:49 [::]     retval = "".join(_run_stack(item) for item in stack)
18-10-24 01:46:49 [::]   File "/usr/src/evennia/evennia/utils/inlinefuncs.py", line 423, in _run_stack
18-10-24 01:46:49 [::]     retval = "" if strip else func(*args, **kwargs)
18-10-24 01:46:49 [::]   File "/usr/src/evennia/evennia/prototypes/protfuncs.py", line 315, in obj
18-10-24 01:46:49 [::]     obj = _obj_search(return_list=False, *args, **kwargs)
18-10-24 01:46:49 [::]   File "/usr/src/evennia/evennia/prototypes/protfuncs.py", line 295, in _obj_search
18-10-24 01:46:49 [::]     raise ValueError("$obj: Query '{}' gave no matches.".format(query))
18-10-24 01:46:49 [::] ValueError: $obj: Query '#1234' gave no matches.
18-10-24 01:46:49 [EE] User input was: '@spawn {'key':'stone(#1234)'}'.
```

It seems like the parsing and its regular expression is incorrect?

**Code lifted from `evennia.prototypes.prototypes.protfunc_parser`**

```python
import re
_RE_DBREF = re.compile(r"(?<!\$obj\()(#[0-9]+)")
v = 'stone(#1234)'
_RE_DBREF.sub('$obj(\\1)', v)
'stone($obj(#1234))'
 ```

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65173364-spawn-traceback-within-evennia-prototypes-prototypes-protfunc_parser?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### PR DESCRIPTION
**Pull request by Henddher**
Fixes #1704 

- `#number` is not automatically converted to object search by `#dbref`. 
- New `$dbref()` protofunc is introduced as alias for `$obj(#Number)`.
- Traceback no longer happens as automatic object search is not done unless explicit `$obj()` or `$dbref()` protofunc is used.

### DIFF
--- a/evennia/prototypes/protfuncs.py
+++ b/evennia/prototypes/protfuncs.py
@@ -37,6 +37,7 @@ prototype key (this value must be possible to serialize in an Attribute).
 
 from ast import literal_eval
 from random import randint as base_randint, random as base_random, choice as base_choice
+import re
 
 from evennia.utils import search
 from evennia.utils.utils import justify as base_justify, is_iter, to_str
@@ -325,3 +326,15 @@ def objlist(*args, **kwargs):
 
     """
     return ["#{}".format(obj.id) for obj in _obj_search(return_list=True, *args, **kwargs)]
+
+
+def dbref(*args, **kwargs):
+    """
+    Usage $dbref(<#dbref>)
+    Returns one Object searched globally by #dbref. Error if #dbref is invalid.
+    """
+    _RE_DBREF = re.compile(r"\#[0-9]+")
+    if not args or len(args) < 1 or _RE_DBREF.match(args[0]) is None:
+        raise ValueError('$dbref requires a valid #dbref argument.')
+
+    return obj(args[0])
--- a/evennia/prototypes/protfuncs.py
+++ b/evennia/prototypes/protfuncs.py
@@ -44,6 +44,8 @@ from evennia.utils.utils import justify as base_justify, is_iter, to_str
 
 _PROTLIB = None
 
+_RE_DBREF = re.compile(r"\#[0-9]+")
+
 
 # default protfuncs
 
@@ -333,7 +335,6 @@ def dbref(*args, **kwargs):
     Usage $dbref(<#dbref>)
     Returns one Object searched globally by #dbref. Error if #dbref is invalid.
     """
-    _RE_DBREF = re.compile(r"\#[0-9]+")
     if not args or len(args) < 1 or _RE_DBREF.match(args[0]) is None:
         raise ValueError('$dbref requires a valid #dbref argument.')
 
--- a/evennia/prototypes/prototypes.py
+++ b/evennia/prototypes/prototypes.py
@@ -5,7 +5,6 @@ Handling storage of prototypes, both database-based ones (DBPrototypes) and thos
 
 """
 
-import re
 import hashlib
 import time
 from ast import literal_eval
@@ -32,8 +31,6 @@ _PROTOTYPE_RESERVED_KEYS = _PROTOTYPE_META_NAMES + (
 _PROTOTYPE_TAG_CATEGORY = "from_prototype"
 _PROTOTYPE_TAG_META_CATEGORY = "db_prototype"
 PROT_FUNCS = {}
-
-_RE_DBREF = re.compile(r"\$dbref\((\#[0-9]+)\)")
 
 
 class PermissionError(RuntimeError):
@@ -575,9 +572,6 @@ def protfunc_parser(value, available_functions=None, testing=False, stacktrace=F
         value = to_str(value, force_string=True)
 
     available_functions = PROT_FUNCS if available_functions is None else available_functions
-
-    # $dbref(#Number) becomes $obj(#Number)
-    value = _RE_DBREF.sub("$obj(\\1)", value)
 
     result = inlinefuncs.parse_inlinefunc(
         value, available_funcs=available_functions,
--- a/evennia/prototypes/prototypes.py
+++ b/evennia/prototypes/prototypes.py
@@ -33,7 +33,7 @@ _PROTOTYPE_TAG_CATEGORY = "from_prototype"
 _PROTOTYPE_TAG_META_CATEGORY = "db_prototype"
 PROT_FUNCS = {}
 
-_RE_DBREF = re.compile(r"(?<!\$obj\()(#[0-9]+)")
+_RE_DBREF = re.compile(r"\$dbref\((\#[0-9]+)\)")
 
 
 class PermissionError(RuntimeError):
@@ -576,7 +576,7 @@ def protfunc_parser(value, available_functions=None, testing=False, stacktrace=F
 
     available_functions = PROT_FUNCS if available_functions is None else available_functions
 
-    # insert $obj(#dbref) for #dbref
+    # $dbref(#Number) becomes $obj(#Number)
     value = _RE_DBREF.sub("$obj(\\1)", value)
 
     result = inlinefuncs.parse_inlinefunc(
--- a/evennia/prototypes/tests.py
+++ b/evennia/prototypes/tests.py
@@ -263,23 +263,36 @@ class TestProtFuncs(EvenniaTest):
                      "key": "ExampleObj"}
 
     def test_RE_DBREF(self):
-        def check_RE_DBREF(value, expected_value):
+        def check_RE_DBREF(value, expected_value,
+                exp_is_match, exp_is_search):
             try:
-                result = (
-                    protlib._RE_DBREF.match(value),
-                    protlib._RE_DBREF.search(value),
-                    protlib._RE_DBREF.sub("$obj(\\1)", value)
-                )
-                assert expected_value == result[2]
-            except Exception:
-                self.fail()
-                pass
+                mm = protlib._RE_DBREF.match(value)
+                ms = protlib._RE_DBREF.search(value)
+                sub1 = protlib._RE_DBREF.sub("$obj(\\1)", value)
+                assert expected_value == sub1
+                assert (exp_is_match and mm) or (not exp_is_match and mm is None)
+                assert (exp_is_search and ms) or (not exp_is_search and ms is None)
+            except Exception as e:
+                self.fail("%r" % ((value, mm, ms, expected_value, exp_is_match, exp_is_search, e),))
 
-        check_RE_DBREF('#1234', '#1234')
-        check_RE_DBREF('(#1234)', '(#1234)')
-        check_RE_DBREF('obj(#1234)', 'obj(#1234)')
-        check_RE_DBREF('$obj(#1234)', '$obj(#1234)')
-        check_RE_DBREF('obj($obj(#1234))', 'obj($obj(#1234))')
+        check_RE_DBREF('1234', '1234', False, False)
+        check_RE_DBREF('#1234', '#1234', False, False)
+        check_RE_DBREF('(#1234)', '(#1234)', False, False)
+        check_RE_DBREF('obj(#1234)', 'obj(#1234)', False, False)
+        check_RE_DBREF('dbref(#1234)', 'dbref(#1234)', False, False)
+        check_RE_DBREF('$obj(#1234)', '$obj(#1234)', False, False)
+        check_RE_DBREF('$dbref(#1234)', '$obj(#1234)', True, True)
+        check_RE_DBREF('obj($obj(#1234))', 'obj($obj(#1234))', False, False)
+        check_RE_DBREF('obj($dbref(#1234))', 'obj($obj(#1234))', False, True)
+        check_RE_DBREF('some #1234 value', 'some #1234 value', False, False)
+        check_RE_DBREF('some (#1234) value', 'some (#1234) value', False, False)
+        check_RE_DBREF('some obj(#1234) value', 'some obj(#1234) value', False, False)
+        check_RE_DBREF('some dbref(#1234) value', 'some dbref(#1234) value', False, False)
+        check_RE_DBREF('some $dbref(#1234) value', 'some $obj(#1234) value', False, True)
+        check_RE_DBREF('some obj($obj(#1234) value)', 'some obj($obj(#1234) value)', False, False)
+        check_RE_DBREF('some obj($dbref(#1234) value)', 'some obj($obj(#1234) value)', False, True)
+        check_RE_DBREF('some dbref($obj(#1234) value)', 'some dbref($obj(#1234) value)', False, False)
+        check_RE_DBREF('some dbref($dbref(#1234) value)', 'some dbref($obj(#1234) value)', False, True)
 
     @mock.patch("evennia.prototypes.protfuncs.base_random", new=mock.MagicMock(return_value=0.5))
     @mock.patch("evennia.prototypes.protfuncs.base_randint", new=mock.MagicMock(return_value=5))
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by Griatch** [comment]
Yes, it fails, as said. Not sure what the purpose of this PR is?

**Comment by Henddher** [comment]
The objective is to "fix" `protofunc_parser()` if it's clear what the expected functionality is to be like.

So far, it's just few tests of different combinations.

Once the requirements are clear, I should be able to push the fix ... unless @Griatch pushes the fixes before me (again) 😉

File: evennia/prototypes/tests.py [hunk 1]
```diff
@@ -312,11 +344,24 @@ def test_protfuncs(self):
...
+        self.assertEqual(protlib.protfunc_parser("#12345", session=self.session), '#12345')
+        self.assertEqual(protlib.protfunc_parser("nothing(#1)", session=self.session), 'nothing(#1)')
+        self.assertEqual(protlib.protfunc_parser("(#12345)", session=self.session), '(#12345)')
         self.assertEqual(protlib.protfunc_parser("$obj(Char)", session=self.session), '#6')
-        self.assertEqual(protlib.protfunc_parser("$obj(Char)", session=self.session), '#6')
+        self.assertEqual(protlib.protfunc_parser("obj(Char)", session=self.session), 'obj(Char)')
         self.assertEqual(protlib.protfunc_parser("$objlist(#1)", session=self.session), ['#1'])
+        self.assertEqual(protlib.protfunc_parser("objlist(#1)", session=self.session), 'objlist(#1)')
+        self.assertEqual(protlib.protfunc_parser("dbref(Char)", session=self.session), 'dbref(Char)')
+        self.assertEqual(protlib.protfunc_parser("$dbref(Char)", session=self.session), '<UNKNOWN>')
```
Henddher: This last test proves that `$dbref()` is expected to have a valid `#dbref` as passed in param.

evennia/prototypes/prototypes.py
```diff
@@ -33,7 +33,7 @@
 _PROTOTYPE_TAG_META_CATEGORY = "db_prototype"
 PROT_FUNCS = {}
 
-_RE_DBREF = re.compile(r"(?<!\$obj\()(#[0-9]+)")
+_RE_DBREF = re.compile(r"\$dbref\((\#[0-9]+)\)")
```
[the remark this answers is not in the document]
Griatch: This is actually incorrect. This custom regex is not needed if we remove the special case of  converting `#dbref`. So this entire regex should be removed and instead the new `dbref()` protfunc be added in `evennia/prototypes/protfuncs.py` where it can be handled by the normal protfunc parser.

evennia/prototypes/prototypes.py
```diff
@@ -576,7 +576,7 @@ def protfunc_parser(value, available_functions=None, testing=False, stacktrace=F
 
     available_functions = PROT_FUNCS if available_functions is None else available_functions
 
-    # insert $obj(#dbref) for #dbref
+    # $dbref(#Number) becomes $obj(#Number)
     value = _RE_DBREF.sub("$obj(\\1)", value)
```
Griatch: This should go away since we don't do custom parsing of `#dbref` here anymore.

evennia/prototypes/tests.py
Griatch: While passing, it's unfortunately passing for the wrong reason since this is not added as a proper protfunc.

evennia/prototypes/prototypes.py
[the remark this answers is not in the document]
Henddher: Well, this is the way I though of it:

`$obj()` takes `#dbref` or `key` or `alias` and does an object search.
The new `$dbref()` ** **enforces** ** that the passed in param is a `#number`. If it isn't, it becomes an invalid invocation of the protofunc.

At any rate, the implementation of `dbref()` would be identical to `obj()`. It would just be another thin wrapper of `search.search_object()`. 

https://github.com/evennia/evennia/blob/175af0715988f90c7898d612bf02f7d0e9bf635f/evennia/prototypes/protfuncs.py#L309-L318

On top of that, if the function was added, passing `key` and/or `alias` to `dbref()` would work just like `obj()` and there will be no enforcement of passing in a `#number`. Which is what this change is about: using the regex to alias `$dbref(#number)` as `$obj(#dbref)`.

As I mentioned https://github.com/evennia/evennia/issues/1704#issuecomment-433700848, `$dbref` would be a subset of `$obj()`. It would just allow to explicitly *typecast* a `#number` string as a `dbref` so object search would occur.

The regex is just to match `$dbref(#number)` **if and only if** the param matches `#number` and the protfunc invoked is `$dbref()`. Once the match is done, it just needs to call `$obj(#number)` which is the one that searches by `#dbref`, `key`, and `alias`es.

The tests I added show when the match occur (when it doesn't) and which protofunc will be invoked when the match happens (ie. invoke `$obj()` when matched).

https://github.com/evennia/evennia/pull/1716/files#diff-cde206fcfe341b1e066d1546d3666ef9R265

```python
    def test_RE_DBREF(self):
        def check_RE_DBREF(value, expected_value,
                exp_is_match, exp_is_search):
            try:
                mm = protlib._RE_DBREF.match(value)
                ms = protlib._RE_DBREF.search(value)
                sub1 = protlib._RE_DBREF.sub("$obj(\\1)", value)
                assert expected_value == sub1
                assert (exp_is_match and mm) or (not exp_is_match and mm is None)
                assert (exp_is_search and ms) or (not exp_is_search and ms is None)
            except Exception as e:
                self.fail("%r" % ((value, mm, ms, expected_value, exp_is_match, exp_is_search, e),))

         # no match
         check_RE_DBREF('1234', '1234', False, False)
        check_RE_DBREF('#1234', '#1234', False, False)
        check_RE_DBREF('(#1234)', '(#1234)', False, False)
        check_RE_DBREF('obj(#1234)', 'obj(#1234)', False, False)
        check_RE_DBREF('dbref(#1234)', 'dbref(#1234)', False, False)
        check_RE_DBREF('$obj(#1234)', '$obj(#1234)', False, False)

        # does match so parser will invoke $obj()
        check_RE_DBREF('$dbref(#1234)', '$obj(#1234)', True, True)
        
        # no match
        check_RE_DBREF('obj($obj(#1234))', 'obj($obj(#1234))', False, False)

        # does match so parser will invoke $obj()
        check_RE_DBREF('obj($dbref(#1234))', 'obj($obj(#1234))', False, True)

        # no match
        check_RE_DBREF('some #1234 value', 'some #1234 value', False, False)
        check_RE_DBREF('some (#1234) value', 'some (#1234) value', False, False)
        check_RE_DBREF('some obj(#1234) value', 'some obj(#1234) value', False, False)
        check_RE_DBREF('some dbref(#1234) value', 'some dbref(#1234) value', False, False)

        # does match so parser will invoke $obj()
        check_RE_DBREF('some $dbref(#1234) value', 'some $obj(#1234) value', False, True)

        # no match
        check_RE_DBREF('some obj($obj(#1234) value)', 'some obj($obj(#1234) value)', False, False)

        # does match so parser will invoke $obj()
        check_RE_DBREF('some obj($dbref(#1234) value)', 'some obj($obj(#1234) value)', False, True)

        # no match
        check_RE_DBREF('some dbref($obj(#1234) value)', 'some dbref($obj(#1234) value)', False, False)

        # does match so parser will invoke $obj()
        check_RE_DBREF('some dbref($dbref(#1234) value)', 'some dbref($obj(#1234) value)', False, True)

```

The other tests demonstrate that plain text `#number` **never** becomes an object search **unless** it's wrapped by `$dbref()` or `$obj()`. When wrapped, the result is the expected output of `$obj()` (i.e. the successful object search and its string `#dbref` representation:

https://github.com/evennia/evennia/blob/175af0715988f90c7898d612bf02f7d0e9bf635f/evennia/prototypes/protfuncs.py#L317

https://github.com/evennia/evennia/pull/1716/files#diff-cde206fcfe341b1e066d1546d3666ef9R347
```python
       # plain text, no search
        self.assertEqual(protlib.protfunc_parser("obj(#1)", session=self.session), 'obj(#1)') 
        self.assertEqual(protlib.protfunc_parser("$obj(#1)", session=self.session), '#1')
        self.assertEqual(protlib.protfunc_parser("dbref(#1)", session=self.session), 'dbref(#1)') # <<< (there is NO $)

       # search
        self.assertEqual(protlib.protfunc_parser("$dbref(#1)", session=self.session), '#1')

        # Bad function invocation
        self.assertEqual(protlib.protfunc_parser("$badfunc(#1)", session=self.session), '<UNKNOWN>')  

       # plain text, no search
        self.assertEqual(protlib.protfunc_parser("stone(#12345)", session=self.session), 'stone(#12345)')   
        self.assertEqual(protlib.protfunc_parser("#1", session=self.session), '#1') 
        self.assertEqual(protlib.protfunc_parser("#12345", session=self.session), '#12345') 
        self.assertEqual(protlib.protfunc_parser("nothing(#1)", session=self.session), 'nothing(#1)')
        self.assertEqual(protlib.protfunc_parser("(#12345)", session=self.session), '(#12345)

         # search by key
        self.assertEqual(protlib.protfunc_parser("$obj(Char)", session=self.session), '#6')

        # plain text, no search
        self.assertEqual(protlib.protfunc_parser("obj(Char)", session=self.session), 'obj(Char)') 

         # search
        self.assertEqual(protlib.protfunc_parser("$objlist(#1)", session=self.session), ['#1'])

        # plain text, no search
        self.assertEqual(protlib.protfunc_parser("objlist(#1)", session=self.session), 'objlist(#1)')
        self.assertEqual(protlib.protfunc_parser("dbref(Char)", session=self.session), 'dbref(Char)')
```

For the last test, I actually disagree:

The test asserts that it's an incorrect invocation of the protofunc `$dbref()`, as the passed in param is **not** a valid `#dbref`. 

```python
        self.assertEqual(protlib.protfunc_parser("$dbref(Char)", session=self.session), '<UNKNOWN>')
```

Notice that `Char` it's a valid key which returns a successful search `#6` when passed onto`$obj()` (one of the previous tests)

```python
# search by key
        self.assertEqual(protlib.protfunc_parser("$obj(Char)", session=self.session), '#6')
```

One thing I realized as I was writing the tests for `$obj()` and `$dbref()` - to validate that `#number` is **never** converted to `#dbref` automatically - is that none of them never assert that the protofunc was actually invoked. 

For instance, these cases:

```python
        self.assertEqual(protlib.protfunc_parser("$obj(#1)", session=self.session), '#1')
        self.assertEqual(protlib.protfunc_parser("$dbref(#1)", session=self.session), '#1')
        self.assertEqual(protlib.protfunc_parser("#1", session=self.session), '#1') 
```

They all return the **same value** although only the first 2 tests actually execute the `$obj()` function which produces `return "#{}".format(obj.id) `. The last one doesn't perform any search (as expected) and the result is a plain string (that looks like a `#dbref` but it isn't).

WADR, adding `def dbref()` to protofunc will do the same thing as long as `dbref()` function has its own regex to validate that the passed in param is a `#number` that represents a `#dbref` that can be searched by `obj()`

Something along these lines in protofunc.py

```python
def dbref(*args, **kwargs):
    if re.match(r'\#'\d+', args[0]):
        return obj(arg[0])
    else:
        return '<UNKNOWN>' # <<< or whatever constant is appropritate
```

evennia/prototypes/tests.py
Henddher: It's passing in the sense that the expected result is not the successful search **but** `"<UNKNOWN>"` which seems to be the keyword to signify failure executing.

evennia/prototypes/prototypes.py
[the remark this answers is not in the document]
Henddher: Hi @Griatch 

With this last commit dab2e3c5776, I hardened many of the existing cases by asserting that no object search happens unless `$obj()`, `$dbref()` and `$objlist()` are explicitly used as `value`.

I also grouped them so it's easier to digest what they do (no search, search, and bad).

Let me know if you really want to throw away `_RE_DBREF` and add `dbref` to `protfuncs.py`. IMHO, it'll be pretty much the same at the end.

Henddher:
- [ ] Update https://github.com/evennia/evennia/wiki/evennia.prototypes.protfuncs

Griatch:
@Henddher - I don't doubt that the unit tests return a consistent result. It's however irrelevant. 

The point of the protfunc system (as with the inlinefunc system it is based on) is to allow the user to customize their profuncs. If you supply a structure on the form `$dbref()` you must be able to identify that as a protfunc and know that you can replace it with your own `$dbref()` protfunc - maybe you only want it to perform a certain type of lookup or also accept dbref input without the '#' ... the point is that protfuncs are extandable by the developer and if *some* of them look like protfuncs but are actually custom, hard-wired regexes, it is not consistent not extendable.

Griatch:
@Henddher Just so you know, I'll be pretty busy in November and may not be so quick to respond in the coming weeks. A heads up so you don't think I'm ignoring you! :smile:

Henddher:
It took me longer to write the previous comment than discard the regex and add the new `dbref` :P

And only one test case failed (as expected):

```python
self.assertEqual(protlib.protfunc_parser("$dbref(Char)", session=self.session), '<UNKNOWN>')
```

It used to return `<UNKNOWN>` I believe because of the stack resolution executing protofuncs barked because `$dbref()` didn't exist - remember that in the previous code, `$dbref()` would be replaced with `$obj(#dbref)` if and only if `#dbref` was a `#number`.

Now with commit 0736485, it raises `ValueError` as `Char` is **not** a valid `#number` and hence an invalid `#dbref`.

Now, when facing the end user, I believe a traceback will appear instead of a text *error* message `<UNKNOWN>`.

Perhaps we'll need to revisit all the other `protfuncs` so if they raise `Exception`, the stack resolution logic should eat the exception, empty the stack, and just push `"<UNKNOWN>"` or another more suitable error string.

Thoughts?

evennia/prototypes/protfuncs.py
```diff
@@ -325,3 +326,15 @@ def objlist(*args, **kwargs):
...
     """
     return ["#{}".format(obj.id) for obj in _obj_search(return_list=True, *args, **kwargs)]
+
+
+def dbref(*args, **kwargs):
+    """
+    Usage $dbref(<#dbref>)
+    Returns one Object searched globally by #dbref. Error if #dbref is invalid.
+    """
+    _RE_DBREF = re.compile(r"\#[0-9]+")
```
Griatch: This can just as well go into the outer scope of this module so it's only compiled once.

evennia/prototypes/protfuncs.py
```diff
@@ -325,3 +326,15 @@ def objlist(*args, **kwargs):
...
+    if not args or len(args) < 1 or _RE_DBREF.match(args[0]) is None:
+        raise ValueError('$dbref requires a valid #dbref argument.')
```
Griatch: This is correct - for a `protfunc` (as opposed to their sibling, the `inlinefunc` which uses the same parsing engine), it's okay to raise an exception - this will abort the spawning and allow the user to correct the issue.

evennia/prototypes/tests.py
```diff
@@ -312,11 +313,83 @@ def test_protfuncs(self):
...
+            assert ('Char',) == mocked__obj_search.call_args[0]
+
+
+        # bad invocation
+
+        with mock.patch("evennia.prototypes.protfuncs._obj_search", wraps=protofuncs._obj_search) as mocked__obj_search:
+            self.assertEqual(protlib.protfunc_parser("$badfunc(#1)", session=self.session), '<UNKNOWN>')
+            mocked__obj_search.assert_not_called()
+
+        with mock.patch("evennia.prototypes.protfuncs._obj_search", wraps=protofuncs._obj_search) as mocked__obj_search:
```
Griatch: Looks good.

evennia/prototypes/protfuncs.py
Henddher: 👍

evennia/prototypes/tests.py
Henddher: Excellent @Griatch !

I just pushed 70eeb6f6 with the regex at module level as you suggested.

I think this PR can be merged and #1704 closed right after 🎉

Henddher:
@Griatch question though:

What about updating https://github.com/evennia/evennia/wiki/evennia.prototypes.protfuncs ?

Should I do it after the merge or will you take care of it?

Let me know please.

Griatch:
@Henddher Finally getting back to this! Thanks for your work on it. 
The wiki api page is auto-updated when I run the source crawler next, so there is no need (indeed it would be pointless) to edit it manually.

Henddher:
Thank you @Griatch!

🍺 🍺 !!

PR #1724, opened 2018-10-28
Adds CharacterListView

strikaco:
Adds a very basic and very generic ListView to display all Character objects for authenticated users.

Also fixes a permissions issue with CharacterDetailView.

Meant to add this in the initial suite but forgot.

### DIFF
--- a/evennia/web/website/views.py
+++ b/evennia/web/website/views.py
@@ -576,6 +576,42 @@ class CharacterMixin(TypeclassMixin):
         return self.typeclass.objects.filter(id__in=ids).order_by(Lower('db_key'))
         
         
+class CharacterListView(LoginRequiredMixin, CharacterMixin, ListView):
+    """
+    This view provides a mechanism by which a logged-in player can view a list
+    of all other characters.
+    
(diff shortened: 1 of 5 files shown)


evennia/web/website/views.py
```diff
@@ -577,6 +577,42 @@ def get_queryset(self):
...
+            queryset (QuerySet): Django queryset for use in the given view.
+        
+        """
+        account = self.request.user
+        
+        # Return a queryset consisting of characters the user is allowed to
+        # see.
+        ids = [obj.id for obj in self.model.objects.all() if obj.access(account, self.access_type)]
+        
+        return self.model.objects.filter(id__in=ids).order_by(Lower('db_key'))
```
Griatch: I suppose we should use the Typeclass mixin base here?

evennia/web/website/views.py
```diff
@@ -654,6 +690,24 @@ class CharacterDetailView(CharacterMixin, ObjectDetailView):
...
```
Griatch: Using `.typeclass` from the mixin would seem consistent here?

strikaco:
@Griatch Fixed!

PR #1728, opened 2018-10-29
Makes webclient/terminal auto-puppet character selected on website

strikaco:
This PR makes it so that if you select a character on the website and then hop into the game, you'll automatically become that character.

Also adds a link to the puppet activation URL as an action on the character management page.

Currently the character quick-select on the website is cosmetic only and does nothing out of the box.

### DIFF
--- a/evennia/commands/default/unloggedin.py
+++ b/evennia/commands/default/unloggedin.py
@@ -181,7 +181,7 @@ class CmdUnconnectedCreate(COMMAND_DEFAULT_CLASS):
 
         username, password = parts
 
-        # everything's ok. Create the new account account.
+        # everything's ok. Create the new account.
         account, errors = Account.create(username=username, password=password, ip=address, session=session)
         if account:
             # tell the caller everything went well.
--- a/evennia/web/utils/general_context.py
+++ b/evennia/web/utils/general_context.py
@@ -71,8 +71,19 @@ def general_context(request):
     if request.user.is_authenticated(): account = request.user
 
     puppet = None
-    if account and request.session.get('puppet'):
-        pk = int(request.session.get('puppet'))
+    if account:
+        # Get puppet previously selected from the website
+        if request.session.get('puppet'):
+            pk = int(request.session.get('puppet'))
+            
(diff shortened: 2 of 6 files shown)


Griatch:
I'd think this depends highly on `MULTI_SESSION_MODE`. Did you test this with a mode > 1? I imagine it has no effect there (which is correct, but might maybe come off as confusing)

strikaco:
Hmm true. On mode 2 it just takes you to the 'select a character' screen in the terminal.

Griatch:
I wonder if one would need to handle this case somehow, maybe turn off this functionality depending on `MULTI_SESSION_MODE`?

strikaco:
I'm not sure that's desirable though. The function of the dropdown system was supposed to be to allow devs to create mail forms, comments, bulletin boards, whatever on the site and let players interact with web resources in the context of a character-- specifically designed *for* the higher multisession modes. 

This particular PR just gives the dropdown more of a function instead of the appearance of doing nothing. "You become ___"-- great, now what?

Griatch:
@strikaco I think the solution to this is to make it very clear that this is just a stub functionality. If nothing happens when you select it, people will mark it as a bug. It would be clearer if instead it lead to the Not-implemented page (or, I don't know, maybe a popup?) saying something to the effect of 
```
This functionality is not implemented by default - it's up to you 
to add the functionality you want for your own game!
You can find a useful turorial for how to do this [here].
```

Griatch:
@strikaco This appears to only need the stub functionality for other multisession-modes. Or do you have another suggestion?

strikaco:
I did not see your Dec 2 suggestion; I think that should be fine. Will try to implement when I get a chance.

Griatch:
@strikaco If this is to go into Evennia 0.9, this needs to be finalized very soon. Please let me know what the status is.

Griatch:
@strikaco Any plans of wrapping this up or should this PR close?

strikaco:
Yeah I'll go ahead and close this; I no longer remember where I was going with it.

PR #1749, opened 2018-11-22
Allow the 'makemessages' and 'compilemessages' commands

vincent-lg:

Since the new evennia launcher, both `makemessages` and `compilemessages` were broken.  This offers a solution to fix it and allow both commands to work.

Translation of the main Evennia server wasn't possible. Notice a few changes though, some of which may not be appropriate:

- In `settings_default`, the `GAMEDIR` can now be the `EVENNIA_DIR` if the user is running from `evennia`.  We'll let another method check whether it's a problem or not.  Having this check performed in the settings AND in the launcher was a understandable redundancy that didn't help here.  This test is still performed, however, so error messages are correctly displayed if the user isn't in a valid game directory and tries to perform a command like `evennia start`.
- In the launcher, `init_gamedir` now takes an additional optional argument to indicate whether this has to be a valid game directory or not.
- Still in the launcher, `makemessages` is handled as a special case: the `-l` option, used by `makemessages`, conflicts with the option to tail log files.  We cannot use the long option name due to the way `makemessages` handles the `-l/--locale` option (as a list, not a simple `str`).  Therefore, I had to add a tweak just for this command.  The syntax now is:

        evennia makemessages <locale-code>

    The `-l` option will be inserted in between.

### DIFF
--- a/evennia/server/evennia_launcher.py
+++ b/evennia/server/evennia_launcher.py
@@ -1641,7 +1641,7 @@ def error_check_python_modules():
 #
 # ------------------------------------------------------------
 
-def init_game_directory(path, check_db=True):
+def init_game_directory(path, check_db=True, need_gamedir=True):
     """
     Try to analyze the given path to find settings.py - this defines
     the game directory and also sets PYTHONPATH as well as the django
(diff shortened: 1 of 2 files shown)


Griatch:
@vincent-lg Ok. Would you mind updating the i18n wiki page to match?

PR #1755, opened 2018-12-11
Fix #1754: WindowsError is replaced with OSError

### LINKED ISSUE #1754: evennia fails to start on non windows systems
### Brief summary of issue / Description of requested feature: ###

When trying to start an evennia server on a non windows system it fails due to the use of the platform specific python exception WindowsError. This regression appears to have been added recently as part of PR #1748. 

Since WindowsError is defined as a subclass of OSError, this should probably be changed to catch OSError and not WindowsError. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67352131-evennia-fails-to-start-on-non-windows-systems?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522101&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vincent-lg:

The `WindowsError` exception was checked for compliance and is now removed to support non-Windows systems.

### DIFF
--- a/evennia/__init__.py
+++ b/evennia/__init__.py
@@ -105,7 +105,7 @@ def _create_version():
         print(err)
     try:
         version = "%s (rev %s)" % (version, check_output("git rev-parse --short HEAD", shell=True, cwd=root, stderr=STDOUT).strip())
-    except (IOError, CalledProcessError, WindowsError):
+    except (IOError, CalledProcessError, OSError):
         # ignore if we cannot get to git
         pass
     return version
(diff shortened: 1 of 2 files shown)


Griatch:
Unclear what this means. Is the WindowsError not used anymore?

vincent-lg:
As far as I can figure: the `WindowsError` exception doesn't exist on Linux.  All my tests seem to indicate that.  Why this doesn't pose a problem for Evennia now, in most cases, is that `WindowsError` is often tested at the end of an `except` clause with many exceptions.  Since no exception of this nature happens I guess Python doesn't check the name.  But entering `WindowsError` in Python on Linux raises a `NameError`.

However, `OSError` is the parent of `WindowsError` and it hopefully exists on all systems.  That's why this PR replaces `WindowsError` by `OSError` in two places. (It should but I have trouble seeing whether I posted the right content)

PR #1756, opened 2018-12-18
BugFix - Delete Prototypes using spawn command

Antrare:
Spawn/delete <prototypekey> was not actually deleting anything.  This enables the command to work as expected.


Griatch:
Thanks for the PR!

PR #1766, opened 2019-01-06
Fixed a loop on the menu_login contrib

vincent-lg:
This fix solves an infinite loop in the menu_login contrib.

Empty keys in menu options didn't use to generate infinite loops.  Although it makes sense, this needed to be fixed.


Griatch:
This fixes the symptom, not the cause. This is a bug in EvMenu and should be fixed there. Not merging this, instead opening https://github.com/evennia/evennia/issues/1769 .

PR #1768, opened 2019-01-07
Batchcode Build.py for the Tutorial World.

CloudKeeper:
Brief overview of PR changes/additions
This is a draft of a proposed Batchcode Build.py for the Tutorial World.

Motivation for adding to Evennia
Provide example Batchcode for a sizeable play space and direct comparison to the Tutorial World bulid.ev

Other info (issues closed, discussion etc)
Uploaded at this stage to get feedback on layout and other stylistic decisions.
Tested but would appreciate a second set of eyes.

### DIFF
--- a/evennia/contrib/tutorial_world/build.py
+++ b/evennia/contrib/tutorial_world/build.py
@@ -1,0 +1,1718 @@
+"""
+Evennia batchfile - tutorial_world
+
+Cloud_Keeper 2018 
+Adapted from build.ev by Griatch 2011, 2015
+
+This batchfile sets up a starting tutorial area for Evennia.
+
+This uses the custom script parents and code snippets found in the
+same folder as this script; Note that we are not using any
+modifications of the default player character at all (so you don't
+have to change anything in any settings files). We also don't modify
+any of the default command functions (except in states). So bear in
+mind that the full flexibility of Evennia is not used to its maximum
+potential here.
+
+To load this file, load the file as user #1 with
+
+    @batchcode contrib.tutorial_world.build
+
+The area we are building looks like this:
+
+    ? 03,04
+    |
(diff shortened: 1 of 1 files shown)


Griatch:
I think this looks good at a first inspection! Since this is a part of the tutorial however, It think it may be an idea to use the full functionality of the batchcode (while mimicing the functionality of the original `build.ev`), by using the batchcode functionality of splitting the build into sections with a `#HEADER` block and several  `#CODE` blocks (maybe one per room?). That way a user could step through the build and also re-run individual parts of it. It would then also serve as something to point to in the batchcode documentation ("look here for a longer example of batchcode"...).

Griatch:
Ping to see if the bot echoes it this time ...

CloudKeeper:
Roger that.

Griatch:
@CloudKeeper Plan to wrap this one up so it can be merged?

CloudKeeper:
How would you like the block separation done? We could have all the rooms created in the first block and then subsequent blocks will just be detailing the rooms individually? Or we could break it into a room each where it will make the rooms one at a time but only connect the exits if the destination room exists, but then each block will be artificially complicated with checks. What do you reckon?

Griatch:
Hm, the batchcmd version already showcases the room-by-room stepping, so maybe it's not too bad an idea to build in one block and then gradually customize in others.

Griatch:
@CloudKeeper Any progress on this?

Griatch:
@CloudKeeper If this is to go into Evennia 0.9, this needs to be finalized very soon. If not, we should probably close this. Let me know.

Griatch:
Since this has not seen any updates in a good while, I'm closing this PR. If you get around to try to finish it in the future, ping me here and I'll re-open.

PR #1765, opened 2019-01-12
Encoding and Python 3

vincent-lg:
#### Brief summary of issue

The current branch running on Python 3.5 of Python has some issues with encoding that most ASCII users wouldn't notice.  Basically, it's either ASCII, or utf-8, the rest doesn't work.  And while I would be happy to have everyone using utf-8, it's clearly not a good choice, particularly with Evennia's settings.

To see the problem:

1. Install a fresh Evennia with the develop-py3 branch.
2. In the `server/conf/connection_screens.py` file, append some utf-8 in the file (see mine below).
3. With a client that uses a different encoding (mine uses latin-1 for instance), connect to localhost:4000.  You will see an error on the accented character, as you should (default encoding isn't changed).
4. Use the `@encoding` command to change your output encoding.  In my case: `@encoding latin-1` .
5. And type `look` to see... the exact same error.

```python
# connection_screen
CONNECTION_SCREEN = """
médor
|b==============================================================|n
 Welcome to |g{}|n, version {}!

 If you have an existing account, connect to it by typing:
      |wconnect <username> <password>|n
 If you need to create an account, type (without the <>'s):
      |wcreate <username> <password>|n

 If you have spaces in your username, enclose it in quotes.
 Enter |whelp|n for more info. |wlook|n will re-show this screen.
|b==============================================================|n""" \
    .format(settings.SERVERNAME, utils.get_evennia_version())
```

(Yes, it's the é in "médor", which is a beautiful dog name).

What happens?  I would like to know.  It's not just an error of the `@encoding` command, it's the encoding system that's a bit broken.  Full debugging didn't allow me to find the proper solution... but I identified some possible root causes and here's my thinking:

1. `to_str` is called A LOT.  This is probably a very bad idea for encoding support.  In Python 3, every string within the program should be utf-8 (str type).  Converting to other encoding (bytes) should only happen before every evennia-related processing (input), and just before submitting to the client (output).  Some functions (like `pickle.dumps` which is used by `amp.dumps`) return bytes.  Is that where the encoding should be handled?  Turning this bytes into a string afterward (look at the `server.portal.amp.dumps` function) might break some things.
2. The only point where I found the ENCODING protocol flag being used is in `server/sessionhandler.py`.  It converts bytes to string following the encoding.  Again, I would think the other way around would be preferable: converting string to bytes before sending them... but I don't know enough of Evennia's structure to be sure.  And since there's the AMP to consider between them, I don't feel like coming up with a solution by myself, without guidance.

In my opinion, the greatest problem is the excessive `to_str` calls, but again, that might not be the reason for the problem.  If anyone wants additional debugging and can orient me in the right direction, I'll take it.

* Evennia branch: develop-py3

### DIFF
--- a/evennia/server/evennia_launcher.py
+++ b/evennia/server/evennia_launcher.py
@@ -638,7 +638,7 @@ class AMPLauncherProtocol(amp.AMP):
         else:
             status = pickle.loads(status)
             callback(status)
-        return {b"status": b""}
+        return {"status": pickle.dumps(b"")}
 
 
 def send_instruction(operation, arguments, callback=None, errback=None):
--- a/evennia/server/portal/amp.py
+++ b/evennia/server/portal/amp.py
@@ -73,11 +73,11 @@ Content-Type: text/html
 # Helper functions for pickling.
 
 def dumps(data):
-    return to_str(pickle.dumps(to_str(data), pickle.HIGHEST_PROTOCOL))
+    return pickle.dumps(data, pickle.HIGHEST_PROTOCOL)
 
 
 def loads(data):
-    return pickle.loads(to_str(data))
+    return pickle.loads(data)
 
 
 @wraps
--- a/evennia/server/portal/telnet.py
+++ b/evennia/server/portal/telnet.py
@@ -243,7 +243,7 @@ class TelnetProtocol(Telnet, StatefulTelnetProtocol, Session):
             line (str): Line to send.
 
         """
-        line = bytes(line, 'utf-8')
+        line = self.try_encode(line)
         # escape IAC in line mode, and correctly add \r\n (the TELNET end-of-line)
         line = line.replace(IAC, IAC + IAC)
         line = line.replace(b'\n', b'\r\n')
@@ -343,7 +343,7 @@ class TelnetProtocol(Telnet, StatefulTelnetProtocol, Session):
                                          strip_ansi=nocolor, xterm256=xterm256)
                 if mxp:
                     prompt = mxp_parse(prompt)
-            prompt = prompt.encode()
+            prompt = self.try_encode(prompt)
             prompt = prompt.replace(IAC, IAC + IAC).replace(b'\n', b'\r\n')
             prompt += IAC + GA
             self.transport.write(mccp_compress(self, prompt))
--- a/evennia/server/portal/telnet.py
+++ b/evennia/server/portal/telnet.py
@@ -243,7 +243,7 @@ class TelnetProtocol(Telnet, StatefulTelnetProtocol, Session):
             line (str): Line to send.
 
         """
-        line = self.try_encode(line)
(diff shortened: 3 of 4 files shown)


vincent-lg:
So I've tried to fix the issue.  The following workflow has been adopted:

1. I've maintained the input bytes-to-str conversion as it was, since it seemed to be working correctly.  Notice that more unity might be preferable but as it is, the fix works and doesn't disturb the code too much.
2. I removed the `to_str` in `amp.dumps` and `amp.loads`.  I didn't notice any problem with that and it seemed to fix quite a lot of bugs but I don't know how wise it was.
3. I added a method in the session, called `try_encode`.  Its job is simple: to try to encode a `str` into `bytes`.  If it can, it uses the protocol flag.  This method is to be called right before writing to the protocol.  In other words, it is only used for output (and only at the very last moment).

> Notice that the calls to `try_encode` are done in `telnet.py` directly, which might not be the best option.  The webclient keeps on working and handling accents correctly, but other protocols might have to be updated.  Before doing that, though, I would like to know if what I'm doing is wise and advisable.  I tried to maintain code compatibility and improve encoding/decoding logic within Python 3's new workflow.

Overall, the fix works: the provided example works as expected (that is, the bytes for é are different for utf-8 and latin-1 as they should be).

Griatch:
That `try_encode` appears to just be patching the symptom rather than fixing the issue. I'll look some more at it, but it seems wrong to edit this at the telnet level. If things are wrong there, they will be wrong before that level too, meaning this will, as you say, not help for other protocols.

vincent-lg:
It's not just a "patch" to deal with the encoding.  It actually moves the encoding process.  The problem is that telnet (and other protocols) call `self.protocol.write`, which is (I think) the very last stage, and `protocol.write` waits for bytes.  So I do the converting from str to bytes just before calling it.  I don't really know how that could be better fixed, perhaps a generic method to call instead of writing to the protocol, which method would both convert to bytes and then send over the wire.  Open debate.

As for the workflow, everything remains a str (including over AMP) until the moment when we write to the actual protocol.  The `amp.dumps` returns bytes, to be sure, but it's a pickled version (pickle now returns bytes, not str).  There's no encoding taking place before sending to the AMP.  So the AMP is responsible for the output encoding at least.

vincent-lg:
I made a few changes to clarify.  The `try_encode` had a somewhat unhelpful (and a bit untrustworthy) name.  I changed it to `encode_output` and clarified the documentation a little.  The good news is, as far as I can see, no other protocol needs to be changed, though I only checked the protocols in `server/portal`.  The `encode_output` can still work with `str` and `butes` (in the latter case, the output is returned as-is) however.  But I still believe this is much cleaner this way, as it happens just before sending the message to the protocol, or at the very least, just before protocol-specific actions that should work on bytes anyway.  My opinion :D .

PR #1771, opened 2019-01-19
Fix accents in Evennia py3, bring compatibility with Evennia 0.8 data

vincent-lg:

I jsut removed yet another `to_str()` call on the `dbserializer`.  Not only did it fix my issue, but it makes old databases (generated with Evennia 0.8) completely usable now.

It seems like a win-win.  Just a tiny change, Evennia works better and maintains compatibility of data, which was a concern for me, with some users having a much more important database than I.  I worried it would generate problems, but it seems more than stable and to actually fix issues.

The thing is to get used to the idea that pickle needs bytes, and to not convert these bytes to str, especially not with `to_str`.  Basically, pickle will return binary data, and using `to_str` on it will attempt encoding the binary which is dangerous and unnecessary.  So now, I let pickle do its magic and if it wants bytes, bytes it should have.  In terms of exchange with the AMP, bytes are used to transfer data, but inside these binary representation, it's plain str.  str are only converted to bytes when receiving data from the user and just before sending data to the user.

### DIFF
--- a/evennia/utils/dbserialize.py
+++ b/evennia/utils/dbserialize.py
@@ -30,7 +30,7 @@ except ImportError:
 from django.core.exceptions import ObjectDoesNotExist
 from django.contrib.contenttypes.models import ContentType
 from django.utils.safestring import SafeString, SafeBytes
-from evennia.utils.utils import to_str, uses_database, is_iter
+from evennia.utils.utils import uses_database, is_iter
 from evennia.utils import logger
 
 __all__ = ("to_pickle", "from_pickle", "do_pickle", "do_unpickle",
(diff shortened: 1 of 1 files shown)


vincent-lg:
`pickle` updated its protocol between py2 and py3.  Python 2.7 has highest protocol of 2, while Python 3.6 (the one I use on Windows for Evennia now) has highest protocol of 4.  So "old" attributes are stored in a protocol 2 pickle object, while new attributes are stored in protocol 3 or 4.  `pickle` will jungle with different versions, as the protocol version is included in the pickled binary if I'm not mistaken.  However, this might cause problems if you want to search in attributes.  What protocol version to choose, and how to match with different protocol versions?  It could be worth forcing all attributes to update to pickle protocol 4, this could be a single database migration.

Griatch:
@vincent-lg Yes, the attribute version is important and we should make sure to lock versions for general deserialization. Haven't looked at migrations yet, but yes we'll need that before we  can use it. I don't worry too much about people losing their data though - this is an unstable development branch, people should not use that with a production system, period. 
As for Attributes, they *should* not be affected since these are using the PickleField which is hardcoded locked to protocol 2. The main things affected would be ServerConfig, which uses the custom deserializer.

PR #1778, opened 2019-02-09
Correcting typo - "prot" to "proto"

CloudKeeper:
Caught by LilCasper


Griatch:
And that's why we need more unit tests :) Thanks for the fix, both of you!

Issue #1780, opened 2019-02-12
Remnants of card-block (should be card-body) in web defaults


poisonbl:
#### Brief summary of issue / Description of requested feature:

Formatting issues (lack of padding) on several page templates remain after the login/logout pages were patched in PR#1463

Affected templates: 404.html, 500.html. evennia_admin.html. tbi.html. flatpages/default.html

Will add a PR for them shortly.

Griatch:
Thanks! Closed since PR merged.

PR #1781, opened 2019-02-12
Switching card-block to card-body.

poisonbl:

Fix the missing padding in the default 404, 500, tbi, admin, and the flatpage templates by setting the correct class on the cards.

Fixes issue #1780


Griatch:
Thanks for the fix!

PR #1783, opened 2019-02-16
command spelling correction in tutorial build.ev @home ghost = to @se…

jaggiJ:
correcting command in tutorial build.ev from @home ghost = tut#11 to @sethome = tut#11

### DIFF
--- a/evennia/contrib/tutorial_world/build.ev
+++ b/evennia/contrib/tutorial_world/build.ev
@@ -836,7 +836,7 @@ archway
 #
 # Set its home to this location
 #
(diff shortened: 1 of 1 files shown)


jaggiJ:
I think the command is redundant since the ghost mob already is created with 
home location and manually assigning him home location as alias to it still
stores the location (not alias) as home. I don't see point ?

Griatch:
The reason for explicitly setting home location was to make sure default home #2 being set if there was an issue with the room creation. So it's redundant, as a sort of extra fail-safe .

Issue #1685, opened 2018-10-13
Feature Request: Ability to save webclient settings


Griatch:
#### Brief summary of issue / Description of requested feature:

Our new webclient has many new cool features such as split-window functionality. Settings however goes away if you reload your browser and you then have to recreate the pane structure anew. 

This feature concerns implementing a save function so that the user can save their webclient settings. This should include settings like help popup as well as pane structure and message-type assignment. 

#### Reasons for adding feature:

This is a minimum required feature for anyone looking to replace their telnet client for the webclient.

#### Expected result of feature

- A 'save' button for saving the settings server-side (on the Account, maybe tied per-session). 
- This requires developing a generic (probably JSON-based) save format for pane structure etc. 
- As part of this, the developer should be able to customize the 'default' format so as to present a default that the user 
  sees when they connect (and which they could then tweak if they wanted to).
- There should also be a 'reset' ability so as to return to the 'factory' default (forget the customization done by the user). 

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:

Griatch:
Bumping this since it's an important aspect of continued web client development.

Griatch:
Closed in #2224.

PR #1785, opened 2019-02-19
Add golden-layout as an alternative to splithandler

friarzen:

This PR introduces an alternative layout tool to the currently enabled splithandler. 

It provides more much more flexibility in the layouts, and retains layout state between page reloads.
In the future this also allows the client to automatically add new panes for "spawns" behavior.

Addresses #1685


Griatch:
I like the look of this; it's a lot easier to create new panes and adjust/drag them around. The user can also make a tab layout if they want! The fact that it remembers the layout is also perfect. Alas, this will be lost if they switch computer/browser but it's still a great step forward. 

Some aspects of this strikes me more as a proof of concept at this point rather than a ready-to-merge PR, but it's very cool and I'd be interested in using this over split.js. This is a bit of a discussion, so do come with counter-arguments if you disagree! 

- The `Drag me` square is useful for me for testing, but it's pretty confusing and big (and in the way) for an end user. It feels like this button could just as well exist at the top right of every pane as a small `[+]`, maybe with a tool-tip saying "Drag me" if you were to single-click on it (which is the natural inclination to do). 
- The main config button could be a part of the primary input pane (maybe as one of the icons in its header?) since this cannot be removed anyway. 
- When I click the top-left ring of a pane I get an empty drop-down menu. I presume this is intended to hold pane-type selection in the future? 
- How will this interact with the dual-button and hot-button plugins? It appears (at least from a layout perspective) it should be easy to have the two co-exist by just producing separate panes for them?

friarzen:
I think some of what you are seeing is due to Firefox and Chrome acting a bit differently.
Specifically, the window-type mapping IS working.  There is a '+' symbol to the right of the tab Title.  Clicking that creates a dropdown with all known types in it for mapping (and clicking it again closes the dropdown and saves the settings).  But Firefox, for whatever reason, is coloring it very closely to the background titlebar color so it becomes invisible.  (The default is set to 'All' content' for new windows).

The scrolling is definitely a bug I will have to figure out, and seems to be related to the length of the output compared to the Tab's display size.  But I think that is what you are seeing with content "not updating"  I think it is tacking it onto the end and just not scrolling it into view.  Let me know if that isn't what you mean.

The top left double ring icon pops up a new text input window to re-title that Tab, clicking it again closes/saves.  I think the focus() issue in the default_in.js plugin is causing this to seem like that dropdown does nothing/is a placeholder, but it does work once the focus behavior is fixed.  I seem to have not included that default_in.js change as part of this merge request.  I changed it for my testing of the dual_input.js plugin, and so I think I just missed that it is still not corrected in your development branch.

As for the DragMe button, that is a bit of a placeholder.  Unfortunately, the only supported way to get Draggable windows out of the golden-layout framework is to have a draggable HTML element somewhere on screen.  Further explanation:

They do provide examples for click-to-add-a-new-window-to-an-existing-html-element and click-to-select-then-click-something-else-to-add-a-tab-to-an-existing-window.   But the simpler click-to-add-window behavior is such that it will severely rearrange the existing on-screen layout... every time.... which is VERY annoying.  And the "select a window then click something else to add a tab" behavior is not at all obvious to the user, it took me several minutes, even looking at the code to figure out what I was supposed to do to be able to add a new window.  Without that code, I'd have simply assumed you can't add windows at all.

The only approach that has worked and was obvious to me was the drag-to-open-new.
But that means that draggable area needs to go SOMEWHERE that makes sense and is big enough to draw the eye.  I tried moving it around to various corners but something always gets covered.  One solution is to put it into its own non-closable window that gets placed down next to the input bar....but then it becomes a question of how big should it be and what can we do to make that space less wasted.

The other thing I just thought of while writing this is to put the Options toolbar button into the Main tab like you suggest, and then add the DragMe button to the Options popup.  That might work.

Thoughts?

Griatch:
@friarzen I did test this in Chrome as well and could not see an obvious difference in behavior there. I definitely mainly used Firefox though. 
- The lack of scrolling to the bottom is the main issue, but I also felt that sometimes windows did not update at all (not even by appending to the bottom). For example, giving some command producing a list (something disninguishable from normal look) made that list only appear in the latest pane I created and after clicking around a bit, it did appear in another pane. I didn't see a clear pattern to when it would display and not. 
- I think the Dragme thing is fine in itself. Once you understand that it works it's quite intuitive to draw around the new window and place it where you want. The main issue is the size and placeme of the dragme button - the name of it suggests you should be dragging around that button. 
- I like the idea of making the options as a tab. Maybe it should not be a popup at all, but just replace the input pane in-place ...? (as if the options is "behind" the input pane). I'm a bit reluctant to popups overall, it sort of breaks the usage flow I think. 
- It could be worth exploring to put the dragme button into the options (but it would need to be better described) but I still think it may be even easier to have a button `[+]` or similar next to the close button in the header of each pane that acts as the drag-target, but if you just click it (which is the first thing everyone will try), you get a popup saying something like "Click and drag me to create a new window". Putting it in the options pane may be easier from an implementation standpoint though, not sure.

friarzen:
OK, give this a try, I think I still like making the Options Popup Control be based off the Input tab instead of Main, since there is so much going on already on the Main tab, but this is workable.

Griatch:
@friarzen I got bogged down with py3 and other stuff, will review shortly.

Griatch:
@friarzen Wow, this is a huge improvement! This feels a lot more usable now and the placement of the tabs makes it feel a lot more streamlined. The CSS of some elements could be improved, with some margins etc, but that's something one can iterate on later. I have only a few blocking problems I noticed this time around: 

- There appears to be cross-over between the tabs of different panes (saw it in Firefox). It happens very rarely, but when I changed the name of one pane and then clicked naming option of Main, it immediately changed the name of Main to that of the other pane. It also once set Main to "Replace" mode in the menu after I changed it to "Replace" in the other pane (but it didn't actually change the functionality of Main, so it appears it was just the dropdown menu that was read from the wrong place? I can't reliably reproduce this unfortunately, but it's quite confusing when it happens. 
- When clicking outside of a dropdown it does not close. At the very least, selecting another dropdown should close the old one. As it is, the dropdowns get stacked on top of each other which, since they have the same background and width in some cases, causes it to look like the newly opened dropdown has entries from the old one.
- The hover-over for the tabs now just shows the name of the tab. For ease of use, it's probably good to give some help here too, like `Main - drag to desired position`.

Other points (not blocking, but worth considering): 
- I think you have a point about letting the options tab be a part of the input tab being better and easier to find. 
- The Newline/Replace/Append" (NRA!) dropdown indicates output formats, but it's nevertheless things that change on the client level (rather than the "all", "untagged" (AU) dropdown, which handles tags of data from the server). I wonder if the NRA would be a good place to also put "input", so as to create a new input pane ...?

friarzen:
Ok, I've been somewhat learning golden-layouts limitations as I go along.  I *finally* figured out how to do multiple input windows last night and it turns out to actually be more of a problem with the default_in.js plugin's reliance on the '#inputfield' ID than anything.

    The way golden-layout works, every time you change the tab/window layout, it destroys all of the HTML involved, and recreates it using its predefined "component" types as templates.  So, you need to avoid '#id' selectors in the components, otherwise you'll end up with multiple HTML elements with the same ID.  This means using CSS classes for everything and picking through to find what you need.

    Anyway, because of all of that, it has taken me awhile to figure out how to get two input tabs working with the framework, and I somehow managed to never notice the 'open a dropdown, now go open other dropdowns' cases....I *think* your first two blocking points are the same close-dropdown bug.  I'll work on fixing that as soon as I can.  As for the third, there's a tutorial for changing the onHover tooltips, so that shouldn't take much at all to fix, just adding a little extra data to the layout.

    The big issue now, given what I've learned of the destroy/recreate behavior, is allowing other plugins to define their own Components.  The hotbuttons.js plugin for instance.  It's a chicken-egg problem with each plugin's init() calls.  Components need to be created in between the golden-layout master object's (myLayout) creation and when golden-layout's myLayout.init() method gets called.  Right now both of those happen in the plugin's init() function, which doesn't give a chance to any other plugins to build their own components.   I think we may need another plugin framework looping method "postInit()" or similar to separate this kind of ordering dependency out.

   I'll see what I can do about adding the 'input' option to the NRA dropdown...that has potential, but I have a bit of functionality tied to the componentID right now for differentiating input windows from output windows, so I'll see if I can set a componentID at some point in time after the component has already been created....no sure what that will break, if anything.

Griatch:
@friarzen I did some tests and find that the dropdowns appear to be closing correctly now and don't 'collide' with the effects of the other dropdowns. Great! It looks like you are still working on this (will be interesting to see what can be done for the inputs etc), so let me know when I should review further.

friarzen:
Well, adding a control for new input windows was actually pretty easy...

friarzen:
Ok, with this set of bug fixes, I think this is now ready for Merge review.

Griatch:
@friarzen Cool, I'll set up and try it soon. I'm no javascript dev myself, but the Codacity review seems to point out a few simple fixes to improve the new code. Those seem like pretty simple changes: https://app.codacy.com/app/Griatch/evennia/pullRequest?prid=3125468

friarzen:
I just fixed several of these warnings, but it doesn't seem to have updated the Quality Review properly.

Griatch:
@friarzen I talked to codacity and they reran the analyzis manually with your updates now; may need to reload browser without cache to see.

friarzen:
There, I've fought with it down to a handful of things I'd say are mostly the lint tool being a typical overzealous lint tool.  I've had all I can take, I'm done.

Griatch:
@friarzen That's cool - thanks for going the distance. I know how overzealous e.g. pylint can be for Python code so it's not surprising some things are just over the top also for JS. :)

Griatch:
@friarzen Merged! Thanks for your work on this. This looks very good now and this is a big improvement over the previous split functionality. 
I did some minor tweaks of CSS before merging. I think some more elaborate tooltips will be needed in the future to explain things like dragging and the use of Shift-up/down for history. But that's a future project :)

PR #1790, opened 2019-02-27
Fixed "mobon ghost" command called during creation of tutorial world.

jaggiJ:
Before: batchcommander executing `mobon ghost` caused error and mobile wasn't activated.
Now: Works without error. Output message confirming execution of the command.

Comment:
In self.args there was whitespace in front plus data type 'unicode' as potential sources of why self.caller.search(self.args) not worked. changing to type str and removing whitespace seem to fix it.
I also added feedback on proper command execution because there was none and it felt wrong to just guess it from mobile behavior alone.

### DIFF
--- a/evennia/contrib/tutorial_world/mob.py
+++ b/evennia/contrib/tutorial_world/mob.py
@@ -43,12 +43,16 @@ class CmdMobOnOff(Command):
         mob = self.caller.search(args)
         if not mob:
             return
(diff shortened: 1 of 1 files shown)


jaggiJ:
Something still isn't right with this command. Some arguments still trigger it like `mobon door` etc. I have to give it more rework. Hold on with that one.

PR #1791, opened 2019-02-27
Update dice.py

ph1234k:
Corrected the install instructions,specifically which name to import from "role_dice" to "CmdDice"

Make installation instructions easier to follow, particularly for those new to coding


Griatch:
Thanks for the PR, but this change is incorrect. The function to import if you want to use the roll ability in your code is indeed `roll_dice` and not `CmdDice`. On the line below that it's described how you use the `CmdDice` command (which in turn makes use of the `roll_dice` function internally).

PR #1809, opened 2019-04-04
Update Dockerfile to use Python3.7

amfl:

Switched the Dockerfile to python3

**evennia/evennia:develop** is still on python2 and can't actually run evennia any more!

- Changed the base image from `alpine` -> `python:3.7-alpine` due to alpine only having python3.6 in repos. Some packages have changed as a result.
- I had to remove the zope `__init__.py` workaround because evennia was trying to alter zope files in the container which were owned by `root`. I don't think this will cause any problems since [python hasn't required a blank `__init__.py` since 3.3](https://stackoverflow.com/questions/37139786/is-init-py-not-required-for-packages-in-python-3).

**There hasn't been much testing on this!** You may prefer to wait until the vanilla alpine container ships python3.7.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,13 +27,13 @@
 # as a base for creating your own custom containerized Evennia game. For more
 # info, see https://github.com/evennia/evennia/wiki/Running%20Evennia%20in%20Docker .
 #
-FROM alpine
+FROM python:3.7-alpine
 
 LABEL maintainer="www.evennia.com"
 
 # install compilation environment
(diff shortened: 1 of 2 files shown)


Griatch:
@amfl Good stuff; I hadn't gotten to updating the dockerfile yet. I think I remember that I needed the zope fix still, but I may be mixing up versions. Will test a bit but this looks promising.

Griatch:
@amfl Merged. Indeed looks like the zope init issue is no longer an issue in Py3, so removing the workaround makes sense.

PR #1810, opened 2019-04-05
Hook examine

volundmush:
@examine now calls a customizable chain of hooks on TypedObject, allowing each Typeclass to customize how @examine will appear to the caller.

I felt that @examine was not a powerful enough tool. It implements all of its formatting and information retrieval in the Command func(), so one has to change the entire command and replace ALL OF THAT TEXT if they want to add more things for it to display.

That's not exactly 'plug in your changes' friendly. @examine is a great command already but with this sort of functionality it's a GREAT command. Now you can have MobileCharacters give different output from PlayerCharacters, and so on.

Due to the way that @examine uses a Deferred to retrieve cmdsets information, I had to pass the cmdset to the ENTIRE hook chain for simplicity. That's kind of stupid, but I'm not sure how else to do it. Maybe set and delete a brief-lived property on the object?


Griatch:
@volundmush Interesting. Will look at it, but *something* has gone wrong with your PR: 

- I don't think you need to change 279 files for this modification ...
- A new feature should be done against `develop` branch, not `master` branch.

These need to be fix first, before getting into the merits of the PR itself :)

PR #1811, opened 2019-04-05
Copy hook

volundmush:
This simply adds an .at_object_creation_copy(source_obj) that's called on the new object after .copy(). It's meant to add more flexibility to @copy.

I'm just going through the system and adding anything that looks like it fits the design criteria of 'ultimate flexibility'.

Nothing that I can see. This is unlikely to be used anytime soon, but it's one of those things we should have.


Griatch:
279 files does not need to be changed for this ... Also, target `develop` branch rather than `master` branch.

PR #1813, opened 2019-04-05
Added at_object_creation_copy hook to DefaultObject.

volundmush:
Just a small addition meant to go along with the existing set of useful overloadable hooks.

Flexibility is good. Since some people (like myself) may be doing weird things with Django or other data sources, having the ability to copy THOSE TOO is most excellent.

Nothing I can think of, apart from it being unlikely to be used by MOST people. There's no pressing need, I just think it fits the philosophy and future building niceties.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -933,7 +933,7 @@ class DefaultObject(with_metaclass(TypeclassBase, ObjectDB)):
 
         return obj, errors
 
-    def copy(self, new_key=None):
+    def copy(self, new_key=None, **kwargs):
         """
         Makes an identical copy of this object, identical except for a
         new dbref in the database. If you want to customize the copy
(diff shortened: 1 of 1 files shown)


evennia/objects/objects.py
```diff
@@ -959,7 +959,22 @@ def find_clone_key():
                       if obj.key.startswith(key) and obj.key.lstrip(key).isdigit())
             return "%s%03i" % (key, num)
         new_key = new_key or find_clone_key()
-        return ObjectDB.objects.copy_object(self, new_key=new_key)
+        new_obj = ObjectDB.objects.copy_object(self, new_key=new_key)
+        new_obj.at_object_creation_copy(self)
+        return new_obj
+
+    def at_object_creation_copy(self, source_obj):
```
Griatch: - I think the name of this hook should be `at_post_copy` since it's called by the `copy` hook, after the copy has been performed. 
- Since this hook is called from within the `source_object.copy` method, it makes more sense to call it on the `source_object` instead and to pass the `copy` in as an argument. You have full flexibility to customize things either way.

evennia/objects/objects.py
```diff
@@ -959,7 +959,22 @@ def find_clone_key():
                       if obj.key.startswith(key) and obj.key.lstrip(key).isdigit())
             return "%s%03i" % (key, num)
         new_key = new_key or find_clone_key()
-        return ObjectDB.objects.copy_object(self, new_key=new_key)
+        new_obj = ObjectDB.objects.copy_object(self, new_key=new_key)
```
Griatch: Hm. It's not strictly part of this PR, but `ObjectDB.objects.copy_object` takes a lot of `**kwargs` for modifying the resulting copy. I see no reason why the `copy` method could not take `**kwargs` too (they are optional, so would not affect current functionality) and pass them on to `copy_object`. Maybe those `**kwargs` could then also be passed into the new hook.

Griatch:
Looking good now, thanks!

PR #1816, opened 2019-04-10
Styles options

volundmush:
THIS IS JUST A STUB/DEMONSTRATIOn do not merge!

Adding a support for per-Account custom coloring of base Evennia commands and whatever commands a user might want to add. Players can change their Account settings with the @style command.

Customization is king. flexibility is king.

This is just a DEMONSTRATION of the concepts. Further development of it would benefit hugely from an adaptation of the Validators/Settings system of Athanor, which solves a lot of D.R.Y. issues. It can work without them, but that would kinda stink.

### DIFF
--- a/evennia/accounts/accounts.py
+++ b/evennia/accounts/accounts.py
@@ -20,6 +20,7 @@ from django.utils.module_loading import import_string
 from evennia.typeclasses.models import TypeclassBase
 from evennia.accounts.manager import AccountManager
 from evennia.accounts.models import AccountDB
+from evennia.accounts.styles import StyleHandler
 from evennia.objects.models import ObjectDB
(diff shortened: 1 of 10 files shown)


Griatch:
Since you say this is for demonstration purposes, I won't review the code yet. Here are my takes on the overall concept: 
- I think the addition of a Command-level helper function to centralize EvTable styles is a pretty good idea; it does homogenizee things nicely. I wonder if it should rather be on the Command level rather than the MuxCommand level though (since it has nothing to do with the specialitites of MuxCommand, the switches etc). 
- Not too keen on the separate `Account.style` handler. I think that if the goal is to make a centralized place to put options (your 'settings' I guess), it should be used to store things like this rather than to use this stand-alone little handler for it.

PR #1821, opened 2019-04-11
Full options

volundmush:
Second round at the Style Handler and ValidHandler stuff. now it's the VALID_HANDLER and the new OptionHandler that goes on Accounts, and a much better demonstration of exactly what it does.

I'm not gonna write this again.

The only problem with this is if it has to store things like datetimes, it doesn't have any code in place for rendering the values. The Athanor version has a Class for each kind of option which has rules for things like displaying various types of data, but I felt that was really overengineered. Still not sure what the best outcome would be. But now you see what this was meant to do! try some commands like @style border=c


evennia/settings_default.py
```diff
@@ -538,6 +560,10 @@
 PROTOTYPEFUNC_MODULES = ["evennia.utils.prototypefuncs",
                          "server.conf.prototypefuncs"]
 
+# Module holding validator functions. functions in later modules will
+# override those in earlier ones.
+VALIDFUNC_MODULES = ['evennia.utils.validfuncs', ]
```
Griatch: "validfunc" makes this sound like another of Evennia's custom functions like protfunc, inlinefunc etc - making them feel a bit more custom than they need to be. I think `OPTIONS_VALIDATOR_MODULES` is clearer (and also helps group the settings-name with the `OPTIONS_ACCOUNT_DEFAULT` suggested above).

evennia/utils/option.py
```diff
@@ -0,0 +1,62 @@
...
+    def __getitem__(self, item):
+        if item not in self.options_dict:
+            raise KeyError("Option not found!")
+        if item in self.options:
+            return self.options[item]
+        import evennia
+        option_def = self.options_dict[item]
+        save_data = self.obj.attributes.get(item, category=self.save_category)
+        if not save_data:
+            return evennia.VALID_HANDLER[option_def[1]](option_def[2])
```
Griatch: I see no reason for having this 'nested handler' in here. Let the option handler itself load the available validators and apply them here. Also - this is the place to catch validation errors, not inside each validator.

evennia/utils/option.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __init__(self, obj, options_dict=None, save_category=None):
+        if not options_dict:
+            options_dict = dict()
+        self.options_dict = options_dict
+        self.save_category = save_category
+        self.obj = obj
+        self.options = dict()
+
+    def __getitem__(self, item):
```
Griatch: I think it would be interesting to experiment with `__getattribute__` and `__setattr__` here instead, then you can do stuff like `.options.option_name` which makes for a much nicer API (cf `.db.attrname`). One must still keep access get/set methods for getting/setting options whose name you don't know at the time of coding of course. 

The drawback with messing with `__getattribute__`/`__setattr__` is that for the code inside this handler you need to be careful to avoid infinite loops (such as when getting things from `self`). The way around this is to make aliases `_GA = object.__getattribute__` and `_SA = object.__setattr__` and use those instead for the code inside the handler. It's a little tricky but it's no so much code in this case.

evennia/utils/valid.py
```diff
@@ -0,0 +1,16 @@
+from django.conf import settings
```
Griatch: IMO, this handler can be removed.

evennia/utils/option.py
```diff
@@ -0,0 +1,62 @@
...
+        found = string_partial_matching(list(self.options_dict.keys()), option, ret_index=False)
+        if not found:
+            raise ValueError("Option not found!")
+        if len(found) > 1:
+            raise ValueError(f"That matched: {', '.join(found)}. Please be more specific.")
+        found = found[0]
+        option_def = self.options_dict[found]
+        if not value:
+            raise ValueError("Value field blank!")
+        new_value = evennia.VALID_HANDLER[option_def[1]](value, thing_name=found)
```
Griatch: Since I suggest validators should also be able to be simple builtins like `int` and `float`, this call would need to inspect the validator callable (`inspect` module) to make sure it can actually take further arguments (or alternatively catch the error and retry)

evennia/utils/validfuncs.py
```diff
@@ -0,0 +1,175 @@
+"""
+Contains all the validation functions.
+
+All validation functions must have a checker (probably a session) and entry arg.
```
Griatch: I think the allowed call signs here should be either `validator(arg)` or `validator(arg, **kwargs)`. Btw - I like that these raise ValueError, makes for a nice and  clean flow - if the validator returns at all, you are golden.

evennia/utils/option.py
volundmush: What if other arbitrary code wants to use those Validators without needing an Account object to reference? They're useful for more than just Options. :)

evennia/utils/option.py
volundmush: Stuff like int() is already covered with some of the provided validators like signed_integer. The reason these are done this way is also so other code can call the validator and use 'thing name' for custom error messages. I'm not sure what your logic here is.

evennia/utils/validfuncs.py
volundmush: My original version of them also receives the Session or Account, so that it could do things such as 'take a datetime string as that specific user favors in their configured timezone, convert it to UTC, and return the UTC version for code purposes.' So I agree with adding something like **kwargs here.

evennia/utils/validfuncs.py
```diff
@@ -0,0 +1,175 @@
...
+def email(entry, thing_name="Email Address"):
+    try:
+        _val_email(entry) #offloading the hard work to Django!
+    except _error:
+        raise ValueError(f"That isn't a valid {thing_name}!")
+    return entry
+
+
+def lock(entry, thing_name='lockstring', options=None):
+    entry = entry.strip()
```
Griatch: Btw, this will fail if entry is not of the right form (for example it'll be an `AttributeError` if entry is a number). For these to be proper validators you Make sure the validator fails deliberately with a ValuError when input is on the wrong form.

evennia/utils/option.py
Griatch: The logic is simple - use the validator that the user supplies, many of which are supplied by the standard library. Using things like int() is standard and expected when validating input.

evennia/utils/option.py
Griatch: Not sure what you mean by this - the validators have no dependency on any Account object? A validator's job is straight forward - parse a value  and either raise `ValueError` or return the expected type and form for that validator.

evennia/utils/validfuncs.py
Griatch: Whereas people can add custom validators on whatever form they like, I don't think formulating custom error messages is the work of a validator. It would be better to have the option handler handle the error handling and reporting  - it has all the information available to build a proper, informative error message. The validator could raise a `ValueError' with a custom message string if one wanted to display any further special information about the error.

evennia/utils/option.py
volundmush: user-supplied validators are already supported as modules you can add to your mygame, as long as they follow the API. Same deal as lockfuncs really.

Griatch: And yet, being able to just add int as a validator is relevant and expected.

evennia/utils/opclasses.py
```diff
@@ -0,0 +1,319 @@
...
+class _BaseOption(object):
+    """
+    Abstract Class to deal with encapsulating individual Options. An Option has a name/key, a description
+    to display in relevant commands and menus, and a default value. It saves to the owner's Attributes using
+    its Handler's save category.
+
+    Designed to be extremely overloadable as some options can be cantankerous.
+
+    Properties:
+        expect_type (str): What users will see this as asking for. Example: Color or email.
```
Griatch: This seems superfluous and also adds extra boiler plate for every option type. We are already setting both a name, a description *and* a named validator-function for this option. Surely one of those resources should allow us to deduct and display what type we are expecting?

evennia/utils/opclasses.py
```diff
@@ -0,0 +1,319 @@
...
+        Exports the current value to an Attribute.
+
+        Returns:
+            None
+        """
+        self.handler.obj.attributes.add(self.key, category=self.handler.save_category, value=self.export())
+
+
+class Text(_BaseOption):
+    expect_type = 'Text'
```
Griatch: Note how this field tends to just be a repetition of the class name ...

evennia/utils/opclasses.py
```diff
@@ -0,0 +1,319 @@
...
+    Designed to be extremely overloadable as some options can be cantankerous.
+
+    Properties:
+        expect_type (str): What users will see this as asking for. Example: Color or email.
+        valid: Shortcut to the loaded VALID_HANDLER.
+        valid_type (str): The key of the Validator this uses.
+    """
+    expect_type = ''
+    valid = _VAL
+    valid_type = ''
```
Griatch: It would be clearer if this was named `validator_key` or similar since the point of it is to look up and call a validator by this name.

evennia/utils/valid.py
```diff
@@ -0,0 +1,25 @@
...
+from evennia.utils.utils import callables_from_module
+
+
+class ValidHandler(object):
```
Griatch: As you pointed out, this is not really handler in the Evennia sense of the word. I don't think "helper" is a good word either though - it's just a container that preloads resources from a module. "Container" may be a pretty good term actually. 

I think overall though, it would be preferable if resources on these containers could be accessed as properties rather than with dict-like operations - the latter is quite different from how other Evennia APIs operate.

evennia/utils/opclasses.py
```diff
@@ -0,0 +1,319 @@
...
+    @property
+    def value(self):
+        if not self.loaded and self.save_data is not None:
+            self.load()
+        if self.loaded:
+            return self.value_storage
+        else:
+            return self.default
+
+    def validate(self, value, account):
```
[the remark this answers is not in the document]
volundmush: Options aren't used only for Accounts. that's the default implementation, but you could just as easily use this for storing Options on a Global Script by adding the OptionHandler to it and providing an options_dict and a save category.

In this case, Account is passed because there are a rare few cases where the Account needs to be known. For instance, if a user is providing a DATETIME in their local time that the system needs to convert to UTC for storage. (This was a highly requested feature at MCM for the scene scheduler.)

evennia/utils/opclasses.py
```diff
@@ -0,0 +1,319 @@
...
+        Args:
+            value:
+            account:
+
+        Returns:
+
+        """
+        return self.valid[self.valid_type](value, thing_name=self.key, account=account)
+
+    def set(self, value, account):
```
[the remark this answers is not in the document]
volundmush: How pass more than one argument to a @setter ? Account is here for a reason.

evennia/utils/opclasses.py
```diff
@@ -0,0 +1,319 @@
+import datetime as _dt
+from evennia.utils.ansi import ANSIString as _ANSI
+from evennia.utils.validfuncs import _TZ_DICT
+from evennia.utils.valid import VALID_HANDLER as _VAL
+
+
+class _BaseOption(object):
```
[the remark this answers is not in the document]
volundmush: How do I prevent it from doing weird circular imports here? I'm admittedly a little unsure about the best way to lay out these modules and where to put things. But that's why I have you!

evennia/utils/opclasses.py
volundmush: You have a point.

evennia/utils/opclasses.py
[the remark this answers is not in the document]
Griatch: Sure, if it's stored on something other than the account, then that will also be accessible through the `.handler` property.

This seems to be a case of confused purposes. An Option describes the state of the object it sits on. It stores and handles that state. If you want to display that in an "looker aware" fashion, it seems this is something you should either handle after the option has been returned or (since I presume you want to make use of all that validation machinery) by passing it to a custom `display` method for that option.

evennia/utils/opclasses.py
```diff
@@ -0,0 +1,319 @@
...
+        Returns:
+
+        """
+        final_value = self.validate(value, account)
+        self.value_storage = final_value
+        self.loaded = True
+        self.save()
+        return self.display()
+
+    def display(self):
```
Griatch: Why not accept `**kwargs` here so the display can be influenced by whomever looking? Then you don't need to keep passing that `account` argument around elsewhere.

evennia/utils/opclasses.py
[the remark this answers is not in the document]
volundmush: It's the other way around. Let's say I was to put this OptionHandler on a GlobalScript. now an admin wants to set a Datetime on one of its options. the admin is in Eastern Standard Time and wants to enter it in their local time. The system needs to convert this to UTC. How does the Option know which Account to pull from in determining the appropriate timezone to convert from?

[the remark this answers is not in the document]
Griatch: It's not up to the option to do this. The option stores the value you give it, in the type specified for that option. If the option is determined to be, say a UTC field, then the admin should supply their datetime with their local time zone embedded in that datetime object - the option will then convert that to whatever format it should hold.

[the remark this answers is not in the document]
volundmush: I will probably need your help coming up with a way for DUMB USERS to easily input dates without tripping over themselves, then, but okay. I'll remove the account parameters.

evennia/utils/opclasses.py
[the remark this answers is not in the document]
Griatch: If you want to avoid a specific circular import you need to put the import into the class so it's not loaded at the time of importing the module. I usually use this pattern:

```python
_NAME = None

class Foo(object):
    # ...
    def bar(self):
        global _NAME
        if not _NAME:
            from library import name as _NAME

        _NAME.method.function() ...
```

This used to carry some performance benefits I have since forgotten what they were. It's necessary in the flat API to delay-load imports to be made available at a global level. In the case of avoiding a circular import I have not explored if this kind of pattern is actually needed from a performance perspective in py3 anymore; possibly just doing the import inside the class is enough since imports should be cached anyway.

evennia/utils/opclasses.py
```diff
@@ -0,0 +1,278 @@
...
+    def display(self, **kwargs):
+        """
+        Renders the Option's value as something pretty to look at.
+
+        Returns:
+            How the stored value should be projected to users. a raw timedelta is pretty ugly, y'know?
+        """
+        return self.value
+
+    def _load(self):
```
Griatch: In light of discussing using `options` also for, what today's `@options` command handles (client flags), then my request to make this private no longer applies - this and `save` would have to be overloaded and should then not be private - sorry. :)

evennia/utils/opclasses.py
```diff
@@ -0,0 +1,278 @@
...
+            account:
+
+        Returns:
+            None
+        """
+        final_value = self.validate(value)
+        self.value_storage = final_value
+        self.loaded = True
+        self._save()
+
```
Griatch: As mentioned in chat, for maximum flexibility, having additional get/setters that takes a `**kwargs` they can pass on into the validator (as well as into save/load for optional saving) would allow this to handle any option storage type.

evennia/__init__.py
```diff
@@ -111,6 +111,7 @@
 CHANNEL_HANDLER = None
```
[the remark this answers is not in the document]
Griatch: Don't forget to add the global singletons to the flat API.

evennia/utils/option.py
Griatch: This remains a thing. Being able to do `obj.options.optionname.value` makes sense for many common cases.

evennia/utils/opclasses.py
```diff
@@ -0,0 +1,278 @@
...
+class BaseOption(object):
+    """
+    Abstract Class to deal with encapsulating individual Options. An Option has a name/key, a description
+    to display in relevant commands and menus, and a default value. It saves to the owner's Attributes using
+    its Handler's save category.
+
+    Designed to be extremely overloadable as some options can be cantankerous.
+
+    Properties:
+        expect_type (str): What users will see this as asking for. Example: Color or email.
```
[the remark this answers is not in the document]
Griatch: But overall I must say this class now is much easier to read and understand! :+1:

evennia/utils/opclasses.py
```diff
@@ -0,0 +1,278 @@
+import datetime as _dt
```
Griatch: For clarity, naming this module  `optionclasses.py` to be in line with the naming scheme of e.g. `inlinefuncs.py` etc which don't abbreviate the first word.

evennia/utils/validfuncs.py
```diff
@@ -0,0 +1,202 @@
+"""
```
Griatch: For consistency with other such modules, please rename to `validatorfuncs.py`. This makes it a little clearer what they do at a glance.

PR #1822, opened 2019-04-13
Added some Signals.

volundmush:
Django Signals are honking awesome. Let's have these available in addition to hooks to make it easier on contribs that want to make use of these events! Like the Event Handler idea earlier but better.

Yeah I think I explained that. But since we're using Django, we should really USE DJANGO.

Not sure if I put these in all the right places, and this is kind of draft quality. I'd like to know what other signals this should track. ... Not that they're hard to add. This system is actually REALLY SIMPLE TO USE.


TehomCD:
Some people feel that signals can be kind of easy to abuse. When adding them in, it might be a good idea to briefly mention in both the wikia and their docstrings about common pitfalls for them so that some newer developer doesn't add in a signal then winds up asking Evennia chat why their game is behaving in specific ways that are impossible to debug without knowing about them.

https://lincolnloop.com/blog/django-anti-patterns-signals/ has a good synopsis of issues with them.

evennia/utils/signals.py
```diff
@@ -0,0 +1,15 @@
+from django.dispatch import Signal
```
Griatch: 1. This module needs a doc string explaining how to use these signals (how to hook into them etc) without having to refer to Django documentation (there needs to be a wiki page in the future too). 
2. Maybe it would be a good idea to gather the default Django signals  (pre/post save etc) in here too, to have them all in one place?
3. If doing the above, I wonder if `evennia/server/signals.py` may not be a better location for this; these are sort of central things sent out by the server. (not sure about that one though)

evennia/utils/signals.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+ACCOUNT_CREATE = Signal(providing_args=['session', ])
```
Griatch: These need at least a line of comment to explain exactly when it fires (is it before/after the account object is created, for example? Which hooks have fired already at this point?). Same goes for the other signals below.

evennia/accounts/accounts.py
```diff
@@ -227,6 +230,8 @@ def disconnect_session_from_account(self, session, reason=None):
         if not _SESSIONS:
             from evennia.server.sessionhandler import SESSIONS as _SESSIONS
         _SESSIONS.disconnect(session, reason)
+        if not self.sessions.all():
+            ACCOUNT_LOGOUT.send(sender=self)
```
Griatch: Would not `SESSION_POST_LOGOUT` be a clearer name for this signal?

evennia/accounts/accounts.py
```diff
@@ -301,6 +306,7 @@ def puppet_object(self, session, obj):
         # re-cache locks to make sure superuser bypass is updated
         obj.locks.cache_lock_bypass(obj)
         # final hook
+        OBJECT_PUPPET.send(sender=obj, account=self, session=session)
```
Griatch: This seems to be in a strange place - just before the `at_post_puppet` hook that may do a lot of things. It seems clearer overall to put hooks like this very last. This should then be called `OBJECT_POST_PUPPET`. Or why not name it after the hook (`AT_POST_PUPPET`)?

evennia/accounts/accounts.py
```diff
@@ -323,6 +329,7 @@ def unpuppet_object(self, session):
                 obj.sessions.remove(session)
                 if not obj.sessions.count():
                     del obj.account
+                OBJECT_UNPUPPET.send(sender=obj, session=session, account=self)
```
Griatch: Moving this below the hook and calling it `OBJECT_POST_UNPUPPET` seems clearer. Or maybe naming these signals the same as the hooks (`AT_POST_UNPUPPET`)?

evennia/server/sessionhandler.py
```diff
@@ -518,6 +518,8 @@ def login(self, session, account, force=False, testmode=False):
                                                              operation=SLOGIN,
                                                              sessiondata={"logged_in": True,
                                                                           "uid": session.uid})
+        if nsess < 2:
+            ACCOUNT_LOGIN.send(sender=account, session=session)
```
Griatch: Move after the hook?

PR #1824, opened 2019-04-15
Login tracker

volundmush:
Adds new Models to server and Accounts for tracking Logins. A lot of games bar multi-playing or have other reasons to keep a close eye on Login records.

While Evennia has the MULTISESSION_MODE to deal with multi-playing in the most directly restrictive sense, I feel that it's still important to track logins so users can check their own records and admin can also keep an eye open for anything fishy.

I'm not 100% sure if this is the best database schema. It seems to be the most efficient in terms of 3rd normal form approach, but that doesn't mean it's the most natural for Django to deal with.

I believe the system should also have a second layer of record cleaning. LoginSource and Login records older than <configurable period> should be periodically removed. I am not sure what the best way to do this is. The duration should be configurable by putting an entry in settings.py.

The current datetime formatting in the tables for @ip and @logins is awful. Does Evennia have a way to deal with this? It should ideally be formatting to the Timezone in your Account.options.timezone. What do?


Zireael07:
> LoginSource and Login records older than should be periodically removed. I am not sure what the best way to do this is. The duration should be configurable by putting an entry in settings.py.

You can get today's date in Python easily by using date.today(), and it returns a normal datetime object so you could easily subtract however many hours/days your settings flag tells it to.

AFAICT Django uses SQL under the hood, it should be possible to do some form of SELECT login_source, login FROM login_table WHERE date < cutoff_date

strikaco:
> Does Evennia have a way to deal with this? It should ideally be formatting to the Timezone in your Account.options.timezone.

The pytz module can convert UTC timestamps to local time.

Griatch:
Since this has not seen any updates for a long while, I'm closing this PR. If you want to try to finish it up, ping me and I can reopen again.

PR #1828, opened 2019-04-15
Added Base Typeclass loading to Evennia API.

volundmush:
Add the BASE_WHATEVER_TYPECLASS properties to evennia API, containing the loaded typeclasses. EG:
evennia.BASE_CHARACTER_TYPECLASS contains the actual class, not a path to the class.

I feel it's silly that these are imported over and over again all over the place using class_from_module()

Having them all in one place for an easy, direct import seems much smarter, more efficient, and cleaner.

Then again, I'm not sure why you haven't done this already. Because of that, I'm just proposing this at the moment instead of CHANGING ALL OF THOSE CALLS. What do you think of this idea?

### DIFF
--- a/evennia/__init__.py
+++ b/evennia/__init__.py
@@ -114,6 +114,16 @@ CHANNEL_HANDLER = None
 GLOBAL_SCRIPTS = None
 OPTION_CLASSES = None
 
+# typeclasses
+BASE_ACCOUNT_TYPECLASS = None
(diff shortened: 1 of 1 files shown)


Griatch:
It's not been done mainly because it used to be a circular import nightmare. But I think these days most of these are imported at the top of various modules anyway so there should probably no longer be an issue with circular imports. So sure, I guess it makes sense.

PR #1831, opened 2019-04-18
Adding tests for gametime module

chason:

Adds basic tests for gametime module

I want to confidently make changes to gametime and not worry about breaking anything

### DIFF
--- a/evennia/contrib/tests.py
+++ b/evennia/contrib/tests.py
@@ -222,6 +222,8 @@ class ForceUTCDatetime(datetime.datetime):
 
 
 @patch('evennia.contrib.extended_room.datetime.datetime', ForceUTCDatetime)
+# mock gametime to return April 9, 2064, at 21:06 (spring evening)
+@patch('evennia.utils.gametime.gametime', new=Mock(return_value=2975000766))
 class TestExtendedRoom(CommandTest):
     room_typeclass = extended_room.ExtendedRoom
     DETAIL_DESC = "A test detail."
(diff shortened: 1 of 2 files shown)


chason:
ok that failed very strangely. These all pass locally.

chason:
ah, it looks like something else in the test suite is affecting the gametime module. I was only testing evennia.utils locally before. I will look into this tomorrow and fix it up.

Griatch:
It's a twisted async situation, it looks like.

chason:
some of the tests themselves are failing too. My guess is that another test suite is changing some of the configuration/constants or monkey patching something that is causing different results. the `gametime.gametime()` function in particular is returning an integer which shouldn't be possible if it is running "naturally"

chason:
Yay, finally all the tests pass

evennia/contrib/tests.py
```diff
@@ -222,6 +222,8 @@ def fromtimestamp(cls, timestamp):
 
 
 @patch('evennia.contrib.extended_room.datetime.datetime', ForceUTCDatetime)
+# mock gametime to return April 9, 2064, at 21:06 (spring evening)
+@patch('evennia.utils.gametime.gametime', new=Mock(return_value=2975000766))
```
Griatch: Nice. :)

evennia/utils/tests/test_gametime.py
```diff
@@ -0,0 +1,99 @@
...
+from unittest.mock import Mock
+
+from django.conf import settings
+from django.test import TestCase
+
+from evennia.utils import gametime
+
+
+class TestGametime(TestCase):
+    def setUp(self) -> None:
```
Griatch: First time this style of annotation is used in the Evennia code. I still have not gotten used to this (I think it looks pretty messy), but I guess it's fine here.

PR #1873, opened 2019-07-22
Opens an interactive console when using the py command with no argument

vincent-lg:

- When typing the `py` command with no argument, a Python console opens in the client.
- The `evennia.utils.evmenu.get_input` function had to be very slightly altered.

This feature adds a great simplicity for Evennia debugging and administrator operations through the `py` command.  If the admin has full privilege, she can use the `py` command with no argument, open a Python console and interactively type commands in, to see the result, create variables and so on:

    > py
    Python 3.7.3 ...
    Type 'exit' to quit this console.
    >>> evennia
    <module evennia>
    >>> var = 5
    >>> if var < 8:
    ...     var = 32
    ...
    >>> var
    32
    >>> for obj in here.contents:
    ...     here.msg("You are in this location.")
    ...
    >>> # And perform more...
    ... exit
    Closing the Python console.

The advantage of this over the editor provided in `py/edit` is that you see the result of each command you type.  For users that are on good terms with the Python interpreter (that is, most of them, including beginners), this is a great way to test in a Python playground, though doing things potentially dangerous of course.

#### Implementation details

This was trivial to add.  This feature makes use of the `yield` keyword in `func()`, so the Python console isn't persistent when Evennia reloads (I believe this is actually a good thing).  The `code.InteractiveConsole` was subclassed for this purpose with a very light wrapping around sessions.  All in all, it didn't take much code and I still find this feature highly readable.

> Warning: please note that `evennia.utils.evmenu.get_input` was modified.  The old function used to `strip()` all input strings from the user.  This created some problems since you have to enter indentation manually in this mode.  So I replaced it with a simple `rstrip(*)`.  And I still find it "too broad" as it is, perhaps a narrower stripping would be appropriate (`.rstrip("\n")` perhaps)?

- Evennia master.

### DIFF
--- a/evennia/commands/default/system.py
+++ b/evennia/commands/default/system.py
@@ -5,6 +5,7 @@ System commands
 """
 
 
+import code
 import traceback
 import os
 import io
@@ -167,17 +168,7 @@ def _run_code_snippet(caller, pycode, mode="eval", measure_time=False,
     if hasattr(caller, "sessions"):
         sessions = caller.sessions.all()
 
-    # import useful variables
-    import evennia
-    available_vars = {
-        'self': caller,
-        'me': caller,
-        'here': getattr(caller, "location", None),
-        'evennia': evennia,
-        'ev': evennia,
-        'inherits_from': utils.inherits_from,
-    }
-
+    available_vars = evennia_local_vars(caller)
     if show_input:
         for session in sessions:
             try:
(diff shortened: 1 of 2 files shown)


Griatch:
@vincent-lg Technically, this belongs in `develop` branch since it's new functionality. But it's such a useful addition that I extended the stdout rerouting to the normal `py` eval as well and added it to master. :+1:

PR #1877, opened 2019-07-29
[clothing contrib] Allow to create new Clothing classes

vincent-lg:

The `clothing` contrib could be used in standalone.  If new clothing types were created (say, if you created a `Shoe` typeclass inheriting from `Clothing`), you couldn't equip objects from the `Shoe` typeclass.

Knowing how a contrib is used can be hard.  But allowing for inheritance does seem important in this case.  Some users find it actually more logical to create a typeclass per equipment type and set their `db.clothing_type` in `at_object_creation`.  Although this wasn't the intended use of the contrib, where one can easily create clothes using the `Clothing` typeclass, allowing to inherit from `evennia.contrib.clothing.Clothing` while still being able to use the wear and remove commands is important, I think.

The reason why this was not possible was due to the use if `is_typeclass` in different places of the contrib.  By default, `exact` is set to `True`, which meant this only allowed object created from `evennia.contrib.clothing.Clothing`.  Inherited objects didn't work.  I had to set the `exact` flag to `False` explicitly.

By the way, is that really a good thing?  Shouldn't this flag be set to `False` by default?  It's the way `isinstance` or `issubclass` work, after all.  That being said, it's an API change and it can have problematic consequences.

- Evennia master.

### DIFF
--- a/evennia/contrib/clothing.py
+++ b/evennia/contrib/clothing.py
@@ -368,7 +368,7 @@ class CmdWear(MuxCommand):
         wearstyle = True
         if not clothing:
             return
-        if not clothing.is_typeclass("evennia.contrib.clothing.Clothing"):
+        if not clothing.is_typeclass("evennia.contrib.clothing.Clothing", exact=False):
             self.caller.msg("That's not clothes!")
             return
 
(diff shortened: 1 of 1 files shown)


Griatch:
One could maybe rework this contrib overall so as to not use `is_typeclass` so strictly. I guess is_typeclass could default to `exact=False` but I suspect this may break quite a lot of backwards-compatibility and would not be something we could change except as part of a new Evennia version (so `develop` branch change). See https://github.com/evennia/evennia/issues/1904

Issue #1789, opened 2019-02-26
Help popup in web client not clearing after close


fang0654:
#### Brief summary of issue / Description of requested feature:

The help popup does not clear the contents when closing the window. When you type "help" again, the new help text is now at the bottom of the window, and you need to scroll down to find it. This can get a little confusing, as what is being displayed isn't necessarily what was requested.

#### Steps to reproduce the issue / Reasons for adding feature:

1. Type 'help'
2. Close the window
3. Type 'help set'
4. Observe that the help for 'set' is at the bottom of the help window.

#### Error output / Expected result of feature

Expected to clear the window between uses, so only the current help command is displayed.

ph1234k:
I'm new to the project so I am still learning the codebase. Is it possible this is caused by the behavior of the function "evmore" which seems to be called for webclient connected sessions?

fang0654:
Just a follow-up on this. Changing line 140 of `web/static/webclient/js/plugins/options.js` like so:

```javascript
$('#helpdialogcontent').append('<div>'+ args + '</div>');
```
to
```javascript
$('#helpdialogcontent').prepend('<div>'+ args + '</div>');
```

at least puts the new help data at the beginning, instead of at the end. This still then keeps a help history. I can put together a proper patch in a week when I get some free time again.

friarzen:
I guess the proper solution is to just use html() instead of append():
$('#helpdialogcontent').html('<div>'+ args + '</div>');

Griatch:
Resolved in #1887.

PR #1887, opened 2019-08-15
Fixed Issue #1789: Help popup in web client not clearing after close

gwa2100:
I modified line 140:  Replacing the .append with .prepend.  This allows for a running log of previous help requests to be pushed to the bottom.


evennia/web/webclient/static/webclient/js/plugins/options.js
```diff
@@ -137,7 +137,7 @@ let options_plugin = (function () {
         // is helppopup set? and if so, does this Text have type 'help'?
         if ('helppopup' in options && options['helppopup'] ) {
             if (kwargs && ('type' in kwargs) && (kwargs['type'] == 'help') ) {
-                $('#helpdialogcontent').append('<div>'+ args + '</div>');
+                $('#helpdialogcontent').prepend('<div>'+ args + '</div>');
```
gwa2100: By changing the append to a prepend it causes it to push down the current "stack" of help information.  Makes for a better experience.

Griatch:
Makes sense. Thanks!

PR #1894, opened 2019-08-21
do not duplicate whispers

aogier:
This PR fixes whisper behaviour when user type twice a whisper recipient, eg

```
whisper foo,foo,bar = asd
```

with a `set` repetitions no longer matter, we don't have any duplication receiver side

ciao!


Griatch:
Fair enough, simple and effective fix!

PR #1895, opened 2019-08-22
fix ssh import error message

aogier:
added a required python library for use SSH. In newer versions, cryptography seems already installed because pip dependencies, but this is not the case for pyasn1 (already present in err message) and bcrypt, tracking it down required a temporary patch on evennia.server.portal.ssh module

I've left the original message untouched but you could probably delete cryptography dependency too

Thank you, ciao!


Griatch:
Fair enough!

PR #1897, opened 2019-08-22
importing `range` no longer needed

aogier:
pylint complained and indeed now that you droppped python2, importing range from builtins is no longer needed:

```py
In [1]: from builtins import range as r

In [2]: r is range
Out[2]: True
```

thanks!


Griatch:
True, thanks!

PR #1926, opened 2019-09-14
Bump min Django version to 2.2.5

gtaylor:
Bug fix release. For more details, see:
https://docs.djangoproject.com/en/2.2/releases/2.2.5/


gtaylor:
Might be worth considering only checking the major+minor version. Patch level versions of Django are safe to move between in all but the most extreme (security-related) cases.

PR #1933, opened 2019-09-15
Condense requirements files down to one

gtaylor:
PEP 508 defines a way to specify platform-specific requirements.
Rather than having to manage two different requirements files, let's
crunch them down to one and use that facility.

For more details on how this works, see:
https://www.python.org/dev/peps/pep-0508/


gtaylor:
Note: Tested this out with success before PR'ing!

PR #1936, opened 2019-09-16
Remove pillow from requirements

gtaylor:
This was originally suggested due to an imagefield, but that is no longer present. Since Pillow is not required to run Evennia, we should remove it and let the users opt into it if they end up wanting/needing it for their custom code.

I can follow up this merge by removing it from the install docs. That should simplify things greatly for Linux and Windows users both.


Griatch:
I removed `pillow` from the getting-started docs.

PR #1938, opened 2019-09-16
Remove builtins imports

### LINKED ISSUE #1931: Remove builtins imports
We have quite a few imports from the `builtins` modules that don't need to exist.

For example, this occurs in a few dozen files:
```
from builtins import object
```
There are some valid uses of things under `builtins`, but we don't seem to have any cases that require them. It confused me initially in that I thought something more complex was going on, whereas there didn't actually appear to be anything.

May as well clear up the noise!

gtaylor:
This module is not intended to be used directly in most cases, per:
https://docs.python.org/3/library/builtins.html

None of our usages warrant the explicit import. We also avoid some
confusion as folks dig to see what we are doing to require
importing builtins directly.

Fixes #1931.

### DIFF
--- a/evennia/accounts/admin.py
+++ b/evennia/accounts/admin.py
@@ -2,8 +2,6 @@
 # This sets up how models are displayed
 # in the web admin interface.
 #
-from builtins import object
-
 from django import forms
 from django.conf import settings
 from django.contrib import admin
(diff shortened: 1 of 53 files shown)


Griatch:
Some nice cleanup there!

PR #1940, opened 2019-09-16
Proposal: code auto-formatting

gtaylor:
*Disclaimer: This PR should be considered as a proposal rather than something that has any strong expectation of being merged.*

*Tip: I'd suggest reviewing these two commits individually rather than hitting the "Files changed" tab above.*

With Evennia now weighing in around 55,000 lines of code, there's quite a lot of room for stylistic differences to wriggle their way in. In absence of more rigid guidelines beyond what PEP8 suggests, different sections of the codebase can vary in format.

Rather than expect manual review and enforcement of formatting, this PR introduces [Black](https://black.readthedocs.io), an increasingly widespread code formatter for Python. We have been enjoying this at work, and I thought I'd suggest it here!

Some nice things about introducing Black:
* We don't need to spend nearly as much time in PR reviews talking style. Black becomes the authority/enforcer for us.
* The code will be formatted consistently. The output may not always reflect our individual personal ideals, but it'll be "good enough". And consistent.
* I've found Black to be safe in my travels. It hasn't butchered or broken reformatted code on me yet.
* While the formatted code will generally use more vertical space (due to the horizontal line length limit), flip through some of the changes in the second commit and notice that there's some really good stuff in here. I (subjectively) think the overall state of the codebase is better having ran through Black.

If this is something we're interested in, feel free to merge. I can follow up by adding a CI call to `make lint-check` and update the wiki to include a `make fmt` call before committing/pushing code.

*Note: The Codacy failure is due to some "low quality" lines being touched, causing them to be linted. These aren't due to new issues being introduced.*

*Note: I'll try to keep this rebased with master, but it may be best to cherry-pick the first commit, drop the second, then re-run `make fmt` and commit that on your own.*


bubthegreat:
Not that my outside opinion makes the difference, but using black has been an overall very helpful thing for a similarly large codebase.  The auto-enforcement means that no matter how hacky the change it will at least be consistent enough that people can jump in and not be thrown by odd stylistic choices, and can instead focus on the logical content.  It has my vote as a non-contributing lurker

Griatch:
@gtaylor I have experimented with Black in the past and whereas I see the argument for enforcing a consistent formatting (as well as solving pep8 compliance all in one go), converted structures like this gives me pause: 

```python
_MUDINFO_CHANNEL = ChannelDB.objects.filter(db_key=settings.CHANNEL_MUDINFO["key"])[
                    0
                ]
```

I'm not sure this is actually making the code easier to read ... 

I find myself more accepting of some of Black's idiosynchacies now than in the past though (such as letting `):` be on its own line when breaking a def/if statement). It's also good that black can be set to a line width of 100 though, that was not possible last I investigated it.

gtaylor:
Your provided example above is definitely funky. But as I was looking through the full set of diffs I saw many more positive changes than the more head-scratching stuff. For example, [this](https://github.com/evennia/evennia/pull/1940/files#diff-61b30844520366700488a280dccc232cR160) is what I'd consider to be a positive outcome. And even in cases where things got weird, there is consistency whereas formatting varies all over the place right now.

In an ideal case scenario, the code formatter would do everything "right" (for some definition of right) 100% of the time. However, we as humans also do some... interesting things with formatting as well. Also important to note that Black is always getting better.

Perhaps the question ends up being, do we feel like this would represent an overall net positive, even if there are a few funky cases? Do we value having a defined style that is enforced and maintained automatically?

Griatch:
@gtaylor Ok, I have cherry-picked your Makefile and extended it a bit before running black on the sources myself. I also added black to travis. So I'm merging part of this PR (which is why it's not auto-closing). But we'll try with black for now, see how it goes. :)

PR #1943, opened 2019-09-19
1928 nested attributes

### LINKED ISSUE #1928: Nested attribute acccess

When using complex data types on object attrs, accessing nested keys/indexes would be awesome.

1. `set here/test_list = ['this', 'that', 'the other']`
2. `set here/test_list/0`
3. 

Expected: `Attribute Location Name/test_list/0 =  this`
Actual: `Attribute Location Name/test_list = ['This', 'that', 'the other'] Location Name has no attribute '0'.`

Looks like `search_for_obj` in `evennia/commands/default/building.py` would be the place to implement this chained attr lookup. Shouldn't be too hard to implement, but would break folks expecting the multiple attr lookup, so it might need some new semantics/command.

I'm happy to attempt a PR if there is consensus on the interface.

aarcro:
Resolves #1928

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -1688,12 +1688,22 @@ class CmdSetAttribute(ObjManipCommand):
 
     def rm_attr(self, obj, attr):
         """
-        Remove an attribute from the object, and report back.
+        Remove an attribute from the object, or a nested data structure, and report back.
         """
-        if obj.attributes.has(attr):
-            val = obj.attributes.has(attr)
-            obj.attributes.remove(attr)
-            return "\nDeleted attribute '%s' (= %s) from %s." % (attr, val, obj.name)
+        for key, nested_keys in self.split_nested_attr(attr):
+            if obj.attributes.has(key):
+                if nested_keys:
+                    del_key = nested_keys[-1]
+                    val = obj.attributes.get(key)
+                    val = self.do_nested_lookup(val, *nested_keys[:-1])
+                    if val is not self.not_found:
+                        del val[del_key]
+                        obj.attributes.add(key, val)
+                    return "\nDeleted attribute '%s' (= nested) from %s." % (attr, obj.name)
+                else:
+                    exists = obj.attributes.has(key)
+                    obj.attributes.remove(attr)
+                    return "\nDeleted attribute '%s' (= %s) from %s." % (attr, exists, obj.name)
         else:
             return "\n%s has no attribute '%s'." % (obj.name, attr)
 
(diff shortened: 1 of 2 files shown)


evennia/commands/default/building.py
```diff
@@ -1515,31 +1518,130 @@ def check_attr(self, obj, attr_name):
...
+                deep = self.do_nested_lookup(lookup_value, *nested_keys[:-1])
+                if deep is not self.not_found:
+                    # To support appending and inserting to lists
+                    # a key that starts with @ will insert a new item at that
+                    # location, and move the other elements down.
+                    # Just '@' will append to the list
+                    if isinstance(acc_key, str) and acc_key[0] == LIST_APPEND_CHAR:
+                        try:
+                            if len(acc_key) > 1:
+                                where = int(acc_key[1:])
```
Griatch: If the user accidentally enters something wrong, like trying (wrongly) to use `[+'key']`, maybe with the intent of "appending" to a dictionary (which is an easy mistake to make if you are not too Python savvy), this will lead to a traceback about not being able to convert to an int. Best to catch the `ValueError` here as well and to give a proper error message rather than the raw traceback.

evennia/commands/default/building.py
```diff
@@ -1515,31 +1518,130 @@ def check_attr(self, obj, attr_name):
...
+        return result
+
     def view_attr(self, obj, attr):
         """
         Look up the value of an attribute and return a string displaying it.
         """
-        if obj.attributes.has(attr):
-            return "\nAttribute %s/%s = %s" % (obj.name, attr,
-                                               obj.attributes.get(attr))
+        for key, nested_keys in self.split_nested_attr(attr):
```
Griatch: If doing something like `set obj/existing_nested_attr['non_existing']`, the `else` clause below leads to a somewhat confusing error `x has no Attribute y`. At this point it is clear that the user _has_ the Attribute - the error is rather than it does not contain the nested key/index. It would be better to distinguish these two error cases.

evennia/commands/default/building.py
aarcro: Nice catch! I'll add a test :)

evennia/commands/default/building.py
```diff
@@ -1515,31 +1518,130 @@ def check_attr(self, obj, attr_name):
...
     def set_attr(self, obj, attr, value):
+        done = False
+        for key, nested_keys in self.split_nested_attr(attr):
+            if obj.attributes.has(key) and nested_keys:
+                acc_key = nested_keys[-1]
+                lookup_value = obj.attributes.get(key)
+                deep = self.do_nested_lookup(lookup_value, *nested_keys[:-1])
+                if deep is not self.not_found:
+                    # To support appending and inserting to lists
+                    # a key that starts with @ will insert a new item at that
```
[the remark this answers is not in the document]
aarcro: Thanks!

evennia/commands/default/building.py
aarcro: OK, I added a note if a nested lookup was attempted.

Issue #1949, opened 2019-09-24
[BUG] TagHandler - AttributeError: `'int' object has no attribute 'strip'`


strikaco:
#### Describe the bug
There is some inconsistency with the way the TagHandler manages integer-ish values. 

Sometimes it works, sometimes it doesn't, it appears to depend on the specific method used and the pre-casting of the integer in question.

#### To Reproduce
```
from evennia.prototypes.spawner import spawn        

class SpawnerAttributesTest(EvenniaTest):
    
    def setUp(self):
        super().setUp()
        
        # Generic example object
        # Extraneous attributes related to other tests; please ignore
        self.prototype = {
            'key': 'bob',
            'age': 56,
            'attrs': (
                ('favorite_flavor', ['vanilla', 'swirl'], 'ice_cream_prefs', None),
            ),
            'tags': [
                ('happy', 'mood', ''),
            ]
        }
        
    def test_numeric_tags(self):
        "Check if tags consisting of numeric values are allowed."
        # Spawn a normal object and add the tag as str via the TagHandler
        obj = spawn(self.prototype)[0]
        self.assertTrue(obj)
        obj.tags.add('10467', category='zip')
        self.assertTrue('10467' in obj.tags.all(), "Stringified integer was not added to tags via tags.add method.")
        
        # Do the same with a numeric tag cast as int
        obj = spawn(self.prototype)[0]
        self.assertTrue(obj)
        obj.tags.add(10467, category='zip')
        self.assertTrue('10467' in obj.tags.all(), "Integer was not added to tags via tags.add method.")
        
        # Try creating numeric tag using spawner workflow both cast as int and str
        new = {k:v for k,v in self.prototype.items()}
        new['tags'].extend([
            ('10467', 'zip', '10467'),
            (10468, 'zip', 10468),
        ])
        obj = spawn(new)[0]
        self.assertTrue(obj)
        self.assertTrue('10467' in obj.tags.all(), "Stringified integer was not added to tags when using spawner.")
        self.assertTrue('10468' in obj.tags.all(), "Integer was not added to tags when using spawner.")
        
```
#### Traceback
```
======================================================================
ERROR: test_numeric_tags (typeclasses.tests.SpawnerAttributesTest)
Check if tags consisting of numeric values are allowed.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/workspace/demogame/typeclasses/tests.py", line 75, in test_numeric_tags
    obj = spawn(new)[0]
[...]

PR #1957, opened 2019-10-01
Update tags.py

Biswa24:
Added typecast where strip function was causing error.


Biswa24:
Added typecast where strip function was causing error

strikaco:
@Griatch this closes #1949.

Griatch:
@strikaco Aha, thanks, forgot about that one and there was no reference to it in the PR. Closing the ticket then. :D

Issue #1951, opened 2019-09-25
[FEATURE] Django Admin: Add id, other core fields to ObjectDB search_fields scope


strikaco:
#### Is your feature request related to a problem? Please describe.
It would be useful if the Admin view for the ObjectDB allowed searching/filtering by more of the core Evennia fields.

For example, it displays the following headers on the list view:
`('id', 'db_key', 'db_account', 'db_typeclass_path')`

...but you can only search on these fields:
`['^db_key', 'db_typeclass_path']`

Despite displaying them, one is currently unable to locate an object by its ID. It would be useful to be able to search for objects associated with a given account as well. 

Ideally `search_fields` should mirror `list_display` where possible/practical.

Also, a question about this:
https://github.com/evennia/evennia/blob/541a6fa85525f0c40d47cd0074149ba49b9b60c9/evennia/objects/admin.py#L85

Is `^db_key` a typo or a Django secret to anchor queries? I've never seen this syntax in this context.

mdaizovi:
Can I take this?

Griatch:
@mdaizovi Give it a shot if you want to!

Anuj-Devrani:
Hi, I am new to contributing on github. Feel free to counsel me.
@strikaco for more info on your question about the usage of  '^',  read the documentation here:
[Model Admin Search Fields](https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.search_fields)

@Griatch Can I give a shot to this issue?

PR #1961, opened 2019-10-02
Add id, other core fields to ObjectDB search_fields

Anuj-Devrani:
Add id, other core fields to ObjectDB search_fields

HactoberFest. This is my first PR on github for an open source project. Please guide me if I made a mistake.

#### Issues Referenced
Issue #1951

### DIFF
--- a/evennia/objects/admin.py
+++ b/evennia/objects/admin.py
@@ -95,7 +95,7 @@ class ObjectDBAdmin(admin.ModelAdmin):
     list_display = ("id", "db_key", "db_account", "db_typeclass_path")
     list_display_links = ("id", "db_key")
     ordering = ["db_account", "db_typeclass_path", "id"]
-    search_fields = ["^db_key", "db_typeclass_path"]
+    search_fields = ["^db_key", "db_typeclass_path", "id", "db_account__id"]
     raw_id_fields = ("db_destination", "db_location", "db_home")
 
     save_as = True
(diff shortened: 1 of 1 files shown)


evennia/objects/admin.py
```diff
@@ -95,7 +95,7 @@ class ObjectDBAdmin(admin.ModelAdmin):
     list_display = ("id", "db_key", "db_account", "db_typeclass_path")
     list_display_links = ("id", "db_key")
     ordering = ["db_account", "db_typeclass_path", "id"]
-    search_fields = ["^db_key", "db_typeclass_path"]
+    search_fields = ["^db_key", "db_typeclass_path", "id", "db_account__id"]
```
Griatch: - Since accounts and objects will have the same range of ids (starting from 1), filtering last on account will probably mean that account-id searching will rarely be relevant (since there are many more objects than accounts in a normal game). It'd say that it's probably better to filter on the account's `^db_key` field instead. 
- The filter order should also be changed, it makes more sense to prioritize `id` first by putting it first in the list.

Anuj-Devrani: Ok. That makes sense. 
So Should we switch to exact search if a number is specifed.
That means if the user supplies a number we should show only exact match.
So id should be replaced by =id .

Griatch: Sure, sounds good, although I wonder if this is not automatically the case when querying with a primary key.

Anuj-Devrani: I have updated my PR.

PR #1962, opened 2019-10-05
add blank=True

mdaizovi:
It wasn't obvious to me how to make migrations in this project, so all I did was add `blank=True`


Griatch:
@mdaizovi Bump.

mdaizovi:
@Griatch sorry, the first time I only skimmed the install so I didn't see that I needed to use `evennia` in order to make migrations. 
Question: The first time I ran the command I hadn't noticed that you said to specify the app, so 2 extra migrations were made. I removed them and did not include them in the PR, but isn't that a matter of concern? Or is that expected? I got one in `accounts` and another in `objects`

PR #1969, opened 2019-10-19
Unit tests for evennia/utils/text2html.

Henddher:
Unit tests for evennia/utils/text2html.

Coverage increase from 47.46% to 98.28%.

#1458 Hacktoberfest

### DIFF
--- a/evennia/utils/tests/test_text2html.py
+++ b/evennia/utils/tests/test_text2html.py
@@ -1,0 +1,92 @@
+"""Tests for text2html """
+
+from django.test import TestCase
+from evennia.utils import ansi, text2html
+import mock
+
+
+class TestText2Html(TestCase):
+
+    def test_re_color(self):
+        parser = text2html.HTML_PARSER
+        self.assertEqual("foo", parser.re_color("foo"))
+        self.assertEqual(
+            "<span class=\"err\">red</span>foo",
+            parser.re_color(ansi.ANSI_RED + "red" + ansi.ANSI_NORMAL + "foo"))
+
+    def test_re_bold(self):
+        parser = text2html.HTML_PARSER
+        self.assertEqual("foo", parser.re_bold("foo"))
+        self.assertEqual(
+            # "a <strong>red</strong>foo",  # TODO: why not?
+            "a <strong>redfoo</strong>",
+            parser.re_bold(
+                "a " + ansi.ANSI_HILITE + "red" + ansi.ANSI_UNHILITE + "foo"))
+
+    def test_re_underline(self):
+        parser = text2html.HTML_PARSER
+        self.assertEqual("foo", parser.re_underline("foo"))
+        self.assertEqual(
+            "a <span class=\"underline\">red</span>" + ansi.ANSI_NORMAL + "foo",
+            parser.re_underline(
+                "a " + ansi.ANSI_UNDERLINE + "red"
+                + ansi.ANSI_NORMAL  # TODO: why does it keep it?
+                + "foo"))
+
+    def test_re_blinking(self):
+        parser = text2html.HTML_PARSER
+        self.assertEqual("foo", parser.re_blinking("foo"))
+        self.assertEqual(
+            "a <span class=\"blink\">red</span>" + ansi.ANSI_NORMAL + "foo",
+            parser.re_blinking(
+                "a " + ansi.ANSI_BLINK + "red"
+                + ansi.ANSI_NORMAL  # TODO: why does it keep it?
+                + "foo"))
+
+    def test_re_inversing(self):
+        parser = text2html.HTML_PARSER
+        self.assertEqual("foo", parser.re_inversing("foo"))
+        self.assertEqual(
+            "a <span class=\"inverse\">red</span>" + ansi.ANSI_NORMAL + "foo",
+            parser.re_inversing(
+                "a " + ansi.ANSI_INVERSE + "red"
+                + ansi.ANSI_NORMAL  # TODO: why does it keep it?
+                + "foo"))
+
+    def test_remove_bells(self):
+        parser = text2html.HTML_PARSER
+        self.assertEqual("foo", parser.remove_bells("foo"))
+        self.assertEqual(
+            "a red" + ansi.ANSI_NORMAL + "foo",
+            parser.remove_bells(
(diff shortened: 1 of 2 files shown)


Henddher:
Another one @Griatch 

Also in the spirit of Hacktoberfest :)

evennia/utils/tests/test_text2html.py
```diff
@@ -0,0 +1,180 @@
...
+            parser.re_color(ansi.ANSI_BACK_RED + "red" + ansi.ANSI_NORMAL + "foo"))
+        self.assertEqual(
+            "<span class=\"bgcolor-001\"><span class=\"color-002\">red</span></span>foo",
+            parser.re_color(ansi.ANSI_BACK_RED + ansi.ANSI_UNHILITE + ansi.ANSI_GREEN + "red" + ansi.ANSI_NORMAL + "foo"))
+
+    def test_re_bold(self):
+        parser = text2html.HTML_PARSER
+        self.assertEqual("foo", parser.re_bold("foo"))
+        self.assertEqual(
+            # "a <strong>red</strong>foo",  # TODO: why not?
```
Griatch: If this is really the output then this is obviously wrong behavior - the visual result of this is a bold `redfoo` rather than a bold `red`.

Henddher: Yes. The current result is "**redfoo**" (with the ANSI_NORMAL char seq still present) although I was expecting it to be "**red**foo".

Sounds like a plan!

Can you 2x check each of the TODOs to make sure I am not misinterpreting the intention of the code? I'll use those as citations in the new bug.

Meanwhile, I'll [unitest.skip](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) each.

Henddher: #1974 created. Waiting for specific citations.

evennia/utils/tests/test_text2html.py
```diff
@@ -0,0 +1,187 @@
...
+
+    @unittest.skip("parser issues")
+    def test_re_blinking(self):
+        parser = text2html.HTML_PARSER
+        self.assertEqual("foo", parser.re_blinking("foo"))
+        self.assertEqual(
+            "a <span class=\"blink\">red</span>" + ansi.ANSI_NORMAL + "foo",
+            parser.re_blinking(
+                "a " + ansi.ANSI_BLINK + "red"
+                + ansi.ANSI_NORMAL  # TODO: why does it keep it?
```
Griatch: Same here, whereas it's not obvious this sub-parser should handle ANSI_NORMAL here, the global result is a blinking `redfoo` which is not what we want. So it's something that needs to be fixed.

evennia/utils/tests/test_text2html.py
```diff
@@ -0,0 +1,187 @@
...
+    @unittest.skip("parser issues")
+    def test_remove_bells(self):
+        parser = text2html.HTML_PARSER
+        self.assertEqual("foo", parser.remove_bells("foo"))
+        self.assertEqual(
+            "a red" + ansi.ANSI_NORMAL + "foo",
+            parser.remove_bells(
+                "a " + ansi.ANSI_BEEP + "red"
+                + ansi.ANSI_NORMAL  # TODO: why does it keep it?
+                + "foo"))
```
Griatch: This is correct behavior and not a bug. An `ANSI_BEEP` is just a single instruction and has no start/end tag. We should just clean it out and not touch the `ANSI_NORMAL` (which is likely terminating another, earlier instruction).

evennia/utils/tests/test_text2html.py
```diff
@@ -0,0 +1,187 @@
...
+            parser.convert_linebreaks("a\n redfoo\n"))
+
+    @unittest.skip("parser issues")
+    def test_convert_urls(self):
+        parser = text2html.HTML_PARSER
+        self.assertEqual("foo", parser.convert_urls("foo"))
+        self.assertEqual(
+            "a <a href=\"http://redfoo\" target=\"_blank\">http://redfoo</a> runs",
+            parser.convert_urls("a http://redfoo runs"))
+            # TODO: doesn't URL encode correctly
```
Griatch: Given the input, it looks like `http://redfoo` is a correct URL output. In what way is the encoding incorrect?

Henddher: It's not calling a formal urlencode function. It's just doing string replacement (which is unsanitized)

evennia/utils/tests/test_text2html.py
Henddher: Test is now `unittest.skip()`ed

evennia/utils/tests/test_text2html.py
```diff
@@ -0,0 +1,187 @@
...
+    @unittest.skip("parser issues")
+    def test_re_underline(self):
+        parser = text2html.HTML_PARSER
+        self.assertEqual("foo", parser.re_underline("foo"))
+        self.assertEqual(
+            "a <span class=\"underline\">red</span>" + ansi.ANSI_NORMAL + "foo",
+            parser.re_underline(
+                "a " + ansi.ANSI_UNDERLINE + "red"
+                + ansi.ANSI_NORMAL  # TODO: why does it keep it?
+                + "foo"))
```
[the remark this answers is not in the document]
Henddher: Test is now `unittest.skip()`ed

evennia/utils/tests/test_text2html.py
Henddher: Test is now `unittest.skip()`ed

evennia/utils/tests/test_text2html.py
```diff
@@ -0,0 +1,187 @@
...
+    @unittest.skip("parser issues")
+    def test_re_inversing(self):
+        parser = text2html.HTML_PARSER
+        self.assertEqual("foo", parser.re_inversing("foo"))
+        self.assertEqual(
+            "a <span class=\"inverse\">red</span>" + ansi.ANSI_NORMAL + "foo",
+            parser.re_inversing(
+                "a " + ansi.ANSI_INVERSE + "red"
+                + ansi.ANSI_NORMAL  # TODO: why does it keep it?
+                + "foo"))
```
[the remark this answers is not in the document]
Henddher: Test is now `unittest.skip()`ed

PR #1983, opened 2019-10-30
Replacing deprecated method from cgi with html.escape()

driscoll:

Replacing deprecated escape function from the cgi module with escape function from the html module (new since Python 3.2). 
https://docs.python.org/3.5/library/html.html#html.escape

Fixes errors printing messages to the web console.


Griatch:
Thanks for the fix!

Issue #788, opened 2015-08-22
External API


Kelketek:
Use the likes of something like the Django REST Frame to create an API for external utilities that don't require being part of the current process. This should be extensible/overridable to allow developers to add their own calls or change the existing ones.

Some example applications:
1. Collecting gamewide stats
2. Externally run scripting languages
3. Integration with external services on a push basis instead of a pull

Griatch:
@Kelketek 

Could you elaborate some more on what you mean by this and what kind of changes your are suggesting and their effects, maybe with links to relevant django docs?

Kelketek:
@Griatch http://www.django-rest-framework.org/ contains information about the Django REST framework. Having this would allow for more straightforward JS integration. It might be possible, since they both use JSON, to reuse OOB code in some way so that a normal game connection isn't required for certain calls. You could have a call that moves an object, sends a message, lists owned items, deletes objects, etc.

Griatch:
@Kelketek 

Hm, that does sounds pretty nifty! 
I'm thinking I might put up a devel branch of my current reworking of the web client infrastructure for input from more people, maybe also for things like this; one still needs OOB for traditional telnet clients but the webclient is using plain JSON in my most recent implementation and who knows what other angles one might catch.

Griatch:
@Kelketek 

I can see this being useful for allowing powerful extra features and effects on the database. I'm unsure about security though. Do you have any experience with how to do this in a safe way (can't allow just anyone to do stuff with the database after all)?

Kelketek:
@Griatch The Django REST Framework, as linked earlier, has authentication support, with proper token handling, etc. You can plug in your own authentication and have permissions checks done based on locks if needed.

whitenoiseoss:
Going through Django REST Framework documentation plus this tutorial here: https://code.tutsplus.com/tutorials/beginners-guide-to-the-django-rest-framework--cms-19786 gives a really nice and easy picture of how you can start building an API on top of all of the Evennia functionality.

Using DRF, everything will be 100% portable to the API. All that seems to be required is writing up the Serializers, defining exportable fields, and then editing will have to be written in. It will be a lot of code, but none of it is particularly difficult.

whitenoiseoss:
Important features for security (looking above at Griatch's concern) are CSRF and CORS. You could also use something like JWT in place of DRF's default choices for some additional security (this would use either a private key or a secret in order to encrypt the token object).

[...]

PR #1984, opened 2019-11-08
Add django rest framework support

TehomCD:
Still WIP!

This starts to add a RESTful API to Evennia. Currently are just retrieve views - I still need to work on update/partial update/destroy/create to determine everything is working properly. I'm open to suggestions on how to handle attributes - probably read-only or only support primitive types that can be evaluated by literal_eval? I feel most of the examples here will probably be didactic in nature due to requiring customization by any client to consume them. I added a custom permission class to allow for using locks for most permissions: I think it'd be much more intuitive to keep with how Evennia does access/permissions rather than branch into something entirely new, but it'd definitely need to be very well documented, as does everything else - I'll want to add a ton of example usages in docstrings and/or in the wiki.

Also need to add tests for everything.

A RESTful API for Evennia allows for more full-featured modern front-end architectures: modern frameworks such as React/Redux assume a REST API backend for their actions. It also supports other clients, such as mobile clients interacting with a game, etc.

#788


TehomCD:
I think how I'm going to try to handle attributes is using a serializer for validation, then defining an action in the viewsets that will explicitly call the attribute handler for adding/removing them. Same with tags.

evennia/web/api/filters.py
```diff
@@ -0,0 +1,29 @@
+from django_filters.rest_framework.filterset import FilterSet
+
+from evennia.objects.models import ObjectDB
+from evennia.accounts.models import AccountDB
+from evennia.scripts.models import ScriptDB
+
+
+class TypeclassFilterMixin(object):
```
Griatch: While a mixin works, I find it just makes this code a bit too complex and hard to read in this case - it's just an object that you then have to de-reference just a few lines below. You could just define this list in a global var at the top of this module and use it below.

evennia/web/api/permissions.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class EvenniaPermission(permissions.BasePermission):
+    """
+    A Django Rest Framework permission class that allows us to use
+    Evennia's permission structure. Based on the action in a given
+    view, we'll check a corresponding Evennia access/lock check.
+    """
+    # subclass this to change these permissions
+    MINIMUM_LIST_PERMISSION = "builder"
```
Griatch: I wonder if one could not link this into the settings file's permission levels instead somehow.

Griatch:
Awesome, looks promising so far. Didn't review anything yet (since it's still WIP), just made some observations on first glance. But this looks nice and like a good start.

TehomCD:
Just as an update, I haven't forgotten about this - I've just been buried between work and trying to get Arx up to date with the latest version of Evennia/python 3/django. I hope to have time to finish this after that.

Griatch:
@TehomCD Is this still ongoing?

TehomCD:
Yeah, I'd like to get back to this when I have time and ask for a review when I finish create/update/delete endpoints for typeclasses. Current version works as a reference implementation for read-only APIs, though.

TehomCD:
As an update, everything is done but the create endpoint and an endpoint for adding/updating attributes. If you want to review what's already there, feel free. I'll work on finishing the rest hopefully soon.

TehomCD:
While working on the action for setting/removing attributes, I found some errors related to deprecated fields that were removed in django 2.0 but are present in the PickleField. I think this doesn't come up normally since they're related to form/serialization data, which wasn't used in the admin window. I'll see about fixing them as part of the PR. One is straightforward: the `_get_val_from_obj()` method has become `value_from_object`. 

The other is more complex and does break a little bit of functionality in Evennia that was easy to overlook: it seems that `get_db_prep_lookup` is no longer called in fields, but now requires registering a 'Lookup' subclass. This means that the check to prevent "__in" queries for Attribute's `db_value` is no longer working. I'll see if I can poke around at it and get that working - I've never seen the Lookup stuff before, so it's kind of interesting to me.

TehomCD:
Okay, the initial version of this is done/working.

- Adds CRUD viewsets for typeclasses
- Adds a `set-attribute` action for adding/removing attributes from typeclasses

It wouldn't be difficult to add a similar action for tags as well.

evennia/web/api/permissions.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class EvenniaPermission(permissions.BasePermission):
+    """
+    A Django Rest Framework permission class that allows us to use
+    Evennia's permission structure. Based on the action in a given
+    view, we'll check a corresponding Evennia access/lock check.
+    """
+    # subclass this to change these permissions
+    MINIMUM_LIST_PERMISSION = settings.REST_FRAMEWORK["DEFAULT_LIST_PERMISSION"]
```
Griatch: Using `.get()` with a default value here would both make the read-in more stable as well as act as a helpful hint when reading the code as to what these permission values may be looking like.

evennia/web/api/permissions.py
```diff
@@ -0,0 +1,59 @@
...
+    view_locks = settings.REST_FRAMEWORK["DEFAULT_VIEW_LOCKS"]
+    destroy_locks = settings.REST_FRAMEWORK["DEFAULT_DESTROY_LOCKS"]
+    update_locks = settings.REST_FRAMEWORK["DEFAULT_UPDATE_LOCKS"]
+
+    def has_permission(self, request, view):
+        """
+        This method is a check that always happens first. If there's an object involved,
+        such as with retrieve, update, or delete, then the has_object_permission method
+        is also called if this returns True. If we return False, a permission denied
+        error is raised.
```
Griatch: Nice docstring, but it is not on the standard form for Evennia docstrings. Here is an example of rewriting it on the Evennia form: 

```
Determine if the permission check is passed for a given view. 

Args: 
    request (Request): The incoming request object. 
    view (View): The django view we are checking permission for. 

Returns:
    bool: If permission is granted or not. If we return False here, a PermissionDenied error will be raised
         from the view. 

Notes:  
        This method is a check that always happens first. If there's an object involved,
        such as with retrieve, update, or delete, then the has_object_permission method
        is called after this, assuming this returns `True`. 
```

evennia/web/api/permissions.py
```diff
@@ -0,0 +1,59 @@
...
+            return True
+        # these actions don't support object-level permissions, so use the above definitions
+        if view.action == "list":
+            return request.user.has_permistring(self.MINIMUM_LIST_PERMISSION)
+        if view.action == "create":
+            return request.user.has_permistring(self.MINIMUM_CREATE_PERMISSION)
+        return True  # this means we'll check object-level permissions
+
+    @staticmethod
+    def check_locks(obj, user, locks):
```
Griatch: This needs a docstring.

evennia/web/api/permissions.py
```diff
@@ -0,0 +1,59 @@
...
+        return any([obj.access(user, lock) for lock in locks])
+
+    def has_object_permission(self, request, view, obj):
```
Griatch: The description is ok, but needs to be on the Evennia docstring form.

evennia/web/api/permissions.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        This method assumes that has_permission has already returned True. We check
+        equivalent Evennia permissions in the request.user to determine if they can
+        complete the action. If so, we return True. Otherwise we return False, and
+        a permission denied error will be raised.
+        """
+        if request.user.is_superuser:
```
Griatch: This is actually not always true in normal lock mechanisms - there is a special provision for lock-checks that is called `no_superuser_bypass`. This is necessary for certain system mechanism (such as on-object commands to be able to be locked to that character only - without it superusers would get all commands from every object/character in the room). Offhand I don't know if it at all makes sense to have a `no_superuser_bypass` on the REST API level, just pointing out that this conditional makes the lock behavior not completely identical.

evennia/web/api/permissions.py
```diff
@@ -0,0 +1,59 @@
...
+            return True
+        if view.action in ("list", "retrieve"):
+            # access_type is based on the examine command
+            return self.check_locks(obj, request.user, self.view_locks)
+        if view.action == "destroy":
+            # access type based on the destroy command
+            return self.check_locks(obj, request.user, self.destroy_locks)
+        if view.action in ("update", "partial_update", "set_attribute"):
+            # access type based on set command
+            return self.check_locks(obj, request.user, self.update_locks)
```
Griatch: no `create` action here?

evennia/web/api/serializers.py
```diff
@@ -0,0 +1,69 @@
+"""
+Serializers in the Django Rest Framework are similar to Forms in normal django.
+They're used for transmitting and validating data, both going to clients and
+coming to the server. However, where forms often contained presentation logic,
+such as specifying widgets to use for selection, serializers typically leave
+those decisions in the hands of clients, and are more focused on converting
+data from the server to JSON (serialization) for a response, and validating
+and converting JSON data sent from clients to our enpoints into python objects,
+often django model instances, that we can use (deserialization).
```
Griatch: Nice description! :+1:

evennia/web/api/serializers.py
```diff
@@ -0,0 +1,69 @@
...
+from rest_framework import serializers
+
+from evennia.objects.models import ObjectDB
+from evennia.accounts.models import AccountDB
+from evennia.scripts.models import ScriptDB
+from evennia.typeclasses.attributes import Attribute
+from evennia.typeclasses.tags import Tag
+
+
+class AttributeSerializer(serializers.ModelSerializer):
```
Griatch: It's very common to not use docstrings on these (and it gets pretty repetitive), so for once I guess I shall not demand it ;)

evennia/web/api/serializers.py
```diff
@@ -0,0 +1,69 @@
...
+    class Meta:
+        model = ObjectDB
+        fields = ["id", "db_key"]
+
+
+class TypeclassSerializerMixin(object):
+    db_attributes = AttributeSerializer(many=True)
+    db_tags = TagSerializer(many=True)
+
+    shared_fields = ["id", "db_key", "db_attributes", "db_tags", "db_typeclass_path"]
```
Griatch: I do think an `aliases` `MethodField` would be useful here, with the tags on the object identified with the `db_tagtype="alias"`. A similar argument could be made for `permissions`, leaving `tags` as only those tags with neither the `alias` or `permission` `db_tagtype`s.

evennia/web/api/serializers.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class ObjectDBSerializer(TypeclassSerializerMixin, serializers.ModelSerializer):
+    contents = SimpleObjectDBSerializer(source="locations_set", many=True, read_only=True)
+
+    class Meta:
+        model = ObjectDB
+        fields = ["db_location", "db_home", "contents"] + TypeclassSerializerMixin.shared_fields
```
Griatch: I think `exits` may also be useful here. They  could then be removed from `contents` based on having a non-null `db_destination` field.

evennia/web/api/serializers.py
```diff
@@ -0,0 +1,69 @@
...
+        read_only_fields = ["id", "db_attributes", "db_tags"]
+
+
+class AccountDBSerializer(TypeclassSerializerMixin, serializers.ModelSerializer):
+    db_key = serializers.CharField(required=False)
+
+    class Meta:
+        model = AccountDB
```
Griatch: Other things that would be useful here would be `is_logged_in` as a boolean and/or an integer count of how many sessions (maybe even the session-ids?) which are connected to this account right now.

evennia/web/api/tests.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def test_retrieve(self):
+        views = self.get_view_details("detail")
+        for view in views:
+            with self.subTest(msg="Testing {} retrieve".format(view.view_name)):
+                view_url = reverse(
+                    "api:{}".format(view.view_name), kwargs={"pk": view.obj.pk}
+                )
+                response = self.client.get(view_url)
+                self.assertEqual(response.status_code, 200)
```
Griatch: It would be good if this test also included an assertion about just what is returned as well. That it has a 200 status only means itreturned _something_ without dying in the process. This is true for some other tests below as well, might not point each one out explicitly.

evennia/web/api/tests.py
```diff
@@ -0,0 +1,133 @@
...
+        for view in views:
+            with self.subTest(msg="Testing {} update".format(view.view_name)):
+                view_url = reverse(
+                    "api:{}".format(view.view_name), kwargs={"pk": view.obj.pk}
+                )
+                # test both PUT (update) and PATCH (partial update) here
+                for new_key, method in (("foobar", "put"), ("fizzbuzz", "patch")):
+                    field = "username" if "account" in view.view_name else "db_key"
+                    data = {field: new_key}
+                    response = getattr(self.client, method)(view_url, data=data)
```
Griatch: We have the response here, why not check that it looks as we expect?

evennia/web/api/views.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class TypeclassViewSetMixin(object):
+    """
+    This mixin adds some shared functionality to each viewset of a typeclass. They all use the same
+    permission classes and filter backend. You can override any of these in your own viewsets.
+    """
+    # permission classes determine who is authorized to call the view
+    permission_classes = [EvenniaPermission]
+    # the filter backend allows for retrieval views to have filter arguments passed to it
```
Griatch: An example of this is suitable.

requirements.txt
```diff
@@ -1,9 +1,12 @@
 # Evennia dependencies
 
 # general
+attrs >= 19.2.0
 django >= 2.2.5, < 2.3
 twisted >= 19.2.1, < 20.0.0
 pytz
+djangorestframework >= 3.10.3
```
Griatch: Would it not be safer to restrict these to an upper version number as well?

evennia/web/api/filters.py
```diff
@@ -0,0 +1,27 @@
+from django_filters.rest_framework.filterset import FilterSet
+
+from evennia.objects.models import ObjectDB
+from evennia.accounts.models import AccountDB
+from evennia.scripts.models import ScriptDB
+
+SHARED_FIELDS = ["db_key", "db_typeclass_path", "db_tags__db_key", "db_tags__db_category"]
```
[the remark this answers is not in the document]
TehomCD: If I remember right, you can define custom filters that map to arbitrary querystrings, so I should be able to do this

requirements.txt
TehomCD: Any preference between `< 3.11` or `< 4.0`?

evennia/web/api/filters.py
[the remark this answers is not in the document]
TehomCD: This is a bit more complex than I thought since django-filters don't support complex queries by default: what we want is `?alias=foo` doing `Q(db_tags__db_tagtype="alias") & Q(db_tags__db_key="foo")`. I think this is probably doable, but it's not something their documentation covers as far as I can tell. I'll go through django-filter's code a bit and see what I can pick out.

requirements.txt
Griatch: I don't know offhand what DRF's version policy is and if 3.11 would be a backwards-incompatible version or that won't happen until 4.0 (if they have such considerations).

evennia/web/api/filters.py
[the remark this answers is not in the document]
Griatch: Maybe it would be useful to just use key but to embed the alias-search within the key search as well? That is after all how other searching works (you give a search term and both keys and aliases are searched together).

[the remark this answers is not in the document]
TehomCD: django-filter works by letting you specify specific GET arguments to be passed along as filtering: it's not like a search bar where there's just one argument. The idea is that for a front-end form you might have a search box that could have drop-downs, radio buttons, or whatever in addition to a CharField, and those arguments all need to be handled separately to further constrain results. As an example from django admin, think of it as not just the search bar, but also the filters you can set on the right side of the admin window.

I did come up with a way to do it by letting them specify an ?aliases or ?permissions argument and establishing a pattern that should be easy enough to extend. It'd be like this:
`mygame.com/api/objects?aliases="blue"&permissions="builder"` to search for any objects that have an alias of 'blue' and builder permissions.

evennia/web/api/permissions.py
TehomCD: Create can't be an object-level permission since there's no object to check. You have to check it in the other method, `has_permission`, where list permissions live also.

requirements.txt
TehomCD: Turns out that I had 3.11 installed, so that looks to be safe. I'll do `< 3.12`

evennia/web/api/permissions.py
TehomCD: Good point, I'll remove that check to make it allow for `no_superuser_bypass` to be obeyed and raised by the access checks themselves.

evennia/web/api/serializers.py
TehomCD: I think what I'll do here is lean on the `obj.tags`, `obj.aliases`, and `obj.permissions` handlers with the `return_tagobj` option for full serialization

evennia/web/api/permissions.py
TehomCD: Interesting quirk here - after removing that, suddenly tests failed for a superuser when trying to access scripts. It seems that since scripts have no default lockstring, unlike other typeclasses, that `access()` will return `False` even for a superuser. I'm unsure if that's a bug or not. I resolved it in tests by adding lockstrings to the script object used by the test, but figured I should mention it.

TehomCD:
I think I made all the changes you asked for. I ran into some strange bug that had happened to me at work before, but I had never quite realized what was causing it. Apparently if you have `SerializerMethodField`s in a superclass, they're not found correctly in the fields of subclasses, which try to inspect the fields of the model in the Meta argument instead. No clue why inheritance doesn't work, but the serializer fields need to be explicitly there rather than inherited.

TehomCD:
Codacy check failed due to framework sadness of needing to define methods that don't use `self`.

evennia/web/api/README.md
```diff
@@ -0,0 +1,102 @@
...
+>>> response.json()
+{"db_key": "A shiny sword", "id": 214, "db_location": None, ...}
+```
+In the above example, you make a POST request to the /objects/ endpoint with
+the name of the object you wish to create passed along as data. Now suppose you
+decided you didn't like the name, and wanted to change it for the newly created
+object:
+```pythonstub
+>>> data = {"db_key": "An even SHINIER sword", "db_location": 50}
+>>> response = requests.post("https://www.mygame.com/api/objects/214",
```
Griatch: This will lead to a 405 `Method not allowed` error. This is a PATCH operation. Currently it also requires an ending slash.

TehomCD:
At a guess, I'd think the 404 from the requests library while it works in postman would be headers that postman adds, like different accept headers:
```
Accept-Language: en
accept: application/json
```
for example. I'll try to duplicate the behavior you're seeing. The 500 error being swallowed is definitely bad, I'll dig into that, and try to resolve the append slash stuff.

I thought I had added all the testing for content equality in the unit tests, let me know if I'm still missing anything there. Oddly, I have tests specifically for getting exits/rooms/etc, so empty responses are strange.

TehomCD:
>>>For some reason, requests.get("http://localhost:4001/api/3", auth=("Griatch", "mypassword")) does not work for me - I consistently get a 404. The very same call in Postman works perfectly. I don't know if this is some issue with requests and localhost, but 127.0.0.1 produces the same result. Since the README uses requests it may be worth investigating.

I think you may have made a typo in requests that you didn't repeat in postman - `api/:id` is not valid, so a 404 is expected.

>>> GET localhost:4001/api/exits, /rooms, /characters return nothing for me. I definitely have examples of each in the database and I access this as a superuser :)

This turned out to be an error in the typeclasses manager `typeclass_search` method. The TypedManagers who called it would first filter results by their own typeclass, preventing any matches from subclasses. Thus, DefaultCharacter, DefaultExit, etc, would always be empty. Easily fixed by `super().filter` rather than `self.filter`, at least.

>>> POST localhost:4001/api/objects (with any data) gives a 500 error because it's lacking an ending backslash (APPEND_SLASH is set). This seems a little unnecessary and at any rate, failing with 500 is pretty ugly. It's also not necessary for GET.

I resolved the trailing slash issues by making it optional (which you would think would be a kwarg you could pass to the router, but I digress). I couldn't reproduce the 500 issue. That's pretty strange and it bugs me that it's not logging errors properly. I'll try to look into it more.

>>> The returns all have the next/prev pagination helpers. These are not mentioned in the README and it's not clear offhand how they are to be used.

Good point, I should link to django's pagination docs in the README.

>>> Attributes appear as a list of ids on the object. I could see it being a pretty expensive operation to retrieve every one on every object when listing all objects, but it would be reasonable to include them inline when retrieving an individual object (similar to tags/aliases). Also, if their ids are listed, there needs to be an endpoint for actually retrieving them, otherwise the list of ids is meaningless.

Yeah, not useful in their current form. Maybe their keys and string representation of the values might be good for a detail view, though I don't know how much of a performance hit that will cause.

So I think I resolved the issues except for 500s being swallowed somehow and coming up with a better format for Attributes. I'll need to think a little about that one, since I'm not sure there's a way that won't look pretty hideous.

TehomCD:
Oh and needing to think about what to do about aliases. Maybe... have a 'name' filter that is used for db_key/aliases and username/aliases? I'd rather not do it implicitly for db_key/username since that's not how it works in game: we present convenience methods for searching by alias/key which are great, but we don't override the behavior of .filter() for example. I'd think that having db_key and username do that would probably be more confusing than having a different search filter entirely.

TehomCD:
Okay, I fixed the attribute serializer so it'll have a sane display, while still letting people update it with the custom endpoint. I added a 'name' filter that will allow for searching by db_key or alias, returning the distinct queryset of results.

Please re-review this at your convenience and let me know if I missed anything, @Griatch .

TehomCD:
I decided to change the serializer to use the attributes/nickshandlers to take advantage of their caching and breaking them into better organized categories.

TehomCD:
I actually really like swagger, and it has an integration already with django rest framework - it takes pretty much minimal work to set up, just add it and go. I thought about adding it in automatically but I didn't want to randomly add dependencies that some people may not want or need. 

The downside of Swagger is that it's very much reliant on you following the boilerplate code - if you don't use a modelview with a serializer it won't know what to list for input/output and will list results as empty and so on.

Griatch:
@TehomCD I've not actually used swagger myself, but everyone tells me it's easy to set up. It sounds like this API should work quite well with it then?

TehomCD:
Yeah, essentially all you do is add the drf library for swagger (drf-swagger or some such, it has a billion stars) to installed_apps/requirements, add an initialization step with some information about your project in your `urls.py` to generate a view function and add that to an endpoint, and it does the rest. By default it'll autodocument everything based on inspecting the classes and seeing what serializers they use for input/output. So it's really great if you're following all the DRF idioms, but it won't do you much good otherwise.

PR #1991, opened 2019-11-16
Task needs to be started if pause() was previously called

### LINKED ISSUE #1990: [BUG] unpause() does not restart the associated task because self.ndb._task is still set
#### Describe the bug
Calling `unpause()` does not restart the task because _start_task() returns immediately if self.ndb._task is set, but `pause()` doesn't remove this object

#### To Reproduce

Within the same running process (i.e. without a reload):
1. attach a repeating script to an object
2. call script.pause()
3. call script.unpause()

Actual: script has "PAUS" set in the scripts display; script doesn't actually start

#### Expected behavior
Expected: script's task restarts with the appropriate timers

    Evennia 0.9.0 (rev 7c6b1d24) (rev 7c6b1d24)
    OS: posix
    Python: 3.7.5
    Twisted: 19.7.0
    Django: 2.2.6

#### Additional context
I have a two-line patch that seems like it might fix this; not sure if this is sustainable but it works for me

borwick:
Hopefully fixes #1990

### DIFF
--- a/evennia/scripts/scripts.py
+++ b/evennia/scripts/scripts.py
@@ -162,9 +162,8 @@ class ScriptBase(ScriptDB, metaclass=TypeclassBase):
         Start task runner.
 
         """
-        if self.ndb._task:
-            return
-        self.ndb._task = ExtendedLoopingCall(self._step_task)
+        if not self.ndb._task:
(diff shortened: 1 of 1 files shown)


Griatch:
This looks fine, but this particular bit of code is prone to cause other side effects, so need to check this a bit before accepting :)

PR #1995, opened 2019-11-29
Allow single quotes and double quotes in lockstrings

GulliblePsychologist:
This PR should (if I didn't mess this up, it's my first PR after all) change the way lockstrings are parsed so that single quotes and double quotes can be used to avoid an issue where comma's can't be used in lockstrings. It will also allow single quotes and double quotes to be used if you escape them correctly with a backslash.

I have made my own "trigger functions" which is a lame copy of lock functions to run when certain events occur in the game.

While developing this, I came across an annoying quirk: you can't use single or double quotes in lockstrings. Because of this, you can't use comma's in lockstrings either because the parser will think it's a new argument.

This is my first PR and I'm not sure how this works. I looked at the wiki pages for contributors but I'm not sure if there are guidelines I should follow here. Please advise.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -2250,7 +2250,6 @@ class CmdLock(ObjManipCommand):
                 caller.msg("You are not allowed to do that.")
                 return
             ok = False
-            lockdef = re.sub(r"\'|\"", "", lockdef)
             try:
                 ok = obj.locks.add(lockdef)
             except LockException as e:
--- a/evennia/locks/lockhandler.py
+++ b/evennia/locks/lockhandler.py
@@ -235,8 +235,8 @@ class LockHandler(object):
                 if not callable(func):
                     elist.append(_("Lock: lock-function '%s' is not available.") % funcstring)
                     continue
-                args = list(arg.strip() for arg in utils.split_arguments(rest) if arg and "=" not in arg)
-                kwargs = dict([arg.split("=", 1) for arg in utils.split_arguments(rest) if arg and "=" in arg])
+                args = split_arguments(rest, False)
+                kwargs = split_arguments(rest, True)
                 lock_funcs.append((func, args, kwargs))
                 evalstring = evalstring.replace(funcstring, "%s")
             if len(lock_funcs) < nfuncs:
--- a/evennia/locks/lockhandler.py
+++ b/evennia/locks/lockhandler.py
@@ -235,8 +235,8 @@ class LockHandler(object):
                 if not callable(func):
                     elist.append(_("Lock: lock-function '%s' is not available.") % funcstring)
                     continue
-                args = list(arg.strip() for arg in rest.split(",") if arg and "=" not in arg)
-                kwargs = dict([arg.split("=", 1) for arg in rest.split(",") if arg and "=" in arg])
+                args = list(arg.strip() for arg in utils.split_arguments(rest) if arg and "=" not in arg)
+                kwargs = dict([arg.split("=", 1) for arg in utils.split_arguments(rest) if arg and "=" in arg])
                 lock_funcs.append((func, args, kwargs))
                 evalstring = evalstring.replace(funcstring, "%s")
             if len(lock_funcs) < nfuncs:
--- a/evennia/locks/lockhandler.py
+++ b/evennia/locks/lockhandler.py
@@ -235,8 +235,8 @@ class LockHandler(object):
                 if not callable(func):
                     elist.append(_("Lock: lock-function '%s' is not available.") % funcstring)
                     continue
-                args = split_arguments(rest, False)
-                kwargs = split_arguments(rest, True)
+                args = utils.split_arguments(rest, False)
+                kwargs = utils.split_arguments(rest, True)
                 lock_funcs.append((func, args, kwargs))
                 evalstring = evalstring.replace(funcstring, "%s")
             if len(lock_funcs) < nfuncs:
--- a/evennia/utils/utils.py
+++ b/evennia/utils/utils.py
@@ -2128,65 +2128,78 @@ def split_arguments(s, named=False):
         arguments (dict): If "named" is True.
 
     """
+    _ESCAPE_SIGN = "\\"
     args = {} if named else []
-    inside = False
-    char = None
+    inside = (False, None)
     escape = False
     tmp = ""
     key = None
     skip = False
 
     for c in s:
-        if named and key is None:
-            if c == ",":
-                tmp = ""
-            elif not skip and c == "=":
-                key = tmp
-                tmp = ""
-            elif c == "\"" or c == "'":
-                skip = True
+        if c == _ESCAPE_SIGN:
+            # Escape sign used.
+            if escape:
+                # Already escaping: print escape sign itself.
+                tmp += _ESCAPE_SIGN
+                escape = False
+            else:
+                # Enter escape mode.
+                escape = True
+        elif escape:
+            # Escape mode: print whatever comes after the symbol.
+            escape = False
+            tmp += c
+        elif inside[0] is True:
+            # Inside single quotes or double quotes
+            # Wait for the end symbol, allow everything else through, allow escape sign for typing quotes in strings
+            if c == inside[1]:
+                # Leaving single/double quoted area
+                inside = (False, None)
             else:
                 tmp += c
-        else:
-            if not named and c == "=" and not inside:
+        elif c == "\"" or c == "'":
+            # Entering single/double quoted area
+            inside = (True, c)
+            continue
+        elif c == "=":
+            if named:
+                if key is None:
+                    # Named parameters mode and equals sign encountered. Record key and continue with value.
+                    key = tmp.strip()
+                    tmp = ""
+                else:
+                    # Invalid location of equals sign in the argument string. Should be placed inside a quoted area.
+                    raise Exception("Invalid syntax.")
+            else:
+                # Not named parameters mode: skip this parameter, because it is a named parameter.
                 skip = True
-            elif c == "," and not inside:
-                if named:
+        elif c == ",":
+            # Comma encountered outside of quoted area.
+            if not skip:
+                # Record value based on whether named parameters mode is set or not.
+                if named and key is not None:
                     args[key] = tmp.strip()
                     key = None
-                elif not skip:
+                elif not named:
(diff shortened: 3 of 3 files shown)


Griatch:
Thanks for the PR; I'm not sure I understand the motivation for wanting commas or quotes in a lockstring however - a lockstring is a (deliberately) very limited structure whose only function is to handle, well, locks. And those should not have need for neither commas nor quotes. 

Could you elaborate on just why you want this? Would not something like Evennia's [inlinefuncs](https://github.com/evennia/evennia/wiki/TextTags#inline-functions) be a better thing to adopt for this - they support quotes and commas just fine. The inlinefunc parser is quite generic and is also used for prototype-parsing, for example.

GulliblePsychologist:
**About using inlinefuncs instead**

I don't know much about inlinefuncs, but from what I could see they seem to be simple functions to manipulate messages being sent to a session.

The wiki doesn't show inlinefuncs with the use of quotes and commas.
The example of inlinefuncs says:

```
say  "This is $pad(a center-padded text, 30,c,-) of width 30."
```

What if I wanted to use a comma in the string "a center-padded text"? Would it work if I wrapped it in quotes?

Then, what if I wanted to use a quote in the string? Can I escape the quote symbol? (\\\")


The other thing I don't see is how an inlinefunc would work in combination with a lockstring. Stripping quotes happens in the CmdLock command class. Wouldn't that keep happening even if an inlinefunc was used here?

Even if it were saved in a lockstring, how would the inlinefunc work when the lock is checked by an NPC with no session? Am I missing something?

**About my use case**

For the game I'm developing I wanted to give builders access to an event system so they could execute some simple functions when an object/player performs a certain action.

These are called "triggers" which can be placed on objects. Since I 'borrowed' code from the lock handler, they now work the same way as lock functions. Except instead of being called to check if an object/player can perform a certain action, they are called when that action has been performed as a way to do something special.

One of my "trigger functions" is called "playmusic" and it allows builders to send a custom message to a client that makes it play music. (I'm using a modified version of the multimedia plugin to make this happen.)

I found that when a builder tries to assign a URL the parser is messing up the string. After looking at it further, I found out that this was because of the way the arguments were being parsed. (Specifically, it was having issues with the equal sign in some URL's, incorrectly adding them to kwargs because of the simplified implementation.)

To fix this problem for my use case (which I admit is unique) I created my own split_arguments function. Since this issue may also occur in other areas (and I would benefit from having my code run on lockstrings as well) I thought it would be a good idea to share it.

**About this PR**

I think the current implementation is not good enough yet for three reasons:

1. There is no backwards compatibility with older Evennia servers that rely on the old behavior. Just like I need this feature, others may rely on the fact that they don't have it. To solve this, I propose adding a flag to `settings_default.py` to define which method to use. That way, both methods can be supported and the old method can eventually be deprecated once everyone has had a chance to check and update their code.

2. There's a feature I'd like to include. In my "trigger functions" I am also using the concept of exposing variables to be used in these functions. I.e., the "get" trigger has a "getter" variable that references the object that picked up the triggered object. With a little more work, this can be supported fully, recognizing quoted strings as proper strings and unquoted text as variable names, converting them to strings when necessary.

3. I'm not sure, but this feature may need to be implemented for stuff like inlinefuncs too, along with the needed backwards compatibility flag in `settings_default.py`.

**Closing arguments**

But the main thing to ensure is that the code isn't foreign. That's why I added comments to the `split_arguments` function which should make it easier to understand what it does.

I'm open to questions about the code or anything. If this isn't accepted, then so be it, but it'd save me some time because I wouldn't have to figure out how to overrule the Evennia `lockhandler.py` file to include this in my own project.

GulliblePsychologist:
I'm going to reimplement this properly in a different branch.

Griatch:
Belatedly, but as mentioned further down in the documentation for inlinefuncs: 
```
A special case happens if wanting to use an inlinefunc argument that itself includes a comma - this would be parsed as an argument separator. To escape commas you can either escape each comma manually with a backslash \,, or you can embed the entire string in python triple-quotes """ or ''' - this will escape the entire argument, including commas and any nested inlinefunc calls within.
```

GulliblePsychologist:
I already looked at the in-game Python contrib. I didn't want to give builders _that_ much access but I did want to give them a simple way to call custom functions when certain events happen.

Inlinefuncs do not solve this problem and I don't see how they would. I might not understand how they work, but I think that's irrelevant as this PR was never about inlinefuncs in the first place.

I understand that lockfuncs were written to be simplistic by design. I don't understand why improvements are not allowed. I feel like I'm being scorned for my contribution.

That's why I haven't offered a new PR yet, despite finishing my `parse_arguments` function. I don't think my contributions are welcome here, and I do not wish to fight other developers over this.

The version I have now (which I will offer if desired) works like this:
```
parse_arguments("caller, \"This is a string.\", other_variable, answer=42", caller="Timmy")
# (['Timmy', 'This is a string.', <server.utils.FunctionArgument object at 0x0000024A145DA788>], {'answer': 42})
```

I am using this to translate the parameters of a "command function" (think about Evennia's "lock functions", but they are called when certain events happen to do stuff rather than to check stuff)

Previously, I encountered the following issues:

1. I couldn't paste URLs because the equals sign messed up the simple parser, which caught it as a named parameter instead.

2. There was no distinction between a variable name or a regular string. Either I would replace every occurrence or I replace none at all. My approach gives you `FunctionArgument` objects which may be converted as desired.

whitenoiseoss:
Firstly, discussion around a PR is perfectly normal. It doesn't mean a contributor is being scorned or that contributions aren't welcome. That is a bit dramatic as this is normal behavior on any project of any size and scale.

Secondly, assuming I understand everything correctly (I just spent several minutes looking over everything here) -- your main goal is an event system that builders can put events or effects on objects. You are trying to use lockfuncs for this, which then led you to need to modify the lock parser to allow for URLs and the like.

Griatch is correct that lockfuncs are not the place for this. I cannot think of a use case where it would be reasonable to have a URL as part of a lock. It's important that these behaviors stay isolated. Lockfuncs should not be doing anything but stating locks. No system, unless it is referencing locks, should be parsing a lockstring. So in that way, this would not be an improvement of the lockstring--it would be a confusion of it.

There are many ways to accomplish an event system. An event system is not unique, every game engine will have one and this is a well-solved problem. Perhaps it would be best to start with a question in Discord, IRC or our Google group around the design of an event system in Evennia. I think you will find the most traction on your issue that way.

GulliblePsychologist:
I don't understand what you mean by using the Observer pattern for this. Builders would still need to specify what functions they would like to execute if a particular trigger hits. I might be able to use the Observer pattern to compile a custom language, but I didn't want to go that far.

I'm not trying to use lockfuncs to this, but instead, I borrowed the lock functions code for this. My thought pattern was like this:

1. I'd like a way for builders to be able to run a set of predefined commands when certain events happen on objects.

2. Lock functions do essentially the same thing, except they only check if an action is allowed rather than performing actions of their own.

3. What if I copied the code for lock functions, renamed them "trigger functions" or "command functions" and then allowed builders to attach those to objects? That sounds like it'd be easy enough to implement.

---

In my project, I have `server/conf/cmdfuncs.py` which is a list of functions that are intended for the event model. Builders can attach "triggers" to objects which is similar to setting "locks" on objects.

"Locks" still work as designed. They check if an object has access to do something.

"Triggers" is a new concept, foreign to Evennia, local to my project. (But I'm happy to offer it as a contrib.) They execute a list of functions when they are triggered.

---

I have a custom system that works and is in use. But when builders started using it, I found out that there were some issues with the way that the arguments are being parsed.

Ultimately, I ran into these two problems:

1. Characters aren't escaped properly, quotes are stripped out. This was preventing URLs with equal signs and messages containing commas from being used.

2. I wanted to expose a set of variables to the builder depending on which trigger was being invoked. I could do this with the exising parser, but it still suffered from the same issue with escaping. The variable names would either have to be unique or you could never use the literal value.

You can't parse arguments with a regular expression as regular expressions aren't designed for that purpose. So I wrote my own.

Then, I wanted to offer my parser to Evennia. My intent is that it can replace the parser used for lockfuncs. This is because my "triggers" is essentially just a copy/paste of the `evennia/locks/lockfuncs.py` file renamed to `contrib/triggerfuncs.py`.

My case for adding this (meaning only the parser, not triggers themselves) to Evennia:

1. It would make my codebase consistent. I concede that this is not a priority for Evennia, but it would be a nice thing to do. On the other hand, I think there is a case for making things consistent with established coding conventions and standards. (i.e., in a function parameter, strings tend to be wrapped in quotes rather than loosely typed)

2. It's better than the current implementation. The current method strips characters from the input and has no support for substituting variables. I'd think that an improvement to something would be welcome, as long as the underlying code is understood and can be maintained well.

3. It probably won't change anything for the majority of users because unquoted strings are still converted to strings in the end, much like setting an attribute with the "set" command converts unquoted text to strings; and

4. it can be easily disabled with a configuration flag in cases where it does conflict.

5. I have no problem taking responsibility for it and helping out if anything needs to be done with it in the future.

---

By the way, you have a bug at [line 614](https://github.com/evennia/evennia/blob/3cc5701f661fcec90ddbec0612e5fb35dd03edc9/evennia/locks/lockfuncs.py#L614) of the default `lockfuncs.py`.

Griatch:
@GulliblePsychologist I'm finally back at the computer after Christmas and can address this better. :) To echo what @whitenoiseoss said, there is no intention of "scorning" anyone's work here. Discussion around a PR is normal and expected, feedback is never intended personally. If my comment above came across as belittling, that was not my intention and I apologize. 

Ok, with your further explanation I understand your intent a little bit better. You want to use the lockhandler parser as a base for your own system rather than use the lock functionality _per se_ for your system. This is fair enough. The reason I pointed you to the inlinefunc parser is because this is intended for more general use than the lockfunc parser and would be more useful to extend and base your work off IMO.  Looking at it now, there may be a case for combining the lockfunc- and inlinefunc- parsers into one parser for people to use as a resource for things like this, supporting the features you suggest. The main thing to be super-wary of is to make sure the code cannot be executed as a python string (this is why the lockfunc parser is so rigid originally, it's for paranoia-reasons ;) )

As for your implementation, this is something that I need to look over some more. It would be (potentially) applied very often as text comes in, so some optimization might be useful to consider - not sure if a per-character Python loop is the most efficient solution here. A regex-based iterator might be more "Pythonic" overall, as well as put more of the action in C under the hood. But not sure offhand.

Issue #1997, opened 2019-12-02
[BUG] TagHandler continues caching tags even when explicitly disabled


strikaco:
#### Describe the bug
When `TYPECLASS_AGGRESSIVE_CACHE` is set to `False`, the TagHandler will ignore the cache on reads but continue caching writes.

#### To Reproduce
See #1996; replace references to `attributes` with `tags`.

#### Expected behavior
The TagHandler should not be touching the cache when disabled.

#### Environment, Evennia version, OS etc
 Evennia 0.9.0 (rev ff7b739) (rev ff7b739)
    OS: posix
    Python: 3.7.5
    Twisted: 19.7.0
    Django: 2.2.6

PR #1998, opened 2019-12-03
Stops Attribute/Tag handlers from caching when TYPECLASS_AGGRESSIVE_CACHE is disabled

### LINKED ISSUE #1996: [BUG] AttributeHandler continues caching attributes even when explicitly disabled
#### Describe the bug
In trying to get to the bottom of some memory leaks, I noticed that when `TYPECLASS_AGGRESSIVE_CACHE` is set to `False`, the AttributeHandler will ignore the cache for reads but will still continue to write to it.

#### To Reproduce
````
from evennia.utils.test_resources import EvenniaTest
from django.conf import settings
from django.test import override_settings
from mock import patch


class AttributeCachingTest(EvenniaTest):

    @override_settings(TYPECLASS_AGGRESSIVE_CACHE=False)
    @patch('evennia.typeclasses.attributes._TYPECLASS_AGGRESSIVE_CACHE', False)
    def test_caching(self):
        "Test to make sure when caching is disabled, it stays disabled."
        # Make sure the setting override actually stuck
        self.assertFalse(settings.TYPECLASS_AGGRESSIVE_CACHE)
        from evennia.typeclasses.attributes import _TYPECLASS_AGGRESSIVE_CACHE
        self.assertFalse(_TYPECLASS_AGGRESSIVE_CACHE)
        
        # Add an example attribute
        self.obj1.attributes.add('phrase', 'xyzzy')
        
        # Cache should not have been touched.
        self.assertFalse(self.obj1.attributes._cache, f"Cache contents (should be empty): {self.obj1.attributes._cache}")
````

#### Expected behavior
Not sure if this was intentional, but it is self-defeating. It is definitely a major factor in the memory leak since some obscenely long user-defined lists managed to get stored in Attributes and are persisted in memory on every write despite my frustrated efforts to purge them.

````
Evennia 0.9.0 (rev ff7b739) (rev ff7b739)
    OS: posix
    Python: 3.7.5
    Twisted: 19.7.0
    Django: 2.2.6
````

#### Additional context
The cache invalidation mechanism seems to be based on a time interval that does not keep up with intensive I/O and leaves the server open to resource exhaustion in the interim. Curiously there is a `LimitedSizeOrderedDict` in the utils module purpose-built for caching but the AttributeHandler is using a plain `dict` for its cache instead, which lends itself to boundless growth.

Django's own caching framework invalidates keys based on both time and number to get around these sorts of issues (can even use the filesystem as the backend for caching, not just in-memory only) but it is not used by Evennia. Either option might be worth switching to.

strikaco:
When TYPECLASS_AGGRESSIVE_CACHE is disabled, the Attribute and Tag handlers will both continue to write/update the cache, but not read from it. This might be manageable under normal conditions but was causing a memory leak when I attempted to export the contents of my database.

Closes #1996 and #1997

### DIFF
--- a/evennia/typeclasses/attributes.py
+++ b/evennia/typeclasses/attributes.py
@@ -237,6 +237,7 @@ class AttributeHandler(object):
 
     def _fullcache(self):
         """Cache all attributes of this object"""
+        if not _TYPECLASS_AGGRESSIVE_CACHE: return
         query = {
             "%s__id" % self._model: self._objid,
             "attribute__db_model__iexact": self._model,
@@ -298,7 +299,7 @@ class AttributeHandler(object):
                 attr = None
                 cachefound = False
                 del self._cache[cachekey]
-            if cachefound:
+            if cachefound and _TYPECLASS_AGGRESSIVE_CACHE:
                 if attr:
                     return [attr]  # return cached entity
                 else:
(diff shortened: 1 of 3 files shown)


strikaco:
One issue that may manifest from this is the fact that the AttributeHandler `all()` method seems to pull from the cache instead of the db. Doesn't fail the test suite but also logically doesn't make any sense. Meant to double check before merge but got distracted.

Issue #1972, opened 2019-10-20
[BUG] object.copy() is not properly copying tags


awranthyr:
#### Describe the bug
Copying an object does not copy its tags.

#### To Reproduce
1. Create an object.
2. Add tags to said object.
3. Copy the object through any method.

#### Expected behavior
There seems to be a clear intent to make an exact copy of the object, including all attached tags.

#### Environment, Evennia version, OS etc
Evennia 0.9.0, rev 09f10db3

#### Additional context
The root of the problem appears to be in evennia/objects/manager.py, at line 593 at the time of this writing.

The line is currently:

`         for tag in original_object.tags.get(return_tagobj=True, return_list=True):`

Which returns only an empty list.

The following line should fix the issue:

`         for tag in original_object.tags.all(return_objs=True):`

Griatch:
@awranthyr I cannot reproduce this. I tested the following from in-game: 

1. `create/drop box`
2. `tag box = foo`
3. `py self.search('box').copy()`     (this creates `box000`)
4. `ex box000`
The copy now has the tags. 

Could you give a more specific process to reproduce the error?

awranthyr:
@Griatch Sorry about that, I should have done more testing before reporting the issue.

The bug doesn't apply to all tags, apparently, but only those with categories.  I just tested the following in a fresh install of Evennia, to make sure it wasn't just some of my own game's code causing it:

1. `create/drop thing1`
2. `tag thing1 = first_tag`
3. `tag thing1 = second_tag:new_category`
4. `py self.search('thing1').copy()`
5. `ex thing1000`

The copy does have the non-categorized tag, but not the one with a category.

strikaco:
@awranthyr A fix for this was implemented; are you still having this issue?

Griatch:
Since there are no further replies here and a fix was already made, I'm closing this issue.

PR #2001, opened 2019-12-04
Fixes failure of copy() method to replicate categorized attributes and tags

strikaco:
Small tweaks to the objectmanager `copy()` method to replicate categorized attributes and tags.

Problem was reportedly with tags only but the same behavior occurred with attributes as well.

#1972

### DIFF
--- a/evennia/objects/manager.py
+++ b/evennia/objects/manager.py
@@ -575,8 +575,8 @@ class ObjectDBManager(TypedObjectManager):
             return None
 
         # copy over all attributes from old to new.
-        for attr in original_object.attributes.all():
-            new_object.attributes.add(attr.key, attr.value)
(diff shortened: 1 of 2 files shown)


evennia/objects/manager.py
```diff
@@ -575,8 +575,8 @@ def copy_object(
             return None
 
         # copy over all attributes from old to new.
-        for attr in original_object.attributes.all():
-            new_object.attributes.add(attr.key, attr.value)
+        attrs = ((a.key, a.value, a.category, a.lock_storage) for a in original_object.attributes.all())
+        new_object.attributes.batch_add(*attrs)
```
Griatch: This is a good optimization :+1: I don't think the `attributes.batch_add` existed when the Attribute manager was originally created :)

PR #2008, opened 2019-12-29
Extend the Webclient Options handling

friarzen:

This is a pretty big feature update for the webclient.
    It provides a new way of dealing with client side options and a simple "help/about" plugin. 
    It provides a new plugin for selectable font and font sizing
    It includes a new plugin for basic client-side message matching/routing/"spawns" options
    It includes a new window that will open the game's main web page for "in-game" viewing
  and
    It extends the plugin architecture so that other plugins can define their own Options UI bits.

These are all features I've found a need for at some point, so others might like them as well.

I am still working on where to have these plugins store/retrieve info as server-side options


Griatch:
Great! All the points you specify sounds like dearly needed additions. Let me know if there is something in particular you want me to test out at this WIP stage.

friarzen:
primarily it needs people to try it out and tell me if it is sufficiently self-explanatory or not.
I know you mentioned the expectation of the new tab opening and then closing again by clicking on the gear icon (unfortunately, that is not nearly as easy to implement as it would seem, but I am already working on that).

Griatch:
@friarzen I ran this branch against the latest master now. Some really cool features here! I have some further notes on this promising WIP: 

# General 

- When connecting I get a long traceback in the log ` "Request.finish called on a request after its connection was lost; "
        builtins.RuntimeError: Request.finish called on a request after its connection was lost; use Request.notifyFinish to keep track of this.` I'm not sure if this due to these changes or why, but I don't see it in pure `master`. 
- It seems strange that you can open multiple Options tabs when the options are global. Having multiple option tabs also highlights that they don't sync with each other - so if you add, say a Regex/Tag pair in one tab it won't appear in the other option tab until it's closed and reopened. This is pretty confusing and it's easy to add duplicate pairings by accident. The left-over Options tabs also makes the whole thing look messy - I have to go through and delete all the extra tabs.
- I keep wanting to close the options by pressing the cog wheel. Instead a new Options tab is opened (see above). If this is possible I think it would be more in line with how global options usually work. 

# Regex/Tag pairs 

The regex/tag matcher is very cool overall. It's awesome that new tags are actually created when adding them to the list. I tested to add some simple regexes and it seems to work fine!  Here are some things I noticed 

- Duplicate Regex/Tag pairs doesn't really make sense, ideally the old pair should be removed when adding the exact same again (changing the order).
- A regex pair does not seem possible to remove. Setting it  to empty strings and closing/opening the options pane will still show the regex (tag remains empty). 
- If having an empty tag (see point above), an actual empty line appears in the tag selector for all panes.
- This is probably best done as a separate development, but there needs to be a way for the server to tell the client in advance all the tags it may ever send (in addition to discovering on the fly like now) ... most likely this needs to be a list in settings. 

# IFrame

Very cool! 

- The name `Browser IFrame` is not too informative for the average player. Maybe calling it "Embed website" or "Open URL in new  tab"? If it was possible to have a tool-tip, the IFrame nature of this (and its possible limitations such as if trying to embed insecure websites etc) could be described there. 
- I think it would be an idea to offer not only a pre-set URL here, but to give the user the ability to add more buttons with their own URLs, similar to how Regex/Tag pairs work - you then click a button next to it to open that IFrame. Some such urls would be default, like the homepage. But the way I understand how people use a feature like this is in order to quickly keep up with various things on the website, like a wiki, the helpfiles or whatever. Allowing for more that one "view" like this sounds like a good idea. 

# Font Size

I have no critique on this, it seems to work well and with a nice selection of options!

# Looks and docs 

I like that the options are a tab now instead of hovering over things in a popup. Very nice with a scroll bar, making it infinitely expandable. 

- I don't think it's needed to have a huge `Welcome to Evennia` up top. Vertical space needs to be conserved. Just a smaller `Options` is enough. Same with the other Headings: They need to be much smaller, just one level above the normal-size text IMO. 
- I have mentioned this before, but is it possible to have Tooltips inside the panel? I think this would be a very important addition if it was possible - not only could it describe stuff like what the heck an IFrame is, it could also give some hints about what a regexes and tags are respectively (giving a link to a regex tutorial even).

friarzen:
> @friarzen I ran this branch against the latest master now. Some really cool features here! I have some further notes on this promising WIP:
> 
> # General
> * When connecting I get a long traceback in the log ` "Request.finish called on a request after its connection was lost; " builtins.RuntimeError: Request.finish called on a request after its connection was lost; use Request.notifyFinish to keep track of this.` I'm not sure if this due to these changes or why, but I don't see it in pure `master`.
> * It seems strange that you can open multiple Options tabs when the options are global. Having multiple option tabs also highlights that they don't sync with each other - so if you add, say a Regex/Tag pair in one tab it won't appear in the other option tab until it's closed and reopened. This is pretty confusing and it's easy to add duplicate pairings by accident. The left-over Options tabs also makes the whole thing look messy - I have to go through and delete all the extra tabs.
> * I keep wanting to close the options by pressing the cog wheel. Instead a new Options tab is opened (see above). If this is possible I think it would be more in line with how global options usually work.

Ok, options cog now opens/closes a single tab (even if that tab has been dragged somewhere else).  It isn't perfect on Chrome, since chrome's focus handling makes it require two clicks, but it works.
I haven't been able to replicate your console error, but I have been seeing a growing number of cross-origin-content warnings with newer versions of Firefox and Chrome, and some people have reported issues with some of their security/anonymity browser extensions.  All of which are due to not hosting the upstream libraries like goldenlayout or jquery directly from the game project.  We'll probably need to fix this in the future.

> # Regex/Tag pairs
> The regex/tag matcher is very cool overall. It's awesome that new tags are actually created when adding them to the list. I tested to add some simple regexes and it seems to work fine! Here are some things I noticed
>
> * Duplicate Regex/Tag pairs doesn't really make sense, ideally the old pair should be removed when adding the exact same again (changing the order).
> * A regex pair does not seem possible to remove. Setting it  to empty strings and closing/opening the options pane will still show the regex (tag remains empty).
> * If having an empty tag (see point above), an actual empty line appears in the tag selector for all panes.
> * This is probably best done as a separate development, but there needs to be a way for the server to tell the client in advance all the tags it may ever send (in addition to discovering on the fly like now) ... most likely this needs to be a list in settings.

I am still working on this and will have another version soon that ignores blank entries and allows for dynamically deleting each key/value pair.  As for telling the client what tags the server supports, I can have that supported very quickly.  I will just need to know which onOptions kwarg to read the list from, everything else is already in place.

> # IFrame
> Very cool!
> 
> * The name `Browser IFrame` is not too informative for the average player. Maybe calling it "Embed website" or "Open URL in new  tab"? If it was possible to have a tool-tip, the IFrame nature of this (and its possible limitations such as if trying to embed insecure websites etc) could be described there.
> * I think it would be an idea to offer not only a pre-set URL here, but to give the user the ability to add more buttons with their own URLs, similar to how Regex/Tag pairs work - you then click a button next to it to open that IFrame. Some such urls would be default, like the homepage. But the way I understand how people use a feature like this is in order to quickly keep up with various things on the website, like a wiki, the helpfiles or whatever. Allowing for more that one "view" like this sounds like a good idea.

Unfortunately, this is where the browser security has been locking things down heavily as time goes by.  The only way to serve 3rd party website content without fatal errors in various cases is by having the server side act as a tcp-passthrough reverse proxy.  That's a pretty big undertaking.  This works for being able to reach anything within the same domain as the game, but that is a difficult limitation to get the average person to understand.

> # Font Size
> I have no critique on this, it seems to work well and with a nice selection of options!
> 
> # Looks and docs
> I like that the options are a tab now instead of hovering over things in a popup. Very nice with a scroll bar, making it infinitely expandable.
> 
> * I don't think it's needed to have a huge `Welcome to Evennia` up top. Vertical space needs to be conserved. Just a smaller `Options` is enough. Same with the other Headings: They need to be much smaller, just one level above the normal-size text IMO.
> * I have mentioned this before, but is it possible to have Tooltips inside the panel? I think this would be a very important addition if it was possible - not only could it describe stuff like what the heck an IFrame is, it could also give some hints about what a regexes and tags are respectively (giving a link to a regex tutorial even).

Adjusted.  Tooltips should just be adding some CSS and some extra text, I'll look into it after I get the regex UI cleaned up and the client-side logging plugin done.

--Brenden

Griatch:
@friarzen 

The option-cog open/close works great and this makes it feel a lot nicer to use I think! :+1: Also great with the ability to delete the tag/regex pairs. Simple but efficient solution. 
I also like the new font sizes, it's now very unintrusive which is very good. 

## Comments/Replies

- As for the IFrame, I was afraid that'd be an issue yes. Let's not go deeper into this now then. It's probably better to just have people open up a new tab in their browsers anyway. The new title of the button is definitely clearer at any rate!   
- The server is today not making the tag list available, so that would need to be added on the server side (basically reading a setting). The onOption kwarg name would be something like `available_server_tags` maybe? 
- I could not break the regex/tag pairs with an empty field anymore, so that's great! I found a few times that it would not properly save my input if I filled it in and closed the options. It seems I have to specifically un-focus the input field before the regex/pair is saved?
- I don't seem able to insert a second regex/tag pair beyond the first. I add it but no matter how I un-select or go about it, after closing and re-opening the option pane, the second pair is gone. Update: just when I retried this now though, it worked. So not it seems a little fickle. 
- I tried a regex pairing like this:  `You paged` -> `tell` and  `.*? pages:`  -> `tell`. I then created a new pane and assigned the `tell` tag to it. Next I tried `tell Griatch = foo`. I expect both the send `You paged Griatch with: 'foo'` and receive `Account Griatch pages: foo` to appear in the same pane, but instead only the send does (the other appear in the untagged main pane). It does not appear like I should have to match the entire line, so I'm unclear as to why the second regex would not work. 

## Further points 

- Tooltips - as you mention, CSS is the way to go here. 
- Something I now thought about - every new pane is named `Untitled`. You can rename this manually but it would be interesting to have a more useful default. Maybe one could base it on the first tag type unless a name is specifically given? For example, it could be named `untagged` or `all` or `tell` for the example above. If having more than one tag, it could be named `mixed`. And if you set a specific name that overrides it ... I guess the `Main` would be the exception since it holds the options button and should always remain. Not sure how much work this would be or if it should be a separate issue.

friarzen:
> * The server is today not making the tag list available, so that would need to be added on the server side (basically reading a setting). The onOption kwarg name would be something like `available_server_tags` maybe?

Sure.  I just added a simple onOption parser for:  available_server_tags = [ "tag1", "tag2", ... ]

> * I could not break the regex/tag pairs with an empty field anymore, so that's great! I found a few times that it would not properly save my input if I filled it in and closed the options. It seems I have to specifically un-focus the input field before the regex/pair is saved?

I've added a small tweak to onFocusOut to also save, so this should help for this case.

> * I don't seem able to insert a second regex/tag pair beyond the first. I add it but no matter how I un-select or go about it, after closing and re-opening the option pane, the second pair is gone. Update: just when I retried this now though, it worked. So not it seems a little fickle.

I can't seem to reproduce this at all.  I'm thinking this might be a browser cache issue.
Please clear your caches and try again.

> * I tried a regex pairing like this:  `You paged` -> `tell` and  `.*? pages:`  -> `tell`. I then created a new pane and assigned the `tell` tag to it. Next I tried `tell Griatch = foo`. I expect both the send `You paged Griatch with: 'foo'` and receive `Account Griatch pages: foo` to appear in the same pane, but instead only the send does (the other appear in the untagged main pane). It does not appear like I should have to match the entire line, so I'm unclear as to why the second regex would not work.

Ahh....heh....server side rendering strikes again!  This is an issue because the message text the regex is matching against has all of the <span ...> bits in it.  Unfortunately, because browsers tend to ignore/add whitespace as they please, this function is always going to be a bit fiddly. Unless we switch the HTML conversion to the client-side.  The jquery.text() tweak I just added should work for most people, but I'm guessing rare edge-cases will still crop up.

> ## Further points
> * Tooltips - as you mention, CSS is the way to go here.
> * Something I now thought about - every new pane is named `Untitled`. You can rename this manually but it would be interesting to have a more useful default. Maybe one could base it on the first tag type unless a name is specifically given? For example, it could be named `untagged` or `all` or `tell` for the example above. If having more than one tag, it could be named `mixed`. And if you set a specific name that overrides it ... I guess the `Main` would be the exception since it holds the options button and should always remain. Not sure how much work this would be or if it should be a separate issue.

I'd say these two enhancements should probably be a separate PR.

Griatch:
> Ahh....heh....server side rendering strikes again! 

This is of course way beyond the scope of this PR, but _should_ we consider moving the http-rendering into the client instead, for the future? That way you do have a lot more flexibility to do whatever you like with it as well. I suppose one could send it with all the mud-markup (`|r` etc) intact in that case. Just a food for thought. 

I think that I otherwise have seen improvements to the usage points I commented on. The Tooltips and default tag-names will go into separate tickets. I see the WIP status is lifted, so this is now ready for code review?

friarzen:
> > Ahh....heh....server side rendering strikes again!
> 
> This is of course way beyond the scope of this PR, but _should_ we consider moving the http-rendering into the client instead, for the future? That way you do have a lot more flexibility to do whatever you like with it as well. I suppose one could send it with all the mud-markup (`|r` etc) intact in that case. Just a food for thought.

I would say, yes, this is probably the correct approach long term.  Since that enables more options for the client to act as whatever presentation layer it feels like being.  And it would allow us to put in a basic defense for the one known MXP vulnerability, since that would no longer come across as trusted HTML.

> I think that I otherwise have seen improvements to the usage points I commented on. The Tooltips and default tag-names will go into separate tickets. I see the WIP status is lifted, so this is now ready for code review?

Yep, feel free to review.  Thanks.

--Brenden

PR #2011, opened 2020-01-08
Adds a new argument parser for processing function arguments in text

GulliblePsychologist:
This PR adds a new parser for function arguments that behaves more naturally, previously discussed in #1995.

Essentially, it's a function that takes a string such as `"attack", caller, msg="The gnome attacks you for 5 damage!", damage=5)` and returns a tuple containing 2 elements:

* A `list` containing the string "attack" and a `FunctionArgument` object with the name "caller"
* A `dict` containing:
* * Key: "msg", Value: "The gnome attacks you for 5 damage!" (string)
* * Key: "damage", Value: 5 (number)

It is up to the developer to post-process the result, replacing `FunctionArgument` objects.

I didn't want to allow Builders to have access to the underlying Python code but I did want to give them some freedom to call their own little routines.

My solution was to implement "command functions" which are like more flexible lock functions. They are called by "triggers" which like locks can be set on objects.

One of my triggers is called "playmusic" and it accepts 2 parameters:

1. The player that receives the song.
2. The URL to the song to play. We have our own repository of songs on a subdomain, running separately from Evennia.

Currently, no parser is used but instead, the function parameters within a lockstring are parsed by splitting the string. There are two problems I have run into:

1. I have no way to provide a reference to an object, such as the player that receives the song. I've addressed this by returning anything that isn't obviously a string or a number as a `FunctionArgument` object instead, which can be converted to an object prior to calling the function.
2. If a URL contains an equal sign, the value is regarded as a named parameter and split up. The current implementation provided me with no way to quote the string or escape the equal sign.

I have successfully implemented this into my "command functions" and given that it's no more than a copy/paste of the lockfuncs.py file I think this parser might be useful for Evennia.

The PR only contains the function itself (modified and cleaned up this time) and some test cases. The implementation of the parser in the existing lockfuncs is not done here, to minimize the impact of this PR.

### DIFF
--- a/evennia/utils/utils.py
+++ b/evennia/utils/utils.py
@@ -35,6 +35,7 @@ _MULTIMATCH_TEMPLATE = settings.SEARCH_MULTIMATCH_TEMPLATE
 _EVENNIA_DIR = settings.EVENNIA_DIR
 _GAME_DIR = settings.GAME_DIR
 ENCODINGS = settings.ENCODINGS
+_ARG_ESCAPE_SIGN = "\\"
 _GA = object.__getattribute__
 _SA = object.__setattr__
 _DA = object.__delattr__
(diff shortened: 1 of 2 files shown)


GulliblePsychologist:
Griatch commented in #1995:
> Ok, with your further explanation I understand your intent a little bit better. You want to use the lockhandler parser as a base for your own system rather than use the lock functionality per se for your system. This is fair enough. The reason I pointed you to the inlinefunc parser is because this is intended for more general use than the lockfunc parser and would be more useful to extend and base your work off IMO. Looking at it now, there may be a case for combining the lockfunc- and inlinefunc- parsers into one parser for people to use as a resource for things like this, supporting the features you suggest. The main thing to be super-wary of is to make sure the code cannot be executed as a python string (this is why the lockfunc parser is so rigid originally, it's for paranoia-reasons ;) )
> 
> As for your implementation, this is something that I need to look over some more. It would be (potentially) applied very often as text comes in, so some optimization might be useful to consider - not sure if a per-character Python loop is the most efficient solution here. A regex-based iterator might be more "Pythonic" overall, as well as put more of the action in C under the hood. But not sure offhand. And this PR is closed anyway, so not sure if I should really be looking at the code or not...

Griatch:
I've not forgotten this, but I need to look into the performance of this solution. Also, this is just a plain parser; for it to make sense to merge as a general implementation (to gain the advantages of code-reuse), the existing systems should first be updated to use this common parser (which means modifying lockfuncs, inlinefuncs, and by extension prototypefuncs parsers). So from that perspective this is a highly invasive change that would affect a large set of critical systems and would need to be done pretty carefully.

evennia/utils/utils.py
```diff
@@ -2160,3 +2161,120 @@ def decorator(*args, **kwargs):
...
+        return ret
+
+    def _add_value(skip, key, args_list, args_dict, temp):
+        if not skip:
+            # Record value based on whether named parameters mode is set or not.
+            if key is not None:
+                args_dict[key] = _parse_value(temp)
+                key = None
+            else:
+                args_list.append(_parse_value(temp))
```
Vikka, line 2223: Some suggestions here:

1. Return early, you will reduce the cognitive complexity of your function [cognitive complexity](https://www.sonarsource.com/docs/CognitiveComplexity.pdf)
2. Do not use the same name for local variable and outer scope variables. Put one trailing underscore to avoid confusion. It can lead to the misbelief that attributing value to these variables change the variable from outer scope.
3. Line 2221, the local variable 'key' value is not used.
This line is superfluous.

```suggestion
    def _add_value(skip_, key_, args_list_, args_dict_, temp_):
        if skip_:
            return
        # Record value based on whether named parameters mode is set or not.
        if key_ is not None:
            args_dict_[key_] = _parse_value(temp_)
            return
        args_list_.append(_parse_value(temp_))
```

evennia/utils/utils.py
```diff
@@ -2160,3 +2161,120 @@ def decorator(*args, **kwargs):
...
+            # Any other character: add to buffer.
+            temp += c
+
+    if inside[0] is True:
+        # Invalid syntax because we are inside a quoted area.
+        return None
+    else:
+        _add_value(skip, key, args_list, args_dict, temp)
+
+    return args_list, args_dict
```
Vikka, line 2280: Some suggestions here:

1. > Don't compare boolean values to True or False using ==
[PEP 8 -- Programming Recommendations](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)
2. >  When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string. It improves readability.
[PEP-8 String Quotes](https://www.python.org/dev/peps/pep-0008/#string-quotes)
3. Return early, you will reduce the cognitive complexity of your function [cognitive complexity](https://www.sonarsource.com/docs/CognitiveComplexity.pdf)
```suggestion
    for c in s:
        if c == _ARG_ESCAPE_SIGN:
            # Escape sign used.
            if not escape:
                # Enter escape mode.
                escape = True
                continue
            # Already escaping: print escape sign itself.
            temp += _ARG_ESCAPE_SIGN
            escape = False
            continue

        if escape:
            # Escape mode: print whatever comes after the symbol.
            escape = False
            temp += c
            continue

        if inside[0]:
            # Inside single quotes or double quotes
            # Wait for the end symbol, allow everything else through, allow escape sign for typing quotes in strings
            if c != inside[1]:
                temp += c
                continue
            # Leaving single/double quoted area
            inside = (False, None)
            continue

        if c == '"' or c == "'":
            # Entering single/double quoted area
            inside = (True, c)
            is_string = True
            continue

        if c == "=":
            if is_string:
                # Invalid syntax because we don't allow named parameters to be quoted.
                return None
            if key is None:
                # Named parameters mode and equals sign encountered. Record key and continue with value.
                key = temp.strip()
                temp = ""
            continue

        if c == ",":
            # Comma encountered outside of quoted area.
            _add_value(skip, key, args_list, args_dict, temp)

            # Reset
            temp = ""
            skip = False
            is_string = False
            key = None
            continue
        # Any other character: add to buffer.
        temp += c

    if inside[0]:
        # Invalid syntax because we are inside a quoted area.
        return None
    _add_value(skip, key, args_list, args_dict, temp)
    return args_list, args_dict
```

evennia/utils/utils.py
```diff
@@ -2160,3 +2161,120 @@ def decorator(*args, **kwargs):
...
+            try:
+                ret = int(ret)
+            except ValueError:
+                try:
+                    ret = float(ret)
+                except ValueError:
+                    if ret != "":
+                        return FunctionArgument(ret)
+
+        return ret
```
Vikka, line 2214: 1. > Flat is better than nested.
[PEP 20 -- Zen of Python](https://www.python.org/dev/peps/pep-0020/)
3. Do not use the same name for local variable and outer scope variables. Put one trailing underscore to avoid confusion. It can lead to the misbelief that attributing value to these variables change the variable from outer scope.

```suggestion
    def _parse_value(temp_):
        ret = temp_.strip()
        if is_string:
            return ret
        try:
            return int(ret)
        except ValueError:
            pass
        try:
            return float(ret)
        except ValueError:
            pass
        return ret if ret == "" else None
```

Vikka (changes requested):
I suggest some changes, do not hesitate to discuss it, this is my first review 😄

Griatch:
I merged this branch to respect the original idea, but I reworked the concept into a whole new central `FuncParser` that is considerably more powerful than what Evennia had previously. Closing this now, but thanks for the inspiration!

PR #2013, opened 2020-01-10
CmdFind syntax checking and query improvements

### LINKED ISSUE #2012: [BUG] CmdFind - Runaway memory usage with improper syntax
#### Describe the bug
Was dealing with a "bug" where the `find` command was returning seemingly random results and/or dumping the entire contents of the database. Like #1994 this caused the entire queryset to be loaded into memory before rendering to the user.

It turns out the syntax I was using was incorrect but it was accepted anyway, which led to these side effects.

Ex: `find/startswith = searchterm`

Given that I did not provide a lhs and the rhs should only be a dbref/range, validating both before executing any queries would be beneficial.

Dev

strikaco:
Makes some changes to the `CmdFind` func method.

* Adds syntax checking for either a search term OR a valid optional dbref range (closes #2012 ) 
* Determines the default dbref range through database aggregation rather than assuming 1 to be the minimum and evaluating a queryset to get the max id (performance/accuracy)
* Forces unfiltered queryset to use `.iterator()` for initial iteration and issues a subsequent query for objects of specified classes by dbref (performance)
* Cleans up heavily-redundant "plurality" code when rendering search results.
* Replaces some `.for

Closes #2012


evennia/commands/default/building.py
```diff
@@ -2778,8 +2778,35 @@ def func(self):
...
+            # bound and do a less-efficient query to find the upper.
+            low, high = 1, ObjectDB.objects.all().order_by("-id")[0].id
+            
         if self.rhs:
+            # Check that rhs is either a valid dbref or dbref range
+            try:
+                # Get rid of # signs, split on hyphen or space and cast all to int.
+                # Then sort by number to get the lowest and highest values
+                # comprising the bounds.
+                bounds = sorted(int(x) for x in re.split('[-\s]+', self.rhs.strip().replace('#', '')))
```
Griatch: I see that this command uses raw stripping of `#` in other places as well, but these days it may be a good idea to centralize parsing like this by using the special helper `evennia.utils.utils.dbref()` which takes a string and checks if it's a valid `#dbref` structure before spitting out the raw `id`. It returns `None` if the dbref structure is invalid (such as `#foo`).

evennia/commands/default/building.py
```diff
@@ -2778,8 +2778,35 @@ def func(self):
...
+            except ValueError:
+                caller.msg("Invalid dbref range provided (not a number).")
+                return
+            
+            low = bounds[0]
+            if len(bounds) > 1:
+                high = bounds[-1]
+                
+            """
```
Griatch: If this is meant to comment out old code, better use `#` for clarity in the diff. Or better yet - remove the code that is no longer used.

evennia/commands/default/building.py
```diff
@@ -2868,51 +2896,45 @@ def func(self):
...
-                    ]
+                    obj_ids.extend([
+                        obj.id for obj in results if inherits_from(obj, ROOM_TYPECLASS)
+                    ])
                 if "exit" in switches:
-                    results = [
-                        obj for obj in results if inherits_from(obj, EXIT_TYPECLASS)
-                    ]
+                    obj_ids.extend([
+                        obj.id for obj in results if inherits_from(obj, EXIT_TYPECLASS)
```
Griatch: There doesn't seem to be any unit tests for this, was this tested in practice? Because if both e.g. `room` and `exit` are in switches you'll be re-using the iterator (`results`) over and over, which is not safe. The generator is exhausted once you run it once. For example: 
```python
gen = evennia.ObjectDB.objects.all().iterator()

# using the generator for the first time
for o in gen: print(o)
Griatch
Limbo
box
...

# using the generator for the second time
for o in gen: print(o)
# no return since generator was exhausted already
```

The only way to handle this (as far as I know) is to re-create the generator every time you want to re-use it. Or turn it into a list, which is of course just what you wanted to avoid here.

evennia/commands/default/building.py
strikaco: Agreed, I meant to delete it before submitting.

evennia/commands/default/building.py
strikaco: Heh, I had some doubts about what I did here as well-- for different reasons.

You are correct about the generator exhaustion (I was under the impression `.iterator()` was renewable), but even if it worked as I had anticipated it would still result in iterating the *entire* queryset up to 4 times to manually filter the results. Ouch.

evennia/commands/default/building.py
strikaco: Easy enough to fix.

strikaco:
@Griatch Ready for re-review.

PR #2014, opened 2020-01-11
Allow X-Forwarded-For to be accepted for the WebSocket connections.

GulliblePsychologist:
This adds a new configuration parameter to the `settings_default.py` file:

```
# Number of trusted web servers (reverse proxies) in front of this server which
# set the X-Forwarded-For header.
WEBSOCKET_TRUST_X_FORWARDED_FOR = None
```

The `WebSocketServerFactory` in `portal.py` is configured to use this value. The library will then accept the `X-Forwarded-For` header and use it internally.

For Evennia, I've had to modify `webclient.py` with the code from `WebSocketServerFactory` to read the value from the `X-Forwarded-For` header.

I've tested this with and without the `X-Forwarded-For` header and it appears to work.

Currently, `X-Forwarded-For` isn't respected for WebSocket connections. All IP addresses in my server show up as "127.0.0.1" which prevents me from moderating players.

Commit d5129b5065a8d2ec12eacb9c0aa030f005b15a89 seems to deal with supporting multiple values in the `X-Forwarded-For` header. This PR does it "the twisted way" and not the way as implemented in that commit. I'll leave it to the maintainers to decide what to do with that.

### DIFF
--- a/evennia/server/portal/webclient.py
+++ b/evennia/server/portal/webclient.py
@@ -29,6 +29,7 @@ _RE_SCREENREADER_REGEX = re.compile(
     r"%s" % settings.SCREENREADER_REGEX_STRIP, re.DOTALL + re.MULTILINE
 )
 _CLIENT_SESSIONS = mod_import(settings.SESSION_ENGINE).SessionStore
+_UPSTREAM_IPS = settings.UPSTREAM_IPS
 
 
 CLOSE_NORMAL = WebSocketServerProtocol.CLOSE_STATUS_CODE_NORMAL
@@ -73,13 +74,18 @@ class WebSocketClient(WebSocketServerProtocol, Session):
         This is called when the WebSocket connection is fully established.
 
         """
-        if 'x-forwarded-for' in self.http_headers and self.trustXForwardedFor:
+        client_address = self.transport.client
+        client_address = client_address[0] if client_address else None
+
+        if client_address in _UPSTREAM_IPS and 'x-forwarded-for' in self.http_headers:
             addresses = [x.strip() for x in self.http_headers['x-forwarded-for'].split(',')]
-            trusted_addresses = addresses[-self.trustXForwardedFor:]
-            client_address = trusted_addresses[0]
-        else:
-             client_address = self.transport.client
-             client_address = client_address[0] if client_address else None
+            addresses.reverse()
+
+            for addr in addresses:
+                if addr not in _UPSTREAM_IPS:
+                    client_address = addr
+                    break
+
         self.init_session("websocket", client_address, self.factory.sessionhandler)
 
         csession = self.get_client_session()  # this sets self.csessid
--- a/evennia/server/portal/webclient.py
+++ b/evennia/server/portal/webclient.py
@@ -73,8 +73,13 @@ class WebSocketClient(WebSocketServerProtocol, Session):
         This is called when the WebSocket connection is fully established.
 
         """
-        client_address = self.transport.client
-        client_address = client_address[0] if client_address else None
+        if 'x-forwarded-for' in self.http_headers and self.trustXForwardedFor:
+            addresses = [x.strip() for x in self.http_headers['x-forwarded-for'].split(',')]
+            trusted_addresses = addresses[-self.trustXForwardedFor:]
+            client_address = trusted_addresses[0]
+        else:
+             client_address = self.transport.client
+             client_address = client_address[0] if client_address else None
         self.init_session("websocket", client_address, self.factory.sessionhandler)
 
         csession = self.get_client_session()  # this sets self.csessid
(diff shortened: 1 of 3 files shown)


friarzen:
AWESOME!  This has been on my To-Do list to track down for awhile now, but I never got to it.
Thanks!

evennia/server/portal/webclient.py
```diff
@@ -73,8 +73,13 @@ def onOpen(self):
         This is called when the WebSocket connection is fully established.
 
         """
-        client_address = self.transport.client
-        client_address = client_address[0] if client_address else None
+        if 'x-forwarded-for' in self.http_headers and self.trustXForwardedFor:
+            addresses = [x.strip() for x in self.http_headers['x-forwarded-for'].split(',')]
+            trusted_addresses = addresses[-self.trustXForwardedFor:]
```
friarzen: I think this might be off-by-one.  If the header value passed is 'x-forwarded-for: 192.168.1.1, 127.0.0.1' and you have WEBSOCKET_TRUST_X_FORWARDED_FOR = 1 (meaning 127.0.0.1 is the trusted proxy), then you end up with 127.0.0.1 as the client_address...?

GulliblePsychologist: It looks like this won't occur, because 127.0.0.1 will never be in the `X-Forwarded-For` header.

According to Wikipedia, the format of `X-Forwarded-For` is:

```
X-Forwarded-For: client, proxy1, proxy2
```

The leftmost IP address is always the original client address, followed by each proxy server that the request passed through, _minus the final proxy server._ I think this solves the off-by-one.

I ran a few test cases with this in mind:

#### Situation 1: a single reverse proxy
```
# Client IP: 192.168.1.1
# Local Nginx Proxy: 127.0.0.1

addresses = ["192.168.1.1"]  # X-Forwarded-For: 192.168.1.1
trust_x_forwarded_for = 1  # We expect 1 proxy server
trusted_addresses = addresses[-trust_x_forwarded_for:]  # ["192.168.1.1"]
client_address = trusted_addresses[0]  # "192.168.1.1"
```

#### Situation 2: reverse proxy and a load balancer
```
# Client IP: 192.168.1.1
# Load Balancer: 140.82.118.4
# Local Nginx Proxy: 127.0.0.1

addresses = ["192.168.1.1", "140.82.118.4"]  # X-Forwarded-For: 192.168.1.1, 140.82.118.4
trust_x_forwarded_for = 2  # We expect 2 proxy servers
trusted_addresses = addresses[-trust_x_forwarded_for:]  # ["192.168.1.1", "140.82.118.4"]
client_address = trusted_addresses[0]  # "192.168.1.1"
```

#### Situation 3: reverse proxy and a 3rd party proxy server
```
# Client IP: 192.168.1.1
# 3rd party proxy server: 140.82.118.4
# Local Nginx Proxy: 127.0.0.1

addresses = ["192.168.1.1", "140.82.118.4"]  # X-Forwarded-For: 192.168.1.1, 140.82.118.4
trust_x_forwarded_for = 1  # We're expecting only our local nginx proxy
trusted_addresses = addresses[-trust_x_forwarded_for:]  # ["140.82.118.4"]
client_address = trusted_addresses[0]  # "140.82.118.4"
```

friarzen: I guess I didn't remember the tidbit about the last proxy being excluded.  Thanks.

evennia/settings_default.py
```diff
@@ -102,6 +102,9 @@
 # the client will itself figure out this url based on the server's hostname.
 # e.g. ws://external.example.com or wss://external.example.com:443
 WEBSOCKET_CLIENT_URL = None
+# Number of trusted web servers (reverse proxies) in front of this server which
+# set the X-Forwarded-For header.
+WEBSOCKET_TRUST_X_FORWARDED_FOR = None
```
Griatch: I think it would be good to give a comment with a short example here,  of what setting to put here for an example proxy chain. This is after all not immediately intuitive (as exemplified by the questions from friarzen).

evennia/server/portal/webclient.py
Griatch: A code comment here clarifying just how the x-forward-for indexing works here would be useful to improve maintainability of this code (similarly as was explained to friarzen)

GulliblePsychologist:
I've modified the implementation to use `UPSTREAM_IPS` instead of a new configuration option.

GulliblePsychologist:
> This looks a lot more compact. :) Even though `UPSTREAM_IPS` is a Django setting, I'd like to see this added to the `default_settings.py` file (comment it out by default maybe) - along with (and this is the important bit) a comment describing what it does. It could also point to the django docs for further info if there is such a section there.

`settings_default.py` already contains a definition for `UPSTREAM_IPS`...?

Griatch:
@GulliblePsychologist Huh, you are right. I looked for it at the time and didn't find it. Alright then!

PR #2021, opened 2020-01-12
Added rudimentary __format__ to ANSIString.

volundmush:
Adding basic string formatting ( \_\_format\_\_ ) support to ANSIString, allowing ANSIStrings to participate in str.format() and f-strings.

Really, who doesn't want better string formatting? This is a no brainer.

It does not yet support numeric formatting. This can only do left/right/center justifying and single character filling, forsaking numeric formatting such as putting commas in thousands places and floating point precision. However, a comprehensive and tested Regexp pattern matcher that implements the official format_spec is included. It's ready to be improved upon.

### DIFF
--- a/evennia/utils/ansi.py
+++ b/evennia/utils/ansi.py
@@ -673,6 +673,7 @@ class ANSIString(str, metaclass=ANSIMeta):
     and taken literally the second time around.
 
     """
+    # A compiled Regex for the format mini-language: https://docs.python.org/3/library/string.html#formatspec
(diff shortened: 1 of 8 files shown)


evennia/utils/ansi.py
```diff
@@ -733,6 +736,23 @@ def __new__(cls, *args, **kwargs):
     def __str__(self):
         return self._raw_string
 
+    def __format__(self, format_spec):
+        format_data = self.re_format.match(format_spec).groupdict()
```
Griatch: This definitely needs a doc string explaining the purpose of this new method. Including some examples.

Griatch:
This looks much better now! But please add a unit test for this as well. That actually works well as a usage example of this as well.

volundmush:
This thing has gotten all screwed up. I need to fix it before I can merge it.

PR #2024, opened 2020-01-14
Fix for WebSocket failing to log the user in when the page is refreshed

GulliblePsychologist:
This PR introduces a series of fixes for bugs related to WebSocket connections in various browsers.

There are a few issues:

---

* According to the spec, an exit code of `1000` means that the WebSocket connection was closed through JavaScript while an exit code of `1001` means that the WebSocket connection was closed because the user navigated away from the page. Some browsers (e.x., Firefox) seem to wrongly send code 1000 when the user navigates away or reloads the tab/window. Other browsers (e.x., Google Chrome, Mobile Safari) send code 1001 which is then considered an error.

The first change is to properly disconnect the WebSocketClient when exit code 1001 is received.

---

* Both Google Chrome and Mobile Safari end up making a new WebSocket connection _before_ the old connection is closed. The `make_shared_login` in `middleware.py` is called first (which sets "webclient_authenticated_uid"), followed by a `disconnect` in `webclient.py` (which erases the value of "webclient_authenticated_uid"). Result: You are no longer logged in when you refresh the page in these browsers.

I had originally fixed this by adding an `beforeunload` event in JavaScript that closes the WebSocket connection. Further testing revealed that Mobile Safari never calls `beforeunload` when the user refreshes the page.

The new fix is to set a "nonce value", an integer which is incremented each time the "webclient_authenticated_uid" attribute is set, and which is stored in each `WebSocketClient` object. (I'm not sure if "nonce" is the right word, in hindsight, as it seems used more often in cryptographic functions.)

When a `WebSocketClient` object's `disconnect` is called it will first check if its `nonce` is the same as the current session nonce. If it is, it'll clear the "webclient_authenticated_uid" value, otherwise, it'll leave it alone.

I've tested this by hammering the F5 button over and over on my machine, and the webclient reconnects successfully every time.

---

* Both `unload` and `beforeunload` is unreliable across various browsers. Mobile Safari seems to ignore them.

The most controversial change is the removal of the `unload` handler in `webclient_gui.js`. This should not be necessary as every browser should close the WebSocket with either an exit code 1000 or 1001.

These issues have been plaguing me for a while now. It'll be good to see them officially fixed in Evennia.

### DIFF
--- a/evennia/server/portal/webclient.py
+++ b/evennia/server/portal/webclient.py
@@ -30,14 +30,23 @@ _RE_SCREENREADER_REGEX = re.compile(
 )
 _CLIENT_SESSIONS = mod_import(settings.SESSION_ENGINE).SessionStore
 
-
+# Status Code 1000: Normal Closure
+#   called when the connection was closed through JavaScript
 CLOSE_NORMAL = WebSocketServerProtocol.CLOSE_STATUS_CODE_NORMAL
+
+# Status Code 1001: Going Away
+#   called when the browser is navigating away from the page
+GOING_AWAY = WebSocketServerProtocol.CLOSE_STATUS_CODE_GOING_AWAY
 
 
 class WebSocketClient(WebSocketServerProtocol, Session):
     """
     Implements the server-side of the Websocket connection.
     """
+
+    # nonce value, used to prevent the webclient from erasing the
+    # webclient_authenticated_uid value of csession on disconnect
+    nonce = None
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


Griatch:
@GulliblePsychologist Hm, on Firefox (72.01, on Linux Mint), reloading the webclient page will still require me to log back in. In Chrome it does not however. I have no Safari to test with.  Was this tested on Firefox as well?

GulliblePsychologist:
I can't reproduce this with Firefox 72.0.1 on Windows 10. I do know Firefox does something different from Chrome. I'll have to run a Linux VM and test it in there.

GulliblePsychologist:
I wasn't able to reproduce the issue on Firefox in Linux Mint. To test, I ran the Linux Mint ISO in a VM and tested both with the Firefox version that comes with the live dvd and the latest version (after `apt-get upgrade`).

There were two issues though:

* The PR quality review noticed that I accidentally used tabs instead of spaces somewhere. This has been fixed.
* In some cases an error could occur when the `webclient_authenticated_nonce` wasn't set in `disconnect`. I've been running with this fix in my production environment and it seems to solve that problem.

Griatch:
This is now merged, thanks for your work. I'm not sure if it was your fixes or the fact that I went through and emptied Firefox' Cookies/Storage manually, but it works just fine for me to reload and stay connected in Firefox as well now. 

For future refrence - for bug fixes, please work against `master`. The two branches are kept up-to-date at this time. We are not quite at the point of starting full new development in `develop` branch just yet.

Issue #2045, opened 2020-01-30
[Feature Request] CmdScripts - Add default/consistent order_by (or configurable key)


strikaco:
#### Is your feature request related to a problem? Please describe.
The `scripts` command returns results in an inconsistent order.

#### Describe the solution you'd like
Ordering by arbitrary key-- perhaps descending `id` by default, but it would be cool if one could just extend the class and change the key using Django convention without having to replicate the entire codeblock.

```
class Foo(Bar):
    ordering = ['-date_created']
```

Griatch:
This should be easy to add and make extensable.

Griatch:
Resolved this in a simple way, by simply making the script return consistent, ordered by id.

Issue #905, opened 2015-12-16
Task: Update internationalization strings


Griatch:
With the changes to the new library-like layout, some of our internationalization (i18n) strings have been lost, changed or not been updated after having moved. These should be added/fixed to fully have i18n cover the core strings (the strings not changeable otherwise by the user), 

# Translation process

- Identify a text string hard-coded somewhere in the core Evennia engine. Common examples are error messags such as "Connection lost" or "This is an error because ...". This part is mainly a matter of going through all of the core modules manually. 
- Note exceptions to what to translate:
    - Dont' translate Command output (we consider Commands something users can easily replace) 
    - Don't translate Typeclass hook strings (also something highly customized by the user). 
    - Don't translate log output or command-line output. This will only be seen by developers. 
- The guideline is that _we translate "hard-coded" strings (strings not easily replaced by the developer) that the end-user (player) may see._
- At the top of the module you have found translatable strings in, do the following import: 

    ```
    from django.utils.translation import gettext as _ 
    ```
    (so import gettext and rename it to _).
- The newly imported _ is actually a function name. Wrap the string you want to put under translation into that function: `_("string to translate")`. 
- If there are template markers, the translation must happen _before_ the templating.  So `_(this is {foo} going on").format(foo="bar")`
- If there are multiple template markers, make sure to name them (like above). Certain languages may have to present them in a different order. 
- We should try to move away from `%s` -style formatting and use `.format()` as much as possible. 
- Python `f-strings` [does unfortunately _not_ work with i18n at this time](https://code.djangoproject.com/ticket/30109). Use `.format()`. 
- Once a string has been marked up like this, then one can run the i18n mechanism to create/update template files. This allows for translating the texts all over Evennia from a single location. This, the actual act of translation, is covered in our documentation, here: https://github.com/evennia/evennia/wiki/Internationalization#translating-evennia
- Note that when you translate, you must not translate the formatting markers, so if the string to translate looks like this: `"this is {thing} going on"`, it must be translated like `"Här är det {thing} på gång"`.

Griatch:
I got a question just how to actually add i18n as described in this issue. This is the process:

 1. Identify a text string hard-coded somewhere in the core Evennia engine. Common examples are error messags such as "Connection lost" or "This is an error because ...". This part is mainly a matter of going through all of the core modules manually. *Note exceptions to what to translate, listed in the original issue*.
[...]

PR #2048, opened 2020-02-02
#905

trhr:
Minimum viable patch for #2045 

Adds utils/commonstrings.py with a collection of extracted hardcoded strings

Began implementing utils/commonstrings.py across priority (user-focused) areas. To be continued.

### DIFF
--- a/evennia/accounts/accounts.py
+++ b/evennia/accounts/accounts.py
@@ -26,7 +26,13 @@ from evennia.commands import cmdhandler
 from evennia.server.models import ServerConfig
 from evennia.server.throttle import Throttle
 from evennia.utils import class_from_module, create, logger
-from evennia.utils.utils import lazy_property, to_str, make_iter, is_iter, variable_from_module
+from evennia.utils.utils import (
+    lazy_property,
+    to_str,
+    make_iter,
+    is_iter,
+    variable_from_module,
+)
 from evennia.server.signals import (
     SIGNAL_ACCOUNT_POST_CREATE,
     SIGNAL_OBJECT_POST_PUPPET,
@@ -38,6 +44,8 @@ from evennia.commands.cmdsethandler import CmdSetHandler
 from evennia.utils.optionhandler import OptionHandler
 
 from django.utils.translation import ugettext as _
+from django.utils.translation import ngettext as _s
+from evennia.utils.commonstrings import _STRING
 from random import getrandbits
 
 __all__ = ("DefaultAccount",)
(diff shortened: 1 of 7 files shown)


trhr:
Needs unit tests, not familiar with your testing methodology yet.

Griatch:
@trhr Hm, it doesn't seem like #2045 is related to this. Are you referring to #905 maybe?

trhr:
Griatch - two branches here.

The majority of the work is in the #905 branch, but earlier commits include #2045.

trhr:
I have other changes for #2045 that I didn't submit as this PR, and it looks like I'll have to resubmit this PR because my dev environment didn't grab the commit that actually compiles. Before I do that, any thoughts to breaking out common strings to ease translation? Your #905 notes say you want to wrap individual strings with gettext, but there are a lot of places where strings are duplicated (and the translation would likewise have to be duplicated). This was an attempt to normalize some common structures, i.e. "Username/password is incorrect" is a fair way to respond to an auth problem on django login, on setpassword, on telnet auth, etc, and only needs to be translated once rather than three times, in this example.

Griatch:
@trhr I have gone back and forth on whether to break out translatable strings into a separate module like this; 
- My main concern is that it makes it a harder to read the code - especially to grep the code for that error or feature you are looking for. Today you can just do `git grep "You have been banned and cannot ocntinue"` to find just where this text is coming from. 
- Admittedly it's just one step away to first find the string in `commonstrings.py` and then find the corresponding constant, but still. 
- So overall I think moving core strings into a common text lib does make things easier. So if you are willing to put in the work to do that, it should be a useful addition :) 

So on that note, I have a few points 

- A PR is one branch only. Aim towards one of the tickets per PR, please. 
- If you can, avoid doing code "prettification" just for the sake of it (like breaking up lists into multiple lines and the like). It tends to make the commit hard to review since one has to look around to find what is actually new functionality. Besides, we run `black` on the sources regularly, so your prettification changes will be lost anyway. 
- There are so many things in `evennia/utils/` already that I'm wondering if it's really the best place to put `commonstrings.py`. I wonder if `evennia/server` may not be a better location. I'm open to arguments though. 
- To reiterate and expand on the specification in #905, No command bodies are covered by i18n, and strings in Commands should _not_ be broken out into separate string constants. Not translating the Commands means that even with a fully fleshed out translation file, common game returns will be in the original language. This is deemed acceptable.  The reason for this is threefold: 
    1. Doing so makes command code more complex and harder to read 
    2. The default commands are intended to be easily replaceable. 
    3. Default game commands may still change and some refactorings of them are due. Spending time to translate them carries a risk of being a wasted effort. 
- There is no need to translate server-ops facing text (such as logging messages, launcher help etc). 
- What _needs_ translation are user-visible strings that are embedded to be returned in the core itself and which are generally not meant to be replaced by any other means. There could be some code here that are _used_ by Commands to output text; that would be ok to put under i18n.

trhr:
Reworking this slowly, sort of mid-burnered for me; translation is a fairly high priority for my intended users, of which I have none yet until the rest of the system is built :)

PR #2049, opened 2020-02-06
Adds .decode() to byte stream to convert to string

blongden:
Decodes bytes into utf-8 string.

Certificate generation causes an error due to response not being a string (on current master).


Griatch:
Looks good to me, thanks for the PR!

PR #2050, opened 2020-02-10
Django3

trhr:
This PR implements non-breaking deprecation recommendations from Django <2.2 and Python 2 support. It does not update the requirements.txt to require Django 3, but this release will run under Django 3.

Django 3 is required for MariaDB support, which I'll be using.

Anyone considering upgrading from 2 to 3 should honestly be okay at this point -- even with heavily customized codebases, evennia support for 3 seems fairly robust now that Python2 is officially deprecated.


.travis/postgresql_settings.py
```diff
@@ -40,7 +40,7 @@
 
 DATABASES = {
     "default": {
-        "ENGINE": "django.db.backends.postgresql_psycopg2",
+        "ENGINE": "django.db.backends.postgresql",
```
Griatch: There is one additional thing we need to do - we need to put a check for this in `evennia/server/deprecations.py` and give a nice error message if you are trying to use Django3 and the old backend name. We can't expdct people to check the django changelog after all.

trhr: My understanding is that they are both aliases of each other, and that is not changing until probably 4.0.

Griatch: No, at least not according to the 3.0 changelog: https://docs.djangoproject.com/en/3.0/releases/3.0/#features-removed-in-3-0 suggests that `postgres_psycopg2` was completely removed since it's been deprecated since 2.0.

Griatch: If it remains as an alias, it's at least not clear from the changelog :)

trhr: That's fair, but also note:

This is my "non-breaking" branch, it does not update the requirements.txt to support an upgrade to Django 3. I have a separate "breaking" branch for changes that would "break" anyone who is currently running a server on 2 who chooses to upgrade the server to 3 in situ. The deprecation of psyocpg is one of many such changes that would have to be considered on an individual, server-level basis, and custom code (i.e. settings.py) will all have to be updated before merging the "breaking" branch. While it might be prudent to throw a warning that "future support for psycopg will be removed," the alias is not yet deprecated as of this commit, which is really about bringing the codebase in line with 2.2

I considered the strategy of this commit to be making it easier for people who start a server _today_ to later migrate to Django 3, which will inevitably be a multi-stage transition with potentially breaking changes. I did not intend to support/require an upgrade of existing servers to Django3.

trhr: > the alias is not yet deprecated as of this commit

I guess more technically, it is "deprecated" as of this commit, but it is not "removed" - that's fundamentally why I disagree with throwing an error for trying to use the backend name, but would agree with throwing a warning or info-level log.

Griatch: Agreed, since `requirements.txt` is not updated, this will not _require_ people to upgrade to Django 3.0+. Nor is this we want to enforce until we are actively starting to see backwards-incompatible effects. 
It's however recommended by Django itself to upgrade for security reasons, and if we support it, we will eventually increase the range of supported Django versions. Then a nicer error is useful to already have in place. You'll notice that Evennia will tell you such warnings already today. There is no need to make it a blocking error - a printout is fine (there are plenty of examples in `deprecations.py`, just add another line).

PR #2058, opened 2020-02-16
Added a prototype flag to typeclass to allow setting existing objects…

biscuitWizard:
… to a new prototype.

Adding new flag to typeclass cmd for applying prototypes over existing objects.
expand the prototype system to be more useful in non-test-driven iterative development environments. Where one off items may be made before they're made into prototypes and there's a desire to either:

A. Turn that one off item into a prototype and back-propagate it for organizational purposes.
B. Update prototype dicts that are provided in hardcode and reapply them to objects with a full stateless refresh.
Some discussion happened about this on Discord.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -1930,9 +1930,12 @@ class CmdTypeclass(COMMAND_DEFAULT_CLASS):
       list - show available typeclasses. Only typeclasses in modules actually
              imported or used from somewhere in the code will show up here
              (those typeclasses are still available if you know the path)
+      prototype - clean and overwrite the object with the specified
+               prototype key - effectively making a whole new object.
 
     Example:
       type button = examples.red_button.RedButton
+      type/prototype button=a red button
 
     If the typeclass_path is not given, the current object's typeclass is
     assumed.
@@ -1954,7 +1957,7 @@ class CmdTypeclass(COMMAND_DEFAULT_CLASS):
 
     key = "typeclass"
     aliases = ["type", "parent", "swap", "update"]
-    switch_options = ("show", "examine", "update", "reset", "force", "list")
+    switch_options = ("show", "examine", "update", "reset", "force", "list", "prototype")
     locks = "cmd:perm(typeclass) or perm(Builder)"
     help_category = "Building"
 
@@ -2038,6 +2041,28 @@ class CmdTypeclass(COMMAND_DEFAULT_CLASS):
 
         new_typeclass = self.rhs or obj.path
 
+        prototype = None
+        if "prototype" in self.switches:
+            key = self.rhs
+            prototype = protlib.search_prototype(key=key)
+            if len(prototype) > 1:
+                caller.msg(
+                    "More than one match for {}:\n{}".format(
+                        key, "\n".join(proto.get("prototype_key", "") for proto in prototype)
+                    )
+                )
+                return
+            elif prototype:
+                # one match
+                prototype = prototype[0]
+            else:
+                # no match
+                caller.msg("No prototype '{}' was found.".format(key))
+                return
+            new_typeclass = prototype["typeclass"]
+            self.switches.append("force")
+            self.switches.append("reset")
+
         if "show" in self.switches or "examine" in self.switches:
             string = "%s's current typeclass is %s." % (obj.name, obj.__class__)
             caller.msg(string)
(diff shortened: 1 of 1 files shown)


evennia/commands/default/building.py
```diff
@@ -2075,6 +2100,9 @@ def func(self):
                 new_typeclass, clean_attributes=reset, clean_cmdsets=reset, run_start_hooks=hooks
             )
 
+            if "prototype" in self.switches:
+                spawner.batch_update_objects_with_prototype(prototype, objects=[obj])
```
Griatch: The `batch_update_objects_with_prototype` returns the number of objects modified by this operation. In this case it will either be 1 or 0, if there was an error. This feels like a good thing to catch so errors can be reported back to the user. 
It may also be good to modify the success message as well, to inform the user that the prototype this-and-that was successfully applied.

biscuitWizard: Good suggestions. Added.

biscuitWizard:
> This should work fine, but I have a few requests for improvements on the concept though.
> 
> What happens when a prototype is applied to an object is that a `diff` is calculated on the object to determine the changes that needs to happen. When just passing the object into `batch_update_objects_with_prototypes` is that this diff will be calculated internally. From the perspective of the user, applying this to an object that may or may not have had a prototype before could potentially lead to some unexpected results.
> 
> I think it would make sense to do a pre-call of `diff, _ = prototype_diff_from_object(prototype, obj)` and display the diff so the user can review what change will happen. To apply it could then be a simple [Y]/N prompt (see e.g. the delete command for how to make a Y/N prompt).
> 
> (More advanced would be to also be able to select/unselect the changes desired (like the OLC menu allows for), but that may be a little too much for this simple command. :) )

To answer your question: prototype flag calls the reset and force switches, so there will never be a diff.

That might be a boring response, and this could be made much more robust, but I'd say it'd be better to make that a separate command or add it to the OLC menu. Per my specified reason for change, this definitely accepts the reset back to a default state as preferred.

Griatch:
> To answer your question: prototype flag calls the reset and force switches, so there will never be a diff.

If there were no diff, there would be no change made to the object. There is a diff - it's just that the diff is the same as the prototype in this case. 

> That might be a boring response, and this could be made much more robust, but I'd say it'd be better to make that a separate command or add it to the OLC menu. Per my specified reason for change, this definitely accepts the reset back to a default state as preferred.

The OLC already handles the the case of applying a new prototype to an existing spawned entity (I believe only for the case of the same prototype though). I don't think it handles the case of applying a prototype to an object that didn't already have one (it will only find and update objects with existing prototypes). 
Herein lies the issue - if you apply this on an object without an existing prototype, I agree that wiping the old is the way to go. But if you apply it on an object with an existing prototype, this behavior can be quite unexpected. I suggest one of two possible solutions here: 

- Keep the default implicit reset, but add an extra warning to the diff-display when you are trying to apply this on an object with an existing prototype. This has the problem of not really offering an alternative at this time (I don't think the OLC nor the spawner supports this case in command form). 
- Don't enforce reset and instead show a warning with the diff-display that you need to set `/reset` explicitly if you want a blank slate. 

In both of these cases, displaying the diff to apply seems to be a good service to the user to avoid unexpected results. The diff is already there, why not use it.

biscuitWizard:
> > To answer your question: prototype flag calls the reset and force switches, so there will never be a diff.
> 
> If there were no diff, there would be no change made to the object. There is a diff - it's just that the diff is the same as the prototype in this case.
> 
> > That might be a boring response, and this could be made much more robust, but I'd say it'd be better to make that a separate command or add it to the OLC menu. Per my specified reason for change, this definitely accepts the reset back to a default state as preferred.
> 
> The OLC already handles the the case of applying a new prototype to an existing spawned entity (I believe only for the case of the same prototype though). I don't think it handles the case of applying a prototype to an object that didn't already have one (it will only find and update objects with existing prototypes).
> Herein lies the issue - if you apply this on an object without an existing prototype, I agree that wiping the old is the way to go. But if you apply it on an object with an existing prototype, this behavior can be quite unexpected. I suggest one of two possible solutions here:
> 
> * Keep the default implicit reset, but add an extra warning to the diff-display when you are trying to apply this on an object with an existing prototype. This has the problem of not really offering an alternative at this time (I don't think the OLC nor the spawner supports this case in command form).
> * Don't enforce reset and instead show a warning with the diff-display that you need to set `/reset` explicitly if you want a blank slate.
> 
> In both of these cases, displaying the diff to apply seems to be a good service to the user to avoid unexpected results. The diff is already there, why not use it.

This has been completed.

PR #2063, opened 2020-02-21
Boto3 / AWS contrib plugin

trhr:
This plugin is a one-stop shop for Cloudflare/S3 hosting of Evennia's Web assets. Copy some access keys and configurations into secret_settings.py, do full server reboot (for safety only, reload/restart/reboot might work but why chance it?), and you're no longer paying for bandwidth.

The primary classes are based on/shamelessly ripped from django-storages which I have used on many projects. It replaces django's default staticstorage class with this storage class. I can't see any reason why it would be any less than fully resilient, as its leveraging a 100% supported and documented Django feature and the official AWS python SDK. I chose reasonable, newbie-friendly options for configuring, but there are a lot more options in the docs that could be included on request.

I'm going to be storing a lot of images, and I can't spare the 200kB.

I did not add a test suite, because... how? S3 is credentialed.


trhr:
@Griatch If you need some AWS creds because you don't already have them, I can provide.

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
+"""
+This plugin migrates the Web-based portion of Evennia,
+namely images, javascript, and other items located
+inside staticfiles into Amazon AWS (S3) for hosting.
+
+INSTALLATION:
+
+1) If you don't have an AWS S3 account, you should create one now.
```
Griatch: I think it would be a good idea with a link or even a short description for how you can do this, and what it will cost (or not).

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+
+Credentials required are an AWS IAM Access Key and Secret Keys,
+which can be generated/found in the AWS Console.
+
+Example IAM Control Policy Permissions, if desired:
```
Griatch: This means nothing to someone not used to AWS. Where would this structure go? I bet the default assumption would be for people to put it in the Evennia settings file ...

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+for initial installation. You can remove it later, or you can
+create the bucket and set the ACL yourself before you continue.
+
+2) This package requires the dependency "boto3," the official
+AWS python package. You can install it with 'pip install boto3'
+while inside your evennia virtual environment (or, simply
+in your shell if you don't use a virtual environment).
+
+3) Customize the variables defined below in secret_settings.py,
+then run 'evennia stop', 'evennia start', 'evennia collectstatic'
```
Griatch: `evennia collectstatic` should not be necessary; the `evennia start` will run this silently. Also, `evennia stop` followed by `evennia start` is the same as calling `evennia reboot` (which will work also if the server was not previously started).

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+
+AWS_ACCESS_KEY_ID = 'EUHUB20BU08AEU7' # CHANGE ME!
+AWS_SECRET_ACCESS_KEY = 'a/uoexauodabuq4j;kmw;kvka0d2' # CHANGE ME!
```
Griatch: Don't you get these values from AWS? This makes it look like you just cook them up yourself ...?

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+AWS_STORAGE_BUCKET_NAME = 'mygame-evennia' # CHANGE ME!
```
Griatch: Again, while supposedly the bucket concept of S3 will be explained in a little more info added at the top, it would be good to explain just why this should be changed, because as it is written it looks just as important as changing the secret ids/keys above. A suggestion may also be good - like naming it after your game's name.

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
+
+from django.core.exceptions import ImproperlyConfigured, SuspiciousOperation, SuspiciousFileOperation
+
+try:
+    from django.conf import settings as ev_settings
+    if not ev_settings.AWS_ACCESS_KEY_ID or not ev_settings.AWS_SECRET_ACCESS_KEY or not ev_settings.AWS_STORAGE_BUCKET_NAME or not ev_settings.AWS_S3_REGION_NAME:
+        raise ImproperlyConfigured("You must add AWS-specific settings to mygame/server/conf/secret_settings.py to use this plugin.")
```
Griatch: FYI, this is far outside of PEP8 style and the 100-character line width used by Evennia. Why auto-formatting tools like `black` will handle most of this, it will not know how to correctly split such super-long strings, so please split the strings across multiple lines.

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+
+    if 'mygame-evennia' == ev_settings.AWS_STORAGE_BUCKET_NAME:
+        raise ImproperlyConfigured("You must customize your AWS_STORAGE_BUCKET_NAME in mygame/server/conf/secret_settings.py; it must be unique among ALL other S3 users")
```
Griatch: Same here.

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+    :type name: str
+    :param default: Value if setting is unfound
+    :returns: Setting's value
+    """
+    return getattr(ev_settings, name, default)
+
+
+def clean_name(name):
+    """
+    Cleans the name so that Windows style paths work
```
Griatch: This description is unclear. It also does not appear to be a name but a file path it's cleaning - converting Windows paths to Unix-style paths. A complete docstring would make this clear.

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+    base_path_len = len(base_path)
+    if (not final_path.startswith(base_path) or final_path[base_path_len] != '/'):
+        raise ValueError('the joined path is located outside of the base path'
+                         ' component')
+
+    return final_path.lstrip('/')
+
+
+def check_location(storage):
+    if storage.location.startswith('/'):
```
Griatch: This and other helper functions need full docstrings.

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+
+
+@deconstructible
+class S3Boto3StorageFile(File):
+
+    """
+    The default file object used by the S3Boto3Storage backend.
+    This file implements file streaming using boto's multipart
+    uploading functionality. The file can be opened in read or
+    write mode.
```
Griatch: This seems to be an internal thing used by the Storage backend. This should be clarified because the text makes it sound like it should be manually accessed (and if so, a usage example is needed - but since a file object is normally opened within a `with` context manager, there is no need to explicitly call `close` on it either).

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
+"""
+This plugin migrates the Web-based portion of Evennia,
+namely images, javascript, and other items located
+inside staticfiles into Amazon AWS (S3) for hosting.
```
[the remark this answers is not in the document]
Griatch: What is also sorely missing here is a usage example of how you would actually use this in-code. For example a few lines where you first open the storage bucket for access, and do a write followed by a read on it.

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+    write mode.
+    This class extends Django's File class. However, the contained
+    data is only the data contained in the current buffer. So you
+    should not access the contained file object directly. You should
+    access the data via this class.
+    Warning: This file *must* be closed using the close() method in
+    order to properly write the file to S3. Be sure to close the file
+    in your application.
+    """
+    buffer_size = setting('AWS_S3_FILE_BUFFER_SIZE', 5242880)
```
Griatch: Yeah this is likely a good enough default :)

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+                self._create_empty_on_close()
+        if self._file is not None:
+            self._file.close()
+            self._file = None
+
+
+@deconstructible
+class S3Boto3Storage(Storage):
+    """
+    Amazon Simple Storage Service using Boto3
```
Griatch: The relationship between this and the `S3Boto3StorageFile` is not clear, nor which of them should be actually used in code (even though it's clear once reading the code that this one is the one actually accessing the settings).

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+    reduced_redundancy = setting('AWS_REDUCED_REDUNDANCY', False)
+    location = setting('AWS_LOCATION', '')
+    encryption = setting('AWS_S3_ENCRYPTION', False)
+    custom_domain = setting('AWS_S3_CUSTOM_DOMAIN')
+    addressing_style = setting('AWS_S3_ADDRESSING_STYLE')
+    secure_urls = setting('AWS_S3_SECURE_URLS', True)
+    file_name_charset = setting('AWS_S3_FILE_NAME_CHARSET', 'utf-8')
+    gzip = setting('AWS_IS_GZIPPED', False)
+    preload_metadata = setting('AWS_PRELOAD_METADATA', False)
+    gzip_content_types = setting('GZIP_CONTENT_TYPES', (
```
Griatch: This is an interesting point - people used to Evennia Attributes may think that they could store anything (including raw Python structures/objects) in this bucket in a similar way. It should be made clear the limitations of S3 in this regard.

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,847 @@
...
+
+If you haven't made changes to your static files (uploaded images, etc),
+you can simply remove the lines you added to secret_settings.py. If you
+have made changes and want to install at a later date, you can export
+your files from your S3 bucket and put them in /static/ in the evennia
+directory.
+
+LICENSE:
+
+aws-s3-cdn contrib is (c) 2020, trhr and released under BSD 3-Clause
```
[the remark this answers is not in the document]
trhr: Gotcha. The vast majority of the code inside the included classes were from django-storages and this is their attached license. Licenses seem compatible on their face, but IANAL and yours was called the "BSD License" and the existing code I used was "BSD 3-Clause License" so I included for clarity. I guess the bottom line is we can delete all the license information and call them identical?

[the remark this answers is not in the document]
Griatch: If the code and license is from Django storages, then it must indeed be included, but moreso their original copyright statement / author must be retained - that is after all the only main requirement of any BSD license. You should add yourself as author of the contrib (see how others do it in the contrib/ folder) - you will then fall under Evennia's general license. But you cannot add your own (c) into Django storages license nor modify their original license text.

Otherwise yes - it's compatible since it's the same license. It's just two copyright owners that need to be retained.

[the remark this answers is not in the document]
trhr: I copied their authors.md, but pulled the full license text since it's the same license as yours, and just noted such.

For the test packages, I guess the way to go is just to grab the existing tests from django-storages, which looks fairly robust:
https://github.com/jschneier/django-storages/blob/master/tests/test_s3boto3.py

evennia/contrib/aws-s3-cdn.py
trhr: Noted, fixing.

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,1033 @@
...
+Shaung Cheng (S3 docs)
+Andrew Perry (Bug fixes in SFTPStorage)
+
+The repurposed code from django-storages is released under BSD 3-Clause,
+same as Evennia, so for detailed licensing, refer to the Evennia license.
+
+VERSIONING:
+
+This is confirmed to work for Django 2 and Django 3.
+'
```
Griatch: left-over quote character?

evennia/contrib/aws-s3-cdn.py
```diff
@@ -0,0 +1,1033 @@
+"""
```
Griatch: To follow the format of other contribs, add something like this to the top:
```
AWS storage system

trhr 2020
```
You should take some credit too, man ;)

evennia/contrib/aws-s3-cdn.py
trhr: necessary to fix buggy code coloring in atom IDE. if it doesn't break anybody else's IDE (assuming they properly interpret """docstrings""" , probably does more good than harm.

Griatch:
@trhr Any progress on this? It should be possible to test with some mocking, or?

trhr:
@Griatch I can't run tests on my box for evennia. Will CI run these now that they're part of the unittest package?

trhr:
@Griatch looks like these tests are going to fail regardless unless we add boto as a dependency for the entire evennia project.

Griatch:
@trhr That's fine. I'll solve this during merging of the PR.

trhr:
I removed the prior authors' deprecations as some functionality I liked had been deprecated to bring the packages more in line with enterprise usage, which I saw as silly and actually made configuring this for newbies harder. That's why the deprecation tests are failing. I'll address these tests by removing them.

I'll dig deeper into the other tests to see why they're failing.

trhr:
https://bitbucket.org/david/django-storages/issues/226/override_settings-does-not-work-for

Looks like this is a known issue for django-storages because of how the settings file is being read. Since test_override_settings only tests whether or not the @override_settings decorator works, which won't be a real use case for evennia, and this is a preexisting issue, I'll remove the test for now and add to the to-do list refactoring how the original author is reading the settings file, hopefully fixing at the source.

strikaco:
@trhr If one already has `django-storages` configured, are there any additional benefits to using this implementation?

trhr:
> @trhr If one already has `django-storages` configured, are there any additional benefits to using this implementation?

no.

PR #2066, opened 2020-02-27
Fixing unintended consequences of proxy_plugins

trhr:
Django returns a 404 for any putChild inside the proxy plugin module.

As written, this actually isn't putting the child into the Evennia Proxy, it's putting it into Webserver, which is explicitly not the point.

These two lines have to go down below the proxy resources are loaded.


trhr:
Disambiguation of three resources all called "web server" or "web root" -

The Evennia proxy: at_webproxy_root_creation(web_root) injects here
Website(), just below: a standard twisted web site
DjangoWebRoot: at_webserver_root_creation injects here

All three support putChild differently once the whole kit and kaboodle is created and served, but only injecting at the proxy level will serve protocols.

PR #2068, opened 2020-03-02
[WIP] Extended CharacterForm, added CRUM (a plugin that may make some exist…

trhr:
…ing code redundant/easier to maintain)

- This ticks off one documentation-only issue from #533 .
- Overall, the goal of this is permission-based front-end character creation and management.

I'm digging into #533; I wanted to get this little piece done before jumping into Admin ModelForms. 

Check out the crum library, as this PR adds it to requirements.txt. A lot of the frontend work from here till the end of time will need to access self.request.user in a function that doesn't pass "requests," and this keeps all that code clean and nice (as you know, trying to like pop a kwarg through multiple levels of super() inheritance is pretty messy.)

### DIFF
--- a/evennia/objects/admin.py
+++ b/evennia/objects/admin.py
@@ -35,12 +35,15 @@ class ObjectTagInline(TagInline):
 
 class CharacterForm(forms.ModelForm):
     def __init__(self, *args, **kwargs):
-        if 'initial' not in kwargs:
-            kwargs['initial'] = {}
-        if 'instance' in kwargs:
-            kwargs['initial'].update({'superuser': kwargs['instance'].is_superuser})
-            kwargs['initial'].update({
-                'db_account': AccountDB.objects.get_account_from_uid(kwargs['instance'].db.creator_id)
+        if "initial" not in kwargs:
+            kwargs["initial"] = {}
+        if "instance" in kwargs:
+            kwargs["initial"].update({"superuser": kwargs["instance"].is_superuser})
+            kwargs["initial"].update(
+                {
+                    "db_account": AccountDB.objects.get_account_from_uid(
+                        kwargs["instance"].db.creator_id
+                    )
                 }
             )
         super().__init__(*args, **kwargs)
@@ -54,14 +57,14 @@ class CharacterForm(forms.ModelForm):
         label="Player",
         help_text="Who should puppet this object?",
         required=False,
-        )
+    )
 
     superuser = forms.BooleanField(
         label="Superuser?",
         required=False,
         disabled=True,
-        help_text="You can't change this setting here; use the Account record."
-        )
+        help_text="You can't change this setting here; use the Account record.",
+    )
 
 
 class ObjectCreateForm(forms.ModelForm):
@@ -111,6 +114,7 @@ class ObjectEditForm(ObjectCreateForm):
     Form used for editing. Extends the create one with more fields
 
     """
+
     class Meta(object):
         fields = "__all__"
 
@@ -218,6 +222,8 @@ class ObjectDBAdmin(admin.ModelAdmin):
             obj (Object, optional): Database object.
 
         """
+        from evennia.objects.objects import DefaultCharacter
+
         defaults = {}
         if obj is None:
             defaults.update(
--- a/evennia/objects/admin.py
+++ b/evennia/objects/admin.py
@@ -7,7 +7,11 @@ from django.conf import settings
 from django.contrib import admin
 from evennia.typeclasses.admin import AttributeInline, TagInline
 from evennia.objects.models import ObjectDB
+from evennia.accounts.models import AccountDB
 from django.contrib.admin.utils import flatten_fieldsets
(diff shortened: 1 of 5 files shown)


trhr:
A bunch of #533 stuff is progressing nicely. I'm sure this will throw all kinds of bugs/errors if you try to mess with anything that's NOT a character object; I haven't finished all those bits yet.

- On the object's Edit view in the django admin, a field db_account is present. Changing the account here supports multisession modes (0, 1). Last-puppet gets set, existing sessions puppet their new characters. Multisession (2) works as well *except* when locks would prevent it. Maybe this is "working as intended" and requires some elbow grease by an admin to fix their locks, rather than inadvertently allowing any character to puppet any other character just because it gets set so in the admin.

- Added a new util, which is still in progress; a get_all_cmdsets function

- Started futzing around with some ideas for better typeclass/cmdset display. The Evennia core can't read the mygame dir, so I think that's why nothing custom is being added to either the typeclasses select or the commandset select. Any ideas?

trhr:
- The superuser flag sits in the "web only" section. This is the only setting that does matter in-game in the same way as on the web, so it should probably be described somehow in that header (this is just a documentation issue).
- For MULTISESSION_MODE=0,1, one should be able to tie a Character to the Player by setting the Object's db_player foreignfield. In MULTISESSION_MODE=2, one might instead want to assign the character to the _playable_characters Attribute on the Player. A custom form is probably needed for this linking.
- There should probably be a way to connect a Player to a Character in the admin, for MULTISESSION_MODE=0,1 to not need to use @ic to connect.

Those are done.

- Possibly there should be some sort of list of suggested Object cmdsets so one doesn't need to remember and write the path every time. This could be a help text or if possible some selector based on settings.CMDSET_*. Maybe one could consider some sort of clickable buttons that fill in some standard presets ...?

This seems sketchy, given the work on get_all_cmdsets(). We could very easily present defaults, but that's not particularly useful.

- When setting a cmdset on a Character in the admin, the cmdset seems to be set to be visible outside the character. The result is that you get multi-command errors when executing commands in the same room as the new Character.

Surely this has been fixed by now? Steps to duplicate?

- There should probably be some help info or preset for how to set the puppet: lock to allow a Player to puppet a given Character, such as puppet:id(<player id>).

Belongs in docs, surely.

Griatch:
@trhr Just to be clear - this is WIP PR, right? Please mark it as WIP in the title so it's clear it's still being worked on. Sounds great overall! The `get_all_cmdsets` function sounds interesting; I suppose it would be possible to fetch default cmdsets on objects (because they'd be stored on the object), but it'd be pretty impractical (or even impossible) to find all of them. 

>> Possibly there should be some sort of list of suggested Object cmdsets so one doesn't need to remember and write the path every time. This could be a help text or if possible some selector based on settings.CMDSET_*. Maybe one could consider some sort of clickable buttons that fill in some standard presets ...?

> This seems sketchy, given the work on get_all_cmdsets(). We could very easily present defaults, but that's not particularly useful.

Implementation may be tricky, but I don't see why the proposed functionality is sketchy. Manually having to enter paths is always going to be error-prone, and if the cmdset path is set wrong you can make the object unusable very easily. That said, it's clear that there is no central repository of cmdset paths, so presenting a list of the most common ones from settings (charactercmdset, accountcmdset etc) may be a good start at least? 

>> When setting a cmdset on a Character in the admin, the cmdset seems to be set to be visible outside the character. The result is that you get multi-command errors when executing commands in the same room as the new Character.

> Surely this has been fixed by now? Steps to duplicate?

This was not tested in a long while, but reproducing it (or not) should be simple - using the admin, create a character with the CharacterCmdSet on it, place it in a location. Then in-game, go to the location and enter `look`. If you get a multi-match error for the look command, it means the lock is not set on the new Character - because you are getting their `look` command as well as your own. 

>> There should probably be some help info or preset for how to set the puppet: lock to allow a Player to puppet a given Character, such as puppet:id().

> Belongs in docs, surely.

There is already ample docs for this; but this is the kind thing that is easy to forget if working in the admin, so some sort of text field about it is not a bad idea I don't think. 

> Started futzing around with some ideas for better typeclass/cmdset display. The Evennia core can't read the mygame dir, so I think that's why nothing custom is being added to either the typeclasses select or the commandset select. Any ideas?

I find it strange that importing e.g. settings.BASE_ACCOUNT_TYPECLASS in a running server would not access the game dir...

Griatch:
Never mind, I'm a dummy and had evennia installed without the `-e` flag after rebuilding this computer :D Let me try again ...

Griatch:
@trhr Bump! This is a very cool feature and it doesn't appear to be far from being wrapped up. Any plans on continuing work on it?

trhr:
> @trhr Bump! This is a very cool feature and it doesn't appear to be far from being wrapped up. Any plans on continuing work on it?

no.

Griatch:
Closing this, since no more work will be done on it, and the admin was reworked independently anyway. I did borrow the concept of the `get_all_cmdsets` helper function from this PR though, so thanks for that!

PR #2078, opened 2020-03-18
Account.create() does not respect settings.START_LOCATION

trhr:
No more spawning in Limbo from .create()!


Griatch:
The failing unit tests appear to be the real deal. There is some difference between the behavior of different dbs for this change.

trhr:
Not prepared to address any database-level complications in this. I'll workaround for my own use case by adjusting home.

Griatch:
:( I guess I'll have to fix this myself then. The error is real, after all.

Issue #1738, opened 2018-11-06
Security issue: Possible exploit from use of line break in channel message 


Griatch:
#### Brief summary of issue / Description of requested feature:

A user privately forwarded this possible exploit example. It is due to the channel message accepting line breaks in the web client.

#### Steps to reproduce the issue / Reasons for adding feature:

In the web client: 

1. public Hello!\n[Public] Griatch: Evennia is being re-written using Java.
2. The output will be  
    ```
    [Public] pruman: Hello!
    [Public] Griatch: Evennia is being re-written using Java.
    ```

... which can be confusing, 

#### Error output / Expected result of feature

We'd need to discuss how to handle this one. Using line breaks is a useful feature for a more advanced formatting of messages, but this kind of look-alike messaging could indeed be exploited, especially since colors are also possible to use in a channel message. 

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
The HTML converter will map line break to `<br>`, which causes this effect: 
```python
 return text.replace(r'\n', r'<br>')
```

This ties to other issues with how we handle the web client parsing - such as when we should pass through raw html or not.

strikaco:
What is an example of "advanced formatting?" 

Is there a reason we couldn't just add an indent to splitlined messages? I don't know how much *security* it provides, but I've seen it done this way in other protocols to differentiate between messages.

```
# untested
return text.replace(r'\n', r'<br>&nbsp;&nbsp;')
```

Should be:
```
[Public] pruman: Hello!
  [Public] Griatch: Evennia is being re-written using Java.
  [Public] Griatch: No, really, I mean it!
```

> such as when we should pass through raw html or not.

There is never a good reason to do this. We really need to be sanitizing all user input from all channels upon receipt. Once you accept and store unsanitized data, there is no accounting for how it will ultimately be interpreted or executed.

Griatch:
I like the idea of indenting, but remember we need to visually match what you get from the telnet client. I haven't checked what that produces. 

> >  such as when we should pass through raw html or not.
> 
> There is never a good reason to do this. We really need to be sanitizing all user input from all channels > upon receipt. Once you accept and store unsanitized data, there is no accounting for how it will 
> ultimately be interpreted or executed.

There actually is a use case - the use of `@py` as mentioned in #1369 (although that behavior should change from now, as discussed in that thread)

strikaco:
[...]

PR #2084, opened 2020-04-06
#1738 strip HTML, newlines, and tabs from Player input text

trhr:
Patched security issue that allowed players to spoof by adding newlines. Will not affect developers, and easily overloaded by server devs who desire different functionality.

1.0 gogogo

I wonder if this addresses MXP security issue? Surely you'd need to include HTML in your MXP link to actually execute arbitrary code from it.

### DIFF
--- a/evennia/server/inputfuncs.py
+++ b/evennia/server/inputfuncs.py
@@ -27,7 +27,8 @@ from evennia.commands.cmdhandler import cmdhandler
 from evennia.accounts.models import AccountDB
 from evennia.utils.logger import log_err
 from evennia.utils.utils import to_str
-
+from django.utils.html import strip_tags
+import re
 BrowserSessionStore = importlib.import_module(settings.SESSION_ENGINE).SessionStore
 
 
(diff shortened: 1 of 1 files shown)


evennia/server/inputfuncs.py
```diff
@@ -41,6 +42,28 @@
...
+        txt (str) : A cleansed input string
+    """
+    if session.account:
+        if session.account.permissions.get("Developer"):
+            return txt
+
+    # Strip unsafe characters from ingest
+    txt = strip_tags(txt)  # Any HTML codes
+    txt = re.sub(r'|\/', '', txt)  # New lines
+    txt = re.sub(r'|-', '', txt)  # Tabs
```
Griatch, line 65: This is highly specific to Evennia's default ANSI formatter and people can replace it at will (there is even a contrib to do it). Hard-coding these sequences here is not useful. 

Besides, these formatting instructions are expected by users - a user entering a description string in any more RP-heavy game will expect the ability to at least be able to divvy up their text into paragraphs with line break. If we remove them here, what's the point of having these formatting operations at all (if you produce this in code you can just use \n \t directly)?

evennia/server/inputfuncs.py
```diff
@@ -41,6 +42,28 @@
 def _NA(o):
     return "N/A"
 
+def _strip_unsafe(session, txt):
```
Griatch, line 45: If this is intended to be overridden, it's best to leave it named without the leading `_`. We have not been very consistent with this everywhere, but overall, "_private" methods tends to be something that is not really a part of the public API, and which can change more willy-nilly (this is the use in e.g. Twisted and Django too).

Griatch:
@trhr What's your view of my comments above? What is the plan forward for this PR?

trhr:
To be addressed at the channel level.

Issue #1709, opened 2018-10-26
RFC: Non-alphanumeric chars in account/character names


strikaco:
#### Brief summary of issue / Description of requested feature:
There is nothing stopping players from creating nonsensical or confusing names such as `+desc`, `@2@@@jds+-cjn` or even names mimicking default commands (`look`, `@ban`, etc.).

I thought I had addressed some of this with the Django ASCIIUsernameValidator but it allows more nonsense than it should in this context. Its criteria is `This value may contain only English letters, numbers, and @/./+/-/_ characters`.

I can fix this myself (considering replacing AUV with a regex validator and making commands/aliases reserved) but before I do, I wanted to see if it would be worth the effort-- under what circumstances do any of you *want* names to allow special characters (excluding hyphens)? Only if it's not the first character, or only if it isn't repeating, or only if it follows a string of three-but-no-more-than-four consonants and the name keeps a 1:3 vowel:consonant ratio? Trying to get a feel for diverse use-cases when coming up with a sane default.

Griatch:
It's a good question. I think it makes sense to to limit nonsensical and particularly command-like names. The main problem may be how to relay the error correctly to the user in an informative way. It's really hard to tell what kind of logic would be acceptable for the most people but it does feel like most people will not be too adventurous with the letter-combinations and would be ok with having those exotic variations restricted.

Griatch:
@jaggiJ: As a hint, a more efficient way to implement this without having to loop over the string character for character (which is slow), you could use sets: 
```python
allowed_set = set(allowed_input)
user_set = set(user_input.split())

# the difference is what is in user_set but is not in allowed_set
forbidden_set = user_set.difference(allowed_set)
if forbidden_set:
    print("{} characters are invalid for name".format(' '.join(forbidden_set))
# ...
```

strikaco:
@jaggiJ Wow, you allow 79-character usernames? I couldn't imagine having to type that out every time I need to `kill` someone-- they'd get multiple hits on me before I finish typing their name :)

Thanks for the contribution. I'm trying to take advantage of Django's existing validation framework, which uses things like regular expressions as parameters rather than custom functions (although your function could certainly be implemented as a standalone Validator). 

I think you might benefit from regex as well with your code, see:

```
import re
regex = '^([a-zA-Z0-9\&\.\-\_\s]{,79})$'    
users = (
    'pickle rick',
[...]

Issue #1445, opened 2017-09-26
Unicode Zero-Width-Space Character Key Oddities


chainsol:
#### Brief summary of issue / Description of requested feature:
One can set their character key to be the Unicode zero-width-space - U+200B - and potentially cause some security risks to other players of an Evennia-based MU*

#### Steps to reproduce the issue / Reasons for adding feature:

1.  At character creation in a stock or near-stock Evennia instance, set your character key to be the zero-width-space, Unicode 200B
2. Use the "emote" command, and output whatever message you want. The zero width space won't show up at all for most clients, and the only visible difference from an emote message and the real thing is the space at the beginning

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:
Checking that a key doesn't contain blacklisted characters and stripping any of those would be a good sanity check for core Evennia. Overall each game is responsible for implementation, but this particular ability is ripe for exploitation - for instance, a player can pretend to be an admin messaging another player.

Griatch:
This is an important observation. It should definitely be implemented as part of the Account creation and Character creation for all the MULTISESSION_MODES too. 

We might want to gather other forbidden characters to protect against in the same way. Alternatively this could be something we offer a setting for.

Kelketek:
The following Stack Overflow question may be instructive here.

https://stackoverflow.com/questions/37903317/is-there-a-python-constant-for-unicode-whitespace <https://stackoverflow.com/questions/37903317/is-there-a-python-constant-for-unicode-whitespace>

Griatch:
@Kelketek Wow, the whole unicode whitespace delio is certainly a lot more complex than one would think.

chainsol:
That answer on SO was confusing, but it appears importing string and checking string.whitespace will work for what our intent would be.

BlauFeuer:
This spoofing tactic is similar to various other names used like "Someone", "The", "A", "An", ".", "You", etc.  Defense against naming characters such things should probably include feature that blocks a regxp to match against forbidden names, and be tested at `@name` / `@rename`, also.

chainsol:
What about storing a settings key, and using that plus the string.whitespace?

Griatch:
This is addressed with latinification hooks in `develop` branch; closing this for better overview.

PR #2085, opened 2020-04-06
#1445 #1709 - Latinify character names in .create()

trhr:
Forces character names to be Latin script, preventing unicode injection

1.0 gogogo

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -2063,9 +2063,15 @@ class DefaultCharacter(DefaultObject):
         # If no typeclass supplied, use this class
         kwargs["typeclass"] = kwargs.pop("typeclass", cls)
 
+        # Normalize to latin characters and validate, if necessary, the supplied key
+        key = cls.normalize_name(key)
+
+        if not cls.validate_name(key):
+            errors.append("Invalid character name.")
+            return obj, errors
+
         # Set the supplied key as the name of the intended object
         kwargs["key"] = key
-        key = cls._validate(key)
 
         # Get home for character
         kwargs["home"] = ObjectDB.objects.get_id(kwargs.get("home", settings.DEFAULT_HOME))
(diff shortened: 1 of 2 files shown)


strikaco:
This is very useful but something of a misnomer. Your `_validate` function is actually *normalizing* the input string, not asserting whether it is valid.

It is a potential source of confusion and ambiguity because implementation on `Account` objects does distinguish between the two.

https://github.com/evennia/evennia/blob/10a6685fdf56c25c3e175df6753506af12f40d11/evennia/accounts/accounts.py#L544-L601

trhr:
Yeah, I agree. It's normalizing.

I was going to hook it into some kind of setting that would allow people to write "no numbers in usernames" but I didn't because the more I looked at it the more I hated giving any kind of default.

strikaco:
Heh, that's the reason I haven't touched #1709 in years :)

The best I've come up with is defining a mixin with a class-level regex and extending `create()` to check it before running the parent hook.

```
import re

class KeyValidator(object):

    key_regex = '^([a-zA-Z][a-zA-Z0-9\-\_\s]{2,30})$'

    @classmethod
    def validate_key(cls, key, *args, **kwargs):
        bool = kwargs.get('bool', False)
        if not re.match(cls.key_regex, key):
            if not bool: raise UserException(f"Name '{key[:6]}...' contains invalid characters.")
            return False
        return True

    @classmethod
    def create(cls, *args, **kwargs):
        key = kwargs.get('key', args[0])

        # Check for valid key
        cls.validate_key(key, **kwargs)

        return super().create(*args, **kwargs)
```

trhr:
I like the idea of making this standard functionality, but I also like the fact that right now, it solves the security issues and won't have a big impact on the playerbase. If there's some Swedish dev out there who writes their character names in squiggles, they can just overload and return txt to make this functionality not exist.

trhr:
@Griatch - the Accounts already have both normalize and validate, but because they have to work with Django, they use Django hooks.

```
    def normalize_username(cls, username):
        """
        Django: Applies NFKC Unicode normalization to usernames so that visually
        identical characters with different Unicode code points are considered
        identical.

        (This deals with the Turkish "i" problem and similar
        annoyances. Only relevant if you go out of your way to allow Unicode
        usernames though-- Evennia accepts ASCII by default.)

        In this case we're simply piggybacking on this feature to apply
        additional normalization per Evennia's standards.
        """
        username = super(DefaultAccount, cls).normalize_username(username)

        # strip excessive spaces in accountname
        username = re.sub(r"\s+", " ", username).strip()

        return username
```

I don't really want to apply custom validators/normalizers to accounts, since that's a Django thing and it could cause maintenance issues down the road. I suspect the code won't break if a character name is slightly-more-normalized than an account name, but I'll also update tests to include invalid names, if it's not already doing it.

Issue #2088, opened 2020-04-08
[BUG] Search function returns both list and queryset.


kovitikus:
#### Describe the bug
`caller.search` from a command returns a list for the item I have on my character, but a queryset for the item on the ground. Both items are of the Object typeclass, the only difference are the names and the locations of the objects. There's only one of each item in existence.

#### To Reproduce
1. Put item on player's character.
2. Put item on ground.
3. Call a search for each item.
4. Print the results.
5. Item on character will return as a list.
6. Item on ground will return as a queryset.

#### Expected behavior
Both searches should return a list.

#### Environment, Evennia version, OS etc
         Evennia version: 0.9.0 (rev dd13a3e2)
         OS: nt
         Python: 3.7.4
         Twisted: 19.2.1
         Django: 2.2.9

#### Additional context
Here's a pastebin showing my print results at the top and the code I was checking below:
https://pastebin.com/Mg2LSuDr

Here are the functioning commands, after I added in the proper len() checks, more specifically the CmdPut(Command):
https://pastebin.com/e1KNiSKC

Griatch:
Note that it's correct that the manager methods return querysets. The issue here is rather than the search function uses a raw manager query in one case and not in the other. So the fix should happen in the search method, not on the manager level.

Griatch:
This was actually already closed in 224ad5322.

PR #2091, opened 2020-04-09
Manager.py is returning a couple of querysets instead of lists. Fix for bug #2088

davidamandoestrada:
Manager.py is returning a couple of querysets instead of lists. Fix for bug #2088

https://github.com/evennia/evennia/issues/2088

### DIFF
--- a/evennia/objects/manager.py
+++ b/evennia/objects/manager.py
@@ -159,7 +159,7 @@ class ObjectDBManager(TypedObjectManager):
             typeclasses (list, optional): Python pats to restrict matches with.
 
         Returns:
-            matches (list): Objects fullfilling both the `attribute_name` and
+            matches (query): Objects fullfilling both the `attribute_name` and
             `attribute_value` criterions.
 
         Notes:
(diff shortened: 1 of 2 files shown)


davidamandoestrada:
Oh sorry. I thought the doc strings said these methods returned lists. It also seems get_objs_with_key_or_alias can return empty lists. I'll update accordingly.

Griatch:
@davidamandoestrada Yeah, it's an easy mistake to make, because there was a time were we *did* enforce a list-return from most of those managers. Homogenizing the api to give only one consistent return is an ongoing project. If doc strings still say they return list those need to be updated.

Griatch:
@davidamandoestrada Is the plan to continue work on fixing issue #2088 here, or should this PR be closed?

davidamandoestrada:
Yes, I'll work on it today.

davidamandoestrada:
@Griatch I think I can just change the results to a list in objects.objects.DefaultObject.search

Griatch:
@davidamandoestrada Yes, I think making sure that the DefaultObject.search converts the `result` to a list in all cases would make it the most consistent.

PR #2092, opened 2020-04-10
Validators and normalization of character names

trhr:
Rebasing #2085 to develop


trhr:
Closes #1445, #1709 unless you think otherwise.

PR #2094, opened 2020-04-10
Add a contents indexer to optimize inventory lookups.

volundmush:
Adds light / mild indexing to the ContentsHandler on ObjectDB, reducing the filtering and sorting that must be done for frequent operations like viewing a room description.

Also changed ContentsHandler's ._pkcache to be a set() instead of a 'Dictionary that only cares about keys and has None for values'. Because using a dictionary that way is weird.

Seeing how DefaultObject.return_appearance() was constantly guessing at what it was trying to display offends my sensibilities. I'm putting this out there as a possible alternate approach.

I'm not 100% sure the Wilderness contrib likes this.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -203,6 +203,8 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
     without `obj.save()` having to be called explicitly.
 
     """
+    # Used for sorting / filtering.
+    _content_types = ("object")
 
     # lockstring of newly created objects, for easy overloading.
     # Will be formatted with the appropriate attributes.
@@ -257,7 +259,7 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
             and not self.db_account.attributes.get("_quell")
         )
 
-    def contents_get(self, exclude=None):
+    def contents_get(self, exclude=None, category=None):
         """
         Returns the contents of this object, i.e. all
         objects that has this object set as its location.
@@ -266,6 +268,7 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
         Args:
             exclude (Object): Object to exclude from returned
                 contents list
+            category (str): A category to filter by. None for no filtering.
 
         Returns:
             contents (list): List of contents of this Object.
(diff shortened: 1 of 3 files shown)


strikaco:
> Because using a dictionary that way is weird.

...but not unprecedented. Sets cannot be JSON-serialized. Dictionaries with null values can.

evennia/objects/objects.py
```diff
@@ -2026,7 +2034,7 @@ class DefaultCharacter(DefaultObject):
     a character avatar controlled by an account.
 
     """
-
+    _content_types = ("character")
```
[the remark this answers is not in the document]
Griatch: It would also be good with a short comment here, explaining just what this is used for.

evennia/objects/objects.py
```diff
@@ -2277,7 +2285,7 @@ class DefaultRoom(DefaultObject):
     This is the base room object. It's just like any Object except its
     location is always `None`.
     """
-
+    _content_types = ("room", "object")
```
Griatch, line 2288: Add a comment to explain what this is for.

evennia/objects/objects.py
```diff
@@ -2427,7 +2435,7 @@ class DefaultExit(DefaultObject):
     exits simply by giving the exit-object's name on its own.
 
     """
-
+    _content_types = ("exit")
```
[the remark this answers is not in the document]
Griatch: Also add a comment.

evennia/objects/objects.py
Griatch: How come this is both a "room" _and_ an "object" when "exits" and "characters" are not?

evennia/objects/objects.py
```diff
@@ -1656,20 +1659,25 @@ def return_appearance(self, looker, **kwargs):
...
+        things_list = filter_visible(self.contents_get(category="object"))
+
+        things = defaultdict(list)
+
+        for thing in things_list:
+            things[thing.key].append(thing)
+        users = [f"|c{user.key}|n" for user in users_list]
+        exits = [ex.key for ex in exits_list]
+
         # get description, build string
```
Griatch, line 1681: Looks like the base `Object` is missing a `_content_types` tuple. It seems reasonable that you should be able to filter based on 'plain' "object" type as well.

evennia/objects/objects.py
volundmush: What happens if you have a room inside another room? The default return_appearance doesn't have a case for it.

Griatch: A room is, by default, the outermost container possible. In this way, one of the definitions of a room is that it has location `None`. But the typeclass is still Room, and an object, so it will appear in the hosting room as such and you will be able to look at it etc.

volundmush: Right, which is why it is considered an 'object' content_type, so that if it IS somehow inside another room, you can see it. :)

Griatch: It feels somewhat ... strange to handle it like this; you do filtering by content_type with the specific intent of filtering out things of a certain type. Having Rooms not _actually_ abide by that feels wrong.

volundmush: Fair enough. I'll make it just 'room' then. If people stick rooms inside other rooms, they'll have to figure out how to handle that. Or not handle it.

PR #2095, opened 2020-04-10
[On hold] Making ObjectSessionHandler easier to subclass with a different backend.

volundmush:
Just making a few adjustments to ObjectSessionHandler so it can load/save data from/to an overloadable location.

EDIT: Actually, let's change that. Making massive overhaul to Sessions and Account/Puppet linking infrastructure to reorganize and streamlined code for future expansion/moddability.

Evennia's fine on its own without it, but with this little tweak, ObjectSessionHandler becomes easier to refactor in the future, and easier for me to sub-class and re-purpose in the present.

EDIT: Actually, let's change that. As usual, I feel that these kinds of refactors lead to cleaner, more extendable and maintainable code.

All tests passed with flying colors. This is pretty painless.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -19,7 +19,7 @@ from evennia.scripts.scripthandler import ScriptHandler
 from evennia.commands import cmdset, command
 from evennia.commands.cmdsethandler import CmdSetHandler
 from evennia.commands import cmdhandler
-from evennia.server.serversession import EntitySessionHandler
+from evennia.server.linksessionhandler import ObjectSessionHandler
 from evennia.utils import create
 from evennia.utils import search
 from evennia.utils import logger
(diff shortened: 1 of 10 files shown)


evennia/objects/objects.py
```diff
@@ -66,14 +66,32 @@ def _recache(self):
         global _SESSIONS
         if not _SESSIONS:
             from evennia.server.sessionhandler import SESSIONS as _SESSIONS
-        self._sessid_cache = list(
-            set(int(val) for val in (self.obj.db_sessid or "").split(",") if val)
-        )
+        self._sessid_cache = self.load()
```
Griatch, line 69: Should this not be `self._load()` ?

evennia/server/serversession.py
```diff
@@ -537,3 +535,128 @@ def ndb_del(self):
...
+        """
+        Initializes the handler.
+
+        Args:
+            obj (Object): The object on which the handler is defined.
+
+        """
+        self.obj = obj
+        self._sessions_dict = dict()
+        self._sessions = list()
```
Griatch, line 556: The original implementation fetches this from the central ServerSessionHandler. Without a change to the ServerSessionHandler, there is a risk these will go out of sync.

volundmush: They do because... I forgot to stick account.sessions.remove(self) in session.at_disconnect(). Quite forgot that Evennia doesn't normally call those methods. Now it does. I'm not having any issues on my end now. Try again?

Griatch:
@volundmush This has seen a lot of change since the originally reviewed PR, to the point of this almost being a completely different PR. Is this the final candidate now?

volundmush:
> @volundmush This has seen a lot of change since the originally reviewed PR, to the point of this almost being a completely different PR. Is this the final candidate now?

Well. Yes and no. It is the final design for a ServerSession that is kept in memory and could use more consistent rules for linking to Accounts and Puppets, though it does alter the basic assumptions of a few tiny other parts of the code, such as it "not really being an Account that Puppets a thing". The impact of that change seems minimal, but does alter some approaches for calculating Multisession Logic which weren't previously an issue.

However, if the goal is to move ServerSessions into database as a SessionDB, then the 'constant linking/unlinking' of a Reload is no longer a concern, and I question the usefulness of my design.

Griatch:
@volundmush Yeah, with a SessionDB, the requirements/need for this would change. Let's pause this one for a while until we see how it goes with the SessionDB stuff.

Griatch:
This is not going to be implemented nor merged.

PR #2104, opened 2020-04-16
i18n stuff

trhr:
i18 for accounts and utils

objects and locks are already done
commands and typeclasses are out of scope
jade is doing comms
kovi is doing scripts

this might be about done.

### DIFF
--- a/evennia/accounts/accounts.py
+++ b/evennia/accounts/accounts.py
@@ -279,7 +279,7 @@ class DefaultAccount(AccountDB, metaclass=TypeclassBase):
             return
         if not obj.access(self, "puppet"):
             # no access
-            self.msg(_(f"You don't have permission to puppet '{obj.key}'."))
+            self.msg(_("You don't have permission to puppet '{key}'.".format(key=obj.key)))
             return
         if obj.account:
             # object already puppeted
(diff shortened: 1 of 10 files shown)


evennia/accounts/accounts.py
```diff
@@ -275,11 +275,11 @@ def puppet_object(self, session, obj):
             raise RuntimeError("Session not found")
         if self.get_puppet(session) == obj:
             # already puppeting this object
-            self.msg("You are already puppeting this object.")
+            self.msg(_("You are already puppeting this object."))
             return
         if not obj.access(self, "puppet"):
             # no access
-            self.msg(f"You don't have permission to puppet '{obj.key}'.")
+            self.msg(_("You don't have permission to puppet '{key}'.".format(key=obj.key)))
```
Griatch, line 282: Mm, this is not how i18n works. The formatting must be applied _after_ the translation. So it must look like this: 

```python
self.msg(_("You don't have permission to puppet '{key}'.").format(key=obj.key))
```

Note the placement of the format outside the `_()` call.

trhr: oh! that makes more sense.

PR #2114, opened 2020-04-19
Traitshandler contrib

Griatch:

This is an adaption of the [TraitHandler and Traits](https://github.com/evennia/ainneve/blob/master/world/traits.py) from Ainneve into a generic Evennia contrib. Original code by @whitenoiseoss and the Ainneve contributors.

While the comunity-driven Ainneve has shown some minor signs of life lately, it is not going anywhere fast and trying to keep backwards-compatibility to it was not feasible. Having a resource for managing persistent RP traits is a good thing to have in contrib. 

This modifies and refactors the original Trait and TraitHandler considerably. While the concepts are the same, the trait data produced is not 1:1 compatible with the original design. Some of the changes: 

- The original `Trait` is a monolithic class responsible for all three types (static, counter, gauge), separating the cases with if-statements in its methods. I split this into four different Trait classes: 
   - `Trait` - very basic parent class implementing input validation and storage as well as trait-arithmetic operations.
    - `StaticTrait` 
   - `CounterTrait`
   - `GaugeTrait`
- Each trait class has a few common class-properties to describe and itself in the `TraitHandler.add()` method. Each value also has a default it can fall back to. 
- The .actual read-only property was renamed to .value for consistency and more generic application.
- A validate_input() static method sits on the Trait itself. It's called by `TraitHandler.add()`  and allows each class to check the input was solid before it's stored in the database.
- Custom Trait classes can be injected by adding a list of class-paths to `settings.TRAIT_CLASS_PATHS`. 
- The TraitHandler was refactored to support this new module-loading functionality. It is now generic since the input arguments needed for a given Trait class is checked and specified by the Trait itself.
- Some of the functionality of Gauge and Counter traits were not documented. For example adding a modifier to gauge traits also modified the current attribute. In this version, Gauge modifier only affects the max value (which is base + mod).
- The original had the ability to set the max value of Counter traits to the string literal "base". This has no purpose for Gauge now (max is always base + mod) and for Counters it turned out to be very  confusing. So the .max value is the only thing affecting the upper bound of a Counter trait now. 
- Doing `del self.traits.propname` will now reset the value of that property back to whatever default is spefified (if no default is set, this will raise an exception). 
- A new `percent` utility function was added for general use (and for getting the percent value of Counter/Gauge traits). 
- I used some of the original unit tests, but most are new and rewritten. The test suite was also greatly expanded. 

There are also some new functionality of Counters and Gauge-traits: 
- `descs` is a property that takes a mapping of `{upper_bound: "text"}`. Calling the `.desc()` method will return the text describing the current value, for example "Badly hurt" if your health is below 10. 
- `rate ` and `ratetarget` makes the Trait changing over time. For example, if your current health is 10 and max is 100, a rate=1 will have that health climb up to the max over 90 seconds. `ratetarget` allows to set a more temporary stopping point within the min..max range. This is implemented completely on-demand and takes no CPU resources until it is checked.


evennia/server/portal/portal.py
```diff
@@ -261,6 +261,7 @@ def shutdown(self, _reactor_stopping=False, _stop_server=False):
     # Start telnet game connections
```
Griatch, line 261: This is just black formatting changes.

evennia/utils/utils.py
```diff
@@ -1687,6 +1687,66 @@ def format_table(table, extra_space=1):
     return ftable
 
 
+def percent(value, minval, maxval, formatting="{:3.1f}%"):
```
Griatch, line 1690: New percent utility made as part of the trait work.

Griatch:
@whitenoiseoss I have adapted your Traits/TraitHandler from Ainneve into a general Evennia contrib. I thought it prudent that you'd take a look at it before it merged to develop. :smile: 
(cc Ainneve devs @feend78, @friarzen, @chainsol  if you are interested).

evennia/contrib/test_traits.py
```diff
@@ -0,0 +1,999 @@
+#!/usr/bin/env python
```
Griatch, line 1: The position of this file may change. The contrib folder structure itself may see an overhaul. This felt too big to cram into the existing tests.py file.

whitenoiseoss:
@Griatch thanks for making this an official contrib! Maybe Effects, Events, and whatever else can come into here as well at some point. This code looks good to me. The separation into separate classes based on type makes it a lot nicer. Probably a lot for a newbie Python developer to digest, but the documentation up top should assist with them using it anyway.

PR #2133, opened 2020-05-11
TravisCI test fix

trhr:
makes tests pass again


trhr:
@Griatch I'm not sure if the fix should be "or 0," as written, vs "or now." IDE suggests these values count upwards from 0.

Also unclear whether it matters except for tests.

Griatch:
Ah, thanks for the PR! I forgot this PR was here so I already pushed a fix in a slightly different place. :/ And yes, it only affected tests. Closing this, sorry about that.

PR #2138, opened 2020-05-17
Make Puzzle's use command user-friendly.

Henddher:
The current help text of the `use` command is confusing to regular users. 

This PR makes the help text a bit user friendlier.


Henddher:
Hi @Griatch !
Another simple one.
Thanks!

PR #2141, opened 2020-05-25
Upgraded EvMenu

esampson:
Modified EvMenu so that node returns can be dict,dict instead of string,dict or tuple,dict. This allows for information to be more clearly conveyed (text is assigned to the text key and help is assigned to the help key) while also allowing for the passing of more complicated formatting instructions (the ability to hide options or to move them free of the main option  table, the ability to add a footer, and the ability to use alternate formats for display)

This upgrade should make it much easier for people to make minor adjustments to how various nodes are displayed without requiring them to re-write node_formatter or options_formatter and will allow the reformatting on a node-by-node basis. 

Final display is more modular. This allows the creation of alternate layouts without affecting legacy menus. Additionally it allows the implementation of a 'suppress' format which prevents the menu from being drawn at all (which is very useful as it allows the transition from one menu to a completely separate one rather than requiring the entirety of the menu to be in a single file.

String and tuples are still supported to allow for backwards compatibility though tuple should be deprecated. This is mentioned in the docstring and I will update the EvMenu tutorials and documentation once the PR goes live.

### DIFF
--- a/evennia/utils/evmenu.py
+++ b/evennia/utils/evmenu.py
@@ -1173,7 +1173,8 @@ class EvMenu(object):
         """
         return dedent(helptext.strip("\n"), baseline_index=0).rstrip()
 
-    def options_formatter(self, options):
+    @staticmethod
+    def options_formatter(options):
         """
         Formats the option block.
 
@@ -1491,7 +1492,8 @@ class EvMenu(object):
 
         return nodedata
 
-    def text_format_no_bars(self, callback, field, **nodedata):
+    @staticmethod
+    def text_format_no_bars(callback, field, **nodedata):
         """
         Text formatter for new methodology. Really a dummy function that
         exists when no special formatted is being done to a field.
@@ -1508,7 +1510,8 @@ class EvMenu(object):
 
         return nodedata
 
-    def node_format_default(selfself, callback, raw_string, **nodedata):
+    @staticmethod
+    def node_format_default(callback, raw_string, **nodedata):
         """
         Node formatter for new methodology. Does the final assembly of
         the node, placing the header at the top, the general text, the
@@ -1538,7 +1541,8 @@ class EvMenu(object):
             result = result + "\n" + nodedata['footer']['contents']
         return result[1:]
 
-    def node_format_invert(selfself, callback, raw_string, **nodedata):
+    @staticmethod
+    def node_format_invert(callback, raw_string, **nodedata):
         """
         Node formatter for new methodology. Does the final assembly of
         the node, placing the header at the top, the options table, the
@@ -1568,7 +1572,8 @@ class EvMenu(object):
             result = result + "\n" + nodedata['footer']['contents']
         return result[1:]
 
-    def node_format_suppress(selfself, callback, raw_string, **nodedata):
+    @staticmethod
+    def node_format_suppress(callback, raw_string, **nodedata):
         """
         Node formatter for new methodology. Sets the resulting nodetext
         to 'suppress' which prevents it from being drawn by
--- a/evennia/utils/evmenu.py
+++ b/evennia/utils/evmenu.py
@@ -858,6 +858,7 @@ class EvMenu(object):
             )
 
         # validation of the node return values
+        helptext = False
         if type(nodetext) == dict:
             if 'help' in nodetext:
                 helptext = nodetext['help']
--- a/evennia/utils/evmenu.py
+++ b/evennia/utils/evmenu.py
@@ -66,39 +66,61 @@ returned as None as well. If the options are returned as None, the
 menu is immediately exited and the default "look" command is called.
     text (str, tuple, dict or None): Text shown at this node. If a dict
         it can also contain various formatting instructions. In such a
-        case the followng keys are already supported:
-            - 'text' (str): The text of the node.
-            - 'help' (str): The help text of the node.
-            - 'header' (str): Header text for the node.
-            - 'footer' (str): Footer text for the node.
-            - 'format' (str): String dictating the layout of the node if
-                the default layout is not being used. It currently
-                supports three options:
+        case the following keys are already supported:
+
+            - 'header' (dict): Dictionary of header data for the node
+            - 'text' (dict): Dictionary of text data for the node
+            - 'options' (dict): Dictionary of options data for the node
+            - 'footer' (dict): Dictionary of footer data for the node
+            - 'help' (dict): Dictionary of help data for the node
+
+            Each of these keys support their own dictionary with the
+            following keys:
+
+                - 'contents' (string or list): Required. String data for
+                    all keys other than 'options'. List in the case of
+                    options.
+                - 'formatter' (string): Optional. Designator of the
+                    function used to format the contents. Unless this
+                    code has been subclassed the only options are 'bars'
+                    or 'no_bars'.
+
+            In the case of 'options' the following additional keys are
+            supported by its dictionary:
+
+                - 'hidekeys' (list): A list of keys. The options for
+                    those keys will not be listed in the options table
+                    though the node will accept them as input.
+                - 'hidedescs' (list): As hidekey but using the desc of
+                    the options instead.
+                - 'movekeys' (list): A list of keys. The options for
+                    those keys will be put into a separate list after
+                    the options table.
+                - 'movedescs' (list): As movekeys but using the desc of
+                    the options instead.
+                - 'rows' (int): Will attempt to make the options table
+                    the specified number of rows.
+
+            - 'node' (dict): A dict for consistency of use. At present
+                it only handles a single key:
+
+                - 'formatter' (str): String designating the function to
+                    be used for final assembly of the node. Unless this
+                    code has been subclasses it currently supports three
+                    options:
                     - 'suppress': The node is not drawn at all
-                    - 'no_bars': The node is drawn similarly to the
-                        default format but without separator bars.
-                    - 'invert': The node is drawn similar to the
-                        default format but the options table will be
-                        above the text.
-            - 'optionformat' (dict): A dictionary containing information
-                for the formating of the options table. Current options
-                are:
-                    - 'hidekeys' (list): A list of keys. The options for
-                        those keys will not be listed in the options
-                        table though the node will accept them as input.
-                    - 'hidedescs' (list): As hidekey but using the desc
-                        of the options instead.
-                    - 'movekeys' (list): A list of keys. The options for
-                        those keys will be put into a separate list
-                        after the options table.
-                    - 'movedescs' (list): As movekeys but using the desc
-                        of the options instead.
-                    - 'rows' (int): Will attempt to make the options
-                        table the specified number of rows.
+                    - 'default': The node is assembled similarly to the
+                        legacy method.
+                    - 'invert': The node is assembled with the header
+                        followed by the options table, the general text,
+                        and then the footer, more or less inverting the
+                        traditional layout.
+
         If a tuple, the second element in the tuple is a help text to
         display at this node when the user enters the menu help command
         there. Deprecated in favor of dict but kept for backwards
         compatibility.
+
     options (tuple, dict or None): If `None`, this exits the menu.
         If a single dict, this is a single-option node. If a tuple,
         it should be a tuple of option dictionaries. Option dicts have
@@ -137,17 +159,23 @@ Example:
     # in menu_module.py
 
     def node1(caller):
-        text = {"text": "This is a node text",
-                "help": "This is help text for this node",
-                "format": "scroll",
-                "optionformat": {"rows": 1} }
+        text = "This is a node text"
+        helptext = "This is help text for this node"
         options = ({"key": "testing",
                     "desc": "Select this to go to node 2",
                     "goto": ("node2", {"foo": "bar"}),
                     "exec": "callback1"},
                    {"desc": "Go to node 3.",
                     "goto": "node3"})
-        return text, options
+
+        display = {'text': {'contents': text,
+                            'formatter': 'bars'},
+                   'helptext': {'contents': helptext,
+                                'formatter': 'bars'},
+                   'options': {'contents': options},
+                   'node': {'formatter': 'invert'} }
+
+        return display
 
     def callback1(caller):
         # this is called when choosing the "testing" option in node1
@@ -607,13 +635,12 @@ class EvMenu(object):
 
     def _format_node(self, nodetext, optionlist):
         """
-        Format the node text + option section
+        Format the node text + option section. Legacy code. Use is
+        deprecated.
 
         Args:
             nodetext (str): The node text
             optionlist (list): List of (key, desc) pairs.
-            optionformat (dict, optional): Instructions to control the
-                layout of the optionlist
 
         Returns:
             string (str): The options section, including
@@ -630,11 +657,7 @@ class EvMenu(object):
         nodetext = self.nodetext_formatter(nodetext)
 
         # handle the options
-        if isinstance(nodetext, dict) and 'optionformat' in nodetext:
-            optionstext = self.options_formatter(optionlist,
-                                                 nodetext['optionformat'])
-        else:
-            optionstext = self.options_formatter(optionlist)
+        optionstext = self.options_formatter(optionlist)
 
         # format the entire node
         return self.node_formatter(nodetext, optionstext)
@@ -848,7 +871,7 @@ class EvMenu(object):
                 nodename = self.nodename
         try:
             # execute the found node, make use of the returns.
-            nodetext, options = self._execute_node(nodename, raw_string, **kwargs)
+            nodedata, options = self._execute_node(nodename, raw_string, **kwargs)
         except EvMenuError:
             return
 
@@ -857,20 +880,31 @@ class EvMenu(object):
                 "_menutree_saved_startnode", (nodename, (raw_string, kwargs))
             )
 
-        # validation of the node return values
-        helptext = False
-        if isinstance(nodetext, dict):
-            if 'help' in nodetext:
-                helptext = nodetext['help']
-        elif is_iter(nodetext):
+            # validation of the node return values
+        if isinstance(nodedata, dict):
+            nodetext = None
+        else:
+            nodetext = nodedata
+        helptext = ""
+        if not isinstance(nodedata, dict) and is_iter(nodetext):
             if len(nodetext) > 1:
                 nodetext, helptext = nodetext[:2]
             else:
                 nodetext = nodetext[0]
-        nodetext = "" if nodetext is None else nodetext
+        nodetext = "" if nodetext is None else str(nodetext)
         options = [options] if isinstance(options, dict) else options
 
         # this will be displayed in the given order
+        if isinstance(nodedata, dict):
+            if ('options' in nodedata
+                    and 'contents' in nodedata['options']):
+                options = nodedata['options']['contents']
+            else:
+
+                # if nodetext is a dictionary the second part of any tuple will
+                # be quashed. No combining of new and old methods. Pick a lane.
+                options = None
+
         display_options = []
         # this is used for lookup
         self.options = {}
@@ -899,17 +933,75 @@ class EvMenu(object):
                                 exec_kwargs,
                             )
 
-        self.nodetext = self._format_node(nodetext, display_options)
+        if isinstance(nodedata, dict):
+            # Use the new method to generate nodetext
+
+            if 'options' in nodedata:
+                nodedata['options']['display'] = display_options
+                optiondict = nodedata['options']
+                if 'formatter' in optiondict:
+                    formatter = getattr(self, 'options_format_'
+                                        + optiondict['formatter'])
+                    if not callable(formatter):
+                        formatter = getattr(self, 'options_format_no_bars')
+                else:
+                    formatter = getattr(self, 'options_format_no_bars')
+                nodedata = self._safe_call(formatter, '', **nodedata)
+
+            for field in ['header', 'text', 'footer', 'help']:
+                if field in nodedata:
+                    fielddict = nodedata[field]
+                    if 'formatter' in fielddict:
+                        formatter = getattr(self, 'text_format_'
+                                            + fielddict['formatter'])
+                        if not callable(formatter):
+                            if format != 'text':
+                                formatter = getattr(self,
+                                                    'text_format_no_bars')
+                            else:
+                                formatter = getattr(self, 'text_format_bars')
+                    else:
+                        if format != 'text':
+                            formatter = getattr(self, 'text_format_no_bars')
+                        else:
+                            formatter = getattr(self, 'text_format_bars')
+                    nodedata = self._safe_call(formatter, field, **nodedata)
+
+            if 'node' in  nodedata and 'formatter' in nodedata['node']:
+                formatter = getattr(self, 'node_format_'
+                                    + nodedata['node']['formatter'])
+                if not callable(formatter):
+                    formatter = getattr(self, 'node_format_default')
+            else:
+                formatter = getattr(self, 'node_format_default')
+            self.nodetext = self._safe_call(formatter, '', **nodedata)
+
+        else:
+            # Use the old method
+
+            self.nodetext = self._format_node(nodetext, display_options)
+
         self.node_kwargs = kwargs
         self.nodename = nodename
 
-        # handle the helptext
-        if helptext and isinstance(helptext, str):
-            self.helptext = self.helptext_formatter(helptext)
-        elif options:
-            self.helptext = _HELP_FULL if self.auto_quit else _HELP_NO_QUIT
+        if isinstance(nodedata, dict):
+            # use the new method to generate the help text
+
+            if 'help' in nodedata and 'contents' in nodedata['help']:
+                self.helptext = nodedata['help']['contents']
+            elif options:
+                self.helptext = _HELP_FULL if self.auto_quit else _HELP_NO_QUIT
+            else:
+                self.helptext = _HELP_NO_OPTIONS if self.auto_quit else _HELP_NO_OPTIONS_NO_QUIT
+
         else:
-            self.helptext = _HELP_NO_OPTIONS if self.auto_quit else _HELP_NO_OPTIONS_NO_QUIT
+            # handle the helptext through the old method
+            if helptext:
+                self.helptext = self.helptext_formatter(helptext)
+            elif options:
+                self.helptext = _HELP_FULL if self.auto_quit else _HELP_NO_QUIT
+            else:
+                self.helptext = _HELP_NO_OPTIONS if self.auto_quit else _HELP_NO_OPTIONS_NO_QUIT
 
         self.display_nodetext()
         if not options:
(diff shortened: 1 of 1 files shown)


evennia/utils/evmenu.py
```diff
@@ -64,10 +68,37 @@ def another_node(caller, input_string, **kwargs):
...
+        case the followng keys are already supported:        
+            - 'text' (str): The text of the node.
+            - 'help' (str): The help text of the node.
+            - 'footer' (str): Footer text for the node.
+            - 'format' (str): String dictating the layout of the node if
+                the default layout is not being used. It currently
+                supports two options:
+                    - 'suppress': The node is not drawn at all
+                    - 'scroll': The node is framed inside the ANSI image
+                        of a scroll.
```
Griatch, line 82: Such a specialized formatting option, no matter how cool-looking, does not belong in core. This is something that will need to be supplied via external means.

evennia/utils/evmenu.py
```diff
@@ -64,10 +68,37 @@ def another_node(caller, input_string, **kwargs):
...
-        node when the user enters the menu help command there.
+    text (str, tuple, dict or None): Text shown at this node. If a dict
+        it can also contain various formatting instructions. In such a 
+        case the followng keys are already supported:        
+            - 'text' (str): The text of the node.
+            - 'help' (str): The help text of the node.
+            - 'footer' (str): Footer text for the node.
+            - 'format' (str): String dictating the layout of the node if
+                the default layout is not being used. It currently
+                supports two options:
```
Griatch, line 79: I like idea of more control over each node's formatting but I think this is a missed opportunity of flexibility. If having this 'format' key, it makes sense to also be able to supply a callable to it to perform the format. That would mean a user could achieve the effect it needs without hard-coding a specific format structure with a given name (which is the opposite of flexible). 

Maybe it would even make sense to be able to supply separate header- text- and footer-formatters when we're at it - falling back to the defaults if not given. Importing and setting the main text formatter to the `scroll` function would then replicate the scroll effect.

evennia/utils/evmenu.py
```diff
@@ -64,10 +68,37 @@ def another_node(caller, input_string, **kwargs):
...
+                    - 'scroll': The node is framed inside the ANSI image
+                        of a scroll.
+            - 'optionformat' (dict): A dictionary containing information 
+                for the formating of the options table. Current options 
+                are:
+                    - 'hidekey' (list): A list of keys. The options for 
+                        those keys will not be listed in the options 
+                        table though the node will accept them as input.
+                    - 'hidedesc' (list): As hidekey but using the desc 
+                        of the options instead.
```
Griatch, line 90: This is a good idea. Would be more consistent to name them `hidekeys` and `hidedescs` though.

evennia/utils/evmenu.py
```diff
@@ -64,10 +68,37 @@ def another_node(caller, input_string, **kwargs):
...
-
-    text (str, tuple or None): Text shown at this node. If a tuple, the
-        second element in the tuple is a help text to display at this
-        node when the user enters the menu help command there.
+    text (str, tuple, dict or None): Text shown at this node. If a dict
+        it can also contain various formatting instructions. In such a 
+        case the followng keys are already supported:        
+            - 'text' (str): The text of the node.
+            - 'help' (str): The help text of the node.
+            - 'footer' (str): Footer text for the node.
```
Griatch, line 76: Why no `header` text when we are at it?

evennia/utils/evmenu.py
```diff
@@ -64,10 +68,37 @@ def another_node(caller, input_string, **kwargs):
...
+                        table though the node will accept them as input.
+                    - 'hidedesc' (list): As hidekey but using the desc 
+                        of the options instead.
+                    - 'movekeys' (list): A list of keys. The options for 
+                        those keys will be put into a separate list 
+                        after the options table.
+                    - 'movedesc' (list): As movekeys but using the desc
+                        of the options instead.
+                    - 'rows' (int): Will attempt to make the options 
+                        table the specified number of rows.
```
Griatch, line 97: This is a good feature I think.

evennia/utils/evmenu.py
```diff
@@ -1126,17 +1226,174 @@ def node_formatter(self, nodetext, optionstext):
...
         total_width = min(screen_width, max(options_width_max, nodetext_width_max))
         separator1 = sep * total_width + "\n\n" if nodetext_width_max else ""
         separator2 = "\n" + sep * total_width + "\n\n" if total_width else ""
-        return separator1 + "|n" + nodetext + "|n" + separator2 + "|n" + optionstext
+        result = separator1 + "|n" + text + "|n" + separator2 + "|n" + optionstext
+        if footer:
+            result = result + '|/' + footer
+        return result
+
+    def scroll(self, text, options, footer):
```
Griatch, line 1279: As cool as the effect looks, this does not belong in the core EvMenu. It may be a good contrib though, for people to import and use as a formatter for a specific node.

evennia/utils/evmenu.py
```diff
@@ -1126,17 +1226,174 @@ def node_formatter(self, nodetext, optionstext):
...
+
+    Returns:
+        result (str): input padded to width
+    """
+
+    result = input
+    t = strip_ansi(input)
+    a = len(t)
+    result = result + ' ' * (width - a)
+    return result
```
Griatch, line 1396: We shouldn't do workarounds of our own library code. This should be a Ticket against ANSIString and possibly a PR to fix the actual issue.

evennia/utils/evmenu.py
esampson: I chose to use this function primarily because I simply had it about. 

It occurs to me that I could very easily replace this with a 'no_bars' format (being able to suppress the bars on certain nodes being one of the original motivations of the project). I'll swap out scroll for no_bars unless you have some other alternate format you would like to use.

evennia/utils/evmenu.py
esampson: I think this is actually three things.

The first is that perhaps the name should be something other than 'format'. Obviously most of the other options have formatting effects as well. The idea for this options was a to have a sort of large scale single word option that would lie over the 'gross results'. 

Using scroll as the example you can probably see how, other than the solid bars above and below the text that are part of the current default look, scroll is essentially the exact same menu with an 'effect' applied.

This is not a requirement of these functions, however. The whole idea behind them is that someone might want to be able to 'invert' certain menus for some reason (putting the options table above the text) so I didn't really want to call it just **effect** or **style**. I would suggest **frame** as long as we are willing to accept something like **suppress** or **invisible** as an option.

The second thing is that I wanted to move to a dictionary setup because we can extend it much more easily. We don't have to add separate header-, text-, and footer-formatters right now because if we can add them relatively easy later with little risk of breaking backwards compatibility due to complex parsing arguments (which allows us to avoid the 'perfect is the enemy of enemy of done' problem). 

The third thing is I don't think we should rely on replacing the main text formatter. The purpose of the (currently named) **format** option is to allow the main text format to be changed on a node-by-node basis. I originally wanted to suppress the top and bottom bars but only on specific nodes. Someone else might have a variety of layouts like **scroll** and want to switch between them (doing a scroll format when a person is in the 'magic' section of the menus and a format where the text is flanked by swords when they are in the 'fighting' section of the menus).

evennia/utils/evmenu.py
esampson: Those actually are their names. The error is in the docstring.

evennia/utils/evmenu.py
esampson: Mostly because I didn't think when I was writing this how ``header`` was any different than something at the beginning of the textfield. Now that you've brought it up I can see how some formats might want to separate it for use as a title (above the top bar line) or something like that.

evennia/utils/evmenu.py
esampson: Thanks. Being able to control this on a node-by-node basis was one of the primary motivators for this.

evennia/utils/evmenu.py
esampson: I wanted to try to include something that had a more visible effect on the layout of a node to illustrate that there's a great deal of ability in the simple ``format`` (or whatever we eventually name it) function. However, I can probably do that in the tutorials and we can replace this with something simpler such as ``no_bars`` or ``invert`` (placing menu options above the text).

Given the simplicity of those routines we could probably even add both unless you are worried that too many alternate layout methods will bloat the routine.

evennia/utils/evmenu.py
esampson: This only was added to support ``scroll`` so we can easily pull it.

esampson: Actually, as I get ready to submit a ticket it looks like the issue might have been an error in my understanding of ANSIString. .ljust does seem to do just fine when the ANSIString is dealing with color data. The issue is when the string has embeded codes to generate a link in the webclient.

Either way, I've already pulled the function from the since it is no longer needed.

evennia/utils/evmenu.py
Griatch: I think it would be better to make this more flexible; I elaborate on this below.

evennia/utils/evmenu.py
```diff
@@ -64,10 +64,41 @@ def another_node(caller, input_string, **kwargs):
...
+            - 'format' (str): String dictating the layout of the node if
+                the default layout is not being used. It currently
+                supports three options:
+                    - 'suppress': The node is not drawn at all
+                    - 'no_bars': The node is drawn similarly to the
+                        default format but without separator bars.
+                    - 'invert': The node is drawn similar to the
+                        default format but the options table will be
+                        above the text.
+            - 'optionformat' (dict): A dictionary containing information
```
Griatch, line 83: If using the `formatter` callable suggestions above, these would all basically become quick-options for the default formatters only (the **kwarg to send into default formatters). So maybe something like `formatoptions` would be a better name in that case. One could also imagine appending these values to all **kwargs sent to custom formatters too.

evennia/utils/evmenu.py
```diff
@@ -64,10 +64,41 @@ def another_node(caller, input_string, **kwargs):
...
+            - 'footer' (str): Footer text for the node.
+            - 'format' (str): String dictating the layout of the node if
+                the default layout is not being used. It currently
+                supports three options:
+                    - 'suppress': The node is not drawn at all
+                    - 'no_bars': The node is drawn similarly to the
+                        default format but without separator bars.
+                    - 'invert': The node is drawn similar to the
+                        default format but the options table will be
+                        above the text.
```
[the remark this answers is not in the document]
esampson: I see what you are saying about the different formatters. I'm still not completely sold on these being necessities because if someone wants to do an inverted node with no bars they can subclass EvMenu, create a formatter function of ``invert_no_bars_format``, and override node_formatter (which is only 47 lines long now) to add 
``............elif nodetext['format'] == 'invert_no_bars':``
``................result = self.invert_no_bars_format(text, optionstext, header, footer)``

My goal wasn't to completely eliminate subclassing but to make it less necessary, easier to implement, and capable of handling things node-by-node without the user having to change the methodology.

With that said, I can still see validity to adding ``header_formatter``, ``text_formatter``, ``help_formatter``, ``options_formatter``, ``footer_formatter`` and ``node_formatter`` since it will expand the flexibility by a great deal without any subclassing. 

At present I think the only formatters for the subsections need to be ``bars`` or ``no_bars`` and then ``traditional``, ``invert`` or ``suppress`` for ``node_formatter``. Default behaviors when not called can be to create nodes in the current layout so that backwards compatibility is preserved.

How does that work for you?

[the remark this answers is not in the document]
esampson: Here's an idea:

Instead of making independent ``node_formatter``, ``text_formatter``, etc. keys what if we make the ``formatter`` key part of the entry specific ``dict``? This would require us to create an additional key for ``contents`` for each of these but it would also keep things more organized.

Additionally, by biting the bullet and making ``return`` accept a ``dict`` instead of making the first part of the ``tuple`` accept a ``dict`` we can avoid something that has been mildly annoying me, mainly that the formatting instructions are separated from the options list.

Usage would look similar to this:
Example

This would allow additional variables for custom formatters to easily be embedded by simply adding keys to the sub-dictionary. (Similar to how the ``node`` ``dict`` contains ``hidekeys`` and ``rows`` keys.

Backwards compatibility can be maintained by examining the return and seeing if it is a ``dict`` and just making standard assumptions if it is not.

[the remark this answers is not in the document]
Griatch: Overriding EvMenu is not quite the same thing, because we are talking about per-node formatting, something you can't currently easily reproduce with overriding. For changing the look of every node, then overriding is indeed the way to go. (This is also another advantage of formatting functions - you could just import and call the formatting function in your override if you wanted to - any such formatters we provide act as a library for people to use.)

I think returning a dict is not a bad idea; if expanding the `text` to return as a dict anyway, being able to include `options` in that return is a rather natural expansion. The original idea was to keep the node return simple, but it's clear there is a need for more complexity here. So this sounds like an ok change. And one can still keep a backwards-compatible handling.

Overall, the various sections of the return looks fine (rather CSS-like in a way). Having each section set aside with `formatter`, `content` etc does make it look quite clean, I like that. 

There is absolutely no reason to not allow `formatter` to accept a callable though - or, to follow the style used everywhere else, a tuple `(callable, {kwargs})`. This needs to be supported, there is no way around it. 

As for using strings like "no_bars" instead of callables, this may be ok if they are not hard-coded exceptions. That is, let the string-form be defined as being the name of a formatter-method on the EvMenu class. (so "no_bars" would mean using the method `EvMenu.format_no_bars`, "invert" would point a a method `EvMenu.format_invert` and so on). This would also make the string-based formatter style extendable - you could conceivably use the overriding of EvMenu to provide more such simple formatters.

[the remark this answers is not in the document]
esampson: I've update the code so that formatters are string based . The function finally called is a modification of the original string (so for options 'no_bars' uses the ``options_format_no_bars`` function while for strings it would use the ``strings_format_no_bars`` function.

evennia/utils/evmenu.py
esampson: I've changed 'optionsformat' to simply be part of the 'options' dict.

e.g.:
Example_Options
This keeps consistency with how all other parts (header, footer, text, etc.) of the node are defined which will hopefully increase ease of use.
(This design methodology means that if someone creates a formatter where they want to pass additional parameters they can easily pass those parameters by just expanding the corresponding dict)

evennia/utils/evmenu.py
esampson: I went ahead and added a ``header`` element. This is too useful for things such as a title at the top of the menu which can be centered by the node formatter if desired.

evennia/utils/evmenu.py
esampson: I think I have addressed all of the issues you elaborated on below.

evennia/utils/evmenu.py
```diff
@@ -64,10 +64,63 @@ def another_node(caller, input_string, **kwargs):
...
+            - 'footer' (dict): Dictionary of footer data for the node
+            - 'help' (dict): Dictionary of help data for the node
+
+            Each of these keys support their own dictionary with the
+            following keys:
+
+                - 'contents' (string or list): Required. String data for
+                    all keys other than 'options'. List in the case of
+                    options.
+                - 'formatter' (string): Optional. Designator of the
```
Griatch, line 83: This is now the name of the formatting function on the EvMenu class; this should be explained somewhere. It also makes sense to support the standard format here - `(formattername, {kwargs})`, where formatter is either a string (in this case, a method on the EvMenu class). 
(or an actual callable I presume, something like  `formattername(caller, evmenuinstance, **kwargs)`?)

evennia/utils/evmenu.py
```diff
@@ -64,10 +64,63 @@ def another_node(caller, input_string, **kwargs):
...
+                    though the node will accept them as input.
+                - 'hidedescs' (list): As hidekey but using the desc of
+                    the options instead.
+                - 'movekeys' (list): A list of keys. The options for
+                    those keys will be put into a separate list after
+                    the options table.
+                - 'movedescs' (list): As movekeys but using the desc of
+                    the options instead.
+                - 'rows' (int): Will attempt to make the options table
+                    the specified number of rows.
```
Griatch, line 102: For consistency, this should be `formatter` like everywhere else, with optional kwargs being possible to send to it.

evennia/utils/evmenu.py
```diff
@@ -64,10 +64,63 @@ def another_node(caller, input_string, **kwargs):
...
+                    be used for final assembly of the node. Unless this
+                    code has been subclasses it currently supports three
+                    options:
+                    - 'suppress': The node is not drawn at all
+                    - 'default': The node is assembled similarly to the
+                        legacy method.
+                    - 'invert': The node is assembled with the header
+                        followed by the options table, the general text,
+                        and then the footer, more or less inverting the
+                        traditional layout.
```
Griatch, line 117: It's unclear where this block belongs ... I presume it belongs in the outermost level (next to `options`, `text` etc), but not sure a separate `node` level is needed. Just have `formatter` at the top level next to `options`, `text` ...) - by its placement I think this would be quite clearly understood as a top-level formatter (and its arguments) that will affect every node.

evennia/utils/evmenu.py
```diff
@@ -857,17 +933,75 @@ def goto(self, nodename, raw_string, **kwargs):
...
+                nodedata['options']['display'] = display_options
+                optiondict = nodedata['options']
+                if 'formatter' in optiondict:
+                    formatter = getattr(self, 'options_format_'
+                                        + optiondict['formatter'])
+                    if not callable(formatter):
+                        formatter = getattr(self, 'options_format_no_bars')
+                else:
+                    formatter = getattr(self, 'options_format_no_bars')
+                nodedata = self._safe_call(formatter, '', **nodedata)
```
Griatch, line 949: This default value should be generalized and not done this deep inside the code. The possible formats for the `formatter` field should IMO be one of the following: 

   - `formatter = "formattername"`  -> Translated to a call `EvMenu.*_formattername(**default_kwargs)`, the `default_kwargs` are the no-bars, etc we want to have as ready, easy-to-use options to the default formatter. 
   - `formatter = ("formattername", {kwargs}")` -> Translated to f `EvMenu.*_formattername(**kwargs)` for supplying custom kwargs, to the default formatter or to your own. 
   - `formatter = myformatter` -> This will called as `myformatter(caller, evmenuinstance)` - it doesn't make sense to have default kwargs for a custom formatter. 
   - `formatter = (myformatter, {kwargs})` -> Will be called as `myformatter(caller, evmenuinstance, **kwargs)`.

The kwargs of this could thus be checked easily at the top, with default kwargs set in one place and then used later in the code. It'd simplify this code a lot, with many fewer if statements.

evennia/utils/evmenu.py
```diff
@@ -1014,7 +1149,7 @@ def display_helptext(self):
 
     def nodetext_formatter(self, nodetext):
         """
-        Format the node text itself.
+        Format the node text itself. Legacy code. Use is deprecated.
```
Griatch, line 1152: Thinking aloud: I wonder if it may be an idea to log a deprecationwarning for this format, to make people convert their existing menus. Maintaining both forms makes EvMenu considerably more complex ...

evennia/utils/evmenu.py
```diff
@@ -1138,6 +1325,272 @@ def node_formatter(self, nodetext, optionstext):
         separator2 = "\n" + sep * total_width + "\n\n" if total_width else ""
         return separator1 + "|n" + nodetext + "|n" + separator2 + "|n" + optionstext
 
+    def options_format_bars(self, callback, field, **nodedata):
```
Griatch, line 1328: As said, I disagree about the use of distinct methods for this, but think it's way too many special cases for the dev to remember. One standard-named `formatter` callable with optional kwargs to specify different formatting options is a lot less mental overhead. If wanting to keep things tidy, you could have the main default options formatter check the kwargs and call different sub-functions depending on what options are given.

evennia/utils/evmenu.py
```diff
@@ -1038,31 +1173,55 @@ def helptext_formatter(self, helptext):
...
+                option related to this node or dict containing the list as well
+                as various formatting instructions such as a list of options
+                to hide or move.
 
         Returns:
             options (str): The formatted option display.
 
         """
+        if isinstance(options, dict):
+            optionlist = options['display']
```
Griatch, line 1192: This is not covered in the docstring. Was this meant to be `contents`?

evennia/utils/evmenu.py
Griatch: Also, what's with the unused `callback` argument everywhere?

evennia/utils/evmenu.py
esampson: No. This is the ``display_options`` list generated in the ``node`` method starting at line 832 in the existing code and 988 in the proposed code. It isn't listed in the docstring because it isn't something that anyone enters (like ``display_options``)

evennia/utils/evmenu.py
esampson: I don't really see how it is less mental overhead for a developer to say that they want to use the ``no_bars`` format which calls ''options_format_no_bars`` over saying they want to use the ``no_bars`` format which calls ``formatter`` and then decides what part of the code to use within that method. Either way the developer is typing the exact same things.

At present the developers only really need to worry about 5 options, three for the node itself (default, invert, and suppress), and 2 for all the sub-entries to the node (bars and no_bars). That's why a modification is appended to the call name (so that ``bars`` for help calls a different formatter than ``bars`` for options)

The advantage to using the individual methods is that if someone wants to create a new formatter such as a ``box`` formatter (one that not only creates bars on the top and bottom but also creates vertical bar glyphs on the sides) they only need to create a new formatter function for their specific override (or two or three functions if they want boxes around text, options, and/or the assembled node itself). 

They do not need to go and copy the existing ``formatter`` function into the override and place their code inside which increases the possibility of breaking something. Also, by making the individual formatters their own cases it makes code sharing easier. Rather than having to cut formats from different sources and paste them into the ``formatter`` method they can simply be pasted `as is` into the override (with the possible exception that a name might need to be changed if there's naming collision between the formats).

As for the unusued ``callback`` argument, I'm using the ``_safe_call`` function you wrote to call the individual formatters. This means that when I add ``text_format_box`` to my override I don't need to go in somewhere else and add ``if field['formatter'] == 'box':``. 

From what I was able to deduce it seems like _safe_call demands that a variable be there in that position for callback to previous nodes. I was going to talk about that in the tutorial I plan to write for EvMenu.

evennia/utils/evmenu.py
esampson: I could certainly add something to log that a deprecated function is being used, but that seems to me to just add another complication to the code rather than removing anything.

At present I think we should continue to support both formats because we don't want to make everyone have to go back and rewrite all their menus. If we do eventually decide that we want to remove the original functions and force rewrites we should probably do that as a sort of phased rollout. Phase 1 is both formats continue to work equally with no warnings, just things in the docstring and tutorials saying this is the new preferred format. Phase 2 would be where we actually start to give gentle nudges with log warnings and the like. That way if people say 'why am I suddenly getting these warnings' the reply can be 'we said six months ago that we plan to deprecate the old method'. Phase 3 would be the final removal of the code.

evennia/utils/evmenu.py
esampson: I can move the default formats up to the top easily enough.

I think I need a little more explanation on ``EvMenu.*_formattername(**default_kwargs)``. Are you proposing that ``*`` be something like 'text_format' and 'formattername' be a variable? I didn't think that would work in python and even if it did I thought the preferred form was ```formatter = getattr(self, <string>)```
```nodedata = formatter(**kwargs)```

I used your ``_safe_call`` function because I thought it would be preferable, though that does require me to pass something to fill in the ``callback`` value.

If the intent isn't to call a variable name I would refer you to the discussion below. I think using individual functions called by variable name would be easier on developers.

evennia/utils/evmenu.py
esampson: I didn't want to refer to it purely as ``formatter`` because I'm concerned that it will become confused with the ``formatters`` for the individual entries, so ``node`` seemed a better option. I also realized that leaving it as a ``dict`` seems like a better choice because it makes it possible for someone to pass variables to a custom node formatter if they want to.

evennia/utils/evmenu.py
esampson: My concern with using kwargs like that is that this is already a dict with sub-dicts. Adding kwargs to the formatter rather than leaving them outside as they are would make sub-sub-dicts.

Also, at that point why aren't things like ``contents`` part of the kwargs? I think perhaps the issue is I need to re-examing the explanation of the methodology. Essentially ``options`` is the instruction to format something ``formatter`` simply one of the kwargs about how to handle it (along with the contents, movekeys, etc.)

Griatch: Yes, it's a more complex return structure. It's still more consistent than these custom variable names IMO.

evennia/utils/evmenu.py
Griatch: "node" is yet another custom key to remember. Better to use the same form as elsewhere: 
- `formatter = callable_or_str`
- `formatter = (callable_or_str, {kwargs})`

evennia/utils/evmenu.py
Griatch: With `EvMenu.*_formattername(**default_kwargs)` I just refer to a method string-name built from the string you pass to the `formatter=` key (two of the four possible forms above), and then passed through the `getattr(self, ...)`.

evennia/utils/evmenu.py
Griatch: Fair enough. We can deprecate later. It's just that the documentation gets ever more complex.

evennia/utils/evmenu.py
Griatch: I can possibly accept your argument on the basis of  an easier overload. If `formatter=` key allows both key, callable and kwargs to be sent as discussed, I suppose it'd be ok to provide some default methods to call even if they are separate methods.

Are you referring to the `EvMenu._safe_call` method? It doesn't require its input to have a callback? It only requires the callable to have a at minimum a `caller` as its first argument, with  an optional range of other possible args/kwargs. It's meant to execute nodes and node-like callables with variable arguments. 
Requiring useless boiler plate to appease `_save_call` goes against the purpose of that method though  - it's meant to give the dev more freedom for how their callables may look, not the other way around. For the formatters you have no need for the `caller` argument so if you want to use the `_safe_call` method you should expand on _safe_call to also handle the possible arguments to a formatter.

evennia/utils/evmenu.py
esampson: I see what you're saying there now. Essentially the entire dict is the 'node' dict so the 'formatter' option would be for the entire body and everything else are components used by that formatter. The each component has its own formatter option along with their specific components.

evennia/utils/evmenu.py
esampson: So basically it is calling by a variable name (as discussed below) you just don't want to use ``_safe_call``. No problem.

Griatch:
@esampson Thanks for your efforts on this. But ... The idea of offering per-node is good, but re-reviewing this after some time away from it, I don't think this is something that will merge. It adds a lot of extra complexity for power that few people will realistically use (the evmenu is likely over-engineered for most users as it is). With the latest EvMenu templating features (which goes towards simplicity rather than increased power), this would also likely need more work to be compatible. For power users, overriding the EvMenu class for custom formatting remains the most flexible solution. 
If pursuing this further, this may be better for a custom EvMenu child class in contribs.

I hope that sounds reasonable. I'm closing this, but you are welcome to comment further if you like.

PR #2153, opened 2020-06-30
Make CmdOpen grant control to creator

### LINKED ISSUE #2152: [BUG] Builders can't control exits created with open command.
#### Describe the bug
When you create an exit using the open command, your id is not added to the exit's lock string for control and delete.
This means that if a builder makes an exit with open, they can not modify or delete that exit.

#### To Reproduce
1. Login as a builder
2. Make an exit with open
3. Try to delete or modify it

#### Expected behavior
The creator's Id should be added to the exit's lock string, as is done with the create command, so that they can modify or delete it without being admin.

Evennia version: 0.9.0 (rev ff11fbd8)
OS: posix
Python: 3.7.3
Twisted: 19.7.0
Django: 2.2.5

#### Additional context
I ran into this while playing around on the demo.

bonnedav:
Edit building.py to make the open command assign control permissions to the creator. 

Ran into bug while playing around in the demo.

Not tested but based on existing code in CmdCreate so I imagine it will work.
Solves issue #2152

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -1407,6 +1407,7 @@ class CmdOpen(ObjManipCommand):
     locks = "cmd:perm(open) or perm(Builder)"
     help_category = "Building"
 
+    new_obj_lockstring = "control:id({id}) or perm(Admin);delete:id({id}) or perm(Admin)"
     # a custom member method to chug out exits and do checks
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks for the PR! This looks ok to me. Merged.

PR #2198, opened 2020-09-18
expose onTabCreate() so other plugins can create blessed goldenlayout…

friarzen:
… tabs

This just exposes the goldenlayout onTabCreate function so other, user-created plugins can use it to create tabs with the
proper 'tagging' drop-downs and retitle support.

Without this, other plugins would have to violate DRY principles.


Griatch:
Looks straightforward enough. Merging.

PR #2212, opened 2020-10-04
Separate User IP from Proxy when webclient uses AJAX

Tanookami:
This change will make it possible to see user IP addresses even while behind a proxy, similar to the websockets implementation.
Keeps expectations similar regardless of which technology is used by the webclient.
Unsure of relevant Issues. Happy Hacktoberfest - inspired me to try fixing some bugs or helping out!

### DIFF
--- a/evennia/server/portal/webclient_ajax.py
+++ b/evennia/server/portal/webclient_ajax.py
@@ -182,6 +182,16 @@ class AjaxWebClient(resource.Resource):
         csessid = self.get_client_sessid(request)
 
         remote_addr = request.getClientIP()
+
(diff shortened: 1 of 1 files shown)


TehomCD:
We hadn't really noticed IP addresses not being forwarded until we had a guest who was unhappy with us telling him his character idea didn't work, so he decided the proper response was to flood our guest channel with racial slurs. Tanookami hastily made a patch for us to fix it, and I encouraged her to send the fix here as a PR.

Griatch (approved):
Looks good. Thanks for the contribution. Cool that this already was helpful on Arx too.

Issue #2218, opened 2020-10-08
[Resizeable webclent text] Allow text within the web client to be resized and recolored independent of browser settings.


Cantir2068:
#### Describe the solution you'd like
An option within the web portal's options settings  to allow font size and type manipulation that does not rely on client browser don't settings, or zoom level.

Additional color settings would benefit for the sake of allowing end users to adjust what contrasts they are comfortable with for the default presentation of text independent of any softcore settings for indavidual tags within game, as some prefer full white on black,others prefer a light grey, beige, and others prefer a light green for aesthetics.

#### Describe alternatives you've considered
A clear and concise description of any alternative solutions
Normally simply bumping zoom or font up would be feasable, however considering the portal window to evennia itself is a separate element to the rest of the page it is in, adjusting global text values within client browser could cause unforseen formatting and readability issues based on client desktop resolution, fonts, windowed vs fullscreen, and so forth.

#### Additional context

With best correction I have 20/300 vision, have worked with text enlargement adaptive technology for close to thirty years,and have been within one mud/mush/etc community or other since roughly the mid 90s using everything from windows 95 terminal emulator, Linux terminal window, putty, mushclient, tintin, and potato.

I don't have all the answers, but a higher than average percentage of MU* players have vision issues of one stripe or another based on how friendly raw text is for visual immparement.
With best correction I have 20/³⁰

friarzen:
Font and Font Size controls were recently added to the settings tab that can be opened via the "Cog" Icon on the top left of "Main" pane.

Griatch:
@friarzen That's a good point, I had forgotten about those. @Cantir2068 You can set your own fonts if you open the settings (cogwheel) of the client. Is this enough to satisfy this request? If not, could you elaborate on what is missing?

Vikka:
> * Colors can be changed through CSS in your browser, in principle, but yeah, it's not as easy as clicking an option, that's for sure.

What do you think about this: An option panel that let you choose the different color used in the webclient ?

The goal is to save the settings and to make the color change easier than the CSS alternative.

Griatch:
@Vikka Yes, that would of course be possible; there are a few issues before one could do something like that though;
- Not all colors are controlled via easy CSS classes. If I'm not mistaken, some refactoring of the CSS structure needs to happen to make things easier to modify (without anything getting hard-coded/set directly in JS). @friarzen probably knows more about the state of this though. 
[...]

PR #2224, opened 2020-10-11
Make the goldenlayout layout work with Evennia options

friarzen:

This puts into place the infrastructure for the webclient layout to be stored and reset via server-side options.

This does introduce a breaking change for anyone who has defined custom webclient components.
Basically, instead of being able to just do the GoldenLayout registerComponent() once in a postInit() function, it will now need to define an onLayoutChanged() callback and do its registerComponent() calls each time.

See the options2.js plugin updates in this PR for an example.

Fixing long standing persistence issues.

Addresses ##1685
Addresses #2226 
Bug fixes address persistence issues of #2218


Griatch (changes requested):
This seems promising. I find some critical errors in the curren implementation though: 
Firefox: 
1. Be logged in or not. 
2. Reload page. With/without cache doesn't matter.
3. I get a completely black screen with no UI whatsoever. Reloading again doesn't help.
4. Manually wiping the localstorage is the only way to restore normal operation. But on the next page reload, the screen is all black again. 

Other points: 
- When changing the layout, the name in the save menu remains `default`. I think we talked about this changing to something like `default_modified` when that happened?
- When  clicking `Clear layout`, also the option window clears. I can see the reasoning, but it's pretty unexpected. Not sure if this is a good or bad thing. 
- What exactly is `Clear layout` clearing to? It does not clear to `default` because it seems possible to save over `default` with your own setting and `Clear layout` always goes back to something with one main pane and one input pane.
- If I save a new layout into `default` (just two vertical horizontal panes), then do `Clear layout`, I cannot select it until I first save a second layout (two horizontal panes) and then go back to `default`.
- Once I have selected my modified `default` setting as per above, I cannot go back to my second layout anymore - I can select it but the pane layout doesn't change.

I think it would make more sense if `default` was un-changeable (name field empty and save-button greyed out when it's selected?). If so, maybe the `Clear layout` is not even necessary - one could just select the default layout and be sure to get the default back. Setting the name (to something other than a default layout) allows you to save it. This would also allow for providing more than one immutable default (like small/medium/big maybe, like we talked about) that always are there in the list to pick if one wants.

friarzen:
I have managed to recreate that black screen issue exactly once...and that disappeared when I full-cleared my browser history, so I think it is something hanging around from the prior implementation.  I haven't been able to reproduce this since.

Clear layout wipes out the localstorage layout values and then force-reloads the page so that it acts exactly as if you were coming to the site clean for the very first time.  This then loads the contents of goldenlayout_default_config.js again as the "Default" layout.

Griatch (changes requested):
Sorry for the delay, but I've now sat down to test this out. It's working a lot better in this second iteration! I see no black-screens anymore and I can save and flip between layouts and it seems to work as expected most of the time. I have noted some issues: 

- The `modified-default` layout is really confusing. I _know_ what we talked about on this and still it confuses me. The issue is that it always exists - it doesn't matter if you modifed the layout while in unloggedin-mode. Heck, even if you rename it, it's still there! 
  - I suggest that the `modified-default` should auto-create only if you tweak the unlogged-in state and that there should be a more normal way to add a layout. For example a + sign to add a new "unnamed layout" line that you can then rename and save. 
- There is no way to remove layouts. When you experiment (or just write the wrong name) you quickly fill up that list. We shouldn't require people to have to wipe the storage and then delete the right key of the on-Account Attribute...
   - I think there should be a matching `-` button for deleting the currently selected layout (all but the `default` layout). 
- Unclear if related to this feature or not: I found that sometimes the `options` pane just stops opening. Often after having created new panes and dragged them around and saved them to a new layout. It's unclear how to reproduce exactly, but clicking the cogwheel will just not do anything until I reload the page. I've looked if the options were simply just already open in some other pane but didn't find it if so.

friarzen:
Yeah, after playing around with it and trying to incorporate the idea of layout save/delete UI elements, I have come to the conclusion that the whole "default-modified" behavior just doesn't match anyone's expectations.  Removing it and just allowing the player to save their current layout using a label seems to be the least confusing approach.

As for the options pane, I haven't ever gotten it to stop working, but I have reproduced the case where it requires up to three clicks to reopen the options pane, due to an annoying interaction between goldenlayout's idea of "active working pane", the open/close state tracking, and how the page reloading works when selecting a new layout.  Unfortunately, when we go to save the layout, the options pane is always open (that is where the save button is), so the saved layout always includes it as the active restored pane.  The only real way to fix that is to have the save button close the Options Pane, so I've added that behavior and changed the wording on the save button a bit to help offest expectations.

Griatch (changes requested):
This is again much better! 
- The saving and loading of layouts seem to work much more reliably now. It makes much more sense to write a name and save it rather than how it worked before. Nice! 
- I like that the option pane auto-closes if switching to a layout which would overlap it. I think you are right - this 3-click is what I hit too. I have not been able to get to a state where the option pane "won't" open with this version. 

There is now only one thing that's dearly missing from this feature - layout deletion. I tried my hardest to delete a layout manually (deleting `_saved_webclient_options` Attribute, wiping the browser storage, reloading the browser with no-cache) ... I can't get my `default3` layout to go away no matter what I do. :smile: That list will keep growing. 

- The final missing feature is the equivalent of a `-` button for deleting the currently active (non-default) layout.

friarzen:
Layout deletion is included in this version, but I set it so that it won't let you delete the default or the currently active layout.  Deleting the currently active layout causes state-weirdness when you close and then reopen the options panel.

Griatch:
@friarzen Hm, I tried to delete all sorts of ways and couldn't figure it out ... should there be a delete button somewhere here? 
DeepinScreenshot_select-area_20201029001518

friarzen:
So, looking at that screenshot, I notice that your splithorizontal button still has a dropdown arrow on the right.
That makes me think you might have older data cached somewhere, but, when I went to double check this last night, I got a somewhat similar result and traced it to a bug with one of the last updates I did to try to cut down on the number of Codacy errors.  I have figured out the solution and submitted another commit to this PR.   Make sure you test with a brand new Evennia database and cleared cache, though, since the bug caused it to store corrupted data into the options and possibly localstorage.
The delete buttons should show up now.

Griatch (changes requested):
This worked much better! Once I wiped things I could see the new interface. Much easier to understand now and saving/loading works fine. I still see an issue with deleting though :sweat_smile: 

- Start fresh (only `default` layout)
- Change panes and save as a new layout, say, `split1`
- Change panes and save it as a third layout, say, `split2`
- Now select `default` again so the custom layouts are available for deletion
- Press the X to delete the `split2` layout. It goes away but this doesn't close the option pane. Click the `default` (or `split1`) layouts to switch to it and close the option pane. 
- Open the option pane again. `split2` is still in the list. Deleting it again and selecting one of the other layouts will now _actually_ remove it (it won't come back when re-opening the options menu).

friarzen:
Ok, wow, this one is due to a pretty crazy interaction with the server.   I consider this to be a server side bug.

The sequence of events is like so:
Client loads, user logs in and creates some layouts, these get saved to the server.
Client deletes a layout, sends the updated layout list to the server, minus the deleted layout.
Server sends webclient_options outputfunc to the client about 8 times in a row.... (Why so many?)
     The first 7 times, it sends the old contents of the webclientLayouts kwarg with all layouts, including the deleted one.
The client duly redefines each layout as provided by the server, including the deleted definition...
     The 8th/final time, it sends the new webclientLayouts kwarg, with the deleted layout having been removed.
The client duly redefines each mentioned layout, but the deleted/recreated one isn't mentioned, so it continues as if undeleted...

Now, the second time the user deletes the layout, it sends the updated list to the server just like last time,
    But, the server's first 7 or so "webclient_options" outputfunc's all match the 8th, so the layout stays gone.
so the client

Griatch:
Hm. interesting. I'll check why it would send so many returns, that's not ideal :)

Griatch (approved):
I fixed what turned out to be a less-than-optimal way for dicts to update - the save operation would be triggered every time for every element of the dict which in turn triggered the monitor of the webclient-options Attribute 9 times in a row. This caused the webclient to be notified as the dict was gradually updated to the new value. A good side effect is that dicts will update better now too. :) With this change the layouts delete properly. 

I found one more quirk of the save operation: 

- Start to edit the `default` layout name, like `default1234232`
- The `Close Options and Save` button becomes active
- Now delete the extra letters, down to `default`
- The `Close Options and Save` buttons remains active and you can now save your modified layout as 'default' 
- This is however not permanent, as loading `default` returns the 'real' default layout. 

This is confusing for the user since it may suggest that they are able to make their own 'default' when they actually cannot. But I will not block this PR merge more over it since it doesn't appear to have a lasting impact and is a rather special case. I will make a new issue for it. 

Thanks for your work on this, this is a very useful feature. For people who wants to change the default, having the ability to change, save and then extract the goldenlayout from the options makes this even more useful to devs. It's something we should document but doc updates are locked right now so we'll have to keep that in mind.

Issue #2227, opened 2020-10-11
[BUG] Lockfuncs - Better handling of cases where `accessing_obj` is a Command instance


strikaco:
#### Describe the bug
(Continuation of #1618; this really is starting to deviate from the scope of the original ticket.)

A side-issue arose from my observation that locks are not being consistently checked on the appropriate object for quelled users-- in some cases the locks are checked using the instance of the Command as the `accessing_obj`, instead of the actual caller or Object.

Given the following container Object and some commands to enter and leave it, `Enter` will always work (albeit incorrectly, since you can recursively enter the same object) and `Exit` is never added to the usable cmdset:

```
class CmdEnter(Command):
    
    key = 'enter'
    locks = 'cmd:not inside()'
    
    def func(self):
        self.msg("You enter.")
        self.caller.move_to(self.obj)
    
class CmdExit(Command):
    
    key = 'exit'
    locks = 'cmd:inside()'
    
    def func(self):
        self.msg("You leave.")
        self.caller.move_to(self.obj.location)

class ContainerCmdSet(CmdSet):
    
    key = "Container"

    def at_cmdset_creation(self):
        self.add(CmdEnter())
        self.add(CmdExit())

class Container(Object):
    
    def at_object_creation(self):
        super().at_object_creation()
        self.cmdset.add(ContainerCmdSet(), permanent=True)
```

Errors are raised in the log concerning the absence of a srcobj when the commands are invoked. Superusers simply bypass this and the lock-checking altogether.

```
20-10-06 23:27:08 [EE] objects.container#581: 'NoneType' object has no attribute 'access'
20-10-06 23:27:10 [EE] objects.room#108: 'NoneType' object has no attribute 'access'
```

In this case, the commands are being added to the usable command set based on whether `CmdExit` itself is inside the container-- instead of the caller.

To work around this, I redefined the `inside` lockfunc and added a check to look for delegate objects:

```
def inside(accessing_obj, accessed_obj, *args, **kwargs):
    """
    Usage:
       inside()
    Only true if accessing_obj is "inside" accessed_obj
    """
    if inherits_from(accessed_obj, 'evennia.commands.command.Command'):
        accessed_obj = getattr(accessed_obj, 'obj', accessed_obj)
            
    return accessing_obj.location == accessed_obj
```

And after looking at the code for other functions, other lockfuncs perform similar checks, though inconsistently:

[...]

PR #2228, opened 2020-10-11
Adds appropriate object checking to objtag lockfunc

strikaco:
Modifies the `objtag` and `inside` lockfuncs to perform checks on the appropriate objects.

Commands locked `cmd:objtag(blah)` were checking for tags on the Command instance instead of the bound Object.
Commands locked `cmd:inside()` were checking if the Command instance was inside the bound Object instead of the caller.

Partial fix for #2227.

### DIFF
--- a/evennia/locks/lockfuncs.py
+++ b/evennia/locks/lockfuncs.py
@@ -539,6 +539,8 @@ def objtag(accessing_obj, accessed_obj, *args, **kwargs):
     Only true if accessed_obj has the specified tag and optional
     category.
     """
+    if hasattr(accessed_obj, "obj"):
+        accessed_obj = accessed_obj.obj
     tagkey = args[0] if args else None
     category = args[1] if len(args) > 1 else None
     return bool(accessed_obj.tags.get(tagkey, category=category))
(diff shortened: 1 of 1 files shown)


Griatch:
This seems fine.

Griatch:
... Actually, I was too fast to close this. You just made the `objtag` lockfunc identical to the `tag` lockfunc which already has the `.obj` check. This is deliberate, since in one case you want to check the object's tag and in the other case not.

strikaco:
> ... Actually, I was too fast to close this. You just made the `objtag` lockfunc identical to the `tag` lockfunc which already has the `.obj` check. This is deliberate, since in one case you want to check the object's tag and in the other case not.
> Without this, there is no way to explicitly target tags on the main object rather than on its `.obj`.

Wait, what? There should still be a difference between the two in terms of whether ´obj´ is pulled from the ´accessing_obj´ or the ´accessed_obj´. The lockfuncs should look identical but have one very subtle change. Otherwise commands locked ´objtag´ look for a taghandler on the command instance, which silently fails.

Did you revert because of material problems encountered or just "that don't look right"? I was selective in only modifying these two locks because I'm not aware of any issues.

strikaco:
@Griatch Would you kindly confirm the above? The lock is broken in its current state; this PR does fix it but do let me know if it raised other issues.

Griatch:
@strikaco You know what - you are right. I was not identifying the right differences between the lockfuncs. I have reinstated your change now, thanks for staying with it. Sorry about that! :facepalm:

strikaco:
No worries, they really do look identical at first glance.

Issue #2223, opened 2020-10-11
[BUG] Reload resets the LOGIN_THROTTLE_LIMIT


TehFamine:
#### Describe the bug
I believe it's the login throttle limit, but whatever prevents a single connection (e.g.: same IP address) from creating more than a couple accounts can be circumvented by simply waiting for the MUD to reload. While this may seem silly, the point of the setting is to prevent a single IP from creating multiple accounts. Reloading seems to reset the count so if you restrict a single IP to only having 2 accounts, the reload means you can now create 4 of them.

#### To Reproduce
1. Create 2 accounts (think default throttle is 2)
2. Wait for a game reload / force a crash
3. Create 2 more accounts above the throttle

#### Expected behavior
Would expect the throttle to be persistent.

#### Environment, Evennia version, OS etc
If unsure, run `evennia -v` or get the first few lines of the `about` command in-game.

#### Additional context
This is not a huge issue, but still seems like a bug because you would expect the setting to be persistent across reloads.

strikaco:
https://github.com/evennia/evennia/blob/367db031fa9e5727c3fbf90e624ab06bd9b6f82f/evennia/accounts/accounts.py#L54-L56

The throttle was never meant to be persistent or to permanently block anything-- only to *slow down* mass account creations. It resets itself automatically after some period of time, no reload necessary.

Until the world at large adopts ipv6, you really don't want to have a system in place that auto-bans ipv4 addresses for long periods of time anyway. It can be exploited to block/DoS other users in the same address pool (i.e. NATted users on mobile networks).

Griatch:
@strikaco I don't think @TehFamine meant that the throttle should be persistent-persistent (as in blocking forever) but rather that it should not be auto-timing out across a reload ...

I can see the argument for consistency, but not sure it's worth the effort to implement ...

strikaco:
> - While this may seem silly, the point of the setting is to prevent a single IP from creating multiple accounts. 
> - Reloading seems to reset the count so if you restrict a single IP to only having 2 accounts, the reload means you can now create 4 of them.
> - Would expect the throttle to be persistent.

@Griatch I did get the impression that the expectation was to hard-limit the number of accounts created per IP.

[...]

PR #2230, opened 2020-10-13
Refactors throttle to use Django caching (partial fix for #2223).

strikaco:
Refactors the Throttle to make use of Django's cache API instead of an internal dictionary.

- User request for persistence support
- Support for different storage backends

Partial fix for #2223.

To make the throttle persist across reloads, a custom cache that specifies disk storage needs to be defined in `settings`:

```
CACHES = {
    'throttle': {
        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
        'LOCATION': '/tmp/ev/throttle/', # you need to specify a directory to dump cache files in
        'TIMEOUT': 60 * 60 * 24 * 365 * 10000,  # 10,000 year timeout, or could be None I suppose
        'OPTIONS': {
            'MAX_ENTRIES': 5000,  # IME more than 10k entries in filesystem cache does not work well
        },
    },
}
```

### DIFF
--- a/evennia/accounts/accounts.py
+++ b/evennia/accounts/accounts.py
@@ -52,10 +52,10 @@ _MUDINFO_CHANNEL = None
 
 # Create throttles for too many account-creations and login attempts
 CREATION_THROTTLE = Throttle(
-    limit=settings.CREATION_THROTTLE_LIMIT, timeout=settings.CREATION_THROTTLE_TIMEOUT
+    name='creation', limit=settings.CREATION_THROTTLE_LIMIT, timeout=settings.CREATION_THROTTLE_TIMEOUT
 )
 LOGIN_THROTTLE = Throttle(
-    limit=settings.LOGIN_THROTTLE_LIMIT, timeout=settings.LOGIN_THROTTLE_TIMEOUT
+    name='authentication', limit=settings.LOGIN_THROTTLE_LIMIT, timeout=settings.LOGIN_THROTTLE_TIMEOUT
 )
 
 
--- a/evennia/accounts/accounts.py
+++ b/evennia/accounts/accounts.py
@@ -55,7 +55,7 @@ CREATION_THROTTLE = Throttle(
     name='creation', limit=settings.CREATION_THROTTLE_LIMIT, timeout=settings.CREATION_THROTTLE_TIMEOUT
 )
 LOGIN_THROTTLE = Throttle(
-    name='authentication', limit=settings.LOGIN_THROTTLE_LIMIT, timeout=settings.LOGIN_THROTTLE_TIMEOUT
+    name='login', limit=settings.LOGIN_THROTTLE_LIMIT, timeout=settings.LOGIN_THROTTLE_TIMEOUT
 )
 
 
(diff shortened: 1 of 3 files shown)


strikaco:
One remaining bug I need to address-- this will adhere to the lower of the default Django cache timeout or the TTL set on the object instance itself. By default I think the TTL is hard-enforced at 30 seconds (default cache) instead of 5 minutes (object instance). 

I think an in-memory `throttle` cache should be implemented for all cases and the timeout values specified by the cache options dict in settings. Then users will have a template they can copy and modify.

strikaco:
> One remaining bug I need to address-- this will adhere to the lower of the default Django cache timeout or the TTL set on the object instance itself. By default I think the TTL is hard-enforced at 30 seconds (default cache) instead of 5 minutes (object instance).

I was incorrect about this; there was no bug. Since timeouts were being specified at the time of the `set` operation, those trump the defaults defined in settings. The default turned out to be 300 seconds anyway.

> I think an in-memory `throttle` cache should be implemented for all cases and the timeout values specified by the cache options dict in settings. Then users will have a template they can copy and modify.

I've implemented this nonetheless, though the timeouts are still specified at the settings/instance level for consistency instead of trying to pull them out of the cache definition.

Griatch (changes requested):
I think the code looks good, this is a nice cleanup and change of the throttle. Since this is a security-related feature it would be good with a few small unit tests of the functionality though (the actual cache could be mocked I guess).

strikaco:
> I think the code looks good, this is a nice cleanup and change of the throttle. Since this is a security-related feature it would be good with a few small unit tests of the functionality though (the actual cache could be mocked I guess).

Maybe I'm still not getting the "mock" thing, but the cache is a Django construct and tested upstream. Since this *is* a security feature that relies on a native library, we'd really want these tests to fail fast if changes to the cache API would interfere with its operation. (The throttle already had Ev unit tests that this refactor continues to meet.)

strikaco:
(Accidentally hit close)

I can mock out the cache, sure, but why would we want to in this case?

Griatch:
Sure, run the cache if it can be run safely during the test, I was just trying to make it easier.

Please note that the develop unit tests fail for other reasons, so for now, just check the specific new tests pass.

strikaco:
> Sure, run the cache if it can be run safely during the test, I was just trying to make it easier.

Ah, now I get where you were going with that line of thought. Sorry for the misunderstanding!

The in-memory cache is instantiated per-process, so the test runner is inherently going to create its own instance of it. In addition the test Throttle does not share the same namespace as the login or creation throttles, and the test doesn't perform any overtly destructive operations either. The default settings are very compartmentalized and safe.

This changes somewhat when using non-default settings. If one insists on going the persistent route-- at worst, and only _when the cache is 100% full_, the three oldest keys in an on-disk cache will be evicted. But in all cases, this is still ephemeral storage being worked with-- those keys were subject to eviction in the next _n_ minutes or next _m_ logon attempts anyway. 

(It does add the possibility of the unit test failing if on-disk storage is specified and the unittest is run twice in a row, since the keys would have persisted. I should purge the test keys at the end of the test and use non-routable IPs altogether.)

PR #2231, opened 2020-10-14
EvMore - Use page_formatter provided in args

strikaco:
The EvMore constructor was ignoring the `page_formatter` arg and hard-coding `str` for that function.

This change makes EvMore will use the formatter provided in args or default to `str` if one is not.

Pagination was broken again.

### DIFF
--- a/evennia/utils/evmore.py
+++ b/evennia/utils/evmore.py
@@ -256,7 +256,7 @@ class EvMore:
 
         self._npages = 1
         self._paginator = self.paginator_index
-        self._page_formatter = str
+        self._page_formatter = page_formatter
 
         # set up individual pages for different sessions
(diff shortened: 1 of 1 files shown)


Griatch:
Nope, the `page_formatter` kwarg is no more, it was a thing I introduced for the previous attempt at improving paging performance but it was the wrong direction. The reason is because the page formatter as a kwarg is simply not flexible enough - for many paginations you need a lot more information from EvMore than you can get in an external callable like that. So `EvMore` is moving in the same direction as `EvMenu`- if you want to implement a custom page formatter, make a child class of EvMore and override `.page_formatter` as a method. You'll find new custom hooks for `paginator` and `page_formatter` as well as `init_pages` for this purpose.

PR #2245, opened 2020-10-28
History bugfix

### LINKED ISSUE #2233: [BUG] Webclient doesn't save new text in textarea before switching to a history up/down event.
#### Describe the bug

New text in progress should be saved before processing history events.

#### To Reproduce

If you type something into the input textarea and then hit shift+up/down-arrow, your text is lost.

#### Expected behavior
The text should be stored and then if you shift+down-arrow back to the end of the history list, it should restore your work.

latest 0.9

friarzen:

Fixes the issues with not saving partial text before hitting shift+up/down for history control.

bugfix

fixes #2233

### DIFF
--- a/evennia/web/webclient/static/webclient/js/plugins/history.js
+++ b/evennia/web/webclient/static/webclient/js/plugins/history.js
@@ -39,7 +39,6 @@ let history = (function () {
             history[history.length-1] = input;
             history[history.length] = "";
         }
-        historyPos = 0;
     }
 
     // Public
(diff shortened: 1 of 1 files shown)


Griatch (approved):
I was momentarily thrown off by the fact that your "abandoned" (non-committed) entry is not *last* in your history - you can do Ctrl-Down "past" it and end up with a blank input line. I guess I would have expected to stop at that entry since it was not yet sent. But I see from how the history works that this is likely the only way to do it without introducing some new 'uncommitted' storage mechanism instead of using the history buffer. 
Worth thinking about, maybe. But this is already much more convenient functionality, so merging!

Issue #2252, opened 2020-11-14
[BUG] Tutorial creation errors with MULTISESSION_MODE = 2


kcoram:
#### Describe the bug
Running 'batchcommand tutorial_world.build' in a new game world configured with MULTISESSION_MODE = 2 generates an unexpected stack trace:

```
Traceback (most recent call last):
  File "/home/kevin/Source/muddev/evennia/evennia/commands/cmdhandler.py", line 640, in _run_command
    yield cmd.parse()
  File "/home/kevin/Source/muddev/evennia/evennia/commands/default/muxcommand.py", line 201, in parse
    self.character = self.caller.get_puppet(self.session)
  File "/home/kevin/Source/muddev/evennia/evennia/accounts/accounts.py", line 376, in get_puppet
    return session.puppet
AttributeError: 'NoneType' object has no attribute 'puppet'
```

#### To Reproduce
1. Generate a new world with `evennia --init mygame`
2. cd mygame
3. evennia migrate
3. Update server/config/settings.py to add `MULTISESSION_MODE = 2` and `MAX_NR_CHARACTERS = 2`
4. evennia start
5. create super user account
6. log into game as super user
7. `ic Superuser` to puppet the character
8. `batchcommand tutorial_world.build`
9. Observe the stack trace

Details from the server.log file:

```
20-11-14 18:49:27 [::] Traceback (most recent call last):
20-11-14 18:49:27 [::]   File "/home/kevin/Source/muddev/evennia/evennia/commands/cmdhandler.py", line 640, in _run_command
20-11-14 18:49:27 [::]     yield cmd.parse()
20-11-14 18:49:27 [::]   File "/home/kevin/Source/muddev/evennia/evennia/commands/default/muxcommand.py", line 201, in parse
20-11-14 18:49:27 [::]     self.character = self.caller.get_puppet(self.session)
20-11-14 18:49:27 [::]   File "/home/kevin/Source/muddev/evennia/evennia/accounts/accounts.py", line 376, in get_puppet
20-11-14 18:49:27 [::]     return session.puppet
20-11-14 18:49:27 [::] AttributeError: 'NoneType' object has no attribute 'puppet'
20-11-14 18:49:27 [EE] User input was: 'unquell'.
```

#### Expected behavior

Running the `batchcommand tutorial_world.build` script should not generate stack traces. The problem does not happen with `
MULTISESSION_MODE = 0`.

#### Environment, Evennia version, OS etc

evennia -v:
```
[...]

PR #2255, opened 2020-11-17
Add null checking when getting puppet from session

kcoram:

When account.execute_cmd() is called without an explicit session
argument for a server using MULTISESSION_MODE 2 or 3, it does not
know which session to use and may leave it None. Downstream, code
attempting to get the puppet from the session can fail with an
error because of the None session.

This change adds in checking for the None session, and sets the puppet
to None if there is no session defined.

Addresses bug #2252

### DIFF
--- a/evennia/accounts/accounts.py
+++ b/evennia/accounts/accounts.py
@@ -373,7 +373,7 @@ class DefaultAccount(AccountDB, metaclass=TypeclassBase):
             puppet (Object): The matching puppeted object, if any.
 
         """
-        return session.puppet
+        return session.puppet if session else None
 
     def get_all_puppets(self):
         """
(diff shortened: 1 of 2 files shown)


kcoram:
I am not sure what should be done about the `test-suite / build (3.7, postgresql)` failure. Advice welcome!

Griatch:
@kcoram: Looks like some temporary issue in the unittest action on github. Looks fine. Thanks for the PR!

kcoram:
Thank you for the merge! I will close out the bug I filed as fixed by this PR. 

Fixes #2252

Griatch:
Thanks, I forgot to close the ticket :smile: A hint for the future is to add the line `Resolves #2252`  to your commit. Github will then close the linked ticket automatically when the PR merges.

kcoram:
Good to know, thank you. We use Bitbucket Enterprise at my day job, so I don't know many GitHub tricks.

I will remember the `Resolves XYZ` message for next time!

PR #2256, opened 2020-11-18
text2html lines starting with a space need nbsp

### LINKED ISSUE #2254: [BUG] ANSI-to-html leading spaces in messages are not converted
#### Describe the bug
In the webclient ansi/html conversion, leading spaces in messages are not converted to `&nbsp;` anymore.
This can lead to alignment issues.

#### To Reproduce
@py self.msg(" this string should be started with a non-breaking space")

#### Expected behavior
Any leading/trailing whitespace on a message sent to the client should be converted to `&nbsp;`'s to force the browser to not ignore them.

This is new in 0.9.5

#### Additional context
This is probably an edge case related to the other space/tab modifications that were made in 0.9.5.

friarzen:

This update corrects the generated HTML where an input string or new line of text (via |/) begins with a space.
Browsers will ignore leading whitespace for new HTML "blocks", so this requires leading non-breaking spaces to fix.

Evennia 0.9.5 introduced this issue as part of the TAB and other text2html updates.

Fixes #2254 

Also, fixes what I believe is a spacing bug with leading TAB characters (normal space + 3 non-breaking spaces yields incorrect indents of only 3 character widths).

### DIFF
--- a/evennia/utils/text2html.py
+++ b/evennia/utils/text2html.py
@@ -97,8 +97,8 @@ class TextToHTMLparser(object):
     re_blink = re.compile("(?:%s)(.*?)(?=%s|%s)" % (blink.replace("[", r"\["), fgstop, bgstop))
     re_inverse = re.compile("(?:%s)(.*?)(?=%s|%s)" % (inverse.replace("[", r"\["), fgstop, bgstop))
     re_string = re.compile(
-        r"(?P<htmlchars>[<&>])|(?P<tab>[\t]+)|(?P<space> +)|"
-        r"(?P<spacestart>^ )|(?P<lineend>\r\n|\r|\n)",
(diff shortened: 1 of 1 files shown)


Griatch (changes requested):
Thanks for the PR! Overall this looks ok, I have some requests for change though: 
- The regex should not assume anything about which marker-format is used. This is supposed to be replaceable in by the user after all. 
- The unit test should be updated accordingly. Probably a new unit test is also needed to make sure the case of the new-line is handled but also that it still correctly works to leave a space in the middle of the line (there may already be unit-tests for this case though, so just make sure there is).

Griatch:
@friarzen Is this PR still in process or should it be closed due to forthcoming changes on the client side?

Griatch:
@friarzen :arrow_up:  ?

Griatch:
@friarzen Presumably this PR is superceeded by the change to move rendering client-side. So I'm closing this PR.

PR #2260, opened 2020-11-27
Adjusting the width of the output in CmdTime

TehFamine:
CmdTime has two table outputs that are not aligned in the same width. This change aligns the tables to the same width of 78.

Just a simple fix for a simple mistake with the table outputs in CmdTime where one table was not the same width as the other.

Because it was simple, yet annoying.


TehFamine:
image

PR #2266, opened 2020-12-07
Evennia Generic Plug-in System

volundmush:
The creation of a new EvPlugin interface class and the infrastructure to support loading a list of them defined in your settings.py. Plugins integrate with Evennia's load process, calling hooks at varying points with plenty of opportunities for a developer to use them in very creative ways.

For an example of how an EvPlugin is implemented, check out: https://github.com/volundmush/athanor/blob/evennia_plugin/athanor/__init__.py#L8

Evennia has long been overdue for a robust solution to installing mods and plugins. Although "what is a plugin" is somewhat vague, this is a structured way of connecting arbitrary code to Evennia and have it load and run hooks at specific points without having to setup a dozen things manually. Huge improvement.

Mild changes had to be made to numerous files in order to make the plugins load smoothly. This has resulted in a few sharp edges that I may need some advice in smoothing out.


evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
+"""
```
Griatch, line 1: Not sure about the position of this module - but it's hard to claim it belongs in utils/ or server/ per-se ... Probably would also make more sense to name it `plugins.py` since it handles all plugins.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+class EvPlugin:
+    """
+    Base class to use for Evennia Plugins.
+
+    NOTE: the constructor and init_settings methods must not import anything or couple itself to code that's
+        server-only. Keep in mind that the Evennia Launcher must also be capable of running __init__ and
+        init_settings() !
+    """
+
+    def name(self) -> str:
```
Griatch, line 18: This seems like it should be a `@property`.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        The name of the plugin. This is how Evennia and other Plugins will refer to it.
+
+        Returns:
+            name (str)
+        """
+        return f"{self.__class__.__module__}.{self.__class__.__name__}"
+
+    def load_priority(self) -> int:
```
Griatch, line 27: Also a `@property`.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+        First hook to be called. Override this to make a plugin alter the Evennia settings.py
+        during the boot-up process.
+
+        WARNING: This is called by the launcher and possibly in other places, and the run state
+            of Evennia is not guaranteed. ONLY use this method to alter the settings.
+
+        Args:
+            settings (settings.py): The module reference to settings.py.
+            plugins (dict of name->EvPlugin classes): Allows plugins to know what
+                plugins will be loading.
```
Griatch, line 48: The type should `module` here, and just put `dict` as the type for `plugins`. This is more importantly now that our API documentation is parsing this automatically to generate the api documentation.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Returns:
+            None
```
Griatch, line 51: When there are no returns, the `Returns:` docstring field does not need to be specified.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        The order that plugin's load hooks will be called in. the lower the number, the sooner
+        it will be called.
+
+        Returns:
+            number
+        """
+        return 0
+
+    def init_settings(self, settings, plugins):
```
Griatch, line 37: For consistency, this sounds like this should be `at_init_settings`.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+        Args:
+            settings (settings.py): The module reference to settings.py.
+            plugins (dict of name->EvPlugin classes): Allows plugins to know what
+                plugins will be loading.
+
+        Returns:
+            None
+        """
+
+    def at_plugin_setup(self):
```
Griatch, line 54: So this is run for every plugin, whereas the next method in the sequence is run once all plugins have all finished initialized; this should be made clearer in this docstring.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        First phase of plugin setup at the end of evennia._init()
+        """
+
+    def at_plugin_collaborate(self):
```
Griatch, line 59: This name is a bit unclear; the critical thing here is that _all_ plugins have been setup at this point. So maybe rename to something like `at_post_all_plugin_setup(self, plugins)`, where `plugins` is the `PLUGINS` dict, passed explicitly.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        Second phase of plugin setup during evennia._init(). At this point, all plugins
+        are available in the evennia.PLUGINS dictionary.
+
+        Use this method for any plugin setup that involves plugins collaborating with each other.
+        """
+
+    def at_plugin_finalize(self):
```
Griatch, line 67: Not sure this needs to be separated from `at_plugin_collaborate` ... especially since it's a bit unclear that this is not called until after _all_ plugins have been initialized.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def at_server_cold_stop(self):
+        """
+        Runs in the reverse order of load_priority, but before the 'mygame' hooks.
+
+        see at_server_startstop module documentation.
+        """
+
+
+class PluginManager:
```
Griatch, line 115: All non-private methods in this handler require full docstrings, as per the Evennia style guide (also for the purpose of proper API docs)

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+    """
+
+    def __init__(self):
+        self.plugins = dict()
+        self.plugins_sorted = list()
+        self.reverse = None
+
+    def add(self, plugin):
+        self.plugins[plugin.name()] = plugin
+        self.plugins_sorted.append(plugin)
```
Griatch, line 127: This is not a sorted structure, so the name is a misnomer. Did you intend to run `.sort()` on this list when you call `.add` mayhaps?

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class PluginManager:
+    """
+    Basic class used for maintaining an index of Plugins installed.
+    """
+
+    def __init__(self):
+        self.plugins = dict()
+        self.plugins_sorted = list()
```
Griatch, line 122: This is fine, but a nitpick is that it's faster and more explicit to assign `{}` and `[]` here rather than calling the functions. Also, it looks like `.reverse = None` goes poorly with the othe defaults; it would lead to a traceback when no plugins were yet added.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+        self.reverse = reversed(self.plugins_sorted)
+
+    def __getitem__(self, item):
+        return self.plugins[item]
+
+    def init_settings(self, settings):
+        for plugin in self.plugins_sorted:
+            plugin.init_settings(settings, self.plugins)
+
+    def at_plugin_setup(self):
```
Griatch, line 140: All these `at_*` methods are mis-named and are confusing. The `at_*` name is used for overloadable hooks that are _called_ at certain times. This is not what happens here - this handler is not itself meant to be overridden and its methods are _calling_ methods (we trigger these at particular server times). So this should really be named `plugin_setup` or maybe `call_plugin_setup` to be very explicit.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+    def at_server_cold_stop(self):
+        for plugin in self.reverse:
+            plugin.at_server_cold_stop()
+
+
+def import_property(path):
+    """
+    Stripped-down version of evennia.utils.utils.class_from_module that gets along with
+    django.conf.settings without triggering its setup prematurely.
+
```
Griatch, line 181: This needs the correct docstring: 

```
Args:
    path (str): Hopefully, a ...
```

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+    else:
+        return importlib.import_module(path)
+
+
+_already_configured = False
+
+
+def plugin_settings(settings_path):
+    """
+    Run when doing any kind of Django management. Just loads basic plugin settings.
```
Griatch, line 199: Docstring argument types needed.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+    """
+    global _already_configured
```
Griatch, line 201: PEP8: This global should be in upper case `_ALREADY_CONFIGURED` and be defined at the _top_ of the module.

evennia/server/evennia_launcher.py
```diff
@@ -2278,8 +2281,11 @@ def main():
         # pass on to the core django manager - re-parse the entire input line
         # but keep 'evennia' as the name instead of django-admin. This is
         # an exit condition.
+
         sys.argv[0] = re.sub(r"(-script\.pyw?|\.exe)?$", "", sys.argv[0])
-        sys.exit(execute_from_command_line(sys.argv))
+        # sys.exit(execute_from_command_line(sys.argv))
+        django.core.management.call_command(*([option] + unknown_args))
```
Griatch, line 2287: This `execute_from_command_line()` call was added as a result of a lot of bug fixing. The `call_command` simply does not act quite the same way as the draw command line call ... The difference is subtle and I don't recall exactly why at this point but I suspect this change will have unexpected side effects/regressions.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+        thing = getattr(module, thing)
+        return thing
+    else:
+        return importlib.import_module(path)
+
+
+_already_configured = False
+
+
+def plugin_settings(settings_path):
```
Griatch, line 197: This name is not very useful. Better call it `init_plugin_settings` because that's what it does.

evennia/plugin.py
```diff
@@ -0,0 +1,217 @@
...
+
+    _plugins = dict()
+    import evennia
+
+    for proto_plugin in set(_settings.PLUGIN_PATHS):
+        evennia.PLUGINS.add(import_property(proto_plugin)())
+    evennia.PLUGINS.sort()
+    evennia.PLUGINS.init_settings(_settings)
+
+    settings.configure(default_settings=_settings)
```
Griatch, line 216: This is an important call and needs a comment, here or in the docstring.

evennia/server/server.py
```diff
@@ -492,6 +500,7 @@ def at_server_start(self):
         This is called every time the server starts up, regardless of
         how it was shut down.
         """
+        evennia.PLUGINS.at_server_start()
```
Griatch, line 503: Should this really run _before_ the server-level `at_server_start`? I'd imagine you'd want the plugin to modify/operate on/extend what the server itself does?

Griatch (changes requested):
I like this, it's a good level of change - not too invasive and adds functionality without messing with existing functionality. There are a few changes needed before this can be added though: 

1. Ccode-coments (see below)
2. Unit tests (it should be possible to test at least the Plugin-handler by mocking the Evennia startup process)
3. Documentation - the nice thing with having our new doc system is that you can write the to-do documentation at the same time as you write your code! You can just make a new `Plugins.md` in `evennia/docs/source/Components/`. You don't have to worry about setting up and building the docs if you don't want to though - if you add the text I can take on handling the formatting and building the docs as I merge.

evennia/plugin.py
volundmush: I originally had it inside of utils, but this caused some strange behavior due to evennia/utils/__init__.py doing some kind of import that appeared to interact badly with the evennia_launcher - side of things. Although, that might have been my earlier attempt using evennia.utils.utils.class_from_module which it turned out is not usable until Django has finished loading.

evennia/server/evennia_launcher.py
volundmush: The reason I changed it to use `call_command()` is because the plugins need to be able to modify the settings before Django begins its operations. if a whole new process is spawned to run the Django management then it will run using the settings specified in DJANGO_SETTINGS_MODULE environmental variable and will not include the settings from the plugins.

The only way around this that I'm aware of would be the previous method of plugins I used - having the `mygame/server/conf/settings.py` include some code which triggers the plugin initializing inline with importing the module. Perhaps at the very bottom.

Griatch:
@volundmush When you think this is ready for re-review, it helps to re-select me as reviewer and make a comment here explaining what you changed and what was addressed. 
Note to self - I should also fix the devenv CI builds so your new unit tests (when created) can be tested properly against the various other databases ...

evennia/server/plugins.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+        name = plugin.name
+        self.plugins[name] = plugin
+        self.plugin_versions[name] = EvPluginVersion(plugin.version)
+        self.plugin_requirements[name] = plugin.requirements
+
+    def __getitem__(self, item):
+        return self.plugins[item]
+
+    def dispatch(self, hook: str, *args, **kwargs):
```
Griatch, line 276: Suggestion - rename to `call_all_plugins`

evennia/server/plugins.py
```diff
@@ -0,0 +1,379 @@
...
+                # this is good.. we made progress!
+                remaining = new_remaining
+                if not remaining:
+                    # hooray! No more plugins to process
+                    break
+            else:
+                # this is bad. we are not making progress.
+                raise Exception("dependency load order is not progressing!")
+
+    def setup(self, settings_module):
```
Griatch, line 354: Maybe `initialize_from_settings` instead?

evennia/typeclasses/models.py
```diff
@@ -832,6 +832,16 @@ def at_rename(self, oldname, newname):
         """
         pass
 
+    @classmethod
+    def get_concrete_content_type(cls):
```
Griatch, line 836: The 'concrete' part of this name is confusing since it's not a term we've used anywhere in Evennia otherwise (it's a Django-thing). It may be clearer to name this `get_dbmodel_content_type`. 

Also, this is not used anywhere ... so if you want to add this (in this otherwise unrelated PR) you'll need to add a docstring explaining why this is useful.

Griatch (changes requested):
Thanks for the changes! I like some, disagree with others. I have a new slew of feedback here (in addition to the needed unit-tests and docs of course).

gtaylor:
As a less informed observer, this causes me to wonder whether we're recreating something like a Django app concept with a few added hooks.

Just to ask the question: do we need a separate notion of a plug-in? Will we cause confusion with the app vs plug-in distinction and their similar but different attributes?

volundmush:
> As a less informed observer, this causes me to wonder whether we're recreating something like a Django app concept with a few added hooks.
> 
> Just to ask the question: do we need a separate notion of a plug-in? Will we cause confusion with the app vs plug-in distinction and their similar but different attributes?

Well... unfortunately, as INSTALLED_APPS is processed long after settings are loaded by django, the Django AppConfig class is not capable of modifying the settings. So, it couldn't  add lockfuncs or add typeclass paths etc.

Griatch:
@volundmush Any progress on this?

volundmush:
Although there's some cool code in here, I think overall this is a bad idea and should be reconsidered from scratch, if at all.

PR #2280, opened 2020-12-09
Feature/add jupyter notebook support

duysqubix:
Adds support for jupyter notebook via `evennia shell_plus --notebook`

I like that you can use bpython or ipython interactively within your game, but it is frustrating at times
to have to redo everything from scratch to make sure your new feature or changed actually happened.

I like the idea of having a persistent code storage for testing and/sandbox environment, as well as 
more easily execute blocks of code.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,6 +12,7 @@ django-extensions >= 3.1.0
 inflect
 autobahn >= 17.9.3
 lunr == 0.5.6
+ipython >= 7.19.0
 
 # try to resolve dependency issue in py3.7
 attrs >= 19.2.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,12 +8,9 @@ pytz
 djangorestframework >= 3.10.3, < 3.12
 django-filter >= 2.2.0, < 2.3
 django-sekizai
-django-extensions >= 3.1.0
 inflect
 autobahn >= 17.9.3
 lunr == 0.5.6
-ipython >= 7.19.0
-jupyter == 1.0.0
 
 # try to resolve dependency issue in py3.7
 attrs >= 19.2.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,6 +13,7 @@ inflect
 autobahn >= 17.9.3
 lunr == 0.5.6
 ipython >= 7.19.0
+jupyter == 1.0.0
 
 # try to resolve dependency issue in py3.7
 attrs >= 19.2.0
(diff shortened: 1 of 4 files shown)


duysqubix:
hmm so besides the errors that sprung up from the checks...

It seems that `evennia shell_plus` works and you can interact with the database as well as the game..

but doing `evennia shell_plus --notebook` does start a jupyter instance, but it seems variables aren't initialized

duysqubix:
alright.. so still investigating this, but i got the notebook to initialize properly by

```python
import evennia
evennia._init()
```

I am not sure why it doesn't do this automatically. Since the 'shell_plus' is found [here](https://github.com/duysqubix/evennia/blob/92a659da6c1eedab98e74ddd48cca51acfca2e40/evennia/server/evennia_launcher.py#L2262). Which sets `check_db == True`

duysqubix:
Alright, so it seems that it does in fact call `evennia._init()` but the results of that doesn't get relayed to the notebook properly..

requirements.txt
```diff
@@ -8,9 +8,12 @@ pytz
 djangorestframework >= 3.10.3, < 3.12
 django-filter >= 2.2.0, < 2.3
 django-sekizai
+django-extensions >= 3.1.0
 inflect
 autobahn >= 17.9.3
 lunr == 0.5.6
+ipython >= 7.19.0
+jupyter == 1.0.0
```
Griatch, line 16: While I can see a point for having `ipython` in `requirements.txt`, the other two are too specialized. Move these to `requirements_extra.txt` instead - this is used for optional requirements like this.

Griatch (changes requested):
Looking good, I have one request for change in the requirements.

requirements.txt
duysqubix: Alright. I will move those requirements to the extra file. 

So this works perfectly on my end, in fact I am using the notebooks to fine-tune my leveling algorithm. 

A couple of caveats and would need to be documented (not sure where) is 1) that you have to run ‘evennia shell_plus —notebook’ in the root of your game folder.  

2) when you create a new notebook you must run

‘’’python
import evennia
evennia._init()
‘’’

Griatch: For now you can add your info as a new section below this one: https://github.com/evennia/evennia/blob/develop/docs/source/Coding/Coding-Introduction.md#explore-evennia-interactively . You don't have to set up/build the docs locally if you don't want to, just add the text to the file and I'll rebuild next time I update them.

duysqubix:
Made the requested changes.

PR #2291, opened 2020-12-25
Fix settings.START_LOCATION behavior

tim-chaplin:

Fixes START_LOCATION behavior for default account creation (ie from the connection screen). Current behavior is to start new characters in the Void, even when START_LOCATION is overridden in settings.

Bug fix.

### DIFF
--- a/evennia/accounts/accounts.py
+++ b/evennia/accounts/accounts.py
@@ -687,6 +687,7 @@ class DefaultAccount(AccountDB, metaclass=TypeclassBase):
             ip=character_ip,
             typeclass=character_typeclass,
             permissions=character_permissions,
+            location=ObjectDB.objects.get_id(settings.START_LOCATION),
             **kwargs,
         )
         if character:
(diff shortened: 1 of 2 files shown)


evennia/accounts/accounts.py
```diff
@@ -687,6 +687,7 @@ def create_character(self, *args, **kwargs):
             ip=character_ip,
             typeclass=character_typeclass,
             permissions=character_permissions,
+            location=ObjectDB.objects.get_id(settings.START_LOCATION),
```
Griatch, line 690: While the idea is good, this is an incorrect fix and cannot be merged as-is. The reason is that `**kwargs` may well contain the `location=<location>` keyword and that should override the default. 

The correct fix of this would be something like this before the object creation (e.g. on line 682):

```python

if 'location' not in kwargs:
    kwargs['location'] = ObjectDB.objects.get_id(settings.START_LOCATION)

```

Griatch (changes requested):
It's a nice idea, but this makes the `create_character` call less useful since you can no longer pass `location` to it as intended. See inline comment.

evennia/accounts/accounts.py
tim-chaplin: Thanks, I made your suggested change. At least, I think I did, I'm new to GitHub.

Griatch (approved):

PR #2292, opened 2020-12-27
Make more commands honor settings.COMMAND_DEFAULT_CLASS

tim-chaplin:

As suggested in mygame\commands\command.py, I'm overriding COMMAND_DEFAULT_CLASS in settings.py to try to make sweeping changes. Having certain commands inherit directly from the base Command class thwarts this.

The only difference between Command and MuxCommand is that MuxCommand defines some custom parsing, but that's overridden by these commands anyway, so there should be no practical difference.

The import change in game_template/commands/command.py was required to fix an exception on server reload caused by the channelhandler.py change (I guess something to do with import order?)

Makes it easier to edit gamewide look/feel, and brings codebase more in line with documentation that suggests all default commands inherit from COMMAND_DEFAULT_CLASS .

### DIFF
--- a/evennia/commands/default/help.py
+++ b/evennia/commands/default/help.py
@@ -25,7 +25,7 @@ _DEFAULT_WIDTH = settings.CLIENT_DEFAULT_WIDTH
 _SEP = "|C" + "-" * _DEFAULT_WIDTH + "|n"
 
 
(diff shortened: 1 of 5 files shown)


Griatch:
Makes sense. Thanks for the PR!

PR #2299, opened 2020-12-30
Fix test_gametime.py for non-UTC servers

tim-chaplin:

This fixes evennia.utils.tests.test_gametime; datetime.fromtimestamp assumes the current system's timezone, so some of the unit tests currently fail if the system's timezone is set to EST (on Windows).

Bug fix.

### DIFF
--- a/evennia/utils/gametime.py
+++ b/evennia/utils/gametime.py
@@ -8,7 +8,7 @@ total runtime of the server and the current uptime.
 
 import time
 from calendar import monthrange
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 from django.conf import settings
 from evennia import DefaultScript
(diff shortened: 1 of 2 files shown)


Griatch (changes requested):
Hm, this seems like an incorrect fix ... You are right in that the unittest should not fail in a non-UTC timezone, but that fix should happen in the unit test and not in the `gametime` module itself. 
So to be clear - the `gametime` module _should_ work with the system time. If the server wants to use UTC, that's up to the server. Instead the failing unit test should be updated to enforce UTC for this test only - for example using `@override_settings(TIME_ZONE='UTC')`

tim-chaplin:
You're right; however, overriding `settings.TIME_ZONE` has no effect since that setting isn't honored in `gametime`. I think doing so would involve some complicated glue code since as far as I can tell `settings.TIME_ZONE` is using SQL-compatible time zone strings, and Python has its own convention.

Instead I replaced the UTC timestamps in `test_gametime` with their equivalent in the local timezone.

Griatch (approved):
Ok, that should work.

Issue #2248, opened 2020-11-07
[Feature Request] Allow for non-interactive createsuperuser


srmeier:
#### Is your feature request related to a problem? Please describe.
When deploying Evennia in a test environment where the DB gets wiped on each deploy, it would be nice if I could specify the super user creds using environmental variables.

#### Describe the solution you'd like
This might be an anti-pattern, so I'm interested in other's opinion, but could we do something like the following?
```python
def create_superuser():
    """
    Create the superuser account

    """
    print(
        "\nCreate a superuser below. The superuser is Account #1, the 'owner' "
        "account of the server. Email is optional and can be empty.\n"
    )
    from os import environ
    if ("DJANGO_SUPERUSER_USERNAME" in environ) and ("DJANGO_SUPERUSER_EMAIL" in environ):
        username = environ["DJANGO_SUPERUSER_USERNAME"]
        email = environ["DJANGO_SUPERUSER_EMAIL"]
        django.core.management.call_command("createsuperuser", "--noinput",
                                            "--username=" + username,
                                            "--email=" + email, interactive=False)
    else:
        django.core.management.call_command("createsuperuser", interactive=True)

    if "DJANGO_SUPERUSER_PASSWORD" in environ:
        password = environ["DJANGO_SUPERUSER_PASSWORD"]
        from evennia.accounts.models import AccountDB
        u = AccountDB.objects.get(username=username)
        u.set_password(password)
        u.save()
```
Of course you could have the username without the email, as the email is optional. Let me know if this seems reasonable and can create a PR.

#### Describe alternatives you've considered
I could be missing something but the alternative seems to be to type in the information manually whenever setting up a new DB.

#### Additional context
N/A

jerryaldrichiii:
Hey! I had the same issue. I ended up doing something like this in my Dockerfile entrypoint.sh

```
EVENNIA_SUPERUSER_USERNAME=${EVENNIA_SUPERUSER_USERNAME:-"changeme"}
EVENNIA_SUPERUSER_EMAIL=${EVENNIA_SUPERUSER_EMAIL:-"nobody@nowhere.com"}
EVENNIA_SUPERUSER_PASSWORD=${EVENNIA_SUPERUSER_PASSWORD:-"changemeorbehacked"}
superuser_script="
from django.contrib.auth import get_user_model;

User = get_user_model()

superuser_username='$EVENNIA_SUPERUSER_USERNAME'
[...]

PR #2306, opened 2021-01-16
Make use of environmental variables for superuser creation

srmeier:
When superuser is created pull details from environmental variables so that no interaction is required. If environmental variables are not present then fallback to typical behavior.

Very helpful when launching via Docker (see feature request #2248).

I am not sure to what extent we would like this to mimic Django's support (see [here](https://docs.djangoproject.com/en/3.1/ref/django-admin/#createsuperuser)). The approach I have proposed here looks for both a username AND email address first, and then will check to see if an environmental variable for the password is present. Django's `createsuperuser` command does not allow for setting the password programmatically so I am leveraging `evennia.accounts.models.AccountDB` to perform that operation.

### DIFF
--- a/evennia/server/evennia_launcher.py
+++ b/evennia/server/evennia_launcher.py
@@ -1423,7 +1423,21 @@ def create_superuser():
         "\nCreate a superuser below. The superuser is Account #1, the 'owner' "
         "account of the server. Email is optional and can be empty.\n"
     )
-    django.core.management.call_command("createsuperuser", interactive=True)
+
+    from os import environ
+    if ("DJANGO_SUPERUSER_USERNAME" in environ) and ("DJANGO_SUPERUSER_EMAIL" in environ):
+        username, email = environ["DJANGO_SUPERUSER_USERNAME"], environ["DJANGO_SUPERUSER_EMAIL"]
+        django.core.management.call_command("createsuperuser", "--noinput",
+                                            "--username=" + username,
+                                            "--email=" + email, interactive=False)
+        if "DJANGO_SUPERUSER_PASSWORD" in environ:
+            password = environ["DJANGO_SUPERUSER_PASSWORD"]
+            from evennia.accounts.models import AccountDB
+            u = AccountDB.objects.get(username=username)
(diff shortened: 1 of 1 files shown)


Griatch (changes requested):
This is a fine idea, but reviewing and testing this I find a few points that need improvement: 

- Might as well name these envvars `EVENNIA_SUPERUSER_*` instead. While it's true that they are used by Django under the hood, django knowledge is not strictly needed (so it's just confusing) and they are only ever useful for the `evennia` launcher anyway.
- The `..SUPERUSER_EMAIL` should not be a required envvar. If not given, this should be assumed to be left blank. This reflects how you can just press `<return>` to bypass this input when doing the interactive setup. 
- Trying to not specify the email by giving `..SUPERUSER_EMAIL=` or `=""` currently gives an error: `django.core.management.base.CommandError: You must use --email with --noinput.`

srmeier:
Sounds good 👍 , Added a few changes to address the points above. Currently is the password required?

Griatch:
> Sounds good , Added a few changes to address the points above. Currently is the password required?

Haven't tested this yet, but to answer your question - the superuser password should not be allowed to be empty. So if this is missing, this should be handled gracefully rather than triggering a traceback.

srmeier:
> Haven't tested this yet, but to answer your question - the superuser password should not be allowed to be empty. So if this is missing, this should be handled gracefully rather than triggering a traceback.

Thanks, I have added a check for the password. If either the username or the password environmental variables are missing then current logic is to fallback to the standard process of interactively requesting that the user enter them.

`os.environ.get('EVENNIA_SUPERUSER_EMAIL')` Will check for the email environmental variable and if it is missing it will pass `None` as the email arguement to the `AccountDB.objects.create_superuser()` function (which is the default for that function anyways).

I wouldn't expect anything here to trigger a stack trace

srmeier:
Made the code a little more clear and I am also now explicitly checking to make sure the length of the password is greater than 0.

Griatch (approved):
Thanks for the changes!

PR #2326, opened 2021-02-28
CommandTest Update

davewiththenicehat:
CommandTest.call, removed unused code that appeared to have been the start of regex support that was not used. Makes method easier to understand. Has a very minor positive impact to performance.

CommandTest, created method CommandTest.call_multi_receivers. Allows for testing multiple messages to multiple receivers. Supports sending None or '' message. Broader overview in the method's doc string. Added a unit tests for this in evennia.commands.default.tests.TestCallMultiReceivers. Tests multiple receives, and that the proper string is returned from the method.
Very minor but I did comment the process. Hopefully making the unit tests easier for other people to understand.

Added support for users using Atom and remote sync. Atom saves a local config file that stores a clear text password. Added that config file to .gitignore. This is a security and ease of use update.

Make CommandTest.call easier for more people to understand.

Add support for making a single unit test command call that tests what messages multiple Objects should receive when the command runs.

Ease of use and security improvement for Atom IDE uesrs

Each step has been tested using a basic evennia game directory unit tests. All tests passed each time.
All github checks passed.


Griatch:
@davewiththenicehat  Thanks for the PR! I wrote these testing tools for internal Evennia use but people have found good use of them since; you make a good point that these should be made more accessible. For one thing I need to have a proper doc string for `.call` :sweat_smile: 

However, while the idea of `call_multi_receivers` is good, this is not the way to implement this. There is no reason to have a duplicate method that does everything exactly the same way that normal `call` does, but with a twist. There's a lot of identical code here which would make this hard to maintain.

Since I had to do some needed cleanup of `.call` anyway, I ended up adopting the concept of your code but refactored `call_multi_receivers` into `call` instead - by allowing `msg` to accept _both_ a string and a dict and switching to your solution in the latter case.

Since this is new functionality, it belongs in `develop` branch, not `master`. While I'd have wanted to retain your PR history, there are so many differences between the branches that rebasing caused a lot of side effects - in the end I had to copy things between the branches and refactor it there: 6a9ed96a0

Closing this PR now. Thanks again!

PR #2388, opened 2021-04-20
Don't allow fuzzy match on db if exact match on module prototype

blongden:

If you have an exact match on a module prototype, the current logic will go on to fuzzy match on database prototypes and potentially match things from the db in addition to the exact match from the module prototype.

The existing behaviour results in many matches for a spawn when there is an exact match available.


blongden:
Testing a comment to see if it's clever enough to show the linked user on the discord side...

PR #2399, opened 2021-04-29
Pass **kwargs to various hooks on move_to

blongden:

There seems to be no reason NOT to pass kwargs passed into move_to to the hooks underneath (I actually had this overridden solely for this purpose, locally).

Makes move_to more flexible, and I suspect it should work this way anyway.


Griatch (approved):
Makes sense. It's something I planned to do for all sorts of methods down the line, so this is one spot I don't need to do. It helps, thanks!

PR #2412, opened 2021-05-10
delay docstring update

davewiththenicehat:
delay docstring update

The old one had not been properly updated or formatted.

An update to utils.delay's docstring

I had it formatted incorrectly. Also it's return message was incorrect.

### DIFF
--- a/evennia/utils/utils.py
+++ b/evennia/utils/utils.py
@@ -1021,44 +1021,34 @@ _TASK_HANDLER = None
 
 
 def delay(timedelay, callback, *args, **kwargs):
-    """
-    Delay the calling of a callback (function).
+    """Delay the calling of a callback (function).
+
+    The task handler (`evennia.scripts.taskhandler.TaskHandler`) will
+    be called for persistent or non-persistent tasks.
+
+    If persistent is set to True.
+        The callback, its arguments and other keyword arguments will be saved
+        (serialized) in the database, assuming they can be. The callback will
+        be executed even after a server restart/reload, taking into account the
+        specified delay (and server down time).
+        Keep in mind that persistent tasks arguments and callback should not
+        use memory references.
+        All task's whose time delays have passed will be called on server startup.
 
     Args:
         timedelay (int or float): The delay in seconds
         callback (callable): Will be called as `callback(*args, **kwargs)`
             after `timedelay` seconds.
         args (any, optional): Will be used as arguments to callback
+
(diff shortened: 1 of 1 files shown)


evennia/utils/utils.py
```diff
@@ -1021,44 +1021,34 @@ def uses_database(name="sqlite3"):
...
+    be called for persistent or non-persistent tasks.
+
+    If persistent is set to True.
+        The callback, its arguments and other keyword arguments will be saved
+        (serialized) in the database, assuming they can be. The callback will
+        be executed even after a server restart/reload, taking into account the
+        specified delay (and server down time).
+        Keep in mind that persistent tasks arguments and callback should not
+        use memory references.
+        All task's whose time delays have passed will be called on server startup.
```
Griatch, line 1036: This sounds like it belongs partly in the `persistent` arg description and in a `Notes` section.

evennia/utils/utils.py
```diff
@@ -1021,44 +1021,34 @@ def uses_database(name="sqlite3"):
...
             after `timedelay` seconds.
         args (any, optional): Will be used as arguments to callback
+
     Keyword Args:
-        persistent (bool, optional): Should make the delay persistent
-            over a reboot or reload. Defaults to False.
-        any (any): Will be used as keyword arguments to callback.
+        persistent (bool, optional): should make the delay persistent
+            over a reboot or reload
+        any (any, optional): Will be used as keyword arguments to callback.
```
Griatch, line 1047: It's actually OK to just put `any:` I think.

Griatch (changes requested):
Not sure why you moved this text up from the `Note` to the top of the function?

davewiththenicehat:
What tool is being used to generate the documentation?
I think I got the kwargs section incorrect also.

I'm happy to read up on it. I don't know where to look this up. The google style docstring documentation I was able to find did not show 'Notes' as a header type so I removed it.

Griatch:
We use sphinx with the Napoleon parser. Easiest is probably to look at existing Evennia docstrings, but the Napoleon parser is documented here: https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html

Griatch:
As far as I can tell the existing doc string was fine, so closing this.

davewiththenicehat:
delay does not return a deferred or int. It returns a TaskHandlerTask instance.

Issue #2031, opened 2020-01-22
[Feature Request] Move text2html into client layer


Griatch:
#### Is your feature request related to a problem? Please describe.

Today the output to the webclient passes through the server-side `evennia.utils.utils.text2html` module in order to convert it it to valid HTML for display in the webclient. There are notable flexibility issues with this, since it does not allow the webclient as much flexibility in displaying its output as it pleases. There are also at least one security issues with MXP links that is due to this server-side conversion to HTML. 

#### Describe the solution you'd like

The functionality of the `text2html` module should be moved into webclient js library for use by the web client. By default this should work the same as now, mimicking exactly the output and look created by telnet and other protocols. 

#### Additional context

The `text2html` conversion is today entirely driven by regexes. This has turned out to become gradually more and more complex - and even so, not all possible tag combinations cannot be properly handled. It's possible that some other algorithm would end up being more efficient to handle all possible edge cases. 

At any rate, implementing an equivalent in javascript is likely a considerable undertaking.

Griatch:
This is paused until 1.0 release.

PR #2419, opened 2021-05-21
WIP - Clientside rendering

friarzen:

This offloads the server overhead of converting/handling ANSI to HTML to the webclient 

Having the server hard-code the HTML for the webclient limits some of the dynamic behavior that the client could otherwise do.
It also means less work for the server to have to do for all clients, when they can simply do it themselves.
Exotic clients become possible, when they don't have to render/interpret HTML in their input.

This is still a work in progress


Griatch:
@friarzen Any status updates on this?

davewiththenicehat:
If it is worth anything. This is a clever idea. I would love to see it completed. If that helps for motivation.

friarzen:
As far as I am aware, this branch is functionally complete other than figuring out client-side MXP link parsing.
The only reason I have this still marked as WIP is there are a number of features on the Evennia side that need updating for this to work as expected.  EvMenu messages, EvTable messages, etc that seem to be sending cooked telnet instead of raw pipecodes.

Griatch:
@friarzen Thanks - yes I remember the discussion now, but since it's not written down anywhere in the issue, it's easy to forget the details. :sweat_smile:

Griatch:
@friarzen I have looked into this. The characters you see in e.g. the `who` output is ANSI escape sequences. This is because who uses `ANSITable` which in turn runs its output through `ANSIString`. I think in order to properly support in-browser rendering like this, we would need to move all markup parsing off `server` and into the `portal/protocol` level. \

It does makes sense to do this, because ANSI is a pure telnet markup that has nothing to do with the webclient. The issue is that ANSIString is used on the server side as a way to conveniently build tables and other structures where you want to measure the visible length (skipping the color markups for `len()` calculations). So to properly do this one would need to 
 - Rework the ANSIParser to be more generic (and ANSI-telnet agnostic) 
 - Rework ANSIString to work differently as well - maybe only have it work on the `|r` etc markup directly rather than converting to ANSI
 - Move all ANSI conversion into the `Telnet` protocol layer. 

So a lot of things to do. We'll see if it's something I'll go forward with before 1.0, it sounds like a big project. :sweat_drops:

mfreidel:
I'd be willing to help however I can. I'm new to contributing (generally) and still getting up to speed on the inner workings of Evennia, but I'll keep an eye on what's going on with this and jump in when I see stuff I know I can do.

Griatch:
@mfreidel 

> I'd be willing to help however I can. I'm new to contributing (generally) and still getting up to speed on the inner workings of Evennia, but I'll keep an eye on what's going on with this and jump in when I see stuff I know I can do.

Thanks! We'll see if there is much help to give here, because this requires quite a bit of restructuring of the code which likely will need to be done all at once or nothing works. :sweat_drops:

Griatch:
@friarzen - I'm closing the `develop` branch, meaning this PR will close as well. We'll still need this, but it's pending resolution of https://github.com/evennia/evennia/issues/2031 before it can be used. Feel free to reopen against the `main` branch.

PR #2422, opened 2021-05-24
Fix priv check when displaying names in CmdWho

phantummm:
- In `who` command, object IDs were showing up next to user names, even when the caller doesn't have dev/admin privileges.
- This is because a local variable was being re-used inside the scope of a loop, causing the access check to be run against the account being looked at, rather than against the account issuing the command.

Example, from a non-dev account:

image

And after fix:

image

And proof, after fix, that privileged view still works:
image


Griatch:
Makes sense. Merged! Thanks for the PR!

Issue #2417, opened 2021-05-18
Task Management Command


davewiththenicehat:
#### Is your feature request related to a problem? Please describe.
No, this is mostly to provide developer's a method to look for rogue tasks.

#### Describe the solution you'd like
A command to show and manage tasks (delays) deferred.
I'd like to make it into a clickable table. Due to the volatile nature of delays, command switches may not be fast enough for managing short tasks.
This will not remove the need for switches, as I will still need to use them in the linking text tag.

I am going to:
use format `tasks/[action] [task_id]`. Where the switches were methods on the task handler.
make this an developer security command. As I could see both developers and admins wanting to use it.
put it in the admin help category.

#### Describe alternatives you've considered
A management window within the web app. I decided against this due to the short nature of a standard task.

#### Additional context
I am going to make this. I figured before I did I would see if there was interest for others to have this.
Than if there is interest, to make it in such a way that everyone is happy with it.

Griatch:
This could be an interesting thing for `develop` branch, sure. I'm not sure how often it would be used, but having a way to get an overview of delays may be instructive. The `script/scripts` command will also get some refactoring but this is a separate functionality.

Note that `delay()` is used in many situations, including in things like the `@interactive` decorator and for `yield` operations inside the `Command.func` method. So there could be delays from a lot of sources.

davewiththenicehat:
I am working off of `master` when I PR to `develop` is it likely to cause merge errors?
If I can I would like to avoid pulling develop.

I doubt the tasks command is used often. I see developers using it to look at tasks when they something occurring without manual calls. It will probably be used mostly just for peace of mind.

Griatch:
Generally, you always should work off the branch you want to merge with, since you want to be able to test properly. So I recommend rebasing your work off develop branch right away. As [per the git recommendations](https://www.evennia.com/docs/latest/Version-Control.html?highlight=git#committing-fixes-to-evennia) I also strongly suggest you make a new branch of develop and work on that rather than working on your local copy of the develop (or master) branch - that's just inviting merge conflicts. Best to keep your local version of master/develop as 'pristine' 1:1 to upstream as possible.

davewiththenicehat:
PR https://github.com/evennia/evennia/pull/2425 is merged, closing issue.

PR #2425, opened 2021-05-28
CmdTasks Created

davewiththenicehat:
Allows command line manipulation of evennia tasks (delays).

https://github.com/evennia/evennia/issues/2417

A single unit test in evennia.commands.default.tests.TestHelp fails. Message pattern mismatch.
It may just be something lingering in my evennia migrations folder.
All other unit tests pass using default run level. Tasks should not be affected by run level.

### DIFF
--- a/evennia/commands/default/system.py
+++ b/evennia/commands/default/system.py
@@ -24,6 +24,7 @@ from evennia.utils import logger, utils, gametime, create, search
 from evennia.utils.eveditor import EvEditor
 from evennia.utils.evtable import EvTable
 from evennia.utils.evmore import EvMore
+from evennia.utils.evmenu import ask_yes_no
 from evennia.utils.utils import crop, class_from_module
 from evennia.scripts.taskhandler import TaskHandlerTask
 
@@ -1217,7 +1218,7 @@ class CmdTasks(COMMAND_DEFAULT_CLASS):
         tasks[/switch] [function name]
           Process the action in the switch to all tasks that are deferring a specific function name.
           This would match the name of the callback you passed to a delay or the task handler.
-        tasks[/switch] [task id], [completion date], [function memory reference]
+        tasks[/switch] [task id]
           Process the action in the switch to a specific task.
 
     Switches:
@@ -1237,17 +1238,16 @@ class CmdTasks(COMMAND_DEFAULT_CLASS):
         Only the action requested on the first switch will be processed. All other switches will
         be ignored.
 
-        Manipulation of a single task is intended to be done via the clickable links or through
-        code directly. Due to generally short life of a task, the inclusion of completion date
-        and function's memory reference guarentees an incorrect task will not be manipulated.
-
         By default, tasks that are canceled and never called are automatically removed after
         one minute.
 
     Example:
         tasks/cancel move_callback
-          Cancel all movement delays from the slow_exit contrib.
-          In this example slow exits creates it's tasks with: utils.delay(move_delay, move_callback)
+            Cancel all movement delays from the slow_exit contrib.
+            In this example slow exits creates it's tasks with:
+                utils.delay(move_delay, move_callback)
+        tasks/cancel 2
+            Cancel task id 2.
 
     """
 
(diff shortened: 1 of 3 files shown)


davewiththenicehat:
@Griatch I am done with this. Let me know if anything else is needed.

evennia/commands/default/system.py
```diff
@@ -1203,3 +1205,183 @@ def func(self):
...
+      one minute.
+
+    Usage:
+      tasks
+        show all current active taks.
+      tasks[/switch] [function name]
+        Process the action in the switch to all tasks that are deferring a specific function name.
+        This would match the name of the callback you passed to a delay or the task handler.
+      tasks[/switch] [task id], [completion date], [function memory reference]
+        Process the action in the switch to a specific task.
```
Griatch, line 1232: This is also the in-game help for this command, so look at how these sections are ordered for other default commands:

```
Short intro

Usage:
    tasks[/switch] [task_id][, completion_date[, functionref]

Switches:

Notes:

Example:

```

evennia/commands/default/system.py
```diff
@@ -1203,3 +1205,183 @@ def func(self):
                 "*" if sub[5] else "-",
             )
         self.caller.msg("|wActive tickers|n:\n" + str(table))
+
+
+class CmdTasks(COMMAND_DEFAULT_CLASS):
+    """
+    Manage active tasks (delays).
+
+    Note:
```
Griatch, line 1214: This note should mention what a 'task' is and that it can only be started from code. Other commands like this have in-game creation equivalents (`scripts` have `script`, `objects` have `create/dig` etc), so reading this one would expect the same for this when reading the help string.

Griatch (changes requested):
@davewiththenicehat Looks like a good implementation at first glance. I added some comments but didn't review in detail yet since I ran into a traceback when trying to run it:

I made a module `testcallable.py` in my `mygame/` folder with a simple `callable` function inside:
```
# in mygame/testcallable.py

def callable(caller):
    caller.msg("Callable called!")
```
 I then ran the following: 
 
     py import testcallable;evennia.TASK_HANDLER.add(10, testcallable.callable, self)

Upon running `tasks` I now get the following traceback:
```
  File "/home/griatch/Devel/Home/evennia/evennia/evennia/commands/cmdhandler.py", line 605, in _run_command
    ret = cmd.func()
  File "/home/griatch/Devel/Home/evennia/evennia/evennia/commands/default/system.py", line 1383, in func
    if width >= 75:
TypeError: '>=' not supported between instances of 'dict' and 'int'
```
Looks like you are extracting the client width from the client flags and not getting back what you expect (it's a screen-height-width dict). It's recommended you use the in-built `Command.client_width()` method instead.

A minor thing I can fix, but your code is wider than 100 characters in places, which is what Evennia (tries to) adhere to.

evennia/commands/default/system.py
davewiththenicehat: I do get a cyclomatic warning on func. I did put in line breaks at major logic changes.
I understand the width logic now. Did replace collection of screen width with client_width.
I changed my linter width warning to 100 and resolve the items that were longer than 100.
I am going to continue forward and commit once completed with the remaining requests.

evennia/commands/default/system.py
davewiththenicehat: Put in a reference in notes and make the header sentence more clear.

I don't see task creation at command line as a value add. Admittingly, I see script creation at command line as a nightmare. So I don't have a very robust grasp of how other people like to manage an environment.

Griatch (changes requested):
The initial crash was fixed  - thanks! This is a good functionality, but while the idea behind the clickable links are good, this implementation won't fly for a default command, unfortunately. 

1. MXP links are not supported by all clients, for example not by tintin++. They can at best be an extra usability bonus, not a fundamental part of the command functionality. Having the 'clickable actions' is a neat concept (it works fine in webclient if using a lot of space) but sorry, they'll have to go.
2. The help string is confusing since it actively mentions links that do not work universally.
3. The `readthedocs` is not the latest version of the docs, it's https://www.evennia.com/docs/latest/Command-Duration.html  (I should really take down the readmedocs version)
4. Using the plain command `tasks/pause <task-id>` seems like the natural way to use this command (first look at the list, then apply the command to the correct line) - but this does not work. I realize there are timing issues if tasks finish and the list changes. So I would suggest that when doing `tasks/switch <identifier>`, the command loads the task/memory-id etc under the hood and presents the user with a yes/no verification question, showing all the info for the task to make sure the user is manipulating the correct one. You can either use `yields` or the [yes_no_question]( https://github.com/evennia/evennia/blob/develop/evennia/utils/evmenu.py#L1736) helper introduced in `develop` branch.

davewiththenicehat:
@Griatch 
Clickable text tags were removed.
A yes no prompt was added. Includes verification that the task was not called while waiting for the yes input. Includes verification that the task was not called than a new one with the same task id was created, while waiting for the yes input.
In your previous comment, I did not understand item 3. Did I happen to fix what you were mentioning when I updated the docstring?

Example of commanding being used:
image

Griatch:
Looking good! I made a few further tweaks to the wording and style of help to match the style elsewhere, but nice work on this!

PR #2429, opened 2021-06-04
file and command help web support, proof of concept

davewiththenicehat:
Adds command help entries (command doc strings) and web.

Discussion: https://github.com/evennia/evennia/discussions/2426

Issues found as of yet:

File and command entries only
gray area above entry text does not show proper "Help Index / Category / Topic" format. Probably something that needs to be added to the context dictionary in get_contect_data.
entries need to convert from ascii to html. Or likely just drop ascii text tags all together.
In detail view category name is not showing above "context["topic_list"]", the gray list of other entries in the current category. Probably something that needs to the context dictionary in get_contect_data.

command entries only
The general category is not sorting with the other categories. Probably just case sensitivity?

Loads of log_infos throughout that need to be removed.
Better unit test HelpDetailTest in web/website/tests. It only tests url_name 'help-entry-detail'. Does not matter what the arguments are. It only capable of catching tracebacks or other code breaking errors.
locks need to be verified working.
HelpMixIn.get_queryset needs to support 'AnonymousUser'.
HelpMixIn.get_queryset needs to get commands from all of the user's Characters.
web.website.veiw.help.collect_topics I don't know enough to know if I should be looking for further command sets on the Character.

### DIFF
--- a/evennia/help/filehelp.py
+++ b/evennia/help/filehelp.py
@@ -244,5 +244,3 @@ class FileHelpStorageHandler:
 
 # singleton to hold the loaded help entries
 FILE_HELP_ENTRIES = FileHelpStorageHandler()
-# Used by Django Sites/Admin
-#get_absolute_url = web_get_detail_url
--- a/evennia/help/filehelp.py
+++ b/evennia/help/filehelp.py
@@ -152,6 +152,18 @@ class FileHelpEntry:
         except Exception:
             return "#"
 
+    def web_get_admin_url(self):
+        """
+        Returns the URI path for the Django Admin page for this object.
+
+        ex. Account#1 = '/admin/accounts/accountdb/1/change/'
+
+        Returns:
+            path (str): URI path to Django Admin page for object.
+
+        """
+        return False
+
     def access(self, accessing_obj, access_type="view", default=True):
         """
         Determines if another object has permission to access this help entry.
--- a/evennia/help/filehelp.py
+++ b/evennia/help/filehelp.py
@@ -67,11 +67,14 @@ An example of the contents of a module:
 
 from dataclasses import dataclass
 from django.conf import settings
+from django.urls import reverse
+from django.utils.text import slugify
 from evennia.utils.utils import (
     variable_from_module, make_iter, all_from_module)
 from evennia.utils import logger
 from evennia.utils.utils import lazy_property
 from evennia.locks.lockhandler import LockHandler
+from evennia.utils.logger import log_info
 
 _DEFAULT_HELP_CATEGORY = settings.DEFAULT_HELP_CATEGORY
 
@@ -114,6 +117,43 @@ class FileHelpEntry:
     @lazy_property
     def locks(self):
         return LockHandler(self)
+
+    def web_get_detail_url(self):
+        """
+        Returns the URI path for a View that allows users to view details for
+        this object.
+
+        ex. Oscar (Character) = '/characters/oscar/1/'
+
+        For this to work, the developer must have defined a named view somewhere
+        in urls.py that follows the format 'modelname-action', so in this case
+        a named view of 'character-detail' would be referenced by this method.
+
+        ex.
+        ::
+            url(r'characters/(?P<slug>[\w\d\-]+)/(?P<pk>[0-9]+)/$',
+                CharDetailView.as_view(), name='character-detail')
+
+        If no View has been created and defined in urls.py, returns an
+        HTML anchor.
+
+        This method is naive and simply returns a path. Securing access to
+        the actual view and limiting who can view this object is the developer's
+        responsibility.
+
+        Returns:
+            path (str): URI path to object detail page, if defined.
+
+        """
+        # log_info('filehelp web_get_detail_url start')
+        try:
+            return reverse(
+                'help-entry-detail',
+                kwargs={"category": slugify(self.help_category), "topic": slugify(self.key)},
+            )
+        except Exception as e:
+            log_info(f'Exception: {getattr(e, "message", repr(e))}')
+            return "#"
 
     def access(self, accessing_obj, access_type="view", default=True):
         """
@@ -207,3 +247,5 @@ class FileHelpStorageHandler:
 
 # singleton to hold the loaded help entries
 FILE_HELP_ENTRIES = FileHelpStorageHandler()
+# Used by Django Sites/Admin
+#get_absolute_url = web_get_detail_url
--- a/evennia/help/filehelp.py
+++ b/evennia/help/filehelp.py
@@ -74,7 +74,6 @@ from evennia.utils.utils import (
 from evennia.utils import logger
 from evennia.utils.utils import lazy_property
 from evennia.locks.lockhandler import LockHandler
-from evennia.utils.logger import log_info
 
 _DEFAULT_HELP_CATEGORY = settings.DEFAULT_HELP_CATEGORY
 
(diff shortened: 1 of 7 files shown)


davewiththenicehat:
@Griatch was able to get file and command help entries to the website.

It is ready for review.
I created what unit tests I could think to create. All web unit test pass for me.

Should commands be grouped separately from file and db help entries? So the list view (index page) had some kind of header separating commands from "Game & World" the way CmdHelp does.
Are there other upgrades that can be done that I am not thing of?

Example image of HelpListView:
image

Example image of command HelpDetailView:
image

Example image of a file HelpDetailView:
image

Example image of a db HelpDetailView:
image

evennia/help/filehelp.py
```diff
@@ -207,3 +244,5 @@ def all(self, return_dict=False):
 
 # singleton to hold the loaded help entries
 FILE_HELP_ENTRIES = FileHelpStorageHandler()
+# Used by Django Sites/Admin
+#get_absolute_url = web_get_detail_url
```
Griatch, line 248: Commented out - should be removed?

evennia/help/models.py
```diff
@@ -56,13 +56,23 @@ class HelpEntry(SharedMemoryModel):
     db_key = models.CharField(
         "help key", max_length=255, unique=True, help_text="key to search for"
     )
+
+    @lazy_property
+    def key(self):
+        return self.db_key
```
Griatch, line 62: No need for these properties to be lazy. Not sure it even works that well her - lazy properties are for handlers that initialize themselves.

evennia/help/models.py
```diff
@@ -56,13 +56,23 @@ class HelpEntry(SharedMemoryModel):
...
+
     # help category
     db_help_category = models.CharField(
         "help category",
         max_length=255,
         default="General",
         help_text="organizes help entries in lists",
     )
+
+    @lazy_property
```
Griatch, line 72: Same here, let this be a regular `@property`.

evennia/help/models.py
```diff
@@ -221,11 +231,14 @@ def web_get_detail_url(self):
...
 
         """
+
         try:
-            return reverse(
+            url = reverse(
                 "%s-detail" % slugify(self._meta.verbose_name),
                 kwargs={"category": slugify(self.db_help_category), "topic": slugify(self.db_key)},
             )
+            # log_info(f'HelpEntry web_get_detail_url url: {url}')
```
Griatch, line 240: leftover debug

evennia/web/website/views/help.py
```diff
@@ -1,17 +1,149 @@
...
+                        continue
+                    # skip the command if the puppet does not have read access
+                    if not can_read_topic(cmd, puppet):
+                        continue
+                    # skip the command if it's help entry already exists in the topic list
+                    entry_exists = False
+                    for verify_cmd in cmd_help_topics:
+                        if verify_cmd.key and cmd.key and \
+                           verify_cmd.help_category == cmd.help_category and \
+                           verify_cmd.__doc__ == cmd.__doc__:
```
Griatch, line 114: Minor style comment - you can just put the whole thing in parentheses to break lines. The `\` to break lines is very rarely needed in Python. 

```python
if (verify_cmd.key and cmd.key
        and verify_cmd.help_category == cmd.help_category
        and verify_cmd.__doc__ == cmd.__doc__):
    entry_exists = True 
    break

File: evennia/web/website/views/help.py [hunk 6]
```diff
@@ -88,19 +224,23 @@ def get_context_data(self, **kwargs):
...
-            .order_by(Lower("db_key"))
-        )
-        context["topic_list"] = queryset
-
-        # Find the index position of the given obj in the queryset
-        objs = list(queryset)
+        # Get the object in question
+        obj = self.get_object(full_set)
+
+        # filter non related caegories from the query set
```
Griatch, line 233: typo

Griatch (changes requested):
Great job on this! This seems to be working fine for the most part. I found one critical regression in the `sethelp` command though: 

1. Create a new db-based help entry in a new category with `sethelp bigtest,modifiable = A test help topic`
2. Refresh the web page and the new category + name pops up - great! Clicking on the `Admin` button does nothing though, even though it's a db-based help entry that should be possible to manipulate in the admin.
3. Reboot the server. The entry is now gone, but adding it again errors out with an error that it already exists (same if going back to `develop branch`.  
4. Trying to create any other entry with `sethelp` (like wjfwefjwpoierjwoei) will now fail with a message in log that `the key already exists`. Same thing works if switching back to `develop` branch, so this appears to be a regression.

Answers to some other questions: 
- I think it's fine to have commands/help entries grouped together here. I separate them in the in-game help, not sure if that's actually needed. 
- I think help entries that are actually modifiable from admin (and if you have the `is_staff` flag and can access the admin in the first place) should be marked in the list somehow, maybe as `(editable)` or somesuch. As it is, there is no way to know which entry can be modified - all you can do is to click the `Admin` button and see if anything happens (that button should probably be removed for file/command-help).

evennia/help/filehelp.py
davewiththenicehat: Was not needed has been removed.

evennia/help/models.py
davewiththenicehat: When I change this to a standard python property the database entries fail to create.
Keeping notes here, will circle back to the issue when I am done with issues with the `sethelp` command entries. Possibly related.

60 - 62 changes to:
```python
@property
    def key(self):
        return self.db_key
```
Causes evennia.web.website.tests.HelpDetailTest to fail with:
image

evennia/help/models.py
davewiththenicehat: Done, changing this property does not cause the unit test to fail.

evennia/help/models.py
davewiththenicehat: Removed

evennia/web/website/views/help.py
davewiththenicehat: Updated.

evennia/web/website/views/help.py
davewiththenicehat: done

evennia/help/models.py
davewiththenicehat: Note changing evennia.help.models.help_category from a lazy_property descriptor to a standard property descriptor made the evennia.web.website.tests.HelpLockedDetailTest fail.
Taking note will resolve later.

davewiththenicehat: Resolved with commit: https://github.com/evennia/evennia/pull/2429/commits/d2f410ee26458d431c6ebb575e653d5791a60d94.

Created entry restarted, created another entry restart:
image

Both entries show up in help list view (side note they both display properly in detail view.):
image

Both entries show up in the admin page:
image

I was also able to delete them without issue:
image

davewiththenicehat:
> * I think help entries that are actually modifiable from admin (and if you have the `is_staff` flag and can access the admin in the first place) should be marked in the list somehow, maybe as `(editable)` or somesuch. As it is, there is no way to know which entry can be modified - all you can do is to click the `Admin` button and see if anything happens (that button should probably be removed for file/command-help).

@Griatch 
Like this?
tempsnip

Where `Test Key` is the only database entry in this list.
Than the `edit` hyperlinks to domain_name/admin/help/helpentry/entry.id/change/

Griatch:
@davewiththenicehat Yeah, something like that. I was just thinking of having an `(editable)` marker to let the user know they can click on the `Admin` button when they go to it, but having it an edit-link directly is an even better idea.

davewiththenicehat:
@Griatch 

Created a `-edit-` link that appears to the right of a help entry that has the `web_get_admin_url` and it returns truthy.
The link goes to the object's admin url.
If the user is not `staff` the edit link will not appear.

I am doing this by using `if object.web_get_admin_url` in `evennia.web.templates.website.help_list.html`
Is there any concern about tracebacks occurring somewhere in the background? Should I make a `web_get_admin_url` method on file and command help entries that returns False?
image


Logged in as staff a db entry appears like this in list view:
image

If I am not logged in as staff a db entry appears like this:
image

Clicking `-edit-` takes the user to the admin entry for that help topic:
image

Griatch:
@davewiththenicehat Sorry for the slow follow-up on this. The contrib work in develop took longer than expected. I'll get down to re-review and hopefully merge this soon.

PR #2433, opened 2021-06-12
Cmd help clickable topics

davewiththenicehat:
Help topics and subtopics are clickable when the CmdHelp command is used.
Includes class attribute to disable clickable links.
unit tests evennia.commands.defaul.tests.TestHelp and evennia.utils.tests.test_utils.TestFormatGrid pass.

evennia.utils.utils.format_grid now has an option to ignore ansi markups when calculating row lengths. Including white spaces between elements.
Verified CmdHelp was only only evennia code (in develop branch) that uses format_grid.

Ease of use. Especially for disabled users.

output example:
image

Verification that formatting is unchanged.

clickable topics off:
image

Clickable topics on:
image

### DIFF
--- a/evennia/utils/utils.py
+++ b/evennia/utils/utils.py
@@ -1837,10 +1837,7 @@ def percentile(iterable, percent, key=lambda x: x):
     return d0 + d1
 
 
-from evennia.utils.ansi import strip_ansi
-
-
-def format_grid(elements, width=78, sep="  ", verbatim_elements=None, ignore_ansi=True):
+def format_grid(elements, width=78, sep="  ", verbatim_elements=None):
     """
     This helper function makes a 'grid' output, where it distributes the given
     string-elements as evenly as possible to fill out the given width.
@@ -1864,9 +1861,6 @@ def format_grid(elements, width=78, sep="  ", verbatim_elements=None, ignore_ans
         like this to make it easier to insert decorations between rows, such
         as horizontal bars.
     """
-    def ansi_len(text):
-        """Use arg ignore_ansi to calculate length of text in element."""
-        return len(strip_ansi(text)) if ignore_ansi else len(text)
 
     def _minimal_rows(elements):
         """
@@ -1875,8 +1869,8 @@ def format_grid(elements, width=78, sep="  ", verbatim_elements=None, ignore_ans
         """
         rows = [""]
         for element in elements:
-            rowlen = ansi_len((rows[-1]))
-            elen = ansi_len((element))
+            rowlen = display_len((rows[-1]))
+            elen = display_len((element))
             if rowlen + elen <= width:
                 rows[-1] += element
             else:
@@ -1888,7 +1882,7 @@ def format_grid(elements, width=78, sep="  ", verbatim_elements=None, ignore_ans
         Dynamic-space, good for making even columns in a multi-line grid but
         will look strange for a single line.
         """
-        wls = [ansi_len((elem)) for elem in elements]
+        wls = [display_len((elem)) for elem in elements]
         wls_percentile = [wl for iw, wl in enumerate(wls) if iw not in verbatim_elements]
 
         if wls_percentile:
@@ -1904,7 +1898,7 @@ def format_grid(elements, width=78, sep="  ", verbatim_elements=None, ignore_ans
             # we use rstrip here to remove extra spaces added by sep
             return [
                 crop(element.rstrip(), width) + " " \
-                     * max(0, width - ansi_len((element.rstrip())))
+                     * max(0, width - display_len((element.rstrip())))
                 for iel, element in enumerate(elements)
             ]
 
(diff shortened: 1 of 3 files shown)


davewiththenicehat:
@Griatch This is ready for review and consideration.
There is a corresponding CmdHelp attribute to disable the clickable topics. This PR has that attribute as True.
I could not imagine not wanting help topics clickable. Especially with subtopic support being added.
It's very easy to change if you are interested but would like it off by default.

Also utils.format_grid has the ability to ignore ansi markups when it calculates row width and white spacing between elements.

Thanks for your time Griatch.

evennia/utils/utils.py
```diff
@@ -1854,21 +1857,26 @@ def format_grid(elements, width=78, sep="  ", verbatim_elements=None):
...
+        ignore_ansi (bool, optional): Ignore ansi markups when calculating white spacing.
 
     Returns:
         list: The grid as a list of ready-formatted rows. We return it
         like this to make it easier to insert decorations between rows, such
         as horizontal bars.
     """
+    def ansi_len(text):
+        """Use arg ignore_ansi to calculate length of text in element."""
+        return len(strip_ansi(text)) if ignore_ansi else len(text)
```
Griatch, line 1869: The way to do this should really be to convert the input to `ANSIString` - you can then use `len()` on them normally without this extra function.

Griatch (changes requested):
This is a good idea and the implementation seems fine. 
The only concern I have is with the custom `ansi_len` function in `format_grid`. We have [evennia.ANSIString](https://github.com/evennia/evennia/blob/develop/evennia/utils/ansi.py#L706) specifically for this purpose - to create strings that can be used normally in length-measurements/tables etc without needing to worry about ansi markers inside them. It probably makes sense to simply just convert the elements of `format_grid` into ANSIStrings and operate on them rather than making a custom solution.

evennia/utils/utils.py
davewiththenicehat: The only reason I had it in a function call was to support disabling ignoring ansi characters.
Should I drop that support?

There is something I am not understanding about ANSIString.
If I change the function ansi:
```python
def ansi_len(text):
        """Use arg ignore_ansi to calculate length of text in element."""
        print(ANSIString(text).clean())
        return len(ANSIString(text).clean()) if ignore_ansi else len(text)
```

I do not receive a string with the ansi markups removed.

My output (top window is the web client) (bottom window is putty with a `evennia restart -l` running)
image

Griatch:
I don't think there is ever a situation where you _don't_ want to ignore the ANSI markers in your text-length evaluation. If you want to print the ansi markup verbatim for some reason you need to escape it manually for that purpose. So yes, I guess the `ignore_ansi` is a little superfluous. 

`print(evennia.ANSIString("|rtext").clean())` correctly gives me the string without any ansi markup. Not sure what you mean there. You'd never just convert it to ANSIString just to do this measurement though, and not in this way - there's no point in doing `len(ANSIString(text).clean())` - that's defeating the purpose of ANSIString. Once you convert the string (with |-markers) into an `ANSIString` the idea is that you operate on it like any string: 

```python
>>> inputtxt = "|rTest|n"
>>> text = ANSIString(inputtxt)  # when you get the string - this can be passed into EvTable etc too.
>>> len(text)
4   # what you want 
>>> len(text.raw())
13   # this is with all the ansi markup (almost never needed)

>>> obj.msg(text) 
Test    # (red)
>>> obj.msg(text.clean())
Test   # no ansi
```

davewiththenicehat:
@Griatch 
Would there be any issue if I used evennia.utils.utils.display_len?
If it is good I will add a unit test or two yet. Mainly to make certain the `evennia.settings_default.HELP_CLICKABLE_TOPICS` variable does not cause people issues.

For ANSIString and len, in sort I am going to assume it is an isolated issue somehow with my code. I will look into it another time.

If for some reason you are interested.
I was able to find that ANSIString is getting passed kwargs `code_indexes`, `char_indexes` and `clean_string`. The `clean_string` is a raw string rather than a clean one.
I know enough to know it is happening at the ANSIMeta stage.
It explains issues I have seen with making clickable links in tables throughout my project.
The only head scratcher here, is that I directly imported ANSIString. So even if I someone had made changes somewhere they wouldn't affect the instances I imported. Not to mention I verified I don't have any ansi setting changed in my settings file. Not that any of them would affect this behavior.

Griatch:
@davewiththenicehat Not sure what you are asking. Are you finding bugs in the ANSIString, this needs to be addressed as a bug report towards `ANSIString`. You should not need to worry about which kwargs are being passed into `ANSIString`, just passing your string with |- markers should be enough. 

I guess you could use `display_len` too, this is also used in the `EvTable` (but EvTable also uses ANSIString).

davewiththenicehat:
@Griatch Just asking about the usage if `display_len`. 

No questions about ANSIString, just pondering in text. I highly doubt it is an evennia bug if you are not seeing the same behavior. I won't walk away from the issue. Now that I know about ANSIString it's far too useful for me not to use. If somehow it is an evennia bug I'll find out.

Griatch:
It seems this was done now, so merged. Thanks for the fixes and work on this!

PR #2453, opened 2021-06-23
Change most links to https

amfl:

This PR changes most links in the site/docs/codebase from http to https.

This is not a blanket find-and-replace. I have made a good effort to check that every link I change still works. (Examples: https://games.evennia.com does not work, and thus has not been changed. References to localhost also remain http.)

The usage of https improves security for users.
References: https://www.cloudflare.com/learning/ssl/why-use-https/, https://www.eff.org/https-everywhere

Most sites automatically redirect the user to https, but some do not. Changing the links in the source means that users don't accidentally end up on an insecure portal.

I noticed the link to blogspot on the homepage used http and thought "Hey, I can fix that..."
I may have become distracted :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # Base docker image for running Evennia-based games in a container.
 #
 # Install:
-#   install `docker` (http://docker.com)
+#   install `docker` (https://docker.com)
 #
 # Usage:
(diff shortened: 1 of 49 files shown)


Griatch:
@amfl Very cool! This looks good at a first glance (sorry for the slow follow-up, I've been heads-down in code). Will review and hopefully merge in the coming days.

Issue #2461, opened 2021-06-28
[DEVEL] [BUG] Typeclass dropdown does not show custom typeclasses


GulliblePsychologist:
#### Describe the bug
In the develop branch, the admin site seems to have a new feature where typeclasses are no longer CharFields but are ChoiceFields instead. However, none of my typeclasses are shown there. Instead, the only typeclasses that are shown are the DefaultAccount/DefaultGuest/Bots and DefaultObject/DefaultRoom/DefaultExit/etc.

#### To Reproduce
1. Start my project.
2. Check out the admin site and open an Account or an Object.
3. Typeclass is wrong and is also reset to the wrong typeclass when you save the entity.

#### Expected behavior
Typeclass field should be a string or a valid list of typeclasses.

#### Environment, Evennia version, OS etc
Evennia 1.0-dev (rev 43fd533) (rev 43fd533)
OS: nt
Python: 3.8.0
Twisted: 21.2.0
Django: 3.2.4

#### Additional context
It does say "If your custom class is not found here, it may not be imported as part of Evennia's startup." but it's unclear how to make the typeclass be imported on Evennia's startup. In this case, I actually wanted to use the default "typeclasses.accounts.Account" typeclass. I also have some custom object typeclasses that should continue to work.

![image](https://user-images.githubusercontent.com/56974578/123567111-5894cb00-d7b9-11eb-8c6b-6762b4de50a0.png)
This should be "typeclasses.accounts.Account"

![image](https://user-images.githubusercontent.com/56974578/123567184-76623000-d7b9-11eb-9d5a-c460bf10493a.png)
This should be "share.objects.Object"

GulliblePsychologist:
I'll close this due to the existence of #2462.

PR #2462, opened 2021-06-28
[DEVEL] Fix typeclass dropdown and make it always show all of the available typeclasses

GulliblePsychologist:
This fixes the issue I reported, #2461 where typeclasses are not shown in the admin site.

If this _isn't_ merged, then any time I open an account or an object or a script from the admin site and dare to make a change it'll overwrite the typeclass with one of Evennia's default typeclasses. This should also occur for a stock project, so it's recommended to make this change.

Apparently, you can pass lambda expressions as the choices in a ChoiceField and this will cause Django to call that in order to get the list of choices. I just added `lambda: ` before the call to `adminutils.get_and_load_typeclasses` and it immediately fixed the problem.

image
(My project uses "share.objects.Object" as the default object typeclass, which wasn't showing up before.)


Griatch:
@GulliblePsychologist This looks like a good change; I'm back to more regular reviewing now (sorry for the delay), will try to review in the coming days.

Issue #2474, opened 2021-07-24
[BUG] Using evennia.prototypes.spawner.spawn merges keys with incorrect precedence 


jerryaldrichiii:
#### Describe the bug

The following example/test fails:

```python
from evennia.prototypes.spawner import spawn
from evennia.utils.test_resources import EvenniaTest

PROTOTYPES = {
    "WEAPON": {
        "typeclass": "evennia.objects.objects.DefaultObject",
        "key": "Weapon",
        "desc": "A generic blade.",
        "magic": False,
    },
    "STING": {
        "prototype_parent": "WEAPON",
        "key": "Sting",
        "desc": "A dagger that shines with a cold light if Orcs are near.",
        "magic": True,
    },
}

class Example:
    def run():
        return spawn(PROTOTYPES["STING"], prototype_parents=PROTOTYPES)[0]

class TestExample(EvenniaTest):
    def test_run(self):
        sting = Example.run()
        self.assertEqual(sting.db.magic, True)
```

I would expect `sting.db.magic` to be True, but it appears that `spawn` has the reverse precedence than I expect.

This can also be seen in the `tutorial_world` contrib when receiving one of the magic weapons at the end. The key is correct, but the description, magic status, etc are all set to the attributes of the parent. Additionally, when traversing multiple parents it only overwrites using the value from the 1st parent. This leads to the magic weapons not having the magic attribute set.

#### Environment, Evennia version, OS etc

```
    Evennia 0.9.0
    OS: posix
    Python: 3.7.6
    Twisted: 19.10.0
    Django: 2.2.9
```

Griatch:
@jerryaldrichiii 
Thanks for the fix! I'm not sure why this would fix the issue at hand though, except for the special case where the prototype-parent uses the "simple" type of Attribute definition (`{attrname: value}`) while the child uses full Attribute definitions (`{'attrs': [(key, value, category, locks), ...]}`). It would seem it would fail in the opposite direction if the Attribute types were reversed between parent and child. 

Am I missing something here?

Griatch:
Resolved in https://github.com/evennia/evennia/commit/0129786c3d5f1b92fef480849beb9a95d2494455.

PR #2475, opened 2021-07-25
Fix merge order with spawn() and prototype_parents

jerryaldrichiii:

Fixes merge order when using spawn() with prototype_parents (see tests for an example)

I <3 you all and want to help :D

Fixes issues described in #2474

### DIFF
--- a/evennia/prototypes/spawner.py
+++ b/evennia/prototypes/spawner.py
@@ -970,7 +970,7 @@ def spawn(*prototypes, **kwargs):
                     (key, init_spawn_value(value, value_to_obj_or_any), None, None)
                 )
 
-        attributes = attributes + simple_attributes
+        attributes = simple_attributes + attributes
         attributes = [tup for tup in attributes if not tup[0] in _NON_CREATE_KWARGS]
 
         # pack for call into _batch_create_object
--- a/evennia/prototypes/tests.py
+++ b/evennia/prototypes/tests.py
@@ -86,6 +86,10 @@ class TestSpawner(EvenniaTest):
             ["goblin grunt", "goblin archwizard"],
         )
 
+    def test_spawn_attr_merging(self):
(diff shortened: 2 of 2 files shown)


Griatch:
@jerryaldrichiii 
Thanks for the fix! While I agree there is a bug somewhere, I'm not sure why this would fix the issue at hand though - except for the special case where the prototype-parent uses the "simple" type of Attribute definition (`{attrname: value}`) while the child uses full Attribute definitions (`{'attrs': [(key, value, category, locks), ...]}`). It would seem it would fail in the opposite direction if the Attribute types were reversed between parent and child. 

Am I missing something here?

jerryaldrichiii:
To be honest, I'm not exactly sure why this fixes things either.

My guess is maybe something somewhere else iterates over this list and order matters for merging? I couldn't find such a thing and the prototype inheritance code confuses me greatly :(

I didn't test a simple + full attributes combination, but the example test uses simple attributes only so I'm not sure if that plays a role or not.

Griatch:
@jerryaldrichiii I fixed this in a different way. The issue is how the custom protparent was added - basically the base prototype was converted to the full attribute form (`{"attrs": [(key, value, category, locks), ...]`) while the protototype-parent remained using the simple-form (`{key:value}`). So when it came to do the update, the prototype had `{"attrs": ("magic", "True", None, ")}` while the parent had `{"magic": False}`. When the prototype-dicts merged, both thus had different keys and both ended up in the dict. Your change made the full one take precedence (True) instead of the simple one (False). The real issue is that there are two representations of the magic value in the prototype; once homogenizing the parent to the same form, the bug is resolved. 
If you are interested, you can see the fix here: https://github.com/evennia/evennia/commit/0129786c3d5f1b92fef480849beb9a95d2494455 

Thanks for the help finding a clue to the solution! Closing this now.

jerryaldrichiii:
Wonderful. Thanks for taking the time to track down the real issue! Also, thanks for the explanation as well! It's very clear.

Every time I interact with this project or its people I learn a ton. You folks are most awesome.

PR #2486, opened 2021-08-09
Update evennia.js

luciensadi:

Fix for misplaced parens-- previous code checked 'is the output of `typeof(listener === 'function')` truthy', which is equivalent of `typeof(boolean)` which is always true. This patch changes it to check for a comparison between the stringified type of listener and the string 'function', which is more likely to be the expected behavior.

Fixing a minor nit.

No issue filed for this yet, it was faster to just write the PR.

### DIFF
--- a/evennia/web/webclient/static/webclient/js/evennia.js
+++ b/evennia/web/webclient/static/webclient/js/evennia.js
@@ -195,7 +195,7 @@ An "emitter" object must have a function
         //     to listen to cmdname events.
         //
         var on = function (cmdname, listener) {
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks for the fix! Small things like that (which probably didn't have much effect for most the time) are easy to miss!

luciensadi:
Absolutely! I noticed it when I was reading the code to find out how to add a plugin, and I figured I'd file a PR and fix it even though it probably caused only minor issues, if any. Appreciate all the work you're doing with Evennia!

PR #2542, opened 2021-10-03
Issue 2261

### LINKED ISSUE #2261: [Feature Request] Lowercase search_prototype.
#### Is your feature request related to a problem? Please describe.
The [evennia.prototypes.prototypes.search_prototype() method](https://github.com/evennia/evennia/blob/master/evennia/prototypes/prototypes.py#L335) will fail to find a matching prototype if the supplied key has any uppercase characters.

#### Describe the solution you'd like
I think that the key given to the method could be forced into all lowercase: `key = key.lower()` 

This should probably be done somewhere at the top, but there's multiple `if key` statements, so I'm uncertain of where this would go.

#### Additional Thoughts
There may be other areas where this would be useful, but I've only used `search_prototype()` recently, so it's the only one that I can confirm would benefit from this change. I haven't used the tags in this method yet either, but perhaps they have the same issue?

Henddher:

Resolves #2261

### DIFF
--- a/evennia/prototypes/prototypes.py
+++ b/evennia/prototypes/prototypes.py
@@ -365,6 +365,10 @@ def search_prototype(key=None, tags=None, require_single=False, return_iterators
         be found as a match.
 
     """
+    # prototype keys are always in lowecase
+    if key:
+      key = key.lower()
+
     # search module prototypes
 
     mod_matches = {}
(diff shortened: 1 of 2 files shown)


Henddher:
Nice! Thanks @Griatch !!

PR #2566, opened 2021-10-14
Components Contrib

ChrisLR:

This adds a contrib for Components, enabling a slightly different style of implementing features while still embracing Evennia's database attributes systems

I was asked to show my implementation in the snippets section but I think having it as a contrib would reach more developers

#### Details

There are a few important classes in this PR
Component - Base Class for every component
ComponentProperty - Descriptor for setting class components 
(Components that are always created with default values at object creation)

ComponentHolderMixin - Mixin that allows typeclasses to use components via a ComponentHandler
ComponentHandler - Wrapper that handles retrieving components both runtime and class defined
DBField and NDBField, essentially AttributeProperties for Components (They wrap to the "host" typeclass)

Prototypes can be used, if you prefix the attribute name by the component name, like
'health__current'

We may want to merge this by squashing, to avoid the ton of commits I made >.>


Griatch:
@ChrisLR This is still marked as WIP but there have been changes to the code. What is the status of this PR?

ChrisLR:
@Griatch I'm slowly cleaning up, writing docs and adding tests.
The majority of the code is there but there's still a few points to investigate.
The callback on modification is currently missing, in our chats you mentioned .db attributehandler being able to do this, so I wanted to replace my current code to instead use it.
Also, we wanted the Components to allow the code to use the DB and NDB by wrapping to the host's db/ndb.
Finally there's the entire cache when there are no host, I wanted to use temporary DB and NDB replacements that can copyover when there's a new host, so it can bulk copy or something...

So uh yeah still working on it :)

ChrisLR:
I am opening up this PR for reviewing, I think this is good enough to work but it surely needs some more refinement.
I need some help carrying this to the finishing line :)

docs/source/Contribs/Components.md
```diff
@@ -0,0 +1,119 @@
...
+class Health(Component):
+    health = DBField(default=1)
+```
+
+Note that default is optional
+
+
+Each typeclass using the ComponentHolderMixin can declare its components
+in the class via the ComponentProperty.
+These are components that will always be present in a typeclass.
```
Griatch, line 63: It may be worth comparing these to how the `AttributeProperty` works in regular (develop) Evennia.

docs/source/Contribs/Components.md
```diff
@@ -0,0 +1,119 @@
...
+
+vampirism_from_elsewhere = character.components.get("vampirism")
+```
+
+Keep in mind that all components must be imported to be visible in the listing
+As such, I recommend regrouping them in a package
+You can then import all your components in that package's __init__
+
+Because of how Evennia import typeclasses and the behavior of python imports
+I recommend placing the components package inside the typeclass package.
```
Griatch, line 92: From experience I know this is not enough info - people won't know what 'typeclass package' means. If you mean "importing your code from `mygame/typeclasses/__init__.py`", write that, preferably with an example of how it would look.

docs/source/Contribs/Components.md
```diff
@@ -0,0 +1,119 @@
...
+
+
+# Technical Stuff you should know about
+
+There is both an DBField and a NDBField for components.
```
Griatch, line 97: This is good info, but it should be mentioned what DBField/NDBField is already when they were first mentioned - or you could point here (your choice). You can also reference the Attribute docs.

docs/source/Contribs/Components.md
```diff
@@ -0,0 +1,119 @@
...
+To do this, call the component's as_template() method and pass it
+values as kwargs to override the component's defaults.
+
+You can then apply this template by calling the component's duplicate method
+to have the new instance created and registered to the new host.
+Something like
+```
+character = self
+health_template = components.Health.create(hp=100)
+new_instance = health_template.duplicate(character)
```
Griatch, line 113: On the face of it, `copy()` seems like it would better match the regular Python method. Is there a reason 'duplicate' was chosen for this?

docs/source/Contribs/Components.md
```diff
@@ -0,0 +1,119 @@
...
+but NDBField will only store it in memory.
+They use AttributeProperty under the hood.
+
+You can make Components stand-alone as a template.
+To do this, call the component's as_template() method and pass it
+values as kwargs to override the component's defaults.
+
+You can then apply this template by calling the component's duplicate method
+to have the new instance created and registered to the new host.
+Something like
```
Griatch, line 109: While an example is given, it's unclear what the benefit of the templating is and why it's called a template. Maybe some basics around components would also be needed in this text?

docs/source/Contribs/Components.md
```diff
@@ -0,0 +1,119 @@
...
+```
+character = self
+health_template = components.Health.create(hp=100)
+new_instance = health_template.duplicate(character)
+character.components.add(new_instance)
+```
+
+Note that duplicate can also accept not having a host,
+this lets it store values that it will later copy to its future host.
+This is a bit slower than giving it the host straight away.
```
Griatch, line 119: Without a host (the db-object), how will you be able to store things persistently? (I know it uses in-memory store, but that's not clear from this text)

evennia/contrib/components/component.py
```diff
@@ -0,0 +1,138 @@
+import itertools
+
+from evennia.typeclasses.attributes import AttributeHandler, InMemoryAttributeBackend
+
+
+class Component:
```
Griatch, line 6: A docstring is needed here.

evennia/contrib/components/component.py
```diff
@@ -0,0 +1,138 @@
...
+            setattr(new, key, value)
+
+        return new
+
+    def cleanup(self):
+        """ This cleans all component fields from the host's db """
+        for attribute in self._all_db_field_names:
+            delattr(self, attribute)
+
+    def duplicate(self, new_host=None):
```
Griatch, line 56: Would `.copy` be a more python-aligned named here maybe? Not sure.

evennia/contrib/components/component.py
```diff
@@ -0,0 +1,138 @@
...
+            if self.host == host:
+                return
+            else:
+                raise ComponentRegisterError("Components must not register twice!")
+
+        self.host = host
+        if self._tdb or self._tndb:
+            self._copy_temporary_attributes_to_host()
+
+    def on_removed(self, host):
```
Griatch, line 86: This works, note that Evennia itself tends to use `at_<event>` for its hooks though (unless in javascript).

evennia/contrib/components/holder.py
```diff
@@ -0,0 +1,136 @@
+from evennia.contrib import components
```
Griatch, line 1: A module docstring is needed.

evennia/contrib/components/holder.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class ComponentProperty:
```
Griatch, line 4: Class docstring is needed. A descriptor like this is also tricky for newcomers to understand, so briefly describing just what it does and how it's used is useful to aid reading. Remember that this docstring will pop up in the API autodocs without users necessarily knowing to reference the README.
(See `AttributeProperty` docstring for an example)

evennia/contrib/components/holder.py
```diff
@@ -0,0 +1,136 @@
...
+    def __set_name__(self, owner, name):
+        class_components = getattr(owner, "_class_components", None)
+        if not class_components:
+            class_components = []
+            setattr(owner, "_class_components", class_components)
+
+        class_components.append((self.component_name, self.values))
+
+
+class ComponentHandler:
```
Griatch, line 25: Doc string needed. See Evennia CODE_STYLE.md

Griatch (changes requested):
Thanks again for the PR and sorry for the slow review of this. :(  I think overall this is looking good. I have made comments for clarifying docs and to add more doc strings (as required by Evennia code standard) but the code itself seems fine (I've not tested edge cases etc though). 

Apart from other extensions of the docs, it may be worth to clarify in your docs how this differentiates itself from the `Traits` contrib and why/when this is used in comparison.

ChrisLR:
Thank you for reviewing! I will correct the PR as soon as I can :)

docs/source/Contribs/Components.md
```diff
@@ -0,0 +1,119 @@
...
+Components may define DBFields or NDBFields at the class level
+Example:
+```
+from evennia.contrib.components import Component, DBField
+
+class Health(Component):
+    health = DBField(default=1)
+```
+
+Note that default is optional
```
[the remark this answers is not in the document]
ChrisLR: Yes, None is used as default, it comes from the underlying AttributeProperty

ChrisLR:
@Griatch I think I will remove the templating and the in memory (no host) storage from this PR.
Components are really useful when they are used to isolate a certain feature and less as containers for data.
Most of the templating can be removed because you can set attributes on prototypes and the components will pick it up.
Just doing that was sufficient for most of my use cases.

Also, use cases that would require templating were better solved using something else.
For example a race component, you would want preset templates that duplicates but that would just create many attributes that ultimately will not change.
It was better solved using a simple DBField key that points to a Race Dataclass, still allowing the character to switch race while in game and without spamming multiple attributes.

As for the in memory (no host) attributes, it offers very little benefit over regular objects or dataclasses and since create accepts kwargs it is very easy to simply build a dict and pass it to components as needed.

What do you think?

Griatch:
> What do you think?

Again, sorry for the late reply. I think this sounds like good changes; there is no need to have features in the contrib that are better done in other ways. :)

Griatch:
@ChrisLR I note that you made changes here, if you are ready for a new review, please re-request the review and make a comment here - since your changes are not echoed to the chat, it's easy to miss otherwise :)

ChrisLR:
@Griatch almost, I was waiting to do some manual testings before requesting a new one but the prototype issue got in the way :)

ChrisLR:
@Griatch okay this seems ready for a 2nd round

Griatch (changes requested):
It's getting there :) This needs more doc strings to match the Evennia CODE_STYLE spec, and I also still think the documentation could be clearer.

ChrisLR:
@Griatch sorry no progress yet but its on top of my TODO list, I'll try to get it done before next week

evennia/contrib/base_systems/components/component.py
```diff
@@ -0,0 +1,79 @@
...
+        return new
+
+    @classmethod
+    def create(cls, host, **kwargs):
+        """
+        This is the method to call when supplying kwargs to initialize a component.
+        """
+        new = cls.default_create(host)
+        for key, value in kwargs.items():
+            setattr(new, key, value)
```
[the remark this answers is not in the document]
ChrisLR: Create is the first step, it must then be added to the Host to persist.

ChrisLR:
@Griatch okay I tried once more, added a bunch of docstrings, rewrote the doc a bit.
The tests run right but I can't get evennia to run with latest develop again, will probably have to fight my venv to death again.
Anyway, should be reviewable, code changes are minimal (only one method I think?) the rest of it is docs

docs/source/Contribs/Components.md
```diff
@@ -0,0 +1,156 @@
+# Components
```
Griatch, line 1: Ehm :point_right::point_left: Since this was started (I guess it would be _Contrib by ChrisLR 2022_ now...) the way contribs are documented has changed. The documentation is to be kept in a `README.md` file in the contrib's main folder instead. The doc-building mechanism will extract the README into a doc page (so it needs to be valid markdown) and place it in the right place in the docs (and also include it in a nice index of contribs). 

So please add this as `evennia/contrib/base_systems/components/README.md` instead.

Griatch (changes requested):
Looking a lot better; I see code doc strings. This is looking good. The doc-changes looks better too. I still have a few un-attended review-commends in the docs, please address those and/or comment on them being addressed (or motivate why you disagree). 

The only other blocker is the contrib's README.md requirement that has come to be since after this contrib started development (see comment). 

Once that is done you can go ahead and squash the commits as well :) I don't want to do it, because then it'd look like I commited your work.

docs/source/Contribs/Components.md
```diff
@@ -0,0 +1,133 @@
...
+to the object at runtime, giving you a one-off destructible object.
+
+Then at any point where you want to try to damage something you can check for the feature like
+
+target_health = target.cmp.health
+if not target_health:
+    return "You cannot attack this!"
+else:
+    damage_value = roll_damage()
+    target_health.damage(damage_value)
```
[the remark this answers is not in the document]
ChrisLR: cmp is just a shortcut to components, simlarly to db and attributes

docs/source/Contribs/Components.md
```diff
@@ -0,0 +1,133 @@
...
+character = self
+vampirism = components.Vampirism.create(character)
+character.components.add(vampirism)
+
+...
+
+vampirism_from_elsewhere = character.components.get("vampirism")
+```
+
+Keep in mind that all components must be imported to be visible in the listing
```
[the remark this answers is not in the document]
ChrisLR: This is to note that any components must be imported before being used.

ChrisLR:
@Griatch I will close this PR and open another from a different branch with the commits squashed

PR #2571, opened 2021-10-19
Replace every space/tab with non-breaking spaces.

### LINKED ISSUE #2532: [BUG] EvTables header skewed when using colors
#### Describe the bug
In certain cases, the evtables header can be skewed. This happens when the header has leading/trailing spaces. They get trimmed which causes the table to look skewed.

#### To Reproduce
1. Use the following example script:
```
from evennia.utils import evtable

table = evtable.EvTable("|cHeading1|n", "|cHeading2 |n", "|cHeading 3|n")
table.add_row("This","Test","test")
print(table)
```
2. Execute said script
3. See the error occur (as shown below)
```
+----------+-----------+-----------+
| Heading1 | Heading2 | Heading 3 |
+~~~~~~~~~~+~~~~~~~~~~~+~~~~~~~~~~~+
| This     | Test      | test      |
```

#### Expected behavior
The table should be straight without any indentations

Evennia version: 1.0-dev (rev 54ebdb5ab)
        OS: nt
        Python: 3.9.7
        Twisted: 21.7.0
        Django: 3.2.7

#### Additional context
This only seems to occur when you are using colors and the space character, otherwise it looks fine.

Henddher:
Resolves #2532

### DIFF
--- a/evennia/commands/default/general.py
+++ b/evennia/commands/default/general.py
@@ -51,13 +51,6 @@ class CmdHome(COMMAND_DEFAULT_CLASS):
             caller.msg("There's no place like home ...")
             caller.move_to(home)
 
-        from evennia.utils import evtable
-        table = evtable.EvTable("|cHeading1|n", "|cHeading2 |n", "|cHeading 3|n")
-        table.add_row("This","Test","test")
-        print(table)
-        caller.msg(str(table))
-
-
 
 class CmdLook(COMMAND_DEFAULT_CLASS):
     """
--- a/evennia/commands/default/general.py
+++ b/evennia/commands/default/general.py
@@ -51,6 +51,13 @@ class CmdHome(COMMAND_DEFAULT_CLASS):
             caller.msg("There's no place like home ...")
             caller.move_to(home)
 
+        from evennia.utils import evtable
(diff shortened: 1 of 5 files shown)


Griatch:
@Hendher: 
This is how it looks in Firefox 93 on Linux with current `develop` branch and two panes echoing a long `say` (to make this, just add a new client tab and drag it out into a new pane to create a split): 
DeepinScreenshot_select-area_20211031203439
This is how it looks with your PR code: 
DeepinScreenshot_select-area_20211031203909
Note the right pane fails to wrap the output and instead it flows out of it. 

It's possible this is something that can be corrected on the javascript side in the web client. But the fact remains that we can't merge this as-is.

Henddher:
Thank you @Griatch 
I'll try that here (using Chrome) and see what other options exist. I don't think I'll be able to resolve this issue for Hacktoberfest :frowning_face: 

With this PR, word-wrapping is not happening. Only char-wrapping happens. Perhaps the sane thing to do here is start from scratch.

Griatch:
@Henddher I think you already fulfill your T-shirt quota with previous contributions, no? Apart from that, don't worry too much if the Hacktoberfest deadline is kept or not :)

Henddher:
Closing this PR because the solution doesn't address all use-cases.

PR #2577, opened 2021-10-23
WIP Issue 2519

### LINKED ISSUE #2519: [Feature Request] (Develop) Attribute `defaultdict` value does not serialize/deserialize correctly
#### Describe the bug
Storing a `defaultdict` to an Attribute results in the values being lost.

#### To Reproduce
```
>>> from collections import defaultdict
>>> d = defaultdict(list)
>>> c.db.lists = d
>>> c.db.lists
[]
>>> d['name'].append('bob')
>>> c.db.lists = d
>>> c.db.lists
['name']
>>> c.db.lists['name']
  File "<console>", line 1, in <module>
  File "/home/ubuntu/workspace/evennia/evennia/utils/dbserialize.py", line 224, in __getitem__
    return self._data.__getitem__(key)
TypeError: list indices must be integers or slices, not str
```

#### Expected behavior
Was mainly curious to see if it works.

Store the object correctly or raise an error about it being unserializable.

Develop

Henddher:
Attempts to fix #2519 

Unfortunately, not working still. See review comments.

Hi @Griatch, any suggestion?

### DIFF
--- a/evennia/utils/dbserialize.py
+++ b/evennia/utils/dbserialize.py
@@ -203,6 +203,10 @@ class _SaverMutable(object):
                 dat = _SaverDict(_parent=parent)
                 dat._data.update((key, process_tree(val, dat)) for key, val in item.items())
                 return dat
+            elif dtype == defaultdict:
+                dat = _SaverDefaultDict(item.default_factory, _parent=parent)
+                dat._data.update((key, process_tree(val, dat)) for key, val in item.items())
+                return dat
             elif dtype == set:
                 dat = _SaverSet(_parent=parent)
                 dat._data.update(process_tree(val, dat) for val in item)
@@ -221,6 +225,7 @@ class _SaverMutable(object):
         return self._data.__iter__()
 
     def __getitem__(self, key):
+        # breakpoint()
         return self._data.__getitem__(key)
 
     def __eq__(self, other):
@@ -303,6 +308,22 @@ class _SaverDict(_SaverMutable, MutableMapping):
     @_save
     def update(self, *args, **kwargs):
         self._data.update(*args, **kwargs)
+
+
+class _SaverDefaultDict(_SaverDict):
+    """
+    A defaultdict that stores changes to an attribute when updated
+    """
+
+    def __init__(self, factory, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._data = defaultdict(factory)
+        self.default_factory = factory
+
+    # @_save
+    def __getitem__(self, key):
+        # breakpoint()
+        return self._data[key]
 
 
 class _SaverSet(_SaverMutable, MutableSet):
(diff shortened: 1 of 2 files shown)


Griatch:
@Henddher As I noted in a comment, I'd think that `self.obj.db.test` should have a type of `_SaverDefaultDict` at the point of your test (not `defaultdict`)? Without that there is no auto-saving to the database. I have not investigated the code any closer though.

Henddher:
Thank you @Griatch

Yes. When `defaultdict` instance is assigned to `self.obj.db.test`, `self.obj.db.test` becomes a `_SaverDefaultDict`. It's saved and retrieved from database correctly.

However, any `defaultdict` operation applied to `self.obj.db.test` *after* the assignment does NOT work. IOW, `_SaverDefaultDict` doesn't honored any of the `defaultdict` operations as I was expecting. I was expecting that `self.obj.db.test` would always behave like a `defaultdict`. 
Is my expectation incorrect?

(I know today's the last day for Hacktoberfest so hopefully we can iron out all this before today's deadline)

Griatch:
The `_SaverDefaultDict` should implement all features of a `defaultdict` since it's meant to be used as one. Exactly what is missing in your implementation is not something I can tell offhand. I'd suggest stepping through the entire setup/access process with a debugger to see exactly what is going on and where it is not acting the way you expect.

Henddher:
So when `self.obj.db.test = X` happens, `self.obj.db.test` is no longer behaves like X does?
 
I'm expecting `self.obj.db.test` will behave like X does after the assignment. If that's not the case, then the current code "works". The current code saves the `defaultdict` as is and allows it to be read back from DB. However, `defaultdict` operations applied to `self.obj.db.test` don't work ... how could one use `self.obj.db.test` as `defaultdict` if it no longer works like one?

Griatch:
@Henddher 
> So when `self.obj.db.test = X` happens, `self.obj.db.test` is no longer behaves like X does?
> 
> I'm expecting `self.obj.db.test` will behave like X does after the assignment. If that's not the case, then the current code "works". The current code saves the `defaultdict` as is and allows it to be read back from DB. However, `defaultdict` operations applied to `self.obj.db.test` don't work ... how could one use `self.obj.db.test` as `defaultdict` if it no longer works like one?

The `defaultdict` should be a `_SaverDefaultDict`, when it sits on `self.obj.db.test` (not a `defaultdict`, even though that's what is serialized into the db). And `_SaverDefaultDict` _should_ work exactly like the `defaultdict` except it saves. So if you check the type of `self.obj.db.test` and it is `defaultdict`, that's where the problem is and why you won't save things.

Griatch:
@Henddher Any plans to continue working on resolving this PR?

Griatch:
@Henddher Thanks for the PR! I fixed the issue with this and merged now.

PR #2589, opened 2021-11-21
get_display_name looker default. Resolves #2415

### LINKED ISSUE #2415: [BUG/Annoyance] Object `get_display_name` - Make `looker=None` by default
#### Describe the bug
The syntax for object method `get_display_name` _requires_ a `looker` reference.

This is unnecessary and annoying, since explicitly passing `None` has the same effect as passing almost any other (non-privileged) object. `None` might as well be the default since the effect is the same in _most_ cases.

It also leads to dbref exposure if a message intended for broadcast is tailored to a privileged caller because one felt compelled to provide _something_ as an argument, in cases where they should have abstained:
```
> self.location.msg_contents(f"{self.caller.get_display_name(self.caller)} opens the gate")
# everybody sees:
johnny(#1) opens the gate 

vs

> self.location.msg_contents(f"{self.caller.get_display_name()} opens the gate")
# everybody SHOULD see:
johnny opens the gate 
```
(I know there are some functions that do variable replacement for tailoring custom messages; this is a simplistic example despite them.)

#### Expected behavior
Just make the method definition use `None` as the default value for `looker`.

The-Archemist:
Defaults object get_display_name() looker parameter to none.

Resolves #2415

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -309,7 +309,7 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
 
     # main methods
 
-    def get_display_name(self, looker, **kwargs):
+    def get_display_name(self, looker=None, **kwargs):
         """
         Displays the name of the object in a viewer-aware manner.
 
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks for the PR, seems to work fine :)

PR #2594, opened 2021-11-24
Add new cooldowns contrib.

owllex:

This PR adds a new cooldowns contrib module that can be used off the shelf for tracking cooldown timers. Code can set a cooldown for some number of seconds using an arbitrary cooldown key. The cooldown will be reported as not ready until that amount of time has passed.

Cooldowns can be easily integrated into commands as a manual check to block repeated uses until the cooldown is ready, and can also be easily integrated into a custom lockfunc if desired (not included in this PR).

Cooldowns are a simple attribute handler under the hood but have lots of varied uses. Many kinds of game require an action rate-limiting mechanism, and this system provides it very simply as a drop-in.

I make heavy use of this in my under-development game, and thought it might be helpful to others.

N/A

### DIFF
--- a/evennia/contrib/cooldowns.py
+++ b/evennia/contrib/cooldowns.py
@@ -15,18 +15,21 @@ state. They do not fire callbacks, so are not a good fit for use cases
 where something needs to happen on a specific schedule (use delay or
 a TickerHandler for that instead).
 
+See also the evennia documentation for command cooldowns
+(https://github.com/evennia/evennia/wiki/Command-Cooldown) for more information
+about the concept.
+
 Installation:
 
-To use, simply add the following property to the typeclass definition of
-any object type that you want to support cooldowns. It will expose a
-new `cooldowns` property that persists data to the object's attribute
-storage. You can set this on your base `Object` typeclass to enable cooldown
-tracking on every kind of object, or just put it on your `Character`
-typeclass.
+To use, simply add the following property to the typeclass definition of any
+object type that you want to support cooldowns. It will expose a new `cooldowns`
+property that persists data to the object's attribute storage. You can set this
+on your base `Object` typeclass to enable cooldown tracking on every kind of
+object, or just put it on your `Character` typeclass.
 
-By default the CooldownHandler will use the `cooldowns` property, but you
-can customize this if desired by passing a different value for the
-db_attribute parameter.
+By default the CooldownHandler will use the `cooldowns` property, but you can
+customize this if desired by passing a different value for the db_attribute
+parameter.
 
     from evennia.contrib.cooldowns import Cooldownhandler
     from evennia.utils.utils import lazy_property
@@ -37,14 +40,16 @@ db_attribute parameter.
 
 Example:
 
-Assuming you've installed cooldowns on your Character typeclasses, you can
-use a cooldown to limit how often you can perform a command:
+Assuming you've installed cooldowns on your Character typeclasses, you can use a
+cooldown to limit how often you can perform a command. The following code
+snippet will limit the use of a Power Attack command to once every 10 seconds
+per character.
 
 class PowerAttack(Command):
     def func(self):
         if self.caller.cooldowns.ready("power attack"):
             self.do_power_attack()
-            self.caller.cooldowns.set("power attack", 10)
+            self.caller.cooldowns.add("power attack", 10)
         else:
             self.caller.msg("That's not ready yet!")
 """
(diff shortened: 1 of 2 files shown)


evennia/contrib/cooldowns.py
```diff
@@ -0,0 +1,188 @@
...
+Example:
+
+Assuming you've installed cooldowns on your Character typeclasses, you can
+use a cooldown to limit how often you can perform a command:
+
+class PowerAttack(Command):
+    def func(self):
+        if self.caller.cooldowns.ready("power attack"):
+            self.do_power_attack()
+            self.caller.cooldowns.set("power attack", 10)
```
Griatch, line 47: Presumably, this won't allow for 10 seconds, may be good to mention this in the descriptive text.

evennia/contrib/cooldowns.py
```diff
@@ -0,0 +1,188 @@
...
+typeclassed Object or Account. A cooldown is a lightweight persistent
+asynchronous timer that you can query to see if it is ready.
+
+Cooldowns are good for modelling rate-limited actions, like how often a
+character can perform a given command.
+
+Cooldowns are completely asynchronous and must be queried to know their
+state. They do not fire callbacks, so are not a good fit for use cases
+where something needs to happen on a specific schedule (use delay or
+a TickerHandler for that instead).
```
Griatch, line 16: May be useful to reference (or at least mention) the Evennia doc on cooldowns as well, for those that want more info about the concept.

evennia/contrib/cooldowns.py
```diff
@@ -0,0 +1,188 @@
...
+    supports DB attributes (like a Character or Account).
+
+    A cooldown is a timer that is usually used to limit how often
+    some action can be performed or some effect can trigger. When a
+    cooldown is first set, it counts down from the amount of time
+    provided back to zero, at which point it is considered ready again.
+
+    Cooldowns are named with an arbitrary string, and that string is used
+    to check on the progression of the cooldown. Each cooldown is tracked
+    separately and independently.
```
Griatch, line 68: This (might) make it sound like cooldowns are global when they are in fact per-object. May be worth to clarify this; something like 

    A cooldown tracked separately and independently from other cooldowns on that same object. It is unique per-object.

evennia/contrib/cooldowns.py
```diff
@@ -0,0 +1,188 @@
...
+    - time_left(name): Returns how much time is left on a cooldown.
+    - set(name, seconds): Sets a given cooldown to last for a certain
+        amount of time. Until then, ready() will return False for that
+        cooldown name.
+    - extend(name, seconds): Like set, but adds time to the given
+        cooldown name. If it doesn't exist yet, calling this is equivalent
+        to calling set.
+    - reset(cooldown): Resets a given cooldown, causing ready() to return
+        True for that cooldown immediately.
+    - clear(): Resets all cooldowns.
```
Griatch, line 86: Overall, a very nice doc string! :+1:

evennia/contrib/cooldowns.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def ready(self, *args):
+        """
+        Checks whether all of the provided cooldowns are ready (expired).
+        If a requested cooldown does not exist, it is considered ready.
+
+        Args:
+            any (str): One or more cooldown names to check.
+        Returns:
+            (bool): True if each cooldown has expired or does not exist.
```
Griatch, line 111: A minor thing, but it matters nowadays when these docstrings are auto-parsed into documentation... 

- `any (str)` should be `*args (str)` these days. We use `any` in several other places, but this should be changed.
- `(bool):` should be just `bool:`.

evennia/contrib/cooldowns.py
```diff
@@ -0,0 +1,188 @@
...
+                ready. Returns 0 if all cooldowns are ready (or don't
+                exist.)
+        """
+        now = time.time()
+        cooldowns = [self.data[x] - now for x in args if x in self.data]
+        if not cooldowns:
+            return 0
+        return math.ceil(max(max(cooldowns), 0))
+
+    def set(self, cooldown, seconds):
```
Griatch, line 134: While `set` does make sense too, it's worth to point out that Evennia's handlers always use `add` here instead. For consistency, I'd suggest changing this method into `add` and maybe let `set ` be an alias to it.

evennia/contrib/cooldowns.py
```diff
@@ -0,0 +1,188 @@
...
+        now = time.time()
+        self.data[cooldown] = int(now) + (max(seconds, 0) if seconds else 0)
+
+    def extend(self, cooldown, seconds):
+        """
+        Adds a specific amount of time to an existing cooldown.
+
+        If this cooldown is already ready, this is equivalent to calling
+        set. If the cooldown is not ready, it will be extended by the
+        provided duration.
```
Griatch, line 154: It feels like the outcome of this operation can vary quite a bit. It's basically either an extend or an add/set; the result of which depends on what existed beforehand. You may not always know this when calling, and could end up with a longer delay than you anticipated. How about returning the resulting time-left for calling code that cares?

evennia/contrib/cooldowns.py
```diff
@@ -0,0 +1,188 @@
...
+        for cooldown in list(self.data.keys()):
+            del self.data[cooldown]
+
+    def cleanup(self):
+        """
+        Deletes all expired cooldowns. This helps keep attribute storage
+        requirements small.
+        """
+        now = time.time()
+        keys = [x for x in self.data.keys() if self.data[x] - now < 0]
```
Griatch, line 186: Rather than looping over the (Attribute) data set, which means a lot of deserializations and re-saving to the database, I'd suggest extracting the data and decouple it from the database at the top of this method. Then loop+delete over the in-memory dict. Afterwards you save the resulting dict back to the Attribute.

This matters little for most code, but when you are making a formal thing like this, one can just as well do it efficiently. :wink:

evennia/contrib/cooldowns.py
```diff
@@ -0,0 +1,188 @@
...
+        Returns:
+            (int): Number of seconds until all provided cooldowns are
+                ready. Returns 0 if all cooldowns are ready (or don't
+                exist.)
+        """
+        now = time.time()
+        cooldowns = [self.data[x] - now for x in args if x in self.data]
+        if not cooldowns:
+            return 0
+        return math.ceil(max(max(cooldowns), 0))
```
Griatch, line 132: While sub-second precision does not make much sense for repeating tasks (echoing messages faster than 1/s is not really that useful to a user), it seems unnecessary to limit this check to second-precision. One of the main uses of cooldowns are for combat systems, and sub-second precision can make use of those (even though you don't need to echo it to the user that fast). 

`ceil` should allow the `.ready` to still operate at sub-second precision (it's 1 until it suddenly flips over to 0), but it's useful to be able to see exactly how much time is left (e.g. by other code using this). So I suggest changing this to return floats by default instead. 

You could add a kwarg `precision` or `use_int` or something maybe?

Griatch (changes requested):
Looks good overall, thanks for the PR! I have some comments and feedback to address before merging.

evennia/contrib/cooldowns.py
owllex: Done!

evennia/contrib/cooldowns.py
owllex: Thanks!

evennia/contrib/cooldowns.py
owllex: Good point. Fixed.

evennia/contrib/cooldowns.py
owllex: Updated the text.

evennia/contrib/cooldowns.py
owllex: Thanks, fixed here and elsewhere in the PR.

evennia/contrib/cooldowns.py
owllex: Good point. Added a return value for time left.

evennia/contrib/cooldowns.py
owllex: Done. Note that I had to add storage for the parent object and db attribute name to enable this. I changed this class to use __slots__ to minimize memory impact.

evennia/contrib/cooldowns.py
owllex: Good point. I added a `use_int` kwarg (default False) and changed everything to be float-first.

owllex:
Addressed comments, please take a look!

Griatch (approved):
Thanks for the changes, I think this looks good now!

PR #2605, opened 2021-12-20
dbserialize fix to support Python 3.10

The-Archemist:
**Overview:**
Update to imports supports Python 3.10 while maintaining backwards compatibility.

For updated collections references, see:
https://docs.python.org/3.10/library/collections.html
https://docs.python.org/3/library/collections.abc.html

### DIFF
--- a/evennia/utils/dbserialize.py
+++ b/evennia/utils/dbserialize.py
@@ -19,8 +19,8 @@ be out of sync with the database.
 
 """
 from functools import update_wrapper
-from collections import defaultdict, MutableSequence, MutableSet, MutableMapping
-from collections import OrderedDict, deque
+from collections import deque, OrderedDict, defaultdict
+from collections.abc import MutableSequence, MutableSet, MutableMapping
 
 try:
     from pickle import dumps, loads


THEGODFATHERRX:
nvm, I replaced "collections" with "collections.abc" everywhere and ended up editing every file(except 1) individually to remove the ".abc"

so only fix the "evennia/evennia/utils/dbserialize.py" to get it to work with 3.10.

it should look like this:
```
from collections.abc import MutableSequence, MutableSet, MutableMapping
from collections import OrderedDict, deque, defaultdict
```

The-Archemist:
> nvm, I replaced "collections" with "collections.abc" everywhere(using sed and find) and ended up editing every file(except 1) individually to remove the ".abc"
> 
> so only fix the "/evennia/evennia/utils/dbserialize.py" to get it to work with 3.10.
> 
> it should look like this:
> 
> ```
> from collections.abc import MutableSequence, MutableSet, MutableMapping
> from collections import OrderedDict, deque, defaultdict
> ```

That's... literally what I did.

kj23decker:
Installed this morning on Windows 11, had to manually perform this step and then worked perfectly.

PR #2606, opened 2021-12-24
Fix gh-pages README.md link text

gtaylor:
Leetle fix!


gtaylor:
Hmm, perhaps this should point to https://www.evennia.com/docs/latest/ now instead?

Griatch:
Thanks, yes I merged your typo fix, but ended up rewriting that README since it was no longer relevant to how the doc structure looks these days. :)

PR #2628, opened 2022-01-15
Adding "mult" to traits contrib

Tegiminis:

Adds "mult" as a stat to the traits contrib, allowing developers to use multiplicative modifiers for their traits. Anywhere that you would receive "`base + mod`" as a return has been edited to "`(base + mod) * mult`". This is the current value for Static and Counter traits, and the max value for Gauge traits.

I use multiplicative math for my game's buffs and stats sometimes, so I figure other people might too.

### DIFF
--- a/evennia/contrib/rpg/traits/traits.py
+++ b/evennia/contrib/rpg/traits/traits.py
@@ -1151,7 +1151,7 @@ class StaticTrait(Trait):
 
     trait_type = "static"
 
-    default_keys = {"base": 0, "mod": 0}
+    default_keys = {"base": 0, "mod": 0, "mult": 1.0}
 
     def __str__(self):
         status = "{value:11}".format(value=self.value)
@@ -1180,9 +1180,19 @@ class StaticTrait(Trait):
             self._data["mod"] = amount
 
     @property
+    def mult(self):
+        """The trait's multiplier."""
+        return self._data["mult"]
+    
+    @mult.setter
+    def mult(self, amount):
+        if type(amount) in (int, float):
+            self._data["mult"] = amount
+
+    @property
     def value(self):
-        "The value of the Trait"
-        return self.base + self.mod
+        "The value of the Trait."
+        return (self.base + self.mod) * self.mult
 
 
 class CounterTrait(Trait):
@@ -1225,6 +1235,7 @@ class CounterTrait(Trait):
     default_keys = {
         "base": 0,
         "mod": 0,
+        "mult": 1.0,
         "min": None,
         "max": None,
         "descs": None,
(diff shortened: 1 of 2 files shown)


Griatch (changes requested):
Thanks for the change! The docstrings for each of the Trait classes must be updated as well though, since these don't mention the multiplier anywhere today and you'd have to read the code to find out it exists. 
It would also be good to see a new unit per trait class to show that this behaves as expected.

Tegiminis:
Updated with documentation that reflects the change as well as properly utilizing multiplier in validation and string methods. Will start on some unit tests.

evennia/contrib/rpg/traits/traits.py
```diff
@@ -171,25 +171,26 @@ class Object(DefaultObject):
 ```
 ## Static trait
 
-`value = base + mod`
+`value = (base + mod) * mult`
 
-The static trait has a `base` value and an optional `mod`-ifier. A typical use
+The static trait has a `base` value and an optional `mod`-ifier and 'mult'-iplier. A typical use
```
Griatch, line 176: Worth pointing out that `mult` defaults to `1.0` and can thus be ignored if you don't need it.

evennia/contrib/rpg/traits/traits.py
```diff
@@ -201,16 +202,17 @@ class Object(DefaultObject):
                                   current    value
                                              = current
                                              + mod
+                                             * mult
```
Griatch, line 205: This looks wrong since it lacks the parenthesis; it looks like mod * mult. Better reformat it to use the parenthesis.

evennia/contrib/rpg/traits/traits.py
```diff
@@ -201,16 +202,17 @@ class Object(DefaultObject):
...
 
 A counter describes a value that can move from a base. The `.current` property
 is the thing usually modified. It starts at the `.base`. One can also add a
-modifier, which will both be added to the base and to current (forming
-`.value`).  The min/max of the range are optional, a boundary set to None will
-remove it. A suggested use for a Counter Trait would be to track skill values.
+modifier, which is added to both the base and to current. '.value' is then formed 
+by multiplying by the multiplier. The min/max of the range are optional, a boundary
```
Griatch, line 210: Again worth to note that `mult` defaults to 1.0.

evennia/contrib/rpg/traits/traits.py
```diff
@@ -1207,8 +1212,9 @@ class CounterTrait(Trait):
                                     current   value
                                               = current
                                               + mod
+                                              * mult
```
Griatch, line 1215: Same here, best to reformat to include the parenthesis.

Griatch (changes requested):
Thanks for the additions. :) I only have some minor comments.

Griatch:
@Tegiminis Any progress with the last changes to this?

Griatch (approved):
Thanks for the changes!

PR #2631, opened 2022-01-16
Set pause-related script attributes to None rather than deleting them

owllex:

Modify DefaultScript to re-use the same attributes for  `_paused_time`, `_paused_callcount`, and `_manually_paused`. These attributes currently are deleted on each script every time it restarts (which occurs for all scripts on server reboot) and then recreated again. For better efficiency, these values can be simply set to None rather than deleted, which allows them to stay in the attr cache and not caused unnecessary creations of new attributes.

Note that I also reformatted the file with the black formatter, which is why you're seeing some other smaller style diffs in here.

DB efficiency.

I posted some comments about this in the discord where I noticed by attribute ID numbers going up quite high (into the 20000's) on a development server with relatively few objects, and tracked it down to attribute creation/deletion thrash. The issue becomes more pronounced for servers running many scripts, even those that do not manually pause.

This change should be safe since all uses of the attributes in question are simply checking truthiness of the attribute value, so a non-existent attribute and one whose value set to None will behave identically.

I couldn't find a simple way to unit test this though without explicitly checking the state of these attributes, and since they are intended to be private, it didn't feel like a useful test to write since it would be testing implementation-specific logic. Existing script tests pass of course.

This issue also exits on master, but this PR cannot be cleanly merged back. The fix is the same in principle.

### DIFF
--- a/evennia/scripts/scripts.py
+++ b/evennia/scripts/scripts.py
@@ -158,8 +158,15 @@ class ScriptBase(ScriptDB, metaclass=TypeclassBase):
         # TODO - restart anew ?
         return ret
 
-    def _start_task(self, interval=None, start_delay=None, repeats=None, force_restart=False,
-                    auto_unpause=False, **kwargs):
+    def _start_task(
+        self,
+        interval=None,
+        start_delay=None,
+        repeats=None,
+        force_restart=False,
+        auto_unpause=False,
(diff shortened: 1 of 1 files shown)


Griatch (approved):
Good catch! Thanks for the PR!

Issue #2620, opened 2022-01-11
[BUG][DEVELOP] Module Prototypes are no longer loaded


ChrisLR:
#### Describe the bug
Module prototypes no longer load, trying to spawn one says there are None

#### To Reproduce
1. Add a prototype in a module
2. Try to spawn said prototype
3. ??
4. See error (No profit here)

#### Expected behavior
The prototypes are loaded

#### Environment, Evennia version, OS etc
This happens on the latest develop commit

#### Additional context
Regression as talked about on Discord

PR #2634, opened 2022-01-18
2620 fix prototype loading

ChrisLR:
This ensures the prototypes are loaded before being searched

Fix issue

https://github.com/evennia/evennia/issues/2620

### DIFF
--- a/evennia/commands/default/tests.py
+++ b/evennia/commands/default/tests.py
@@ -1405,6 +1405,7 @@ class TestBuilding(EvenniaCommandTest):
             "Could not find 'Obj'.| (Objects to destroy "
             "must either be local or specified with a unique #dbref.)",
         )
+        settings.DEFAULT_HOME = f"#{self.room1.dbid}"
         self.call(
             building.CmdDestroy(), settings.DEFAULT_HOME, "You are trying to delete"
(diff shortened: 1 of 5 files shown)


Griatch (changes requested):
Thanks for the PR! Curiously, there seems to be some failing unit tests though ... while there are some issues with the mysql/postgres builds in develop, the sqlite3 test should pass (see the previous https://github.com/evennia/evennia/runs/4833101042?check_suite_focus=true). There seems to be some failing object lookups from this PR. Unclear why that would arise from this change but it seems to be new.

Griatch:
That looks better. The reason the sqlite build fails now is due to a bug in the CI script that I fixed (it fails since you of course don't have the evennia docker creds).

ChrisLR:
@Griatch there's still a remaining issue hidden somewhere, some of the tests override settings and let the teardown clean it but somehow it doesn't. The temporary work around was to enforce the DEFAULT_HOME back but the override_settings and the teardown are supposed to rollback the changes...

Griatch:
@ChrisLR That's fine, it's already a step forward and the test suite works when I built it locally. If you want to dig further into the testing logic, best to do so in a separate PR :) I merged this now.

PR #2662, opened 2022-03-02
Develop - Allow customization of server reset/restart/shutdown messages.

aMiss-aWry:
Added the following to default_settings, retaining the same form as evennia currently has:
```
SERVER_RESTARTED_MESSAGE = " ... Server restarted."
SERVER_RESET_MESSAGE = " Server reset initiated ..."
SERVER_RESTART_MESSAGE = " Server restart initiated ..."
SERVER_SHUTDOWN_MESSAGE = " Shutting down server ..."
SERVER_SHUTDOWN_FINAL_MESSAGE = " Server is being SHUT DOWN!"
```

In lieu of hardcoded text, reload & similar commands now pull from the above messages.

Allow customization of text displayed on server reset.

Comment for the future: Would suggest breaking out custom messaging into a different file than 'settings', but no such file presents itself. I make use of a 'messages.py' in my own game, something like that might be valuable eventually if default_settings gets noisy. Would effectively be the same thing, but might make it easier to alter displayed strings instead of sifting through all settings.

### DIFF
--- a/evennia/utils/funcparser.py
+++ b/evennia/utils/funcparser.py
@@ -1153,7 +1153,7 @@ def funcparser_callable_search(*args, caller=None, access="control", **kwargs):
     if not targets:
         if return_list:
             return []
-        raise ParsingError(f"$search: Query '{query}' gave no matches.")
+        raise ParsingError(f"$search: Query '{args[0]}' gave no matches.")
 
     if len(targets) > 1 and not return_list:
         raise ParsingError(


Griatch (commented):
Thanks for the PR! Hm, I'll need to think some more about this one, because the strings you replace are under i18n and making them dynamic like this will break existing language translations. And as you say, maybe one should consider some other way to do this than to add a big block of text alternatives to the settings. Finally, we don't usually add string replacements in commands; since commands can be replaced from the game dir easily enough. 

So holding off on this one a little.

The-Archemist:
System.py will yield a spacing area between system message and reason on line 81.

The-Archemist:
> Thanks for the PR! Hm, I'll need to think some more about this one, because the strings you replace are under i18n and making them dynamic like this will break existing language translations. And as you say, maybe one should consider some other way to do this than to add a big block of text alternatives to the settings. Finally, we don't usually add string replacements in commands; since commands can be replaced from the game dir easily enough.
> 
> So holding off on this one a little.

In regards to not taking string replacement for commands, I think this one warrants an exception. While you can easily exchange the string "System Reboot Initiated..." for something else, the return prompt when it completes is handled internally and not via the command itself. Therefore, the current process only half accounts for customization. It's a bit odd in practice to see "A storm blows through..." and for the next string to be something immersion breaking such as "Server reloaded."

Griatch:
I agree with this particular error probably being worthy of an exception, just not 100% sure about putting these are separate settings variables; it may be more appropriate to have some sort of `DEFAULT_MESSAGES` dict or something. Not sure yet.

Griatch:
As an update, I have not forgotten about this, but I'll hold off on merging it for a little. I'm considering if it's better to organize strings like this differently in the settings file.

PR #2667, opened 2022-03-06
Components Contrib (Pt1)

ChrisLR:

This adds a contrib to support Components.
This implementation of Components are made in a way that reuses many core features of Evennia.

It was requested a few times in Discord, so I thought this could help.

There are a few important classes in this PR
Component - Base Class for every component
ComponentProperty - Descriptor for setting class components
(Components that are always created with default values at object creation)

ComponentHolderMixin - Mixin that allows typeclasses to use components via a ComponentHandler
ComponentHandler - Wrapper that handles retrieving components both runtime and class defined
DBField and NDBField, essentially AttributeProperties for Components (They wrap to the "host" typeclass)

Prototypes can be used, if you prefix the attribute name by the component name, like
'health__current'

**There are still a few features missing, such as Component Signals.
I will open a separate PR for it.
The automagical import nature of components may be improved at a later date, I'm still unsure about it.**

Squashed PR of #2566

### DIFF
--- a/evennia/contrib/base_systems/components/README.md
+++ b/evennia/contrib/base_systems/components/README.md
@@ -1,0 +1,170 @@
+# Components
+
+_Contrib by ChrisLR 2021_
+
+# The Components Contrib
+
+This contrib introduces Components and Composition to Evennia.
+Each 'Component' class represents a feature that will be 'enabled' on a typeclass instance.
+You can register these components on an entire typeclass or a single object at runtime.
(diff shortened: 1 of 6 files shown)


Griatch:
@ChrisLR Thanks for your work on this, and also for adding the README.md. Merging. 
(btw, there's no need to make a new PR when you squash - you can just force-push your squashed commit to the PR branch).

PR #2682, opened 2022-03-10
Allow at_object_leave to abort a move.

owllex:

Check the return value of `at_object_leave` when moving an object in `move_to`. If it returns False, abort the move, similar to how `at_pre_move` behaves.

For backwards compatibility, any return value other than False is considered a success, since code today returns None.

More hook flexibility for object moves. This allows a container to block movements on one of its children, for example.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ docs/build
 
 # Visual Studio Code (VS-Code)
 .vscode/
+
+# Obsidian 
+.obsidian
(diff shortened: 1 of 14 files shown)


Griatch:
Thanks for the PR; As it happens, I'm adding new `at_pre_object_receive` and `at_pre_object_leave` hooks, both of which can abort the move by returning False. So it'll not be necessary to also abort in this hook ... keeping this unmerged for now.

PR #2686, opened 2022-03-12
component-contrib:add component name as tag on host

Thrag-Zan:

Add component names as a tag with the category "component" on host

Searching objects by their components can be useful

Example:
    Applying the logic of a global Weather script to all rooms that contain the component "weather"

This is my first PR ever and also the first time i use tests so i hope i did it the right way ^^'

### DIFF
--- a/evennia/contrib/base_systems/components/holder.py
+++ b/evennia/contrib/base_systems/components/holder.py
@@ -64,7 +64,7 @@ class ComponentHandler:
         """
         self._set_component(component)
         self.db_names.append(component.name)
-        self.host.tags.add(component.name, category="component")
+        self.host.tags.add(component.name, category="components")
         component.at_added(self.host)
 
     def add_default(self, name):
@@ -85,7 +85,7 @@ class ComponentHandler:
         new_component = component.default_create(self.host)
         self._set_component(new_component)
         self.db_names.append(name)
-        self.host.tags.add(name, category="component")
+        self.host.tags.add(name, category="components")
         new_component.at_added(self.host)
 
     def remove(self, component):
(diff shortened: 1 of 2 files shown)


evennia/contrib/base_systems/components/tests.py
```diff
@@ -107,3 +107,37 @@ def test_can_access_component_regular_get(self):
...
+        handler.remove(rct)
+        assert not self.char1.tags.has(key="test_c", category="component")
+
+    def test_host_remove_by_name_component_tags(self):
+        rct = RuntimeComponentTestC.create(self.char1)
+        handler = self.char1.components
+        handler.add(rct)
+        assert self.char1.tags.has(key="test_c", category="component")
+        handler.remove_by_name("test_c")
+        assert not self.char1.tags.has(key="test_c", category="component")
```
ChrisLR, line 143: Missing newline here

evennia/contrib/base_systems/components/holder.py
```diff
@@ -64,6 +64,7 @@ def add(self, component):
         """
         self._set_component(component)
         self.db_names.append(component.name)
+        self.host.tags.add(component.name, category="component")
```
ChrisLR, line 67: I would use the plural for category
category="components"

ChrisLR (commented):
I like the idea of this PR!
At some point I believe we could use tags as the mechanism for loading components instead of the component names attribute.

Thrag-Zan:
Thank you for your review ^^ 
I applied your changes. (I'm sorry for the extra commit)

We should probably open a discution somewhere about your contrib. 
It is actually really useful and i'm sure we could discuss common features some of us want to add like signals, prototype support, bound command-sets and so on. 

I also found out that i needed to add tags in basetype_posthook_setup(), not really an issue but probably more an unexpected behavior. 

On a side note i think at_added() should also be ran when creating typeclass defined component with ComponentProperty. 
In the end there should be no difference between runtime and typeclass defined components from the user point of view ^^.

ChrisLR (approved):
👍

Griatch (approved):
Looking good, thanks for the PR! @ChrisLR already gave some feedback that was adressed and the unit tests pass too, adding further confidence.

PR #2692, opened 2022-03-22
Add TagField to Contrib Components

ChrisLR:
This allow components to specify tags on the class, using the new descriptor.
It supports providing a default value, having multiple tags or enforcing a single value.
It does this by using the TagHandler and setting the category as component_name__field_name.

Tags set this way will be cleared if the component is removed or if 'del' is called on the field.

It is also possible to set values in the prototypes, like
```
'tags': (
        ('axe', 'melee__melee_type', ''),
    )
```

Similar to regular tags, it can be much more efficient to use than attributes when many entities
use the same 'value', such as Enums like melee_weapon_type.

I also added some doc related to #2686

### DIFF
--- a/evennia/contrib/base_systems/components/README.md
+++ b/evennia/contrib/base_systems/components/README.md
@@ -50,7 +50,29 @@ class Health(Component):
     health = DBField(default=1)
 ```
 
-Note that default is optional and will default to None
+Note that default is optional and will default to None.
+
+Adding a component to a host will also a similarly named tag with 'components' as category.
+A Component named health will appear as key="health, category="components".
+This allows you to retrieve objects with specific components by searching with the tag.
+
+It is also possible to add Component Tags the same way, using TagField.
+TagField accepts a default value and can be used to store a single or multiple tags.
+Default values are automatically added when the component is added.
+Component Tags are cleared from the host if the component is removed.
+
+Example:
+```python
+from evennia.contrib.base_systems.components import Component, TagField
+
+class Health(Component):
+    resistances = TagField()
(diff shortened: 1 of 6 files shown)


strikaco:
It may be too late to address this, but be aware that `component_name__field_name` syntax will conflict with Django queryset filters. It uses the same format (`field_name__filter`...`tags__in=[]`, `hp__gte=23`, etc.).

ChrisLR:
Hmm but in this case its a text value, not a field, so it should not conflict right?

strikaco:
This isn't specific to this PR-- I saw it applied elsewhere in the Components contrib too but hadn't gotten around to pointing it out.

They're strings...for now. You don't know that Django will always handle them that way, or in what contexts they will be used as inputs behind the scenes. Remember the log4j drama? Those were just strings, too ;)

It's not a showstopper, just a vulnerability.

Consider simply not using double-underscores as a delimiter. Try use `_FIELDNAME_` or something as the delimiter (`component_name_FIELDNAME_field_name`). Ugly, but safer IMO.

ChrisLR:
@strikaco something like component_name::field_name ?

strikaco:
> @strikaco something like component_name::field_name ?
> 
> I'm trying to keep it easy for a developer to modify prototypes or while in game. Having the same easy delimiter everywhere helps with that.

Clever! That solves the prospective Django problem, as long as nothing converts colons to underscores. One colon should be enough.

I think that's viable. Try it out :)

Griatch:
An interesting addition. I may consider adding something like a `TagProperty` similar to `AttributeProperty` to echo a similar functionality outside of components. :)

Issue #2677, opened 2022-03-08
[BUG - Develop] Unreliable EvMenu option offset in the webclient


InspectorCaracal:
#### Describe the bug
When viewing an EvMenu in the web client, the first line of the options is offset - in my case, to the right by one space. I tested it in the web client using the current versions of Firefox and Edge, as well as in my non-web mu* clients. The issue occurred only in the web client and appeared the same in both browsers.

#### To Reproduce
1. Create an EvMenu node with options
2. Log in using the web client
3. Access the EvMenu in game

#### Expected behavior
The option lines should be lined up evenly, as they are using a non-web client.

#### Develop-branch commit
473f2e379

InspectorCaracal:
I went deeper into my character creation menu testing things out and discovered that the web client doesn't _always_ offset the first line. The nodes that have the offset versus the ones that don't are consistent - it doesn't change which are which if I navigate back and forth between them - but I can't see any rhyme or reason to it.

Here's a node that has the offset:
```python
def menunode_welcome(caller):
	"""Starting page."""
	text = dedent("""\
		|wWelcome to |mSunnyhill|w, the slice-of-life fantasy game.|n

		The character creation process is somewhat long, but if you have to
		stop or are interrupted/disconnected, you'll be able to resume from
		where you left off by using 'charcreate' with the same character
		name.

		If you need to take a break, enter 'q' to quit. Or, if you need more
		information, try entering 'help'.

		Before you begin, you must choose your |corigin|n. Choose an option
		below for a brief overview of its background and game mechanics.
	""")
	help = fill("In the world of |mSunnyhill|n, there are many different types "
				"of people. Due to the unique nature of the city, you can find "
				"examples of people who hail from virtually anywhere else in "
				"the world.")
	options = []
	options.append({"desc": "Learn about |cDragons|n", "goto": "menunode_dragon_info"})
	options.append({"desc": "Learn about |cElves|n", "goto": "menunode_elf_info"})
	options.append({"desc": "Learn about |cPixies|n", "goto": "menunode_pixie_info"})
	options.append({"desc": "Learn about |cShifts|n", "goto": "menunode_shifter_info"})
	return (text, help), options
```

And here is one that does _not_ have the offset:
```python
def menunode_shifter_info(caller):
	"""Information on playing a Shifter."""
	text = dedent("""\
		|bShifts|n

		Unlike the other three groups, |cShifts|n have little to no cultural
[...]

PR #2711, opened 2022-04-03
Properly display whitespace in the webclient

### LINKED ISSUE #2701: [BUG] (webclient conversion of blank spaces unmanageable in EvForm)
#### Describe the bug
The webclient does not show blank spaces the same way that telnet does, due to the conversion to html. An EvForm that looks perfect from mudlet might look strangely jogged around in the webclient. This is a problem majorly with EvTable and EvForm because in other situations, it can be managed by using different symbols for blank spaces. But when it comes to more dynamic content, it is difficult to predict what management might be necessary.

#### To Reproduce
1.  Use this form and command: https://pastebin.com/YiHLXEA0
2.  Connect through the webclient and run the command.
3.  Connect through mudlet and run the command.

#### Expected behavior
Blank spaces and blank rows inside of EvTables should be properly represented in EvForm format in order not to run into problems with the html conversion for the webclient.

         Evennia version: 0.9.5 (rev 76af934fd)
         OS: nt
         Python: 3.7.9
         Twisted: 20.3.0
         Django: 2.2.5

#### Additional context
The html conversion for the webclient could be fixed to not have this problem: it seems that `|_` works better than tabs or blank spaces.

InspectorCaracal:
Currently, Evennia attempts to control the display of white-space in the webclient by doing server-side parsing. Not only is this resulting in numerous inconsistent spacing issues, but controlling content display should be done on the client, not the server.

This removes the white-space parsing from the server entirely and instead adds a CSS rule to the webclient that forces display of white space.

Closes #2701 (if backported), #2677

### DIFF
--- a/evennia/utils/tests/test_text2html.py
+++ b/evennia/utils/tests/test_text2html.py
@@ -120,13 +120,6 @@ class TestText2Html(TestCase):
         )
         # TODO: doesn't URL encode correctly
 
-    def test_re_double_space(self):
-        parser = text2html.HTML_PARSER
-        self.assertEqual("foo", parser.re_double_space("foo"))
-        self.assertEqual(
-            "a &nbsp;red &nbsp;&nbsp;&nbsp;foo", parser.re_double_space("a  red    foo")
-        )
-
     def test_sub_mxp_links(self):
         parser = text2html.HTML_PARSER
         mocked_match = mock.Mock()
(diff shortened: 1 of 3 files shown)


Griatch (approved):
A nice fix, and going a ways towards moving parsing onto the client as well. Good job, I tested

PR #2717, opened 2022-04-06
Added support for createsuperuser --noinput

faangbait:
If the createsuperuser command detects the three required environment
variables, it will bypass the interactive createsuperuser prompt.

https://docs.djangoproject.com/en/3.2/ref/django-admin/#createsuperuser


Griatch:
Thanks for the PR! This looks fine, but it is already implemented in `develop` branch (in a slightly different way, but with functionally same result). That will merge back into `master` branch once 1.0 is released. So I will close this PR without merge.

PR #2720, opened 2022-04-07
`rpsystem` contrib: cache regex tuples by ID instead of key+aliases

### LINKED ISSUE #2705: [BUG - Develop] `rpsystem` object search can't handle duplicate names
#### Describe the bug
When there are two or more objects of the same name in a list of candidates (e.g. the default of a character's contents and location's contents), only the first one is located.

#### To Reproduce
1. Inherit from the `ContribRP` classes on your `Object` and `Character` class.
2. `create test object`
3. `create test object` again
4. `py self.contents` to confirm both are there
5. `py self.search("test object")` and `py self.search("test object", quiet=True)`
6. See error

#### Expected behavior
It should give a multi-match error, or return all matching objects if `quiet=True`.

#### Develop-branch commit
f093c8bcb

InspectorCaracal:

The search system for the `rpsystem` contrib stores cached regex for non-Character objects under a key+aliases string, so all objects with a given key+aliases match to the first such object cached.

This PR changes the cache identifier to the object's own ID to prevent false references.

Closes #2705

### DIFF
--- a/evennia/contrib/rpg/rpsystem/rpsystem.py
+++ b/evennia/contrib/rpg/rpsystem/rpsystem.py
@@ -321,7 +321,7 @@ def regex_tuple_from_key_alias(obj):
 
     if obj.id not in _REGEX_TUPLE_CACHE:
         permutation_string = " ".join([obj.key] + obj.aliases.all())
-        _REGEX_TUPLE_CACHE[permutation_string] = (
+        _REGEX_TUPLE_CACHE[obj.id] = (
             re.compile(ordered_permutation_regex(permutation_string), _RE_FLAGS),
             obj,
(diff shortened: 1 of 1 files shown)


InspectorCaracal:
I just realized a bit belatedly that this doesn't regenerate the regex after renames or additional aliases, so I'm marking this as a draft until I work that out.

InspectorCaracal:
I've thought about it for a while and I think prefixing the object ID to the beginning of the cache key is the simplest way to disambiguate, as it keys each cached regex tuple by the relevant information: the key and aliases being processed for the regex, and the object being referenced by them.

Griatch:
Thanks for the PR!

PR #2725, opened 2022-04-12
[Components Contrib] Component signals

ChrisLR:

This adds a free form Signals system for components, plugging the existing components code.
It allows decoupling components from each other and also from their host to a certain degree.

A brief example would be a level up signal.
Several components could register to this signal "level_up" using a method decorator.
The host typeclass itself could register a method to this signal.
Then, when the "experience" component decides its time for level up, it can
`self.host.signals.trigger("level_up", new_level)`
Which will in turn call every method that was registered as listener, passing the provided args/kwargs.

Similarly, methods can be marked as "responder", which allows querying a particular signal.
This can be useful to something like retrieving bonuses for a stat across an unknown amount of components.
Something like
`self.host.signals.query("stat_bonuses", stat="strength", aggregate_func=sum)`
Would retrieve all the responses, which in this instance should be a number, pass it to the sum function, and return the total.

This system allows plugging in dynamic callbacks but it does not persist such calls.
This means that components themselves must handle reconnecting the signal when initializing or gracefully
handle the fact that the signal is disconnected on server restart.

It was the feature I removed from the previous PR to avoid bloating and delaying.

_This PR is opened as a draft to get early feedback._

I still need to rewrite README for this addition.
I will probably add a simple priority for listeners and responders

### DIFF
--- a/evennia/contrib/base_systems/components/holder.py
+++ b/evennia/contrib/base_systems/components/holder.py
@@ -5,6 +5,7 @@ This file contains the classes that allow a typeclass to use components.
 """
 
 from evennia.contrib.base_systems import components
+from evennia.contrib.base_systems.components import signals
 
 
 class ComponentProperty:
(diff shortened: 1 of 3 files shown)


Griatch (approved):
You are really building on this contrib! Looking good, I also appreciate the nice doc strings.

PR #2726, opened 2022-04-15
Only initialize AttributeProperty and TagProperty in init_evennia_properties

owllex:

`init_evennia_properties()` attempts to fetch all object properties with a `__set_name__` attribute in order to ensure that AttributeProperty and TagProperty fields are initialized. In python 3.10, this has the side effect of also initializing any fields defined with `@property` (and similar decorators), which is not desired. As this initialization occurs on a new object before any attributes from a `create` call are applied, `@property` fields that cache attribute data will cache the wrong value.

Fixes bug on new object initialization.

The problem only occurs in python 3.10, where the `__set_name__` attribute now exists. Python 3.9 and older do not appear to have attribute (or `__dict__`) on `@property`.

### DIFF
--- a/evennia/typeclasses/models.py
+++ b/evennia/typeclasses/models.py
@@ -332,7 +332,7 @@ class TypedObject(SharedMemoryModel):
         by fetching them once.
         """
         for propkey, prop in self.__class__.__dict__.items():
-            if isinstance(prop, (AttributeProperty, TagProperty)):
+            if inherits_from(prop, AttributeProperty) or inherits_from(prop, TagProperty):
                 try:
                     getattr(self, propkey)
                 except Exception:
--- a/evennia/typeclasses/models.py
+++ b/evennia/typeclasses/models.py
@@ -39,11 +39,12 @@ from django.utils.text import slugify
 from evennia.typeclasses.attributes import (
     Attribute,
     AttributeHandler,
+    AttributeProperty,
     ModelAttributeBackend,
     InMemoryAttributeBackend,
 )
 from evennia.typeclasses.attributes import DbHolder
-from evennia.typeclasses.tags import Tag, TagHandler, AliasHandler, PermissionHandler
+from evennia.typeclasses.tags import Tag, TagHandler, AliasHandler, PermissionHandler, TagProperty
 
 from evennia.utils.idmapper.models import SharedMemoryModel, SharedMemoryModelBase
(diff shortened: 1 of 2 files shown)


evennia/typeclasses/models.py
```diff
@@ -331,7 +332,7 @@ def init_evennia_properties(self):
         by fetching them once.
         """
         for propkey, prop in self.__class__.__dict__.items():
-            if hasattr(prop, "__set_name__"):
+            if isinstance(prop, (AttributeProperty, TagProperty)):
```
Griatch, line 335: Please use `inherits_from` here instead; we want to support child classes.

Griatch (changes requested):
I wasn't aware of `@property` being affected in 3.10, so this is a needed fix ... but one of the initial reasons I didn't tie this to `AttributeProperty/TagProperty` explicitly (I considered it), is that we also want to support children of these properties ... because it's pretty much certain someone will want to add their own functionality to them eventually. :sweat_smile: 
So the better choice here would be to use `utils.inherits_from` instead.

ChrisLR:
> I wasn't aware of `@property` being affected in 3.10, so this is a needed fix ... but one of the initial reasons I didn't tie this to `AttributeProperty/TagProperty` explicitly (I considered it), is that we also want to support children of these properties ... because it's pretty much certain someone will want to add their own functionality to them eventually. 😅 So the better choice here would be to use `utils.inherits_from` instead.

can confirm, component contrib DBField inherits from AttributeProperty

owllex:
Updated to use `inherits_from` instead, and added an explicit test to make sure we pick up subclasses.

evennia/typeclasses/models.py
owllex: It looks like `inherits_from` chokes on anything that doesn't have an mro() (like functions), which raises exceptions and causes a bunch of test failures.

I believe isinstance() does check the entire class heirarchy (see docs: https://docs.python.org/3.9/library/functions.html#isinstance) so I think it might be the right choice here. Subclasses like DBField get picked up correctly.

owllex: I set this PR back to using isinstance for now since inherits_from is breaking tests. Let me know what you think.

Griatch (approved):
Interesting about `isinstance`; I don't think this used to consider subclasses in the past :thinking: Merging to see how it goes :)

Issue #2724, opened 2022-04-12
[Feature Request] Improve performance of rpsystem search


owllex:
#### Is your feature request related to a problem? Please describe.

RPSystem's `ordered_permutation_regex` runs very slowly for for objects with long names since it is a factorial-time algorithm. This can causesnoticeable multi-second hangs to compile all of the regexes in preparation for a search.

#### Describe the solution you'd like

If we sacrifice the ability to match words out of order (e.g. matching "man tall" for a tall man), we can remove the need to generate permutations and can do a much simpler search that also does not require writing cached regexes to persistent attributes.

One way to do this (what I ended up doing in my codebase) is to generate only a single regex for an incoming query. We then run this regex against each candidate object's recog (if available), key, and aliases (if necessary) to find matches. This inverts the regex matching behavior from what is done today and only generates one regex per query. This nd runs much much faster against all objects (especially those that are not RPSystem-aware).

Below is the code I use to handle the search portion of this (where the performance bottleneck lies). It's not a direct drop-in as I built my own RPSystem-analogue that decouples the parsing of emote markers from the searching for objects that match them, but the algorithm is here and could be ported.

```py
def search_candidates_by_appearance(caller, query, candidates, exact=False):
    """
    Search the appearance of provided set of candidates for one or more that
    match a given query.

    This search takes into account appearance attributes set on candidate
    objects if present, as well as remembered names from the caller, if
    available. It specifically excludes the true name (key) of candidates
    hidden by appearances that are unknown to the caller, unless the caller
    is a builder.

    Args:
        caller (Object): The object performing the search.
        query (str): A string to search for.
        candidates (list): The list of possible candidates to search.
        exact (bool): True to require an exact match of the query against any
            valid appearance string.
    Returns:
        list: A list of zero or more matches from the provided candidates.
    Notes:
        Each candidate can be matched either by its appearance, remembered
        name, or a numerical separator attached to either ("2-elf")
    """
    # Strip and extract numerical match prefix, if any.
    index = None
    if index_match := _SEARCH_MULTIMATCH_REGEX.search(query):
        index = int(index_match.group("number"))
        query = index_match.group("name")
    is_builder = caller.permissions.check("Builder")

    # Add display name and appearance for each candidate. Recognized names are
    # handled by get_display_name if supported.
    candidate_text = {}
    for item in candidates:
        texts = list(item.aliases.all())
        is_character = hasattr(item, "is_typeclass") and item.is_typeclass(
[...]

PR #2728, opened 2022-04-15
Streamline `rpsystem` sdesc processing

InspectorCaracal:
Moves the SdescHandler to the base `ContribRPObject` class and streamlines the recog/sdesc fetching and processing for ease of customization.

It also rewrites the sdesc parsing for searches and emotes, moving the regex processing onto the search/emote text and eliminating the regex calculations and cache for sdescs, as per #2724

I've been using the rpsystem contrib for a while and it has a lot of quirks which have made basic customizations a bit difficult. Most of the streamlining changes are intended to make that customization simpler by minimizing how much code you have to modify in order to overload display options, such as sdesc vs recog coloring.

Additionally, the current ordered permutation regex generation makes loading multiple objects with multi-word names cause sizeable hangs in the game (see issue above), which can be a real problem in any game with descriptive object names.

#### Other information
~~This is a draft because it's currently a port of some of my own customizations back into the base contrib, and while I won't have an opportunity to test it and review the unit tests for at least a couple days, I'd like to make it available for feedback.~~

Please feel free to critique/correct any regex that could be improved; regular expressions are not my forte.

### DIFF
--- a/evennia/contrib/rpg/rpsystem/rpsystem.py
+++ b/evennia/contrib/rpg/rpsystem/rpsystem.py
@@ -694,7 +694,7 @@ class SdescHandler:
         self.sdesc = self.obj.attributes.get("_sdesc", default=self.obj.key)
         sdesc_regex = self.obj.attributes.get("_sdesc_regex", default="")
         if not sdesc_regex:
-            permutation_string = " ".join([self.key] + self.aliases.all())
+            permutation_string = " ".join([self.obj.key] + self.obj.aliases.all())
             sdesc_regex = ordered_permutation_regex(permutation_string)
         self.sdesc_regex = re.compile(sdesc_regex, _RE_FLAGS)
 
--- a/evennia/contrib/rpg/rpsystem/rpsystem.py
+++ b/evennia/contrib/rpg/rpsystem/rpsystem.py
@@ -284,7 +284,7 @@ def parse_language(speaker, emote):
         # the key is simply the running match in the emote
         key = f"##{imatch}"
         # replace say with ref markers in emote
-        emote = "{start}{key}{end}".format( start=emote[:istart], key=key, end=emote[iend:] )
+        emote = "{start}{{{key}}}{end}".format( start=emote[:istart], key=key, end=emote[iend:] )
         mapping[key] = (langname, saytext)
 
     if errors:
@@ -408,6 +408,8 @@ def parse_sdescs_and_recogs(sender, candidates, string, search_mode=False, case_
                 # save current index as end point of matched text
                 iend = i
 
+            # save search string
+            matched_text = "".join(tail[1:iend])
             # recombine remainder of emote back into a string
             tail = "".join(tail[iend+1:])
 
@@ -469,7 +471,7 @@ def parse_sdescs_and_recogs(sender, candidates, string, search_mode=False, case_
             # multimatch error
             refname = marker_match.group()
             reflist = [
-                "{name}{sep}{num} ({text}{key})".format(
+                "{num}{sep}{name} ({text}{key})".format(
                     num=inum + 1,
                     sep=_NUM_SEP,
                     name=_RE_PREFIX.sub("", refname),
@@ -581,7 +583,7 @@ def send_emote(sender, receivers, emote, anonymous_add="first", **kwargs):
         receiver_sdesc_mapping = dict(
             (
                 ref,
-                obj.get_display_name(receiver, noid=True),
+                obj.get_display_name(receiver, ref=ref, noid=True),
             )
             for ref, obj in obj_mapping.items()
         )
@@ -1109,7 +1111,7 @@ class CmdRecog(RPCommand):  # assign personal alias to object in room
             caller.msg(_EMOTE_NOMATCH_ERROR.format(ref=sdesc))
         elif nmatches > 1:
             reflist = [
-                "{sdesc}{sep}{num} ({recog}{key})".format(
+                "{num}{sep}{sdesc} ({recog}{key})".format(
                     num=inum + 1,
                     sep=_NUM_SEP,
                     sdesc=_RE_PREFIX.sub("", sdesc),
@@ -1461,7 +1463,7 @@ class ContribRPObject(DefaultObject):
         
         # add dbref is looker has control access and `noid` is not set
         if self.access(looker, access_type="control") and not kwargs.get("noid",False):
-            sdesc = f"{sdesc}{self.id}"
+            sdesc = f"{sdesc}(#{self.id})"
         
         return self.get_posed_sdesc(sdesc) if kwargs.get("pose", False) else sdesc
 
--- a/evennia/contrib/rpg/rpsystem/rpsystem.py
+++ b/evennia/contrib/rpg/rpsystem/rpsystem.py
@@ -526,7 +526,7 @@ def send_emote(sender, receivers, emote, anonymous_add="first", **kwargs):
     # if anonymous_add is passed as a kwarg, collect and remove it from kwargs
     if "anonymous_add" in kwargs:
         anonymous_add = kwargs.pop("anonymous_add")
-    self_refs = (f"{skey}{ref}" for ref in ('t','^','v','~',''))
+    self_refs = [f"{skey}{ref}" for ref in ('t','^','v','~','')]
     if anonymous_add and not any(1 for tag in obj_mapping if tag in self_refs):
         # no self-reference in the emote - add to the end
         if anonymous_add == "first":
--- a/evennia/contrib/rpg/rpsystem/rpsystem.py
+++ b/evennia/contrib/rpg/rpsystem/rpsystem.py
@@ -150,7 +150,6 @@ Extra Installation Instructions:
 """
 import re
 from re import escape as re_escape
-import itertools
 from string import punctuation
 from django.conf import settings
 from evennia.objects.objects import DefaultObject, DefaultCharacter
@@ -159,8 +158,6 @@ from evennia.commands.command import Command
 from evennia.commands.cmdset import CmdSet
 from evennia.utils import ansi, logger
 from evennia.utils.utils import lazy_property, make_iter, variable_from_module
-
-_REGEX_TUPLE_CACHE = {}
 
 _AT_SEARCH_RESULT = variable_from_module(*settings.SEARCH_AT_RESULT.rsplit(".", 1))
 # ------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


InspectorCaracal:
All right, I've tested this now and while I think they're good changes on their own, the search lag delay is still very much a thing. The main differences now is that it frontloads the hang to when you log in while carrying a number of longer-name objects, or you enter a room that contains a number of longer-named objects, rather than happening on the search command.

I could postpone updating the unit tests and wrap a rewrite of the search to work as per owllex's suggestion (in the issue linked above) into this PR. If no one objects, I'll do that next.

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -619,66 +534,41 @@ def send_emote(sender, receivers, emote, anonymous_add="first", **kwargs):
     # if anonymous_add is passed as a kwarg, collect and remove it from kwargs
     if "anonymous_add" in kwargs:
         anonymous_add = kwargs.pop("anonymous_add")
-    if anonymous_add and not any(1 for tag in obj_mapping if tag.startswith(skey)):
+    self_refs = [f"{skey}{ref}" for ref in ('t','^','v','~','')]
+    if anonymous_add and not any(1 for tag in obj_mapping if tag in self_refs):
         # no self-reference in the emote - add to the end
-        obj_mapping[skey] = sender
         if anonymous_add == "first":
+            skey = skey + 't'
```
Griatch, line 541: `+=` ?

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -619,66 +534,41 @@ def send_emote(sender, receivers, emote, anonymous_add="first", **kwargs):
...
             possessive = "" if emote.startswith("'") else " "
             emote = "%s%s%s" % ("{{%s}}" % skey, possessive, emote)
         else:
             emote = "%s [%s]" % (emote, "{{%s}}" % skey)
+        obj_mapping[skey] = sender
 
     # broadcast emote to everyone
     for receiver in receivers:
         # first handle the language mapping, which always produce different keys ##nn
         receiver_lang_mapping = {}
```
Griatch, line 551: It seems this is no longer needed since `receiver_lang_mapping` is now created in-place on the lines below.

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -1539,27 +1397,26 @@ def get_display_name(self, looker, **kwargs):
 
         Keyword Args:
             pose (bool): Include the pose (if available) in the return.
+            ref (str):   Specifies the capitalization for the displayed name.
```
Griatch, line 1400: Why `ref` as a name for this? Make the arg more readable and relevant. This also seems like it could do with an example for how it should look.

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -1539,27 +1397,26 @@ def get_display_name(self, looker, **kwargs):
...
         else:
             try:
-                recog = looker.recog.get(self)
+                sdesc = looker.get_sdesc(self, ref=ref)
             except AttributeError:
-                recog = None
-            sdesc = recog or (hasattr(self, "sdesc") and self.sdesc.get()) or self.key
-        pose = " %s" % (self.db.pose or "") if kwargs.get("pose", False) else ""
+                sdesc = self.sdesc.get()
+        pose = " %s" % (self.db.pose or "is here.") if kwargs.get("pose", False) else ""
```
Griatch, line 1419: The `"is here."` string should be lifted out of this code and put in a more easily modifiable location. For example, it can be a property at the class level.

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -1635,21 +1487,23 @@ def get_display_name(self, looker, **kwargs):
...
             characters stand out from other objects.
 
         """
-        idstr = "(#%s)" % self.id if self.access(looker, access_type="control") else ""
+        idstr = "(#%s)" % self.id if self.access(looker, access_type="control") and not kwargs.get("no_id",False) else ""
+        ref = kwargs.get("ref","~")
+    
         if looker == self:
-            sdesc = self.key
+            sdesc = self.process_recog(self.key,self)
```
Griatch, line 1498: Why would you `recog` yourself?

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -1635,21 +1487,23 @@ def get_display_name(self, looker, **kwargs):
...
             try:
-                recog = looker.recog.get(self)
+                sdesc = looker.get_sdesc(self, process=True, ref=ref)
             except AttributeError:
-                recog = None
-            sdesc = recog or (hasattr(self, "sdesc") and self.sdesc.get()) or self.key
+                sdesc = self.sdesc.get()
         pose = " %s" % (self.db.pose or "is here.") if kwargs.get("pose", False) else ""
-        return "|c%s|n%s%s" % (sdesc, idstr, pose)
+        return "%s%s%s" % (sdesc, idstr, pose)
```
Griatch, line 1505: Why not get rid of the `%s` formatting when you are at it? Also, if the intent of this refactor is (in part) to make this easier to override and customize, maybe it would be an idea to have these arguments pass through a helper method you can override instead?

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -1679,8 +1533,36 @@ def at_pre_say(self, message, **kwargs):
...
+        """
+        Single hook method to handle getting recogs with sdesc fallback in an
+        aware manner, to allow separate processing of recogs from sdescs.
+        Gets the sdesc or recog for obj from the view of self.
+
+        Args:
+            obj (Object): the object whose sdesc or recog is being gotten
+        Keyword Args:
+            process (bool): If True, the sdesc/recog is run through the
+                appropriate process_X method.
```
Griatch, line 1549: This is not very informative. What does it mean to run through the `process_X` methods? Why would you not want to use them? Are they just hooks to customize the output? Also, there are only two `process_X` methods: `.process_recog` and `.process_sdesc` - you can spell this out in full  (note the use of period to indicate they are on this class, but it can be good to also mention this explicitly).

evennia/contrib/rpg/rpsystem/tests.py
```diff
@@ -96,7 +96,7 @@ def test_obfuscate_whisper(self):
 recog02 = "Mr Receiver2"
 recog10 = "Mr Sender"
 emote = 'With a flair, /me looks at /first and /colliding sdesc-guy. She says "This is a test."'
-case_emote = "/me looks at /first, then /FIRST, /First and /Colliding twice."
+case_emote = "/Me looks at /first, then /FIRST, /First and /Colliding twice."
```
Griatch, line 99: This seems like a regression? You should be able to use `/me` in every situation?

evennia/contrib/rpg/rpsystem/tests.py
```diff
@@ -156,12 +126,8 @@ def test_recog_handler(self):
         self.speaker.recog.add(self.receiver2, recog02)
         self.assertEqual(self.speaker.recog.get(self.receiver1), recog01)
         self.assertEqual(self.speaker.recog.get(self.receiver2), recog02)
-        self.assertEqual(
-            self.speaker.recog.get_regex_tuple(self.receiver1)[0].pattern,
-            "/[0-9]*-*Mr\\ Receiver(?=\\W|$)+|/[0-9]*-*Receiver(?=\\W|$)+|/[0-9]*-*Mr(?=\\W|$)+",
-        )
         self.speaker.recog.remove(self.receiver1)
-        self.assertEqual(self.speaker.recog.get(self.receiver1), sdesc1)
+        self.assertEqual(self.speaker.recog.get(self.receiver1), None)
```
Griatch, line 130: This is a backwards-incompatible change, for what seems to be rather limited gain. Is there a reason why we can't just return the sdesc/key if no recog was registered?

evennia/contrib/rpg/rpsystem/tests.py
```diff
@@ -212,18 +178,18 @@ def test_send_emote(self):
         rpsystem.send_emote(speaker, receivers, emote, case_sensitive=False)
         self.assertEqual(
             self.out0,
-            "With a flair, |bSender|n looks at |bThe first receiver of emotes.|n "
+            "With a flair, |mSender|n looks at |bThe first receiver of emotes.|n "
```
Griatch, line 181: Is there a purpose to changing this (apart from you preferring one color over the other :wink: )

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -1539,27 +1397,26 @@ def get_display_name(self, looker, **kwargs):
...
                 including the DBREF if this user is privileged to control
                 said object.
 
-        Notes:
-            The RPObject version doesn't add color to its display.
-
         """
-        idstr = "(#%s)" % self.id if self.access(looker, access_type="control") else ""
+        idstr = "(#%s)" % self.id if self.access(looker, access_type="control") and not kwargs.get("no_id", False) else ""
+        ref = kwargs.get("ref","~")
```
Griatch, line 1410: It would be an idea to also update the `%s` when you are at it (the original code is old enough to predate things like .format and f-strings)

Griatch (changes requested):
Thanks for your work on this! I have not run the code yet, but it looks like a reasonable refactor all considered. As expected for a large suite of code, I do have some comments and feedback to address.  :)

(A hint for the future is that you can comment your PR to clarify changes you did for the benefit of the reviewer.)

InspectorCaracal:
Thanks for the tip and all the feedback! It'll take me a while to work through all of it but I will. :)

evennia/contrib/rpg/rpsystem/tests.py
InspectorCaracal: The original code had `/me` explicitly preserving case every time, but I removed the exception.

I can guess at the logic behind preserving case, but it only holds up from the perspective of the sender - since they always see their own key. If you did e.g. the first emote test with case sensitivity but excluding the case sensitivity for self-references, the output to Receiver1 would look like this:
```
"With a flair, |bA nice sender of emotes|n looks at |mReceiver1|n and "
'|banother nice colliding sdesc-guy for tests|n. She says |w"This is a test."|n',
```

And if your sdesc was `a nice sender of emotes` and you tried to explicitly capitalize with `/Me` like you would for the other references, it wouldn't work.

evennia/contrib/rpg/rpsystem/tests.py
InspectorCaracal: I admit I hadn't considered backwards compatability, but yes. Using the sdesc/key as a fallback for recog makes the separation of `process_sdesc` and `process_recog` completely meaningless, as the recog handler takes precedence to the sdesc handler. As a result, everything would processed via `process_recog` if the viewing object has a recog handler, regardless of whether it was a recog or an sdesc, and `process_sdesc` would only be used by viewing objects that have no recog handler.

The issue isn't noticeable in the original code because `process_recog` just calls `process_desc`, making them indistinguishable, but caused me a lot of headaches when I tried to add a separate format for recogs versus unrecognized sdescs. Returning `None` if there's no recog and explicitly coding in a fallback in the new `get_sdesc` hook are how I ultimately resolved that.

evennia/contrib/rpg/rpsystem/tests.py
InspectorCaracal: It's intended to highlight the difference between a recog and an sdesc by using different colors to show which is which. (I just picked magenta because it wasn't in use yet xD)

evennia/contrib/rpg/rpsystem/rpsystem.py
InspectorCaracal: It's the processing code, not the actual recog handler - the logic there being, you recognize yourself, so it should be processed like other recognized character names. Although I can see an argument against differentiating between the two for the name line in `return_appearance` - it'd only take a small amount of rewriting, if you think that would be a better result.

evennia/contrib/rpg/rpsystem/rpsystem.py
InspectorCaracal: The purpose of it is to include a mapping for the self-reference sdesc key that was just added via `anonymous_add`. I just moved it to the end of the `if` block instead of the beginning in order to account for the potentially added case ref.

evennia/contrib/rpg/rpsystem/tests.py
Griatch: Ok, I can see that argument. I would then strongly suggest to also add unit tests for `/me` and `/Me` respectively.

evennia/contrib/rpg/rpsystem/tests.py
Griatch: Hm. Yes, I can see the argument of giving the implementer more power to separate between recog/non-recog'ed desciptors for games that want to do that. Thanks for elaborating! 
It's not ideal from a compatibility perspective though, so this will need to be properly documented in the CHANGELOG (I can do that when this is to be merged, or you can do it if you can sum changes into max two lines).

evennia/contrib/rpg/rpsystem/rpsystem.py
Griatch: I don't see what you mean? This declaration of an empty `receiver_lang_mapping` is immediately followed by an `if...else` block where both conditions generate `receiver_lang_mapping` from scratch.

evennia/contrib/rpg/rpsystem/tests.py
Griatch: Since there is a change to separate `recog`'d and un-recoged `sdesc`s now, there needs to be new unit tests to also cover the non-recog'd case. 
Also, since `RecogHandler.get()` has changed in purpose, its docstring needs to be updated' both its `Returns:` stanza as well as the general instruction on how this is to be used.

Griatch:
@InspectorCaracal Btw, I see you are making changes - please make a comment here when you feel you are ready for a new review, so I don't go back over stuff before you feel it's ready. :wink:

InspectorCaracal:
I'm definitely not there yet, but I'll post here when I am!

evennia/contrib/rpg/rpsystem/rpsystem.py
InspectorCaracal: Ohhhh, I see where I got confused, you're suggesting I take the unnecessary declaration out, not remarking on the change I made. Good catch, I'll do that.

InspectorCaracal:
@Griatch Okay, I think this is ready for you to take another look.

One thing I realized while replacing % formatting: the current `.return_appearance` methods are coded from before all the new hooks (e.g.`.get_visible_contents`) were added. Should I update them to the new structure? I would just go for it, but it'll make it more difficult to backport so I thought I'd ask first.

Griatch:
@InspectorCaracal Sorry for being slow to follow up on this; I've been away and then distracted by RL stuff :/ I have skimmed over the code and it looks better now; I'll need to run it here a bit first before I do the final review of it though. It's a good point about the `return_appearance`, it's not updated for 1.0-dev changes ... I won't require it for this PR of course, but if you want to take on updating that as well, that would be useful indeed. :+1:

InspectorCaracal:
No worries about the delay, it's a big PR and this isn't like your only job or anything.

> I won't require it for this PR of course, but if you want to take on updating that as well, that would be useful indeed. 👍

I don't think it'll be particularly difficult working in an update on the appearance, so I'll go ahead and do that over the next couple days.

InspectorCaracal:
Actually, going in to update `return_appearance` has me thinking about some effects of the current design (such as a box containing multiple objects without a pose outputting multiple lines of `<object> is here.`) which I think I'd like to consider in more depth and possibly file a separate PR, so I won't include that update here.

Griatch:
@InspectorCaracal Ok then. :+1:

Griatch:
Another comment to test discord integration sync

Griatch (approved):

PR #2739, opened 2022-05-09
Rewrite text2html color and style parsing

### LINKED ISSUE #2244: [BUG] Underline is not terminated properly in EvForm Webclient
#### Describe the bug

When mapping into an EvForm on the webclient and using the text tag |u to start underlining, |n does not end it and return to normalized text.

#### To Reproduce
1. Create an EvForm with something simple like xxxxxx1xxxxxx xxxxxx2xxxxxx
2. Map values 1 and 2, including |u in the first map. Ex: "|uUnderline|n" and "Not Underline"
3. The underline will not end in the displayed form in the webclient. (Have not tested other clients.) It will extend and keep the underline throughout the entirety of the rest of the form. (If color is used it will also taint the "default color" of the rest of the form. For example if you used "|y|uUnderline|n" all further text in the form not contained in text tags will be turned yellow.

#### Expected behavior
|n should remove all text formatting. (Theory is that some HTML tags in web client are not closed properly)

Evennia 0.9.5 (rev d4053b3bc) (rev d4053b3bc)
OS: posix
Python: 3.7.5
Twisted: 20.3.0
Django: 2.2.16

#### Additional context
Not sure how to solve, but I'm sure the complexity arises and is a problem of maybe not knowing what HTML tags are currently "open" beforehand to automatically close them with their counter-tag when using |n within.

It should be noted that in other areas like using SAY for example:
"|uUnderline|n Not Underline
Yields the expected result and properly terminates the Underline.

InspectorCaracal:
This replaces the current regex-substitution parsing with state-tracking iterated substring substitution.

The regex method is based on the assumption that ANSI codes function like HTML tags, with a start marker and a stop marker. However, ANSI codes act as state switches and don't strictly speaking have end markers at all, but instead a "restore" switch (`|n`). This difference in design and use is also what resulted in the trailing/unclosed style classes, such as underlines and blinks.

With that in mind, I also opted to replace the nested `span` approach (which for this method would require keeping track of how many spans there were, and what kind) with consecutive `span` elements that are styled with the applicable classes for that segment of text, where the CSS class list represents the ANSI style switches.

Lastly, as part of this rewrite, I revised the inverse code (`|*`) to behave the expected way of swapping the foreground and background colors, rather than the current dotted outline styling. This PR does not alter the CSS file, however, so the `inverse` class is still there, just unused.

The underline bug was bothering me and the only way to fix it was a full rewrite of the parser, so I did a full rewrite of the parser. >.>

Closes #2244

### DIFF
--- a/evennia/utils/tests/test_text2html.py
+++ b/evennia/utils/tests/test_text2html.py
@@ -7,20 +7,20 @@ import mock
 
 
 class TestText2Html(TestCase):
-    def test_re_color(self):
+    def test_format_styles(self):
         parser = text2html.HTML_PARSER
-        self.assertEqual("foo", parser.re_color("foo"))
+        self.assertEqual("foo", parser.format_styles("foo"))
         self.assertEqual(
             '<span class="color-001">red</span>foo',
-            parser.re_color(ansi.ANSI_UNHILITE + ansi.ANSI_RED + "red" + ansi.ANSI_NORMAL + "foo"),
+            parser.format_styles(ansi.ANSI_UNHILITE + ansi.ANSI_RED + "red" + ansi.ANSI_NORMAL + "foo"),
         )
         self.assertEqual(
             '<span class="bgcolor-001">red</span>foo',
-            parser.re_color(ansi.ANSI_BACK_RED + "red" + ansi.ANSI_NORMAL + "foo"),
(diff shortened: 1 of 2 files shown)


Griatch (commented):
This is a very cool change, and one which I have been posponing doing; thanks for taking it on. I'll need to run it a bit more, but at first glance this looks very nice.

PR #2748, opened 2022-05-31
Bugfix 2747 get_by_tag category list fails on None

### LINKED ISSUE #2747: [BUG - Develop] `get_by_tag` category list fails on None
#### Describe the bug
When using a list for the `category` kwarg that contains an entry of `None` (which is a valid category), it fails with a traceback. Presumably due to the use of `sorted` here <https://github.com/evennia/evennia/blob/develop/evennia/typeclasses/managers.py#L289>

#### To Reproduce
1. Add tags to an object (e.g. a room), both with categories and without. Let's say:
```
tag here = test
tag here = tag1:cat1
tag here = tag2:cat2
```
2. `get_by_tag(key=["tag1","tag2"], category=["cat1","cat2"])` returns the current room
3. `get_by_tag(key=["tag1","test"], category=["cat1",None])` _should_ return the current room as well, but instead will traceback.

#### Expected behavior
Since `None` is a valid tag category, the function should be capable of searching for tags with categories of strings _and_ of `None`.

#### Develop-branch commit
7f4769bd9

#### Additional context
Traceback:
```
File "./evennia/evennia/typeclasses/managers.py", line 289, in get_by_tag
    unique_categories = sorted(set(categories))
TypeError: '<' not supported between instances of 'NoneType' and 'str'
```

ChrisLR:
This removes a sorted call on unique categories that is only used for building OR cases in the query

Fixing issues

Fixes #2747

### DIFF
--- a/evennia/typeclasses/managers.py
+++ b/evennia/typeclasses/managers.py
@@ -286,7 +286,7 @@ class TypedObjectManager(idmapper.manager.SharedMemoryManager):
         categories = make_iter(category) if category else []
         n_keys = len(keys)
         n_categories = len(categories)
-        unique_categories = sorted(set(categories))
+        unique_categories = set(categories)
         n_unique_categories = len(unique_categories)
 
         dbmodel = self.model.__dbclass__.__name__.lower()


Griatch:
Looks good. Thanks for the fix!

PR #2749, opened 2022-06-01
Update and simplify `options_formatter`

InspectorCaracal:
Updates `EvMenu.options_formatter` to use fstrings and changes the "manual" row/column padding to use `EvColumn` instead.

This also updates the docstring to remove the erroneous `caller` arg.


Griatch:
Thanks for the PR - looks like a good change. I don't recall if `EvColumn` was a thing in `EvTable` yet when this was first written, but either way I didn't think to use it to make the columns of EvMenu. :)

PR #2750, opened 2022-06-01
Fix webclient error on server-side echo

### LINKED ISSUE #2742: [BUG - Develop] LOCALECHO flag has no effect when using the webclient
#### Describe the bug
When `option LOCALECHO=True` and using the webclient, the input text isn't echoed back to the client

#### To Reproduce
1. Set up an account in a game and connect with the webclient
2. set the `LOCALECHO` option to True
3. Enter any commands

#### Expected behavior
The flag is supposed to cause the portal to echo back whatever you sent, regardless of protocol.

#### Develop-branch commit
7f4769bd9

InspectorCaracal:
Adds a check to the webclient to make sure a cmdid has a stored callback before trying to retrieve it.

Closes #2742

### DIFF
--- a/evennia/web/static/webclient/js/evennia.js
+++ b/evennia/web/static/webclient/js/evennia.js
@@ -149,7 +149,7 @@ An "emitter" object must have a function
         //   kwargs (obj): keyword-args to listener
         //
         emit: function (cmdname, args, kwargs) {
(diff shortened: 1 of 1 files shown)


Griatch:
Thanks! Good catch!

PR #2755, opened 2022-06-07
[Bugfix] 2736 Allow cmdsets to remove and retrieve by key

### LINKED ISSUE #2736: [BUG - Develop] CmdSet.remove() does not take string args
#### Describe the bug
Not sure if it's a documentation error or a code error - this claims that you can pass it the key of a command to remove, but causes a traceback.
https://github.com/evennia/evennia/blob/7f4769bd9d3403b8011998b0c903a490caa07be9/evennia/commands/cmdset.py#L566-L568 

#### To Reproduce
1. Attempt to remove a command from any CmdSet with the key, e.g. `self.remove("charcreate")`
2. The code gives a traceback because it cannot handle a string.

#### Expected behavior
Either it should remove the command in the set with that key or the docstring is inaccurate.

#### Develop-branch commit
7f4769bd9

#### Additional context
The traceback:
```
  File ".../evennia/evennia/commands/cmdsethandler.py", line 208, in import_cmdset
    cmdsetclass = cmdsetclass(cmdsetobj)
  File ".../evennia/evennia/commands/cmdset.py", line 222, in __init__
    self.at_cmdset_creation()
  File "./accounts/commands.py", line 27, in at_cmdset_creation
    self.remove("charcreate")
  File ".../evennia/evennia/commands/cmdset.py", line 572, in remove
    if cmd.key.startswith("__"):
AttributeError: 'str' object has no attribute 'key'
```

ChrisLR:
This allows removing and retrieving commands in a CmdSet by using the key

Fixing bugs

This fixes #2736 

I also changed the add's handling of the remove duplicates as it seemed to be  removing duplicates at every iteration instead of just once at the end.


Griatch (changes requested):
I don't see why this complexity is needed in order to implement this ... you now have two parallel stores (`.commands` and `.commands_by_key`) that must be maintained separately and which could theoretically go out of sync with one another. 

It would seem much simpler to just implement changes in `remove` and `get` methods here and simply loop over the commands to see if a key matches - potentially caching the result as an optimization.

ChrisLR:
@Griatch true, I was gradually aiming at replacing .commands to only use .commands_by_key, since we'll be iterating all the time, why not simply get by key all the time... but yeah I don't mind I'll make the change

Griatch:
@ChrisLR It's not going to be used all the time; there is a reason noone noticed this missing - all internal operations are done by-command instance. The main use of this is as a convenience when you manually remove a default command in `default_cmdsets`.

ChrisLR:
I meant we iterate over commands often, while every command has a key.. it makes sense to retrieve by key internally.
Anyway, it can wait (or just not be done)

evennia/commands/cmdset.py
```diff
@@ -559,6 +556,11 @@ def add(self, cmd, allow_duplicates=False):
                 except ValueError:
                     system_commands.append(cmd)
 
+        self.commands = commands
+        if not allow_duplicates:
+            # extra run to make sure to avoid doublets
+            self.commands = list(set(self.commands))
+
```
ChrisLR, line 563: This was originally nested, so it would try to make it unique multiple times, this will only do it once at the end.

Griatch:
@ChrisLR Thanks for the change! It looks like the unit tests are not passing here though, and it's cmd-related.

Griatch:
@ChrisLR As a reminder, the unit tests still don't pass on this.

ChrisLR:
@Griatch sorry for the delay, RL getting in the way again.

Griatch:
@ChrisLR As this was merged, it appears to have introduced a regression: The `NO_INPUT` syscommand stopped working; for example you can't accept deleting an object in-game (e.g. `del box? [yes]/no?` prompt) just by pressing `<return>` anymore. 
Not sure why this PR would cause this, but I stepped back in GIT history and the functionality returns if checking out a PR just before merging this particular PR. Will inevestigate further later.

Griatch:
For  completeness: The regression was resolved.

PR #2764, opened 2022-06-09
Fix ANSI_NORMAL override in text2html

InspectorCaracal:
This patch resets style states on ANSI_NORMAL flag regardless of `clean` status.

Fixes the webclient displaying undesired styling that should be cleared by the ANSI_NORMAL flag.


avalonhope (approved):

Griatch:
Looks reasonable :+1:

Issue #2785, opened 2022-06-28
[Feature Request] Specify _no_ endsep for `iter_to_str`


InspectorCaracal:
#### Is your feature request related to a problem? Please describe.
Currently, the `iter_to_str` function uses a comma as the final separator when you give it a blank string as the `endsep` parameter. Since otherwise specifying an end separator completely replaces the comma, this is inconsistent and very unintuitive behavior - and more importantly, makes it absolutely impossible to use the function without a final separator.

#### Describe the solution you'd like
I would prefer if the blank string were treated as such, rather than interpreted as something it isn't. If you want your final separator to be a comma, having to specify it as a comma (the way you would have to put the comma in any other final separator) would be more consistent.

#### Describe alternatives you've considered
It doesn't look as complicated to do as I thought so I'll put in a PR with a proposed change.

PR #2787, opened 2022-06-28
Update separator handling for `iter_to_str`

InspectorCaracal:
Changes the behavior when `endsep=''` to actually have no final separator instead of presuming a comma. While I feel this is a valuable change, it would affect any code that uses the empty string as a way to simply erase the "and" from the default format, causing a minor break in backwards compatibility.

Also, since I was there, I expanded the functionality slightly to allow for defining a custom separator. _This_ change has no effect on existing usage and would be fully backwards-compatible.

#### Motivation for adding to Evennia and Other Info
Fulfilling #2785

### DIFF
--- a/evennia/utils/tests/test_utils.py
+++ b/evennia/utils/tests/test_utils.py
@@ -57,13 +57,15 @@ class TestDedent(TestCase):
 class TestListToString(TestCase):
     """
     Default function header from utils.py:
-    list_to_string(inlist, endsep=", and", addquote=False)
+    list_to_string(inlist, sep=",", endsep=", and", addquote=False)
 
     Examples:
-     with default endsep:
+     with defaults:
         [1,2,3] -> '1, 2, and 3'
      with endsep==',':
         [1,2,3] -> '1, 2, 3'
+     with sep==';' and endsep==';':
+        [1,2,3] -> '1; 2; 3'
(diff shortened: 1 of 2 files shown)


evennia/utils/tests/test_utils.py
```diff
@@ -57,21 +57,27 @@ def test_dedent(self):
...
      with endsep=='and':
         [1,2,3] -> '1, 2 and 3'
-     with addquote and endsep
+     with addquote and endsep="and"
         [1,2,3] -> '"1", "2" and "3"'
     """
 
     def test_list_to_string(self):
-        self.assertEqual("1, 2, 3", utils.list_to_string([1, 2, 3], endsep=""))
-        self.assertEqual('"1", "2", "3"', utils.list_to_string([1, 2, 3], endsep="", addquote=True))
```
Griatch, line 73: These should not be removed, but updated to match the new behavior.

Griatch (changes requested):
I think the change should be fine, as you say it's more intuitive rather than assume a hidden default in the function. I have some comments before this can be approved: 
- This case needs coverage by unit tests using `sep=''` though, since this is still a valid input.
- The docstrings can also be updated to mention this case.

Griatch (approved):

PR #2788, opened 2022-06-29
Fix text + prompt msg function

### LINKED ISSUE #2783: [BUG - Develop] `text` overrides `prompt` when in the same `msg` call
#### Describe the bug
According to [this How-To](https://www.evennia.com/docs/1.0-dev/Howtos/Command-Prompt.html), it's intended that you should be able to do `char.msg("a text message", prompt="some prompt text")` and have the server send both the normal message and also a prompt message.

However, all of the portals appears to be designed so that, if `prompt` is in the kwargs, it _replaces_ the `prompt` string with `text` and sends that as a prompt.

#### Expected behavior
I'd expect it to behave as described in the documentation, where it sends the text field as a regular text message and the prompt field as a prompt message.

Even if only one type of message could be sent at a time, in _no_ situation would I expect the `text` kwarg to replace the `prompt` kwarg and be sent in its place.

#### Develop-branch commit
6edf9064c

#### Additional context
It doesn't look too difficult to fix this, just a bit tedious, so I'll take a stab at it over the next week.

InspectorCaracal:
Fixes a pass-by-reference bug that resulted in msg text overriding prompt text. Also makes sure that msg text is processed before other cmd types in the msg (such as prompt).

Bug fixing

Closes #2783

### DIFF
--- a/evennia/server/sessionhandler.py
+++ b/evennia/server/sessionhandler.py
@@ -270,11 +270,11 @@ class SessionHandler(dict):
             else:
                 rkwargs[key] = [[_validate(data)], {}]
             rkwargs[key][1]["options"] = dict(options)
-        # make sure that any "text" message will be processed first
-        # by putting it at the beginning
-        if "text" in rkwargs:
-            text = rkwargs.pop("text")
-            rkwargs = { "text": text } | rkwargs
+        # make sure that any "prompt" message will be processed last
+        # by moving it to the end
+        if "prompt" in rkwargs:
+            prompt = rkwargs.pop("prompt")
+            rkwargs["prompt"] = prompt
         return rkwargs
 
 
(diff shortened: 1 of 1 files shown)


Griatch (commented):
Nice catch of the reference! 
Is there a purpose to forcing "text" to be processed first though? The whole point of `outputfuncs` is that "text" is just one of the types of messages that Evennia sends ...

InspectorCaracal:
It's a minor QOL change - people playing MU* expect the message part of an action to be first and any other things to happen after it. It doesn't make any apparent difference for the webclient, but there's been at least a couple people who've come into the help channel frustrated by how `msg(text, prompt=prompt)` always resulted in the prompt showing up in telnet clients before the text instead of the desired outcome of the other way around.

Griatch:
A `prompt` is meant to appear and update in-place in a fixed location in the client UI, not to appear free-floating before/after the text-flow itself. Every mud client supports this so the order sent should not matter. :confused: 

But I guess if one hasn't set up one's client properly, it would be annoying. I would say that it would better to change this to "prompt comes last" rather than "text comes first" though.

InspectorCaracal:
Works for me, I'll change it.

Griatch (approved):

PR #2793, opened 2022-07-06
Update evennia-docker-start.sh

blightbow:
Fixed an unquoted shell test in the docker entrypoint script. 

This was generating a "sh: start: unknown operand" message in the startup log when the default "evennia start --log" args were passed in. This has no impact on the return status of the test itself, as it was always succeeding on an empty string and failing on a non-empty one.


blightbow:
I misspoke in the "Motivation" blurb: `evennia start --log` is not a default set of arguments, but the recommended way to run Evennia in the foreground that is suggested in the tutorial. This has no bearing on the patch itself. Passing any arguments to this script via `docker run` will cause the test to generate errors.

PR #2794, opened 2022-07-06
rpsystem: Add special-case handling for self-references.

InspectorCaracal:
This changes `parse_sdescs_and_recogs` to actually use the self-reference regex to identify self-references. It also splits the logic for getting case refs out into a helper function.

Bug fixing.

An issue with self-references which I'd patched out but apparently regressed during parser rewrites arose on discord, where `/me` would multimatch if there was anyone else with an sdesc or recog that starts with "me" (such as Melody, or merchant). Originally I'd resolved this by matching full words only, but that's not a desirable limitation.

To resolve things so you don't have multimatches on `/me` but can still reference a merchant with `/merch`, I've added an initial self-ref pass to the sdesc/recog parser, using the `_RE_SELF_REF` regex. This has the added benefit of actually making the `@` self-ref marker work, since as it turned out, the existing sdesc-matching method of finding the reference marker and iterating forward meant it was completely ignored and useless (since it doesn't _have_ the prefix marker in it at all).

### DIFF
--- a/evennia/contrib/rpg/rpsystem/rpsystem.py
+++ b/evennia/contrib/rpg/rpsystem/rpsystem.py
@@ -202,7 +202,7 @@ _RE_RIGHT_BRACKETS = re.compile(r"\}+", _RE_FLAGS)
 _RE_REF = re.compile(r"\{+\#([0-9]+[\^\~tv]{0,1})\}+")
 
 # This regex is used to quickly reference one self in an emote.
-_RE_SELF_REF = re.compile(r"/me\b|@", _RE_FLAGS)
+_RE_SELF_REF = re.compile(r"(/me|@)(?=\W+)", _RE_FLAGS)
 
 # regex for non-alphanumberic end of a string
 _RE_CHAREND = re.compile(r"\W+$", _RE_FLAGS)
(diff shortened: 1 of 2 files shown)


evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -202,7 +202,7 @@ class Room(ContribRPRoom):
 _RE_REF = re.compile(r"\{+\#([0-9]+[\^\~tv]{0,1})\}+")
 
 # This regex is used to quickly reference one self in an emote.
-_RE_SELF_REF = re.compile(r"/me|@", _RE_FLAGS)
+_RE_SELF_REF = re.compile(r"/me\b|@", _RE_FLAGS)
```
Griatch, line 205: I think the problem I did not use `\b` here is because in English you want to self-reference with possessive form, so you want to be able to write something like `/me's`: 

    There is a frown as the other person tries to touch /me's sword. "Don't do that, he hisses."

In the old style of using the algorithm, this worked out due to how it was split out into components, here one probably needs to handle this specifically.

Griatch, line 205: I think the problem I did not use `\b` here is because in English you want to self-reference with possessive form, so you want to be able to write something like `/me's`: 

    There is a frown as the other person tries to touch /me's sword. "Don't do that, he hisses."

In the old style of using the algorithm, this worked out due to how it was split out into components, here one probably needs to handle this specifically.

Griatch (changes requested):
I haven't tested it, but I suspect the regex change will leave possessive forms of self-reference not working with this new algorithm. ...

Griatch (approved):
Thanks for the updates!

PR #2799, opened 2022-07-09
Refactored CommandHandler

volundmush:
Rewrote the CommandHandler into a replaceable Class, and get_and_merge_cmdsets to be naive of the kinds of objects (session, account, object) it is working with.

Shifted logic for gathering "local object" cmdsets from the get_and_merge_cmdset function to the cmd-bearing typeclasses as obj.get_extra_cmdsets()

Noticed that CmdHandler isn't very flexible or easy to extend and that the cmdset merge process could be made far cleaner, simpler, and more flexible.

Would like to add more kinds of objects that can bear Commands in user-code and this helps a ton with that.

I poked at the tests but I'm not sure that they're all truly fixed.


volundmush:
After more futzing I can tell this particular approach is not going to work.

Still curious about your thoughts on the overall notion.

evennia/commands/cmdhandler.py
```diff
@@ -377,63 +320,29 @@ def _get_cmdsets(obj):
                 _msg_err(caller, _ERROR_CMDSETS)
                 raise ErrorReported(raw_string)
             try:
-                returnValue((obj.cmdset.current, list(obj.cmdset.cmdset_stack)))
+                if extra:
+                    extra = yield obj.get_extra_cmdsets(caller, current, cmdsets)
+                    returnValue(extra)
```
Griatch, line 325: I think in latest Twisted, `returnValue` is not actually needed anymore, and a regular `return value` now works from inside an `@inlineCallbacks`-decorated function.. But I've not actually tested it ...

Griatch (commented):
I don't mind making the CmdHandler more pluggable by all means, but as you notice, it's not so easy to just change it around. Having individual entities be responsible for finding cmdsets on objects is an interesting approach too. But overall, it's probably a good idea to make sure unit tests pass before making the PR ;)

Griatch:
With no more comments or activity after that last comment, I'm closing this.

PR #2803, opened 2022-07-10
Added move_type to obj.move_to and its many hooks and what calls it.

volundmush:
Add a small kwarg to obj.move_to() for handling "move_type". Adjusted announce_move_from and announce_move_to so it generates text-tuples for .msg which use it for {"type": move_type}.

Although move_to is already quite flexible because it passes **kwargs, there is so much ambiguity about the KINDS of movements that exist which might be useful for varied messaging in hooks and calling other code that I believe it's unwise to NOT have this in the move_to operation. Yes, at_give and at_drop and similar hooks do exist but this approach opens up a lot of possibilities.

To my knowledge, this should not break the 0.9 API, because the new argument will just get absorbed into the **kwargs.

Still hammering at tests but so far everything seems good.


evennia/objects/objects.py
```diff
@@ -1472,13 +1477,14 @@ def at_access(self, result, accessing_obj, access_type, **kwargs):
...
 
-    def at_pre_move(self, destination, **kwargs):
+    def at_pre_move(self, destination, move_type="move", **kwargs):
         """
         Called just before starting to move this object to
         destination.
 
         Args:
             destination (Object): The object we are moving to
+            move_type (str): The type of move. "give", "traverse", etc.
```
Griatch, line 1487: This makes it sounds like there are only a few allowed types ... clarify that this is an _arbitrary_ name chosen by the event that triggers the move.

evennia/objects/objects.py
```diff
@@ -1624,27 +1632,29 @@ def at_post_move(self, source_location, **kwargs):
...
-    def at_object_receive(self, moved_obj, source_location, **kwargs):
+    def at_object_receive(self, moved_obj, source_location, move_type="move", **kwargs):
         """
         Called after an object has been moved into this object.
 
         Args:
             moved_obj (Object): The object moved into this one
             source_location (Object): Where `moved_object` came from.
                 Note that this could be `None`.
+            move_type (str): The type of move. "give", "traverse", etc.
```
Griatch, line 1657: Expand here too.

evennia/objects/objects.py
```diff
@@ -1605,7 +1613,7 @@ def announce_move_to(self, source_location, msg=None, mapping=None, **kwargs):
             }
         )
 
-        destination.msg_contents(string, exclude=(self,), from_obj=self, mapping=mapping)
+        destination.msg_contents((string, {"type": move_type}), exclude=(self,), from_obj=self, mapping=mapping)
 
     def at_post_move(self, source_location, **kwargs):
```
Griatch, line 1618: This also needs `move_type` and docstring, no?

Griatch (changes requested):
I don't know if it's strictly _necessary_, but this is the kind of thing `**kwargs` was intended for and I can see the use-case for letting the move operations more aware of what is happening if one wants to customize things there. :+1: I think this looks fine, but it needs some extra documentation clarifications.

Griatch (approved):
Thanks for updating the PR!

PR #2804, opened 2022-07-12
Buff contrib

Tegiminis:
Addition of a flexible buff system to the contribs folder, including documentation and unit tests.
Buff systems are a common request of MUDs and other virtual worlds.

This contrib aims to give developers a starting point to set up their own buff systems. Out of the box, with the proper setup, this system can modify any value through a modifier check, trigger arbitrary code based on an event string, and tick periodically.


Griatch (changes requested):
Thanks for the PR! This looks comprehensive and useful! :) 

Before reviewing though, this needs some restructuring: 
- For 1.0 and forward, we auto-generate contrib-documentation, meaning contribs must be on a specific form, including a separate `README.md` file. Please check out the [guidelines for making a contrib here](https://www.evennia.com/docs/1.0-dev/Contributing.html#guidelines-for-making-a-contrib). I suggest moving most of what you have in the header of `buff.py` into the readme instead, just leaving a shorter summary in the header. 
- It would be a little easier if you could git-squash your commits to one; a few is fine, but this PR has a truckload of them.

Tegiminis:
Moved most of the docstring to a README.md, edited down the buff module docstring, squashed the pre-review commits, and added a few more tests. This should be ready for review! @Griatch

evennia/contrib/game_systems/buffs/buff.py
```diff
@@ -0,0 +1,1100 @@
...
+            mod:        The modifier the buff applies. "add" for add/sub or "mult" for mult/div
+            value:      The value of the modifier
+            perstack:   How much is added to the base, per stack (including first)."""
+        self.stat = stat
+        self.modifier = modifier
+        self.value = value
+        self.perstack = perstack
+
+
+class BuffHandler(object):
```
Griatch, line 297: As a note, in Python3 you don't need to add `(object)` anymore; it's a remnant from Py2. Just `BuffHandler:` is enough.

evennia/contrib/game_systems/buffs/tests.py
```diff
@@ -0,0 +1,366 @@
+# in a module tests.py somewhere i your game dir
```
Griatch, line 1: ? This looks like it should be removed and a proper module docstring added.

evennia/contrib/game_systems/buffs/tests.py
```diff
@@ -0,0 +1,366 @@
...
+    @lazy_property
+    def buffs(self) -> BuffHandler:
+        return BuffHandler(self)
+
+    def at_init(self):
+        self.stat1, self.buffs
+        return super().at_init()
+
+
+class TestBuffsAndHandler(EvenniaTest):
```
Griatch, line 107: Looks comprehensive! :+1:

evennia/contrib/game_systems/buffs/tests.py
```diff
@@ -0,0 +1,366 @@
...
+    key = "ttib"
+    name = "ttib"
+    flavor = "timerbuff"
+    maxstacks = 1
+    tickrate = 1
+    duration = 5
+    mods = [Mod("timetest", "add", 665)]
+
+    def at_tick(self, initial=True, *args, **kwargs):
+        self.owner.db.ticktest = True
```
Griatch, line 92: I would expect this to mock out the actual `delay` call. We don't want timers going off inside the unit test suite.

Griatch (changes requested):
This is looking good. I have only a few inline comments to address, and one additional point: 
- Categorization is hard, but since `Traits` are in `contribs/rpg/` folder, I think this belongs in the same place.

Tegiminis:
@Griatch requested changes made. thanks for the tip on further branching, will carry it forward

Griatch (approved):
Thanks for the changes, looks like things are in their right place now! Approved.

PR #2807, opened 2022-07-14
Added msg_type to rpsystem.send_emote()

volundmush:
the rpsystem contrib is currently stripping out the {"type": "say"} data which the default evennia say command flow creates.

Was easy to add this back in.

Currently working on a project which depends on the text-tuple's data for things like speech.

And just think this should be restored.

Easy peasy.

### DIFF
--- a/evennia/contrib/rpg/rpsystem/rpsystem.py
+++ b/evennia/contrib/rpg/rpsystem/rpsystem.py
@@ -513,7 +513,7 @@ def parse_sdescs_and_recogs(sender, candidates, string, search_mode=False, case_
     return string, mapping
 
 
-def send_emote(sender, receivers, emote, anonymous_add="first", **kwargs):
+def send_emote(sender, receivers, emote, msg_type = "pose", anonymous_add="first", **kwargs):
     """
     Main access function for distribute an emote.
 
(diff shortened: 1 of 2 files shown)


InspectorCaracal:
I don't believe "pose" is a msg type used by anything inside or outside of the contrib, so including it seems out of scope.

I also would recommend making the kwarg optional and only sending the message to the receivers as a tuple if the `msg_type` is set, for more consistency with default `msg` behavior.

volundmush:
> I don't believe "pose" is a msg type used by anything inside or outside of the contrib, so including it seems out of scope.
This is true; however, there are generally two different kinds of actions that the RP system will be covering. Dialogue, and gestures/socials/roleplayed "actions" which are typically known as emotes, poses, or posts in the community. "say" is established and is covered by CmdSay so I set this as a default because it's highly likely to be appropriate for everything else it needs to do.

> I also would recommend making the kwarg optional and only sending the message to the receivers as a tuple if the `msg_type` is set, for more consistency with default `msg` behavior.
As long as it CAN generate the tuple I'm not very picky about the specifics.

Griatch (approved):
This seems like a consistent change, and 'pose' is a reasonable type-name to use; nothing else uses it, but since the RPSystem is a contrib, that's not too surprising. I also think it's fine to pass the msg-tuple here, it's how we do it for `look` as well, for example - it _is_ the full call sign form of `msg` after all.

PR #2809, opened 2022-07-21
enable dict subclasses serialization

### LINKED ISSUE #2808: [BUG] cannot serialize dict subclasses
#### Describe the bug
Dear evennia developers, I've found that I'm not able to persist a dict subclass and after tracking it down a bit I think there's a generic problem when working w/ base types subclasses.

#### To Reproduce
A simple PoC using a dict:
1. put this in `/typeclasses/foo.py`:
```python
from evennia import CmdSet, Command, DefaultObject


class MaskedDict(dict):
    pass


class CmdFoo(Command):

    key = "cf"

    def func(self):
        self.obj.db.foo = MaskedDict({"foo": 1, "bar": 2})


class CmdSetFoo(CmdSet):
    def at_cmdset_creation(self):
        self.add(CmdFoo())


class FooObject(DefaultObject):
    def at_object_creation(self):
        self.cmdset.add_default(CmdSetFoo)
```
2. reload and create a foo object
```
reload
create foo :foo.FooObject
```
3. try to persist a dict subclass in foo obj:
```
cf
  File "/home/oggei/dev/evennia/evennia/commands/cmdhandler.py", line 621, in _run_command
    ret = cmd.func()
  File "/home/oggei/dev/evennia/dict-bug/typeclasses/foo.py", line 13, in func
    self.obj.db.foo = MaskedDict({"foo": 1, "bar": 2})
  File "/home/oggei/dev/evennia/evennia/typeclasses/attributes.py", line 1414, in __setattr__
    _GA(self, _GA(self, "name")).add(attrname, value)
  File "/home/oggei/dev/evennia/evennia/typeclasses/attributes.py", line 1251, in add
    self.backend.create_attribute(keystr, category, lockstring, value, strattr)
  File "/home/oggei/dev/evennia/evennia/typeclasses/attributes.py", line 697, in create_attribute
    attr = self.do_create_attribute(key, category, lockstring, value, strvalue)
  File "/home/oggei/dev/evennia/evennia/typeclasses/attributes.py", line 1033, in do_create_attribute
    kwargs["db_value"] = to_pickle(value)
  File "/home/oggei/dev/evennia/evennia/utils/dbserialize.py", line 663, in to_pickle
    return process_item(data)
  File "/home/oggei/dev/evennia/evennia/utils/dbserialize.py", line 650, in process_item
    return item.__class__([process_item(val) for val in item])
ValueError: dictionary update sequence element #0 has length 3; 2 is required

An untrapped error occurred.
(Traceback was logged 22-07-19 15:38:57).
```
4. confirm that nothing has indeed ben set
```
examine foo
--------------------------------------------------------------------------------
Name/key: foo (#4)
Typeclass: FooObject (typeclasses.foo.FooObject)
Location: oggei (#1)
Home: oggei (#1)
Locks:   call:true(); control:id(1) or perm(Admin); delete:id(1) or perm(Admin);
  drop:holds(); edit:perm(Admin); examine:perm(Builder); get:all();
  puppet:pperm(Developer); tell:perm(Admin); view:all()
Stored Cmdset(s):
  typeclasses.foo.CmdSetFoo [Unnamed CmdSet] (Union, prio 0)
Merged Cmdset(s):
  typeclasses.foo.CmdSetFoo [Unnamed CmdSet] (Union, prio 0)
Commands available to foo (result of Merged Cmdset(s)):
  cf
Persistent Attributes:
  desc=You see nothing special.
--------------------------------------------------------------------------------
```

#### Expected behavior
I'd like to be able to persist those subclasses :)

```
    Evennia 1.0-dev (rev ae5b353ef) (rev ae5b353ef)
    OS: posix
    Python: 3.9.13
    Twisted: 21.7.0
    Django: 4.0.6
```

the problem occurs in 0.9.5 too

#### Moving toward a solution

At a glance, it's clear that an eg. `dict`subclass will never match the correct [condition here](https://github.com/evennia/evennia/blob/ae5b353ef245450a05fba08e863b36d407600031/evennia/utils/dbserialize.py#L620).  
The first solution that come to mind is to check `isinstance` calls rather than `type` ones, by reordering the checks following inheritance chains bottom-up, which does not seems to pose particular diamond problems:

```mermaid
classDiagram

int <|-- bool


str <|-- django_utils_safestring_SafeString

django_utils_safestring_SafeData <|-- django_utils_safestring_SafeString



evennia_utils_dbserialize__SaverMutable <|-- evennia_utils_dbserialize__SaverList

collections_abc_MutableSequence <|-- evennia_utils_dbserialize__SaverList
collections_abc_Sequence <|-- collections_abc_MutableSequence
collections_abc_Reversible <|-- collections_abc_Sequence
collections_abc_Iterable <|-- collections_abc_Reversible

collections_abc_Collection <|-- collections_abc_Sequence
collections_abc_Sized <|-- collections_abc_Collection

collections_abc_Iterable <|-- collections_abc_Collection

collections_abc_Container <|-- collections_abc_Collection


evennia_utils_dbserialize__SaverMutable <|-- evennia_utils_dbserialize__SaverDict

collections_abc_MutableMapping <|-- evennia_utils_dbserialize__SaverDict
collections_abc_Mapping <|-- collections_abc_MutableMapping
collections_abc_Collection <|-- collections_abc_Mapping
collections_abc_Sized <|-- collections_abc_Collection

collections_abc_Iterable <|-- collections_abc_Collection

collections_abc_Container <|-- collections_abc_Collection

dict <|-- collections_defaultdict

evennia_utils_dbserialize__SaverDict <|-- evennia_utils_dbserialize__SaverDefaultDict
evennia_utils_dbserialize__SaverMutable <|-- evennia_utils_dbserialize__SaverDict

collections_abc_MutableMapping <|-- evennia_utils_dbserialize__SaverDict
collections_abc_Mapping <|-- collections_abc_MutableMapping
collections_abc_Collection <|-- collections_abc_Mapping
collections_abc_Sized <|-- collections_abc_Collection

collections_abc_Iterable <|-- collections_abc_Collection

collections_abc_Container <|-- collections_abc_Collection


evennia_utils_dbserialize__SaverMutable <|-- evennia_utils_dbserialize__SaverSet

collections_abc_MutableSet <|-- evennia_utils_dbserialize__SaverSet
collections_abc_Set <|-- collections_abc_MutableSet
collections_abc_Collection <|-- collections_abc_Set
collections_abc_Sized <|-- collections_abc_Collection

collections_abc_Iterable <|-- collections_abc_Collection

collections_abc_Container <|-- collections_abc_Collection

dict <|-- collections_OrderedDict

evennia_utils_dbserialize__SaverMutable <|-- evennia_utils_dbserialize__SaverOrderedDict

collections_abc_MutableMapping <|-- evennia_utils_dbserialize__SaverOrderedDict
collections_abc_Mapping <|-- collections_abc_MutableMapping
collections_abc_Collection <|-- collections_abc_Mapping
collections_abc_Sized <|-- collections_abc_Collection

collections_abc_Iterable <|-- collections_abc_Collection

collections_abc_Container <|-- collections_abc_Collection


evennia_utils_dbserialize__SaverMutable <|-- evennia_utils_dbserialize__SaverDeque

```

I could work on that but maybe I'm missing an obvious, already existing way of doing things so what do you think? 

Many thanks, regards

aogier:

This PR tentatively fix #2808 by letting un/pickling functions correctly identify dict subclasses and properly serialize them.

close #2808

### DIFF
--- a/evennia/typeclasses/tests.py
+++ b/evennia/typeclasses/tests.py
@@ -2,6 +2,7 @@
 Unit tests for typeclass base system
 
 """
+
 from django.test import override_settings
 from evennia.typeclasses import attributes
 from evennia.utils.test_resources import BaseEvenniaTest, EvenniaTestCase
@@ -26,13 +27,24 @@ class TestAttributes(BaseEvenniaTest):
         self.obj1.db.testattr = value
         self.assertEqual(self.obj1.db.testattr, value)
 
+        # "plain" subclasses
         value = DictSubclass({"fo": "foo", "bar": "bar"})
         self.obj1.db.testattr = value
         self.assertEqual(self.obj1.db.testattr, value)
 
-        value = DictSubclass({"fo": "foo", "bar": "bar", "obj": self.obj2})
-        self.obj1.db.testattr = value
+        self.obj1.db.testattr["fo"] = "foo2"
+        value.update({"fo": "foo2"})
         self.assertEqual(self.obj1.db.testattr, value)
+        self.assertEqual(self.obj1.attributes.get("testattr"), value)
+
+        # nested subclasses
+        value = DictSubclass({"nested": True, "deep": DictSubclass({"fo": "foo", "bar": "bar"})})
+        self.obj1.db.testattr = value
+
+        self.obj1.db.testattr["deep"]["fo"] = "nemo"
+        value["deep"].update({"fo": "nemo"})
+        self.assertEqual(self.obj1.db.testattr, value)
+        self.assertEqual(self.obj1.attributes.get("testattr"), value)
 
     @override_settings(TYPECLASS_AGGRESSIVE_CACHE=False)
     @patch("evennia.typeclasses.attributes._TYPECLASS_AGGRESSIVE_CACHE", False)
@@ -47,14 +59,25 @@ class TestAttributes(BaseEvenniaTest):
         self.assertEqual(self.obj1.db.testattr, value)
         self.assertFalse(self.obj1.attributes.backend._cache)
 
+        # "plain" subclasses
         value = DictSubclass({"fo": "foo", "bar": "bar"})
-        self.obj1.db.dict_subclass = value
-        self.assertEqual(self.obj1.db.dict_subclass, value)
+        self.obj1.db.testattr = value
+        self.assertEqual(self.obj1.db.testattr, value)
+
+        self.obj1.db.testattr["fo"] = "foo2"
+        value.update({"fo": "foo2"})
+        self.assertEqual(self.obj1.db.testattr, value)
+        self.assertEqual(self.obj1.attributes.get("testattr"), value)
         self.assertFalse(self.obj1.attributes.backend._cache)
 
-        value = DictSubclass({"fo": "foo", "bar": "bar", "obj": self.obj2})
-        self.obj1.db.dict_subclass = value
-        self.assertEqual(self.obj1.db.dict_subclass, value)
+        # nested subclasses
+        value = DictSubclass({"nested": True, "deep": DictSubclass({"fo": "foo", "bar": "bar"})})
+        self.obj1.db.testattr = value
+
+        self.obj1.db.testattr["deep"]["fo"] = "nemo"
+        value["deep"].update({"fo": "nemo"})
+        self.assertEqual(self.obj1.db.testattr, value)
+        self.assertEqual(self.obj1.attributes.get("testattr"), value)
         self.assertFalse(self.obj1.attributes.backend._cache)
 
     def test_weird_text_save(self):
--- a/evennia/typeclasses/tests.py
+++ b/evennia/typeclasses/tests.py
@@ -3,8 +3,8 @@ Unit tests for typeclass base system
 
 """
 from django.test import override_settings
-from evennia.utils.test_resources import BaseEvenniaTest, EvenniaTestCase
 from evennia.typeclasses import attributes
+from evennia.utils.test_resources import BaseEvenniaTest, EvenniaTestCase
 from mock import patch
 from parameterized import parameterized
 
@@ -13,12 +13,24 @@ from parameterized import parameterized
 # ------------------------------------------------------------
 
 
+class DictSubclass(dict):
+    pass
+
+
 class TestAttributes(BaseEvenniaTest):
     def test_attrhandler(self):
         key = "testattr"
         value = "test attr value "
         self.obj1.attributes.add(key, value)
         self.assertEqual(self.obj1.attributes.get(key), value)
+        self.obj1.db.testattr = value
+        self.assertEqual(self.obj1.db.testattr, value)
+
+        value = DictSubclass({"fo": "foo", "bar": "bar"})
+        self.obj1.db.testattr = value
+        self.assertEqual(self.obj1.db.testattr, value)
+
+        value = DictSubclass({"fo": "foo", "bar": "bar", "obj": self.obj2})
         self.obj1.db.testattr = value
         self.assertEqual(self.obj1.db.testattr, value)
 
@@ -33,6 +45,16 @@ class TestAttributes(BaseEvenniaTest):
         self.assertEqual(self.obj1.attributes.get(key), value)
         self.obj1.db.testattr = value
         self.assertEqual(self.obj1.db.testattr, value)
+        self.assertFalse(self.obj1.attributes.backend._cache)
+
+        value = DictSubclass({"fo": "foo", "bar": "bar"})
+        self.obj1.db.dict_subclass = value
+        self.assertEqual(self.obj1.db.dict_subclass, value)
+        self.assertFalse(self.obj1.attributes.backend._cache)
+
+        value = DictSubclass({"fo": "foo", "bar": "bar", "obj": self.obj2})
+        self.obj1.db.dict_subclass = value
+        self.assertEqual(self.obj1.db.dict_subclass, value)
         self.assertFalse(self.obj1.attributes.backend._cache)
 
     def test_weird_text_save(self):
--- a/evennia/utils/dbserialize.py
+++ b/evennia/utils/dbserialize.py
@@ -644,7 +644,15 @@ def to_pickle(data):
                 # classmethods) and instances
                 pass
 
-        if hasattr(item, "__iter__"):
+        if hasattr(item, "items"):
+            # we try to conserve the iterable class, if not convert to dict
+            try:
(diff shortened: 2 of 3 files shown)


aogier:
There currently is a limitation as it phases out origin class in favor of `_SaverDict` when coping with an [exception in `process_tree`](https://github.com/evennia/evennia/pull/2809/files#diff-e2bb450a91e32b35e4ddad632999b0bb59adb482225cf9e6894d4ca2c293ef74R804), I'm not able to think about a solution that does not resort fiddling with metaclass programming and other brittle paths

Griatch:
Thanks for the PR! I'll try to take a look at it (and other PRs in the pipe) this weekend :)

evennia/typeclasses/tests.py
```diff
@@ -22,6 +26,14 @@ def test_attrhandler(self):
...
         self.assertEqual(self.obj1.db.testattr, value)
 
+        value = DictSubclass({"fo": "foo", "bar": "bar"})
+        self.obj1.db.testattr = value
+        self.assertEqual(self.obj1.db.testattr, value)
+
+        value = DictSubclass({"fo": "foo", "bar": "bar", "obj": self.obj2})
+        self.obj1.db.testattr = value
+        self.assertEqual(self.obj1.db.testattr, value)
+
```
Griatch, line 36: I'd like to also see tests of on-object inserts and reads here; so checking that you can do `self.obj1.db.testattr['fo'] = 'foo2'` and read data out from nested structures. 
Also would be useful to make sure that you can store a `DictSubclass` with another `DictSubclass` (and/or a regular dict by all means) nested inside it.

evennia/utils/dbserialize.py
```diff
@@ -644,7 +644,15 @@ def process_item(item):
                 # classmethods) and instances
                 pass
 
-        if hasattr(item, "__iter__"):
+        if hasattr(item, "items"):
```
Griatch, line 647: I'm  a little reluctant to base item identification on the existence of an `items` property; it's not unthinkable that a user may add a property `.items` to their non-dict entity (to represent a nested inventory, for example). Using dunder methods and duck typing seems better. I don't think there's a unique dunder for dict (`__getitem__` is in both list and dict), but maybe something like this? (pseudocode):

```
if item has __iter__ property: 
    try to process key, val as if a dictionary 
    if failing to do so (iterating does not generate key,val pairs):
        try it as a list
```

Griatch (changes requested):
Thanks for the PR! I think overall this looks like a correct approach to fixing this issue. I'm however reluctant to base the child of a `dict` based on it having an `.items` property like done here, I suspect it could be too fragile; see the comment with a suggested alternative solution.

evennia/typeclasses/tests.py
aogier: ah great, thanks: I wasn't totally sure on how to cover all the cases

evennia/utils/dbserialize.py
aogier: looks good to me!

evennia/typeclasses/tests.py
aogier: done

aogier:
Ok another iteration is ready with the following changelog:

 * **unit tests**: they now hopefully cover all the edits/nesting scenarios that are needed, expanding them led me to rework deserialization logic as per next bullet
 * **deserialization logic**: during deserialization we now return a `_Saver` where we'd normally return it with base types, so using a base dict/list or a subclass should be idempotent now (and retaining custom logic obv). Attribute proxing in `_SaverMutable` is already good and can be extended, should any need arise.  
Having reached this point I see no reason not to extend the functionality to all other types which provide a `_Saver` class, I will work on that in the next iteration.

evennia/utils/dbserialize.py
aogier: done

evennia/typeclasses/tests.py
aogier: ("resolving" so the red note disappear below :P)

aogier:
I'd also be glad to hear what @InspectorCaracal would think about it since we've talked a bit in #2808 :)

InspectorCaracal:
> I'd also be glad to hear what @InspectorCaracal would think about it since we've talked a bit in #2808 :)

Oh hey, this looks great! I think it'll resolve the issues we were talking about really neatly.

aogier:
> Oh hey, this looks great! I think it'll resolve the issues we were talking about really neatly.

great, what do you think about extending the solution to other _Saver classes?

evennia/utils/dbserialize.py
```diff
@@ -645,11 +645,20 @@ def process_item(item):
...
-            except (AttributeError, TypeError):
-                return [process_item(val) for val in item]
+                # we try to conserve the iterable class, if not convert to dict
+                try:
+                    return item.__class__(
+                        (process_item(key), process_item(val)) for key, val in item.items()
+                    )
+                except (AttributeError, TypeError):
+                    return {process_item(key): process_item(val) for key, val in item.items()}
+            except:
```
Griatch, line 656: Never use a bare `except` ... use at least `except Exception` here.

evennia/utils/dbserialize.py
```diff
@@ -714,11 +723,20 @@ def process_item(item):
...
```
Griatch, line 733: Same here.

evennia/utils/dbserialize.py
```diff
@@ -780,13 +798,30 @@ def process_tree(item, parent):
...
+                        (process_item(key), process_tree(val, dat)) for key, val in item.items()
+                    )
+                    return dat
+                except (AttributeError, TypeError):
+                    dat = _SaverDict(_parent=parent)
+                    dat._data.update(
+                        (process_item(key), process_tree(val, dat)) for key, val in item.items()
+                    )
+                    return dat
+            except:
```
Griatch, line 814: Same

evennia/utils/dbserialize.py
```diff
@@ -830,6 +867,34 @@ def process_tree(item, parent):
...
+                        (process_item(key), process_tree(val, dat)) for key, val in data.items()
+                    )
+                    return dat
+                except (AttributeError, TypeError):
+                    dat = _SaverDict(_db_obj=db_obj)
+                    dat._data.update(
+                        (process_item(key), process_tree(val, dat)) for key, val in data.items()
+                    )
+                    return dat
+            except:
```
Griatch, line 885: Again

Griatch (approved):
Thanks for the updates!

PR #2810, opened 2022-07-22
add pypy3.9 to CI

### LINKED ISSUE #1472: Preparing Evennia for Pypy
There seems to be two separate issues keeping Evennia from running on Pypy, one of which is the `cStringIO` module apparently being overshadowed by Pypy's default `cStringIO.py` module which returns `StringIO`, with `evennia_launcher.py` setting the `PYTHONPATH` environment variable for use by a `Popen`ed instance of `evennia_runner.py` which then `Popen`s `twistd`:

```
  File "pyenv-pypy/bin/twistd", line 11, in <module>                                            
    sys.exit(run())                                                                                                     
  File "pyenv-pypy/site-packages/twisted/scripts/twistd.py", line 29, in run                    
    app.run(runApp, ServerOptions)                                                                                      
  File "pyenv-pypy/site-packages/twisted/application/app.py", line 662, in run                  
    runApp(config)                                                                                                      
  File "pyenv-pypy/site-packages/twisted/scripts/twistd.py", line 25, in runApp                 
    _SomeApplicationRunner(config).run()                                                                                
  File "pyenv-pypy/site-packages/twisted/application/app.py", line 382, in run                  
    self.logger.start(self.application)                                                                                 
  File "pyenv-pypy/site-packages/twisted/application/app.py", line 191, in start                
    observer = self._getLogObserver()                                                                                   
  File "pyenv-pypy/site-packages/twisted/scripts/_twistd_unix.py", line 172, in _getLogObserver 
    return logger.textFileLogObserver(logFile)                                                                          
  File "pyenv-pypy/site-packages/twisted/logger/_file.py", line 86, in textFileLogObserver      
    return FileLogObserver(outFile, formatEvent)                                                                        
  File "pyenv-pypy/site-packages/twisted/logger/_file.py", line 34, in __init__                 
    if ioType(outFile) is not unicode:                                                                                  
  File "pyenv-pypy/site-packages/twisted/python/compat.py", line 395, in ioType                 
    from cStringIO import InputType, OutputType                                                                         
ImportError: cannot import name 'InputType'
```

Commenting out [this line](https://github.com/evennia/evennia/blob/a994fd3c63bd0120f5c8745fad97fedb0d849ee3/evennia/server/evennia_launcher.py#L662) seems to resolve this issue, but I'm wondering if there's maybe a cleaner way to handle this? That is, if having the `PYTHONPATH` environment variable set is really useful at all for this project. At the very least, we can check if the user is running Pypy and not copy `sys.path` into there.

I haven't been able to reproduce this bug outside of Evennia using minimal test cases which use `PYTHONPATH` to overshadow `cStringIO` through a `Popen`ed subprocess of Pypy, so I'm not sure if it originates from somewhere within Evennia itself, Pypy, or even Twistd or Django. Tracking down its source has been difficult.

The next problem keeping Evennia from booting in Pypy is much simpler.

```
  File "pyenv-pypy/site-packages/twisted/application/app.py", line 662, in run
    runApp(config)
  File "pyenv-pypy/site-packages/twisted/scripts/twistd.py", line 25, in runApp
    _SomeApplicationRunner(config).run()
  File "pyenv-pypy/site-packages/twisted/application/app.py", line 380, in run
    self.application = self.createOrGetApplication()
  File "pyenv-pypy/site-packages/twisted/application/app.py", line 445, in createOrGetApplication
    application = getApplication(self.config, passphrase)
--- <exception caught here> ---
  File "pyenv-pypy/site-packages/twisted/application/app.py", line 456, in getApplication
    application = service.loadApplication(filename, style, passphrase)
  File "pyenv-pypy/site-packages/twisted/application/service.py", line 412, in loadApplication
    application = sob.loadValueFromFile(filename, 'application')
  File "pyenv-pypy/site-packages/twisted/persisted/sob.py", line 177, in loadValueFromFile
    eval(codeObj, d, d)
  File "evennia-dev/evennia/server/server.py", line 535, in <module>
    EVENNIA = Evennia(application)
  File "evennia-dev/evennia/server/server.py", line 186, in __init__
    channelhandler.CHANNELHANDLER.update()
  File "evennia-dev/evennia/comms/channelhandler.py", line 258, in update
    self.add(channel)
  File "evennia-dev/evennia/comms/channelhandler.py", line 218, in add
    locks="cmd:all();%s" % channel.locks,
  File "evennia-dev/evennia/utils/utils.py", line 1679, in __get__
    value = self.func(obj)
  File "evennia-dev/evennia/typeclasses/models.py", line 269, in locks
    return LockHandler(self)
  File "evennia-dev/evennia/locks/lockhandler.py", line 184, in __init__
    self.reset()
  File "evennia-dev/evennia/locks/lockhandler.py", line 421, in reset
    self._cache_locks(self.obj.lock_storage)
  File "evennia-dev/evennia/utils/idmapper/models.py", line 175, in fget
    def fget(cls): return _get(cls, fieldname)
  File "evennia-dev/evennia/utils/idmapper/models.py", line 116, in _get
    if _GA(cls, "_is_deleted"):
exceptions.TypeError: unbound method __getattribute__() must be called with newobject instance as first argument (got Channel instance instead)
```

Pypy doesn't seem to like the use of `object.__getattribute__` on instances which are not `object`, which appears to be a simple enough fix. Changing [this line](https://github.com/evennia/evennia/blob/a994fd3c63bd0120f5c8745fad97fedb0d849ee3/evennia/utils/idmapper/models.py#L30) from:

```python
_GA = object.__getattribute__
```

..to..

```python
def _GA(obj, key):
    return object.__getattribute__.__get__(obj)(key)
```

..appears to resolve these sort of issues. There are several instances in different files where `_GA` appears, and they would all have to be made compatible, likely `_SA` and `_DA` as well, as they follow suit in using `object` methods on non-object instances.

These are the initial blockers at least, to getting Evennia up and running with Pypy. There may be more issues, but this is what I've discovered so far in my research. If anyone more experienced with Python can offer suggestions, I'd appreciate it.

ArchLinux, Linux 4.12.13-1-ARCH x86_64
pypy 5.9.0-1, using python-virtualenv 15.1.0-1
Evennia 0.7.0 (rev a994fd3c)

aogier:

After some local tests, this PR adds pypy3.9 to CI so to close #1472


Griatch:
Hm, if you can initialize and run/test the server under pypy locally, it would seem you have some setup than what is used for github actions. Maybe try to duplicate exactly the virtualenv running in github action locally, to see if you see the same error? 

Fun to see someone experimenting with getting this working!

aogier:
yeah I'm trying to figuring out how to figure out this, I warmed up some caches github side so doing tries will not take forever but it's definitely time to bring back the beast to my machine!

aogier:
what is very strange is the failure is about a builtin module presence, maybe some configuration on ubuntu-latest runners, oh well, we will see :rocket:

dvoraen:
The test suite is installing Django 4.0+, is that intended?

aogier:
I'd say yes, given the base branch. The problem is about an unresolved, builtin, `cgi` module atm

dvoraen:
> I'd say yes, given the base branch. The problem is about an unresolved, builtin, `cgi` module atm

Ugh, sorry.  I completely missed this was going to develop and not master.  I was looking at master's requirements.txt and going "WHY IS THIS INSTALLING THE WRONG VERSION" thinking that was the root cause.   Interestingly enough, the current main branch of django's repository has removed cgi from its import in that particular file, because it's deprecated in Python 3.11 apparently, but the most recent release retains that import.

aogier:
> Ugh, sorry. I completely missed this was going to develop and not master. 

don't worry, and thanks for your help :)

> I did a little digging and PyPy's installation is what I'm eyeballing the most. How sure are you that `actions/setup-python@v4` is installing PyPy correctly? That's the only thing that logically makes sense, but the evennia_launcher.py should be complaining about imports if other issues were arising, yet it's the only thing I can think of.

As the other steps are less or more irrelevant to that issue, that is also my suspect. The problem is, I've tried to debug print some stuff and:

```
python -V
pypy -V
python -c 'import cgi'
```

does indeed return the expected results (executables are the same, no import error), you can find them [here](https://github.com/evennia/evennia/runs/7481876954?check_suite_focus=true) so a deeper investigation is still needed

aogier:
So, I've narrowed it a bit getting evidence something happens inside `evennia` command: I've added a custom command
```python
def foobar(*args, **kwargs):
    import cgi
    print("FOOBAR: ", args)
```
and, whereas a normal python invocation importing cgi module does not fails, this command indeed fail. Here's [the evidence](https://github.com/evennia/evennia/runs/7482019852?check_suite_focus=true#step:15:56). This is good news as it can alleviate the painful guesswork against obscure system setup stuff in github runners, and we can concentrate on something we can better observe and control yay!

aogier:
after a very trivial bisect (putting `import cgi` everywhere in the main tool code path, meh) I've found something interesting: placing an `import cgi` in script stub lead to another failure with [another builtin module](https://github.com/evennia/evennia/runs/7482082728?check_suite_focus=true#step:14:78).

aogier:
the problem seems like `sys.path` gets rewritten at a certain point and it loose the standard library part, evidence [is here](https://github.com/evennia/evennia/runs/7482345186?check_suite_focus=true#step:6:22) where I print it in the launcher stub first (PRE) and in first lines of actual launcher main() (POST). Since nothing explicitly happens between the two I'll have to find out some automagic stuff around

aogier:
[This piece in evennia_launcher.py](https://github.com/evennia/evennia/blob/ae5b353ef245450a05fba08e863b36d407600031/evennia/server/evennia_launcher.py#L100) look like a good candidate:
```python
try:
    sys.path[1] = EVENNIA_ROOT
except IndexError:
    sys.path.append(EVENNIA_ROOT)
```

in fact [1] entry is what is get overwritten, by CI output. 

Why it doen not happen in my machine?

My machine isn't very clean at the end of the day, and in [1] I have a useless entry:

```
evennia foobar bar
PRE ['/home/oggei/.pyenv/versions/evennia-pypy/bin', '/home/oggei/dev/pythonpath', '/home/oggei/.pyenv/versions/pypy3.9-7.3.9/lib/pypy3.9', '/home/oggei/.pyenv/versions/evennia-pypy/lib/pypy3.9/site-packages', '/home/oggei/dev/evennia']
POST ['/home/oggei/.pyenv/versions/evennia-pypy/bin', '/home/oggei/dev/evennia', '/home/oggei/.pyenv/versions/pypy3.9-7.3.9/lib/pypy3.9', '/home/oggei/.pyenv/versions/evennia-pypy/lib/pypy3.9/site-packages', '/home/oggei/dev/evennia']
```

unsetting local PYTHONPATH env var indeed reproduce the problem.

@Griatch  do you perhaps remember why this was needed? It was introduced in 60ff4b8515874830e5047b9b235bbaf934b6a822, so a long time ago. 

In the meanwhile I'm cleaning everything up and launch a run which will hopefully go green :smile_cat:

aogier:
Ok here we have a great milestone: everything fails with the same results
```
FAILED (failures=6, errors=38, skipped=36)
```

which is cool because we can now concentrate on cpython->pypy specific errors and not random quirks. 

Now, if there could be somebody interested in helping out, that would be a great moment to do so :)

aogier:
FAILED (errors=6, skipped=36)

here we go!

aogier:
```
FAILED (errors=6, skipped=36)
```

here we slowly go!

Griatch:
> [This piece in evennia_launcher.py](https://github.com/evennia/evennia/blob/ae5b353ef245450a05fba08e863b36d407600031/evennia/server/evennia_launcher.py#L100) look like a good candidate:
> 
> ```python
> try:
>     sys.path[1] = EVENNIA_ROOT
> except IndexError:
>     sys.path.append(EVENNIA_ROOT)
> ```
> @Griatch do you perhaps remember why this was needed? It was introduced in [60ff4b8](https://github.com/evennia/evennia/commit/60ff4b8515874830e5047b9b235bbaf934b6a822), so a long time ago.

The reason for adding the root is clear, but I don't recall why it replaces the index. I think we could probably get away with an `sys.path.insert(1, EVENNIA_ROOT)` here instead.

aogier:
> The reason for adding the root is clear, but I don't recall why it replaces the index. I think we could probably get away with an `sys.path.insert(1, EVENNIA_ROOT)` here instead.

ok looks good to me thx

aogier:
Ok I think we're ready for a review!  
A brief changelog:
 * introduced pypy-3.9 in CI matrix
 * cache dependencies because scipy compilation take ages under pypy
 * installed a psycopg drop-in working replacement for pypy, it load its compat layer at startup
 * yak shavings:
   * fix `sre_constants` name mangling w/ a safer approach
   * fixed a pypy quirk about [dict unpacking when calling functions](https://bugs.python.org/issue39694#msg408101)
   * fixed sys.path mangling so it does not overwrite other entries
   * fixed couple tests w/ minor compat/docs following changes

Things left to do:
 * disable jit and try to find other speedup cause at the moment everything is very slow. Waiting for [this](https://foss.heptapod.net/pypy/pypy/-/issues/3148) and maybe other ideas.

As a side note, I am wondering if it makes sense to release official support for pypy in evennia: there are many quirks and the life of an end user/developer could be quite difficult. Despite this I see value in keeping the tests working on another python interpreter because of the small bugs that this spots, like in this PR. Maybe we could warn in the documentation that pypy works but that the road to using it is a bit more uphill.

aogier:
first iteration on times reduction went well, we will see the others!

aogier:
mysql need a bit more massage but --parallel FTW

Griatch (changes requested):
Thanks a lot for your work on this! 

Hm, overall I think there are some overall improvements to the github actions that could be of general use here, but if performance is _worse_, that in itself is not really making the endeavor of supporting pypy very promising ... it's another thing to support and which may cause test to fail etc ... doing so only for the small gain of some fixed issues is not really worth it (better static linting would be better than that). 

That said, PyPy is of course promising, and I'm open to be convinced ... but even before considering adding pypy support here we'd need to fully document it, e.g. as a file `docs/source/Setup/PyPy.md`. 

To contribute a doc page it doesn't need to be perfect and you don't need to set up the doc build locally; but it sounds like you have looked into this enough to be the one to write up what is needed - and the caveats - of running Evennia under PyPy right now.

aogier:
> Hm, overall I think there are some improvements to the github actions that could be of general use here, but if performance is _worse_, that in itself is not really making the endeavor of supporting pypy very promising ... it's another thing to support and which may cause test to fail etc ... doing so only for the small gain of some fixed issues is not really worth it (better static linting would be better than that).

I'm also unconvinced that a true pypy support should be released, other than some setup notes you suggest. I'm not even convinced about in-matrix CI runs: even w/ jit disabled, pypy will always run w/ [worse times](https://github.com/evennia/evennia/actions/runs/2782303086) than its cpython counterparts. Because how it works, and I'm pretty sure that's not going to change now or ever.  
However, a [`schedule`d](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) run, which isn't tied to people's work nor harms them could still be beneficial: take the [BPO-39694 fix](https://github.com/evennia/evennia/pull/2810/files#diff-259251a64ee1f7a3fe277d4e13a76251cbe189f9775f962ab6d631e70d0928d5R819) or the soon-to-be-dropped `sre_parse` shenanigans. It could represent an (unsupported?) status evennia is in any point in time in terms of passing/failing tests, so people which know wants to use pypy can start from something which already tell them if they're gonna patch - maybe contribute - stuff from the beginning :D

My proposal is to make this PR a:

 * CI scheduled run against pypy w/ current three backends
 * existing minor code fixes so it runs green :P 
 * (unsupported?) setup docs about current pypy status ie. where people can start from if they know they want to use pypy in their setups

Then stem couple PRs from the code here:

 * cache dependencies in CI for the original non-pypy matrix
 * only run coverage against 3.10/sqlite


What do you think? :)

aogier:
I've added the couple infos needed in a new doc bit as requested. Also, I have refactored the IC in the least intrusive way possible so that it can also be used with a night schedule against pypy!

docs/source/Setup/PyPy.md
```diff
@@ -0,0 +1,24 @@
...
+## Notes about PyPy support
+
+Although not actively supported PyPy should work without problems, or at least not too many. Every 
+night the official test suite is run against PyPy so if you want to use it you can start checking
+the latest [CI results against master branch](https://github.com/evennia/evennia/actions/workflows/github_action_pypy_test_suite.yml).
+
+## Additional requirements
+
+Based on your setup, you could need additional packages to be installed along the normal
+requirements.
```
Griatch, line 12: This seems pretty inadequate. I'd expect to at least see full install instructions for getting Evennia running with pypy on a debian-based Linux distro, for example.

Griatch (changes requested):
I think we need a lot more docs on this ... at least to the point that someone with decent Linux knowledge can install Evennia with pypy and run it without needing to dive into the pypy docs (but we can link to those as extra reading too).

docs/source/Setup/PyPy.md
aogier: ok, I can do that. The reason I didn't wrote so much there is existing install instructions are perfectly fine for pypy too

aogier: ok, I've missed the part where docs actually explain how to install python, and what. Professional deformation I guess :D I'll do that

aogier:
I've detailed the pypy installation process which should work for either linux, macos and I'd say wsl. A minor refactor on "installation-git" allowed me to refer to it for non-pypy-specific steps after install

aogier:
I'll move forward w/ docker stuff, rewritten dockerfile will allow for building against a 3.9/3.10/pypy base so we will have this other possibility too

PR #2812, opened 2022-07-25
faster CI via parallel tests

aogier:

This PR drastically reduce tests time by exploiting django tests' parallel function. Having a 2/3x faster tests is good as people don't have to wait a lot and everybody is more happy!

Rationale behind that: this stems from another (pypy) branch where something was needed to reduce the extremely long pypy times, mainly because scipy compilation and jit being enemy of a typical testcase run. Django parallel tests require tests are separated one from another, ie. does not cross-depend and this is something we already have, so exploiting that is a no-brainer.  
Other optimizations could be done, starting from those simple which django docs lists [for psql](https://docs.djangoproject.com/en/4.0/ref/databases/#speeding-up-test-execution-with-non-durable-settings) and probably for others engines.

With the default parallel configuration we already shaved 50% of runtime from ~20' to ~10' so it looks very promising!

### DIFF
--- a/.github/workflows/github_action_test_suite.yml
+++ b/.github/workflows/github_action_test_suite.yml
@@ -127,7 +127,7 @@ jobs:
           ../bin/unix/evennia test \
             --settings=settings \
             --keepdb \
-            --parallel auto \
+            --parallel 4 \
             --timing \
             evennia
         coverage xml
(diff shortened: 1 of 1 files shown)


Griatch (approved):
Looking good :+1:

PR #2814, opened 2022-07-25
Alter ObjectDB.objects.get_objs_with_key_or_alias() to use icontains

volundmush:
ObjectDB.objects.get_objs_with_key_or_alias() now uses db_key__icontains instead of __istartswith when searching for inexact matches. This makes it match the behavior of how searching by the aliases works.

So, with my (Dragon Ball) project, I tried doing @tel/loc me=Roshi and hey wow, no Roshi found. Then I tried @tel/loc me=Master and got a TON of search results.

But when I @search Roshi, I get plenty of Roshi-related stuff.

Turns out that @search is using db_key__icontains while object.search() ends up calling ObjectDB.objects.get_objs_with_key_or_alias(), which uses db_key__istartswith when handling fuzzy matches - so, it first gathers via startswith, and THEN does a string_partial_match on each word... this seems inconsistent and like an oversight to me.

Simple tweak and now searches handle things when the word isn't the first word in the search.

Also, inconsistent search logic bothers the heck out of me.

Shouldn't be any issues, nothing really comes to mind. All tests pass and this is unlikely to surprise many people for a 1.0 release.

### DIFF
--- a/evennia/objects/manager.py
+++ b/evennia/objects/manager.py
@@ -324,7 +324,7 @@ class ObjectDBManager(TypedObjectManager):
             search_candidates = (
                 self.filter(
                     type_restriction
-                    & (Q(db_key__istartswith=ostring) | Q(db_tags__db_key__istartswith=ostring))
+                    & (Q(db_key__icontains=ostring) | Q(db_tags__db_key__icontains=ostring))
                 )
                 .distinct()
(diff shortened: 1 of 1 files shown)


Griatch (approved):
Seems reasonable. Merging.

PR #2815, opened 2022-07-25
Improving at_server_startstop modules

volundmush:
Altered AT_SERVER_STARTSTOP_MODULE to alternatively take a list of module paths instead of a string.

Added an at_server_init function which is always called first before any other "start" variants to handle specialized loading. (It's weird that reload_start or cold start is called before normal start but there's nothing that happens first for initial loading of resources or config or whatnot...)

Main reason I thought of it was for helping Athanor - it has its own load process it handles, but it'd be great if this then called the mygame at_server_startstop... at that point I realized, it'd be even better if Evennia just had one more hook and also the ability to optionally call multiple modules in sequence... and that it can be done without overcomplicating anything.

Everything seems to work.

### DIFF
--- a/evennia/game_template/server/conf/at_server_startstop.py
+++ b/evennia/game_template/server/conf/at_server_startstop.py
@@ -7,6 +7,7 @@ allows for customizing the server operation as desired.
 
 This module must contain at least these global functions:
 
+at_server_init()
 at_server_start()
 at_server_stop()
 at_server_reload_start()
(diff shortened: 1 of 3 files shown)


Griatch (changes requested):
Looks like it'd be a good change. I have some minor comments to address first.

Griatch (commented):
I updated the settings string and merged the change along with updates to the changelog. Thanks for the PR!

PR #2818, opened 2022-07-26
object.at_traverse() now passes the exit being traversed as a kwarg

volundmush:
The title kinda says it all.

Because it's useful for rendering messages and other checks in the other hooks, and I cannot think of a single reason why this information is NOT being made available to the move_to chain via kwargs.

Easy peasy.


Griatch (approved):
Approved, but I'm changing the kwarg to `exit_obj` instead, which is more in line with the rest of the API.

PR #2819, opened 2022-07-27
Improved CmdParser

volundmush:
Streamlined the cmdparser so that it has less code when running build_matches, and moved the command matching logic
over to the default Command class using cmd.match(cmdname, include_prefixes: bool).

The default Command class had some unused optimization and caching which are now used properly, accordingly.

When running tests I noticed that this new branch completes `evennia test evennia.commands` about .8 seconds quicker but haven't exactly crunched to figure out if that was just random chance.

Making heavy modifications of the command system easier by exposing more API to within dev reach.

I can now much more easily make a partial match adjustment on Athanor, for instance, using this approach.

All is well!

### DIFF
--- a/evennia/commands/cmdparser.py
+++ b/evennia/commands/cmdparser.py
@@ -61,38 +61,14 @@ def build_matches(raw_string, cmdset, include_prefixes=False):
     """
     matches = []
     try:
-        if include_prefixes:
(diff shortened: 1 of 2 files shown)


Griatch (approved):
I think these look like reasonable refactorings!

Issue #2795, opened 2022-07-07
[BUG] SET command strips capitals from dictionary keys


Tegiminis:
#### Describe the bug
The default SET command strips capitals from dictionary keys, which causes it to fail when attempting to set a dictionary value which has them, instead creating a new attribute with the name `attr['key']`

#### To Reproduce
1. Create a dictionary attribute on yourself named `foo` with the key `FooCap` and the value `3` through `py` or code
2. Attempt to set the dictionary value using the SET command (`set self/foo['FooCap'] = 5`)
3. Observe creation of new attribute instead of modification of existing attribute

#### Expected behavior
SET should not strip capitals from dictionary keys.

#### Environment, Evennia version, OS etc
1.0 dev

#### Additional context
You have to go through `py` or a code path when creating the CamelCase dictionary key in order to repro this bug, as using SET to create the attribute in-client strips the capitals.

Griatch:
I can see the issue since we do enforce Python-style for dicts, and Python is case-sensitive. So this needs to be fixed. 

(As a side note though - Python style is to use lower case for all keys, properties and variables ... and the same can be be said for keys stored in Attributes ... there's very rarely a need to use upper-case letters in any dict keys unless you are deliberately storing some item name in there ...)

aogier:
fix implemented in #2820, should add some more tests maybe

aogier:
fixed in 0b8c6e294dc0b8aa2ca7e908529ab35516af9a1e

PR #2820, opened 2022-07-27
case sensitive set dict keys

aogier:
This PR allow preserve key case when setting them in existing key attributes

close: #2795

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -35,6 +35,8 @@ COMMAND_DEFAULT_CLASS = class_from_module(settings.COMMAND_DEFAULT_CLASS)
 _FUNCPARSER = None
 _ATTRFUNCPARSER = None
 
+_KEY_REGEX = re.compile(r"(?P<attr>.*?)(?P<key>(\[.*\]\ *)+)?$")
+
 # limit symbol import for API
 __all__ = (
     "ObjManipCommand",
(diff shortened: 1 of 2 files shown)


evennia/commands/default/building.py
```diff
@@ -126,7 +128,17 @@ def parse(self):
...
+                    _attrs = []
+                    for match in (
+                        match
+                        for part in map(str.strip, attrs.split("/"))
+                        if part and (match := _KEY_REGEX.match(part.strip()))
+                    ):
+                        attr = match.group("attr").lower()
+                        if match.group("key"):
+                            attr += match.group("key")
+                        _attrs += [attr]
```
Griatch, line 140: Please use `_attrs.append(attr)` here, it's not only faster than using the add operator, it's also more Pythonic. :)

evennia/commands/default/building.py
```diff
@@ -126,7 +128,17 @@ def parse(self):
                     aliases = [alias.strip() for alias in aliases.split(";") if alias.strip()]
                 if "/" in objdef:
                     objdef, attrs = [part.strip() for part in objdef.split("/", 1)]
-                    attrs = [part.strip().lower() for part in attrs.split("/") if part.strip()]
```
Griatch, line 129: I'm not sure I follow why this regex complexity is needed ... could we not just skip the `.lower()` here? :thinking:

Griatch (changes requested):
Thanks for the PR. I'm not sure I follow why the extra regexes are needed to resolve the issue though ... we want to preserve case in the dict of the command, so it seems removing `.lower()` would be enough? Could you eleaborate? 

Either way, we'd need new unit tests to showcase the case-sensitive scenario.

evennia/commands/default/building.py
aogier: What I did here is to preserve a design choice (lowercase attribute names)

```
create box
You create a new Object: box.
set box/CasedAttribute = "foo"
Created attribute box/casedattribute [category:None] = foo
```

fixing a lack of expected functionality when coping w/ dicts

```
set box/dict_attr = {'foo': 'bar'}
Created attribute box/dict_attr [category:None] = {'foo': 'bar'}
set box/dict_attr['FOO'] = 'new'
Modified attribute box/dict_attr [category:None] = {'foo': 'new'}
examine box/dict_attr
-------------------------------------------------------------------------------
Attribute box/dict_attr [category=None] [type: <class 'dict'>]:

{'foo': 'new'}
```

so it has been necessary to extract the `key` part and mangling it separately from the rest of attribute name. Of course we could simply remove the `lower()`, I simply did not want to take the responsibility of introducing a breaking(ish) change :smile_cat:

evennia/commands/default/building.py
aogier: done!

evennia/commands/default/building.py
Griatch: Ah, that's true; the rest of the attribute name does assume case-insensitivity. Fair enough, I buy that. Thanks for elaborating!
 
I would suggest adding a code comment to explain this in the code; always good to help the reader where needed. 
Also still needs unit tests for this case ;)

Griatch (approved):
Thanks for the updates. Looking good now.

Issue #2697, opened 2022-03-24
[BUG - Develop] AttributeProperty and NAttributeProperty default mutable values are reused


ChrisLR:
#### Describe the bug
When using a Mutable default value such as a List or a Dict with AttributeProperty or NAttributeProperty,
the same object will be reused for all instances.

#### To Reproduce
1. Add either AttributeProperty or NAttributeProperty to a typeclass, like a character
2. Add a value in one instance
3. Verify the presence of the value in the second instance

#### Expected behavior
Each instance should have its own list (or whatever) mutable object.

#### Develop-branch commit
 Evennia 1.0-dev (rev 177f4a01d) (rev 177f4a01d)

#### Additional context
I know this is similar to the issue where using a default arg with a mutable object will have the same problem.
Perhaps we could support taking in a callable instead, such as the builtins list or dict.
Having them all share the same object can be difficult to track down

Here is a TestCase for this issue

```
from evennia import DefaultCharacter
from evennia.typeclasses.attributes import AttributeProperty, NAttributeProperty
from evennia.utils.test_resources import EvenniaTest


class CharA(DefaultCharacter):
    test_att_list = AttributeProperty([])
    test_natt_list = NAttributeProperty([])


class TestMutableAttribute(EvenniaTest):
    character_typeclass = CharA

    def test_attribute_default_mutable_value_should_be_distinct(self):
        self.char1.test_att_list.append("test_1")
        assert "test_1" not in self.char2.test_att_list

    def test_nattribute_default_mutable_value_should_be_distinct(self):
        self.char1.test_natt_list.append("test_1")
        assert "test_1" not in self.char2.test_natt_list

```

strikaco:
Good catch!

> Perhaps we could support taking in a callable instead, such as the builtins list or dict.

Yes-- would recommend following Dataclass conventions here wrt `default` and `default_factory`. One won't *always* want an empty list as the default value, but you usually don't want a global mutable either.

i.e.

```
default=6
default_factory=list
default_factory=lambda:['pen','pineapple','apple','pen']
```

ChrisLR:
There is also the fact that users will expect to be able to modify lists and dicts as if it was the _Saver equivalent.
Currently, even with auto create, the first get will not return the object because it does not retrieve it after set.
We could fix the auto create and consider auto create as True unless specified otherwise when it is a list or a dict?

strikaco:
[...]

PR #2824, opened 2022-08-02
AttributeProperty tries spawning child instances of default value (#2697)

strikaco:
Makes AttributeProperty try using default value as a factory.

Lets dev use lambdas or other callables as default value.

#2697

### DIFF
--- a/evennia/typeclasses/attributes.py
+++ b/evennia/typeclasses/attributes.py
@@ -176,7 +176,7 @@ class AttributeProperty:
 
     attrhandler_name = "attributes"
 
-    def __init__(self, default=None, default_factory=None, category=None, strattr=False, lockstring="", autocreate=True):
+    def __init__(self, default=None, category=None, strattr=False, lockstring="", autocreate=True):
         """
         Initialize an Attribute as a property descriptor.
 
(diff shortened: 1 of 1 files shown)


evennia/typeclasses/attributes.py
```diff
@@ -202,6 +202,21 @@ def __init__(self, default=None, category=None, strattr=False, lockstring="", au
...
+    @property
+    def _default(self):
+        """
+        Tries returning a new instance of default if callable.
+        
+        """
+        try:
+            return self.__default()
+        except:
+            return self.__default
```
Griatch, line 215: Never use a 'bare' `except` like this, it's frowned upon in PEP8 standard and is not considered good practice  (always use at least `except Exception`). 

But in this case I think it's also rather motivated to do a `callable` check here instead; it's probably more likely to not have a callable than having one, and relying on catching the exception every time is comparably expensive (even if 50% are callable, it's still about 10x faster to check `callable()` than to use `try...except` here). 

```python 
if callable(self.__default):
    return self.__default()
else:
    return self.__default
```

Griatch (changes requested):
Thanks for the PR! While I'm not convinced the issue in #2697 is really a bug, I think it would be nice to support a callable factory for the utility of it. I have some comments on the implementation below.

strikaco:
Just noticed there's a stray arg in there I decided not to use. Correction pending.

Griatch (approved):

PR #2827, opened 2022-08-03
speedier format/_map()s

aogier:

I'll promise I'll not become a `timeit` zealot but after that:

```
In [3]: %timeit '{foo}'.format(**{'foo': 'bar', 'a': 1})
817 ns ± 6.67 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [4]: %timeit '{foo}'.format_map({'foo': 'bar', 'a': 1})
495 ns ± 3.46 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [5]: %timeit '{foo}'.format(foo='bar', a=1)
620 ns ± 3.25 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

I've did a quick grep and made couple edits :)

thank you, regards

### DIFF
--- a/evennia/contrib/game_systems/crafting/crafting.py
+++ b/evennia/contrib/game_systems/crafting/crafting.py
@@ -625,7 +625,7 @@ class CraftingRecipe(CraftingRecipeBase):
         mapping["outputs"] = iter_to_str(self.output_names)
 
         # populate template and return
-        return message.format(**mapping)
+        return message.format_map(mapping)
 
     @classmethod
     def seed(cls, tool_kwargs=None, consumable_kwargs=None, location=None):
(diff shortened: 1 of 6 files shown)


Griatch:
Interesting. Can't hurt. :) 

_Sidenote: While f-strings are not applicable for these cases, it's interesting to compare `.format` to an f-string too. While this is not something to spent a lot of time on, it's worth to change the old `.format` to f-string as part of doing other refactorings._

```
In [10]: %timeit '{foo}'.format(foo="bar")
177 ns ± 0.184 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [11]: foo = "bar"

In [12]: %timeit f'{foo}'
16.6 ns ± 0.104 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)
````

Griatch (approved):

PR #2830, opened 2022-08-07
row colors preserving paging

### LINKED ISSUE #2646: [BUG - Develop] EvMore breaks inline coloring for HELP command.
#### Describe the bug

Color markup gets lost in HELP output when run through `EvMore`. Only occurs if help output is long enough to require paging.

color_bleed_bug

#### To Reproduce

1. Add many help topics or commands.
2. Run command HELP
3. Observe color bleed.

#### Develop-branch commit

    Evennia 1.0-dev (rev c2f4f8568) (rev c2f4f8568)
    OS: posix
    Python: 3.10.2
    Twisted: 21.7.0
    Django: 4.0.2

aogier:

This PR is a take on #2646. It introduces an unobtrusive new parameter to format_grid so callers can be sure we don't loose colors when paginating.

The simplicity here has the tradeoff of sending some more useless bytes when paging because every line is indeed prefixed with this optional parm, I'm not able to find a solution which does not require some major work across many codebase pieces though.

close #2646


aogier:
e poi suca, coglione :D

PR #2833, opened 2022-08-09
python3.11 removals fixes

aogier:

python 3.11 rc1 is out: this PR fixes the bare minimum needed in order to successfully pass a test run against it by fixing removals/moves in stdlib

HTH, regards!

### DIFF
--- a/.github/workflows/github_action_test_suite.yml
+++ b/.github/workflows/github_action_test_suite.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10']
+        python-version: ['3.9', '3.10', '3.11.0-beta.5']
         TESTING_DB: ['sqlite3', 'postgresql', 'mysql']
 
     steps:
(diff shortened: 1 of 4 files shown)


aogier:
@Griatch is scipy hard required to be before 1.9? It does not compile against 3.11 because includes changes fixed in https://github.com/scipy/scipy/issues/16262 (aka >=1.9)

Griatch:
We don't and won't support Python release candidates so this will not merge. But it's worth to keep this PR on ice for when 3.11 is actually released. 

We pin the version of scipy, but scipy (which is an optional requirement) is really only needed for xyzgrid contrib I think. We can probably up the version as long as it's backwards compatible.

aogier:
Totally ok on keeping this a draft, my question was about if something blocking would be there in next minor. I also pin versions using `<` like in [starlette-authlib](https://github.com/aogier/starlette-authlib/blob/master/pyproject.toml)

ciao!

aogier:
e poi suca, coglione :D

PR #2835, opened 2022-08-14
html and text2html plugins

friarzen:

Add a pair of plugins to expand the webclient's HTML-message support

The first plugin is just a quicky "html" outfunc message capture plugin, in the same vein as the multimedia plugin.  It allows server-side-generated HTML to be displayed "natively" in a pane.

The second plugin is a rework of my old text2html client-side rendering engine, but this time as a simple text2html outfunc plugin handler.  This allows 1.0-devel to have *some* client-side rendering support for those devs that want to go through the effort of updating their commands to use it instead of the server-side renderer.

### DIFF
--- a/docs/source/Components/Webclient.md
+++ b/docs/source/Components/Webclient.md
@@ -2,7 +2,7 @@
 
 Evennia comes with a MUD client accessible from a normal web browser. During development you can try
 it at `http://localhost:4001/webclient`. The client consists of several parts, all under
-`evennia/web/webclient/`:
+`evennia/web`:
 
 `templates/webclient/webclient.html` and `templates/webclient/base.html` are the very simplistic
 django html templates describing the webclient layout.
@@ -18,7 +18,7 @@ be used also if swapping out the gui front end.
 various plugins, and uses the Evennia object library for all in/out.
 
 `static/webclient/js/plugins` provides a default set of plugins that implement a "telnet-like"
-interface.
+interface, and a couple of example plugins to show how you could implement new plugin features.
 
 `static/webclient/css/webclient.css` is the CSS file for the client; it also defines things like how
 to display ANSI/Xterm256 colors etc.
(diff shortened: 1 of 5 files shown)


Griatch (changes requested):
This sounds like useful stopgap solution, but I feel we need to document this if people are to actually use it. 

It seems suitable to expand `evennia/docs/source/Components/Webclient.md` with these new plugins. Maybe a brief text on what one can do with the text2html plugin (an example?) would be good as well.

> A note on doc commits - there is no need to be able to build the docs locally - just add the markdown text above as well as possible and I'll fix it up after merging. :)

friarzen:
@Griatch Docs updated

docs/source/Components/Webclient.md
```diff
@@ -108,8 +110,16 @@ from the server and display them as inline HTML.
...
+* `options2.js` Defines most callbacks. Provides a goldenlayout-based version of the options/settings tab.
+Integrates with other plugins via the custom onOptionsUI callback.
 * `popups.js` Provides default popups/Dialog UI for other plugins to use.  
+* `text2html.js` Provides a new message handler type: `text2html`, similar to the multimedia and html
+plugins. This plugin provides a way to offload rendering the regular pipe-styled ASCII messages
+to the client.  This allows the server to do less work, while also allowing the client a place to
+customize this conversion process.  To use this plugin you will need to override the current commands
+in Evennia, changing any place where a raw text output message is generated and turn it into a
+`text2html` message. For example: `target.msg("my text")` becomes:
+`target.msg(text2html=("my text", {"type": "tag"}))`
```
Griatch, line 122: When reading this, I can think of several questions a casual reader may ask: 
- What HTML does this produce on the client side?
- What is the intent of the `tag` type -- is this the same for all HTML elements or can you provide a type of `span` or `div` separately, for example? 

It would be good with a brief explanation of what this means now that it's mentioned.

Griatch (changes requested):
Overall, this fixes a lot of legacy issues with the web docs, so nice work. I still think the actual plugin info could be a little more detailed.

friarzen:
Added a section with an example for how html and text2html plugins could be used.

Griatch (approved):

PR #2837, opened 2022-08-25
Character Creator contrib

InspectorCaracal:
This adds a light customization of the `charcreate` command and interface, as well as a character creation menu with extensive functionality examples.

Expanding the contribs!


evennia/contrib/base_systems/character_creator/README.md
```diff
@@ -0,0 +1,87 @@
...
+        self.add(ContribCmdCharCreate)
+```
+
+In your game folder `typeclasses/accounts.py`, import and inherit from `ContribChargenAccount` on your Account class.
+
+(Alternatively, you can copy the `at_look` method directly into your own class.)
+
+Example:
+```py
+from evennia.contrib.base_systems.character_creator.character_creator import ContribChargenAccount
```
Griatch, line 27: I'm not sure this belongs in `contribs/base_systems`. I can see the argument for this be a sort of 'base system' for character generation, but on the other hand a character generation system is very much a RPG-staple, so wouldn't `contribs/rpg` be more suitable?

evennia/contrib/base_systems/character_creator/README.md
```diff
@@ -0,0 +1,87 @@
...
+class Account(ContribChargenAccount):
+    # your Account class code
+```
+
+Lastly, in your `settings.py` file, define `CHARGEN_MENU` to your character creation menu module's location.
+
+Example:
+```py
+CHARGEN_MENU = "evennia.contrib.base_systems.character_creator.example_menu"
+```
```
Griatch, line 38: Since this is a setting only introduced for this contrib, I would suggest making this setting optional, and have the code fall back to the contrib-example if nothing is given. This makes the install shorter and easier to test out. Something like

----

By default, the new `charcreate` command provided by this contrib will use the char-creation menu found in `evennia.contrib.base_systems.character_creator.example_menu`. Look at this to get an idea for how it's set up. To tell the contrib code to use your own menu, point to it with the `CHARGEN_MENU` setting. So if your custom chargen menu nodes are found in `mygame/word/chargen_menu.py`, you'd add: 

    CHARGEN_MENU = "world.chargen_menu"

evennia/contrib/base_systems/character_creator/character_creator.py
```diff
@@ -0,0 +1,201 @@
...
+        # this gets called every time the player exits the chargen menu
+        def finish_char_callback(session, menu):
+            char = session.new_char
+            if not char.db.chargen_step:
+                # this means character creation was completed - start playing!
+                # execute the ic command to start puppeting the character
+                account.execute_cmd("ic {}".format(char.key))
+
+        EvMenu(session,
+               settings.CHARGEN_MENU,
```
Griatch, line 100: I'd suggest loading this at the top, with an `AttributeError` leading to falling back to the example menu.

evennia/contrib/base_systems/character_creator/example_menu.py
```diff
@@ -0,0 +1,444 @@
+"""
+An example EvMenu for the character creator contrib.
```
Griatch, line 2: I like the instructional nature of this! :+1:

evennia/contrib/base_systems/character_creator/example_menu.py
```diff
@@ -0,0 +1,444 @@
...
+    return "menunode_categories"
+
+
+#########################################################
+#                  Multiple Choice
+#########################################################
+
+# it's not as important to make this a separate list, but like all the others,
+# it's easier to read and to update if you do!
+_SKILL_OPTIONS = [ "alchemy", "archery", "blacksmithing", "brawling", "dancing", "fencing", "pottery", "tailoring", "weaving" ]
```
Griatch, line 241: This works fine, but it's worth to note that `evennia.utils.evmenu.list_node` decorator is specifically there to auto-generate a multiple-choice menu node. So it could be worth to point to that since it would lead to a lot less code.

evennia/contrib/base_systems/character_creator/example_menu.py
```diff
@@ -0,0 +1,444 @@
...
+        
+        Especially for roleplaying-centric games, being able to choose your
+        character's name after deciding everything else, instead of before,
+        is really useful.
+        
+        Enter a name here to check if it's available.
+        """)
+    help = "You'll have a chance to change your mind before confirming, even if the name is free."
+    # since this is a free-text field, we just have the one
+    options = [ { "key": "_default", "goto": "menunode_check_charname" } ]
```
Griatch, line 396: Minor detail - but FYI, if you have only one option, you don't need to use a list here, but can just return the dict as-is. Keeping the list can be said to be more style-consistent though.

evennia/contrib/base_systems/character_creator/README.md
```diff
@@ -0,0 +1,87 @@
...
+by InspectorCaracal
+
+This contrib is designed to be used in MULTISESSION_MODE = 2 or higher, where characters are not automatically created to match the account. To use this with lower modes, you'll need to implement your own solution for preventing the built-in automatic character creation.
+
+## Installation
+
+In your game folder `commands/default_cmdsets.py`, import and add `ContribCmdCharCreate` to your `AccountCmdSet`.
+
+Example:
+```py
```
Griatch, line 11: Unlike in discord, the full `python` language name should usually be used here.

evennia/contrib/base_systems/character_creator/example_menu.py
```diff
@@ -0,0 +1,444 @@
...
+def menunode_check_charname(caller, raw_string, **kwargs):
+    """Check and confirm name choice"""
+    
+    # strip any extraneous whitespace from the raw text
+    # if you want to do any other validation on the name, e.g. no punctuation allowed, this is the place!
+    charname = raw_string.strip()
+
+    # check to make sure that the name doesn't already exist
+    candidates = Character.objects.filter_family(db_key__iexact=charname)
+    if len(candidates):
```
Griatch, line 408: Why echo the same node? Would it not make more sense to say "This was already taken", and go back to the `menunode_choose_name` node directly - that one likely contains info about the name you can pick (it could fail validation for different reasons after all, such as containing invalid characters etc - things that could be informed of in the first node). 

An alternative to doing two nodes like this would thus be to use a storage-helper (useful for all nodes in the menu) and a validation-helper (for the name node only) and have a single menu node that repeats. This is more a suggestion (if you are not familiar with the pattern) rather than a request for change though. 

```python
def _store_data(caller, raw_string, **kwargs): 
    # generic storage pass-through (useful for all nodes)
    data = kwargs.pop("data", None) 
    next_node = kwargs.pop("next_node", None)  
    if data: 
        character.ndb.chargen_data[data['key']] = data['value']
    return next_node, kwargs 

def _validate_name(caller, raw_string, **kwargs): 
    # validate input 
    if valid: 
        return None, {"name": name}
    else: 
        caller.msg("This name is invalid because ...") 
        return None, {} 

def node_name(caller, raw_string, **kwargs): 

     new_name = kwargs.get("name")

     if new_name: 
         text = "Are you sure?" 
         options = (
                {
                    "key": ("Yes", "y"), 
                    "goto": (_store_data, {"data": {"key": "name", "value": new_name}})
                 },{
                     "key" ("No", "n"),
                     "goto": "node_name"
                }
         )
     else: 
         text = "What name do you want?" 
         options = {"key": "_default",  
                          "goto": _validate_name}
    return text, options

```

evennia/contrib/base_systems/character_creator/example_menu.py
```diff
@@ -0,0 +1,444 @@
...
+        # the name is already taken - loop this node with new input to enter another name
+        text = f"|w{charname}|n is unavailable.\n\nEnter a different name."
```
Griatch, line 410: One thing (not in the original chargen command either) is the normalization of the name. This is to avoid people making names that are visually similar to other names (by using unicode trickery). The helper is `Account.normalize_username`. May be relevant to add here (and probably to the original chargen command too).

Griatch (changes requested):
I think this will be an appreciated contrib. I have some suggestions and comments, but overall this looks very useful and informative. :+1:

InspectorCaracal:
I've looked over the comments but haven't had a chance to start going through them in detail yet. Just wanted to drop a note that this hasn't fallen off my radar. 👍

Griatch:
Ping :)

InspectorCaracal:
> Ping :)

Thanks for the ping! I've finally recovered from all the nonsense involved in having family visiting for two weeks, so I'll get through this on Monday. :+1:

InspectorCaracal:
Okay, I think that covers all the comments! To summarize:

- moved to `contrib.rpg` (I wasn't sure where to put it, so I was hoping if it was the wrong place you'd suggest something better xD)
- used the example menu as a fallback setting
- rewrote the name selection nodes to be structured more cleanly, I don't know *what* I was thinking with how I did it the first time
- added a comment block mentioning list nodes to the example menu

I also didn't know you could assign a single option dict without putting it in a list! For my own code I like the consistency, but I changed those to just the dictionary of the option for the example, since the point of it is to demonstrate what's possible.

InspectorCaracal:
Okay, done with tweaks for real this time.

InspectorCaracal:
I was just reminded! This needs to be updated for the new character settings, I will get that patched in today.

Griatch (approved):

PR #2841, opened 2022-09-01
#2840 Add tzdata to requirements.txt to prevent stack traces

### LINKED ISSUE #2840: [BUG] tzdata missing from requirements.txt
#### Describe the bug
`tzdata`, while required for Timestamps on `[MudInfo]` events, is not in the requirements.txt file. As a result, unless the system already had it installed before evennia was installed, there are stack traces seen in the game whenever a `[MudInfo]` event occurs.
#### To Reproduce
1. Install evennia on a system where `tzdata` is not installed.
2. Start evennia
3. Login with a superuser that recieves `[MudInfo]` events.

#### Expected behavior
`tzdata` should be explicitly declared in requirements.txt as a dependency.

Tested on a fresh Ubuntu 22.04 container with Python 3.10 installed.

Routhinator:
This PR addresses the findings of #2840 ; wherein on systems with Python 3.9+ that do not have a timezone set (containers) python attempts to fallback to use of `tzdata` for timestamps on `[MudInfo]` emits. Adding tzdata to dependencies ensures that Evennia works consistently on these systems.

Added `tzdata` to requirements.txt with no version pinning as tzdata is a timezone data package for Python and should never have compatability issues. Rather we would always want Python to leverage the latest tzdata changes in the event of timezone changes (for example, some timezones dropping daylight savings).

Reducing friction for installs on container systems.

This will resolve #2840 once merged.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,6 +15,7 @@ lunr == 0.6.0
 simpleeval <= 1.0
 uritemplate == 4.1.1
 Jinja2 < 3.1
+tzdata
 
 # try to resolve dependency issue in py3.7
 attrs >= 19.2.0


Griatch:
Thanks for the PR!

PR #2849, opened 2022-09-10
Add 'search' lock for obj.search() and obj.get_visible_contents()

CloudKeeper:
The PR adds checks for a 'search' lock for obj.search() and obj.get_visible_contents() which identifies whether the object should be searchable.

Vanilla Evennia does not allow true hidden objects by default.

The 'view' lock will prevent the object being displayed in a room's description and stop the look command with "Could not view 'object(#9)'", where as attempting to look at a non-existant object returns 'Could not find '<object>''. Likewise, the 'get' lock will disallow getting with "Could not get 'object(#9)" instead of 'Could not find '<object>''. Both of which give away the existance of a hidden object.

The PR adds checks for a 'search' lock for obj.search() and obj.get_visible_contents() which identifies whether the object should be searchable. Where search = false, then the search command will ignore it, achieveing what the 'view' and 'get' locks do not. Because the get command uses search, search=false objects will not be gettable BUT because rooms use self.contents rather than search, they will still display search = false objects unless the 'view' lock is also false. I can see times when a visible object should not be subject to search and so they should be separate locks. However, something that is not searchable will
ordinarily not be visible. Therefore, obj.get_visible_contents() will remove objects that don't pass either test.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -387,6 +387,7 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
         quiet=False,
         exact=False,
         candidates=None,
+        use_locks=True,
         nofound_string=None,
         multimatch_string=None,
         use_dbref=None,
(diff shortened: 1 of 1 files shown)


CloudKeeper:
Pause on this. obj.get_visible_contents() not working as expected.

Griatch (approved):
Thinking out loud: I guess one can discuss whether one should let the 'view' lock serve this purpose instead, since it's probably hard to picture a case where you want 'view' but not 'search' in tandem ... find the item but not visually inspect it (if so, you'd want to have some custom error message anyway) ... 

... But I suppose using `search` as the lock-type is pretty clear; opening up `view` for more custom uses, maybe. It's also more backwards-compatible. So let's go with it. :+1:

PR #2859, opened 2022-09-14
Basic Worming Map Contrib

aMiss-aWry:
The basic map is a simple contrib that works with any evennia project that hasn't diverged from the basic room/exit structure (as long as the exits used in rooms have ordinal directions as their names or aliases). More unusual exits can be easily added by altering the direction dictionary. 

It adds a 'map' command that shows the local map to the user, checking for traversal locks along the way.

An example map display:
```
       |
     -[-]-
       |
       |
-[-]--[-]--[-]--[-]
  |    |    |    |
       |    |    |
     -[-]--[-]  [-]
       | \/ |    |
     \ | /\ |
     -[-]--[-]
```

Everyone loves a simple mapper.

### DIFF
--- a/evennia/contrib/grid/basicmap/basicmap.py
+++ b/evennia/contrib/grid/basicmap/basicmap.py
@@ -1,0 +1,242 @@
+"""
+Basic Map - helpme 2022
+
+This adds an ascii `map` to a given room which can be viewed with the `map` command.
+You can easily alter it to add special characters, room colors etc.
+
+If you don't expect the map to be updated frequently, you could choose to save the
+calculated map as a .ndb value on the room and render that instead of running mapping
+calculations anew each time.
+
+This ignores non-ordinal exits (i.e. exits must be compass directions or up/down to be
+shown on the map).
+
+An example map:
+```
+       |
+     -[-]-
+       |
+       |
+-[-]--[-]--[-]--[-]
+  |    |    |    |
+       |    |    |
+     -[-]--[-]  [-]
+       | \/ |    |
+     \ | /\ |
+     -[-]--[-]
+```
+
+Installation:
+
+Adding the `BasicMapCmdSet` to the default character cmdset will add the `map` command.
+
+Specifically, in `mygame/commands/default_cmdsets.py`:
+
+```
+...
+from evennia.contrib.grid.basicmap import basicmap  # <---
+
+class CharacterCmdset(default_cmds.Character_CmdSet):
+    ...
+    def at_cmdset_creation(self):
+        ...
+        self.add(basicmap.BasicMapCmdSet)  # <---
+
+```
+
+Then `reload` to make the new commands available.
+
+Additional Settings:
+
+In order to change your default map size, you can add to `mygame/server/settings.py`:
+
+BASIC_MAP_SIZE = 5
+
+This changes the default map width/height. 2-5 for most clients is sensible.
+
+If you don't want the player to be able to specify the size of the map, ignore any
+arguments passed into the Map command.
+"""
+import time
+from django.conf import settings
+from evennia import CmdSet
+from evennia.commands.default.muxcommand import MuxCommand
+
+_BASIC_MAP_SIZE = settings.BASIC_MAP_SIZE if hasattr(settings, 'BASIC_MAP_SIZE') else 2
+
+# _COMPASS_DIRECTIONS specifies which way to move the pointer on the x/y axes and what characters to use to depict the exits on the map.
+_COMPASS_DIRECTIONS = {
+    'north': (0, -3, ' | '),
+    'south': (0, 3, ' | '),
+    'east': (3, 0, '-'),
+    'west': (-3, 0, '-'),
+    'northeast': (3, -3, '/'),
+    'northwest': (-3, -3, '\\'),
+    'southeast': (3, 3, '\\'),
+    'southwest': (-3, 3, '/'),
+    'up': (0, 0, '^'),
+    'down': (0, 0, 'v')
+}
+
+
+class Map(object):
+    def __init__(self, caller, size=_BASIC_MAP_SIZE, location=None):
+        self.start_time = time.time()
+        self.caller = caller
+        self.max_width = int(size * 2 + 1) * 5   # This must be an odd number
+        self.max_length = int(size * 2 + 1) * 3  # This must be an odd number
+        self.has_mapped = {}
+        self.curX = None
+        self.curY = None
+        self.size = size
+        self.location = location or caller.location
+
+    def create_grid(self):
+        # Create an empty grid of the configured size
+        board = []
+        for row in range(self.max_length):
+            board.append([])
+            for column in range(int(self.max_width/5)):
+                board[row].extend([' ', '   ', ' '])
+        return board
+
+    def exit_name_as_ordinal(self, ex):
+        exit_name = ex.name
+        if exit_name not in _COMPASS_DIRECTIONS:
+            compass_aliases = [direction in ex.aliases.all() for direction in _COMPASS_DIRECTIONS.keys()]
+            if compass_aliases[0]:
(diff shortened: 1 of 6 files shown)


evennia/contrib/grid/basicmap/basicmap.py
```diff
@@ -0,0 +1,242 @@
...
+        self.caller = caller
+        self.max_width = int(size * 2 + 1) * 5   # This must be an odd number
+        self.max_length = int(size * 2 + 1) * 3  # This must be an odd number
+        self.has_mapped = {}
+        self.curX = None
+        self.curY = None
+        self.size = size
+        self.location = location or caller.location
+
+    def create_grid(self):
```
Griatch, line 94: As per the Evennia [style guide](https://github.com/evennia/evennia/blob/master/CODING_STYLE.md), every non-private method needs a Google-style docstring. This is needed for all methods/functions and classes.

evennia/contrib/grid/basicmap/basicmap.py
```diff
@@ -0,0 +1,242 @@
...
+        exit_name = ex.name
+        if exit_name not in _COMPASS_DIRECTIONS:
+            compass_aliases = [direction in ex.aliases.all() for direction in _COMPASS_DIRECTIONS.keys()]
+            if compass_aliases[0]:
+                exit_name = compass_aliases[0]
+            if exit_name not in _COMPASS_DIRECTIONS:
+                return ''
+        return exit_name
+
+    def update_pos(self, room, exit_name):
```
Griatch, line 113: Example of the needed doc string to add: 
```python
Update the position pointer.

Args: 
    room (Room): The current location. 
    exit_name (str): The name of the exit to to use in this room. This must 
        be a valid compass direction, or an error will be raised. 
Raises: 
    KeyError: If providing a non-compass exit name.

```

This is parsed automatically and becomes part of the documentation API docs, so the style needs to be followed.

Griatch (changes requested):
Thanks for the PR! I think this looks like an interesting and useful addition. I have a few points of feedback before this can be merged: 

- All class/methods/functions must have full doc strings (see comments below).
- I suggest renaming this contrib to `ingame_map_display`. I fear `basicmap` is too generic a name. It's confusing to understand what it does.  We also have two other contribs named "map" something, both of which involves building maps dynamically or from ascii, and it should be made clear that this is not doing the same thing.
- Per Evennia policy, all contribs must have unit tests. You can add a `tests.py` in this repo (see other contribs for examples).

evennia/contrib/grid/ingame_map_display/ingame_map_display.py
```diff
@@ -0,0 +1,322 @@
...
+        return board
+
+    def exit_name_as_ordinal(self, ex):
+        """
+        Get the exit name as a compass direction if possible
+
+        Args:
+            ex (Exit): The current exit being mapped
+        Returns:
+            exit_name (String): The exit name as a compass direction or an empty string
```
Griatch, line 123: The format for a `Return:` statement is `type: Text.`. In the past, Evennia was a bit lax with that, but with the move to full API-doc generation, this would lead to an error.

evennia/contrib/grid/ingame_map_display/ingame_map_display.py
```diff
@@ -0,0 +1,322 @@
...
+        self.curX += _COMPASS_DIRECTIONS[exit_name][1]
+
+    def has_drawn(self, room):
+        """
+        Checks if the given room has already been drawn or not
+
+        Args:
+            room (Room) - Room to check
+        Returns:
+            bool
```
Griatch, line 160: This needs to be on the form `type: Text.`

evennia/contrib/grid/ingame_map_display/ingame_map_display.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def render_room(self, room, x, y, p1='[', p2=']', here=None):
+        """
+        Draw a given room with ascii characters
+
+        Args:
+            room (Room) - The room to render
+            x (int) - The x-value of the room on the grid (horizontally, east/west)
+            y (int) - The y-value of the room on the grid (vertically, north/south)
+            p1 (char) - The first character of the 3-character room depiction
```
Griatch, line 245: Ther is no such thing as a `char` type, the `chr` type is something else. These are `str`.

evennia/contrib/grid/ingame_map_display/tests.py
```diff
@@ -0,0 +1,35 @@
...
+    [ ]--[ ]
+    """
+    def setUp(self):
+        super().setUp()
+        self.west_room = create_object(rooms.Room, key="Room 1")
+        self.east_room = create_object(rooms.Room, key="Room 2")
+        create_object(exits.Exit, key="east", aliases=["e"], location=self.west_room, destination=self.east_room)
+        create_object(exits.Exit, key="west", aliases=["w"], location=self.east_room, destination=self.west_room)
+
+    def west_room_map_room(self):
```
Griatch, line 27: These tests will not run - each test method must start with the string `test_`. Doing so and running this test suite shows that these tests actually both fail.

evennia/contrib/grid/ingame_map_display/ingame_map_display.py
```diff
@@ -0,0 +1,322 @@
...
+
+            self.update_pos(room, ex_name.lower())
+            self.draw_room_on_map(ex.destination, max_distance - 1)
+
+    def draw_exits(self, room):
+        """
+        Draw a given room's exit paths
+
+        Args:
+            room (Room) - The room to draw exits of
```
Griatch, line 195: This needs to be `room (Room): Description ending with period.`

evennia/contrib/grid/ingame_map_display/ingame_map_display.py
```diff
@@ -0,0 +1,322 @@
+"""
+Basic Map - helpme 2022
+
+This adds an ascii `map` to a given room which can be viewed with the `map` command.
+You can easily alter it to add special characters, room colors etc.
```
Griatch, line 5: It would be relevant to mention already up here that the map is dynamically generated on the fly from the existing grid and that it supports all compass directions + up/down.

evennia/contrib/grid/ingame_map_display/README.md
```diff
@@ -0,0 +1,48 @@
+# Basic Map
+
+Contribution - helpme 2022
+
+This adds a `map` to a given room which can be viewed with the `map` command. You can
```
Griatch, line 5: #It would be relevant to mention already up here that the map is dynamically generated on the fly from the existing grid and that it supports all compass directions + up/down. This is the key defining feature of this compared to other map contribs.

Griatch (changes requested):
Thanks for the changes! I was testing this locally and was going to fix the remaining formatting issues myself before merging, but the unit tests are not actually working and even currently fails. 

More things to address: 
- A 'worming' algorithm like this scales poorly to large depth-searches if the grid is large. Currently there seems to be no limit to the `size` you can provide to the `map` command. This is a potential user exploit - they can give 1000 here and kill the server. I strongly suggest introducing a `_MAX_MAP_SIZE` property. It can be applied  as `size = min(size, _MAX_MAP_SIZE`) in the `func()` of `CmdMap`. 
- It could be a good idea to import your `MapDisplayCmdSet` into `__init__.py`; that way your import looks a little nicer; you can do `from evennia.contrib.grid ingame_map_display import MapDisplayCmdSet` instead of `from evennia.contrib.grid.ingame_map_display.ingame_map_disiplay import MapDisplayCmdSet`.

evennia/contrib/grid/ingame_map_display/tests.py
aMiss-aWry: Fixed!

Griatch (approved):
Thanks for the changes!

PR #2874, opened 2022-09-28
Add an API guide

InspectorCaracal:
The REST API is a pretty under-appreciated tool, so I'm giving it some extra attention by writing a guide for it. (Also Wendy asked me to. >.>)

This is a very rough draft, but I don't have a lot of experience writing guides and tutorials so I'd like some input from people on how easy to follow it is before I do formatting and cleanup and all that.

### DIFF
--- a/docs/source/Howtos/Extending-the-REST-API.md
+++ b/docs/source/Howtos/Extending-the-REST-API.md
@@ -1,0 +1,287 @@
+# Extending the REST API
+
+*Note: This tutorial assumes you have a basic understanding of the `web` folder structure. If you don't, please read [the page on the website](https://www.evennia.com/docs/1.0-dev/Components/Website.html) and make sure you understand it before you come back!*
+
+For this tutorial, we'll be adding an `inventory` action to the `characters` endpoint, showing all objects being _worn_ and _carried_ by a character. The first thing you should do is review the [REST API](https://www.evennia.com/docs/1.0-dev/Components/Web-API.html) documentation page. It's not very long, but it covers how to turn the API on and which parts of Django you should be familiar with.
+
+Once you've read that and visited `/api/characters/`, it's time to get started!
+
+## Creating your own viewset
+
+The first thing you'll need to do is create your own views module: `<mygame>/web/api/views.py`
+
+The default REST API endpoints are controlled by classes in `evennia/web/api/views.py` - you could copy that entire file directly and use it, but we're going to focus on changing the minimum.
+
+To start, we'll reimplement the existing `characters` endpoint: a child view of the `objects` endpoint that can only access characters.
+
+```python
+"""
+<mygame>/web/api/views.py
+
+Customized views for the REST API
+"""
+# we'll need these from django's rest framework to make our view work
+from rest_framework.decorators import action
+from rest_framework.response import Response
+from rest_framework import status
+
+# this implements all the basic Evennia Object endpoint logic, so we're inheriting from it
+from evennia.web.api.views import ObjectDBViewSet
+
+# and we need this to filter our character view
+from evennia.objects.objects import DefaultCharacter
+
+# our own custom view
+class CharacterViewSet(ObjectDBViewSet):
+	"""
+	A customized Character view that adds an inventory detail
+	"""
+	queryset = DefaultCharacter.objects.all_family()
+```
+
+## Setting up the urls
+
+Now that we have a viewset of our own, we can create our own urls module and change the `characters` endpoint path to point to ours. The API routing is more complicated than the other `urls.py` files, so you need to copy evennia's from `evennia/web/api/urls.py` to your folder, `<mygame>/web/api/urls.py` and open it in your editor.
+
+Import your new views module:
+
+```python
+from . import views as my_views
+```
+
+Then, find the `characters` path. The original line should look like this:
+
+```python
+router.register(r"characters", views.CharacterViewSet, basename="character")
+```
+
+You'll want to change that `views` to `my_views` to use your new viewset.
+
+```python
+router.register(r"characters", my_views.CharacterViewSet, basename="character")
+```
+
+**TODO: should this have a copy/paste of the final edited file?**
+
+We've almost got it pointing at our new view now. The last step is to add your own API urls - `web.api.urls` - to your web root url module.
+
+Open `web/urls.py` in your editor and add a new path for "api/", pointing to `web.api.urls`. The final file should look something like this:
+
+```python
+from django.urls import path, include
+
+# default evennia patterns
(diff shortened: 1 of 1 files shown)


aMiss-aWry:
Hey this was super helpful! By messing around and just looking at evennia I managed to get like 70% there, the literal examples are great though especially the serializers - it's a nice straightforward guide. I might recommend pointing out that the .db.worn attribute is connected (I THINK) to the contrib.clothing module, not 'base evennia' - that might be different these days but I don't think so - so this would only work if someone uses the clothing contrib or the same method to separate worn/carried. But that's a super minor point. 👍

InspectorCaracal:
> Hey this was super helpful! By messing around and just looking at evennia I managed to get like 70% there, the literal examples are great though especially the serializers - it's a nice straightforward guide. I might recommend pointing out that the .db.worn attribute is connected (I THINK) to the contrib.clothing module, not 'base evennia' - that might be different these days but I don't think so - so this would only work if someone uses the clothing contrib or the same method to separate worn/carried. But that's a super minor point. 👍

Yeah, I actually use a totally different system so when I was adapting it for this I needed a simple example people could easily test themselves, since core evennia doesn't actually have anything like it built in.... :sweat_smile: I figured an attribute is easy enough, since it doesn't require building a whole system or anything.

If you have any other suggestions of something that'd be a useful endpoint action and also a good example of using game attributes or methods to build the fields that *is* built in, I'm all ears!

docs/source/Howtos/Extending-the-REST-API.md
```diff
@@ -0,0 +1,287 @@
...
+class CharacterViewSet(ObjectDBViewSet):
+	"""
+	A customized Character view that adds an inventory detail
+	"""
+	queryset = DefaultCharacter.objects.all_family()
+```
+
+## Setting up the urls
+
+Now that we have a viewset of our own, we can create our own urls module and change the `characters` endpoint path to point to ours. The API routing is more complicated than the other `urls.py` files, so you need to copy evennia's from `evennia/web/api/urls.py` to your folder, `<mygame>/web/api/urls.py` and open it in your editor.
```
Griatch, line 44: Worth repeating what the `urls.py` module actually does - routing the URL onto a python view. Also an idea to link to the corresponding Django docs for those that want to dive deeper.

docs/source/Howtos/Extending-the-REST-API.md
```diff
@@ -0,0 +1,287 @@
+# Extending the REST API
+
+*Note: This tutorial assumes you have a basic understanding of the `web` folder structure. If you don't, please read [the page on the website](https://www.evennia.com/docs/1.0-dev/Components/Website.html) and make sure you understand it before you come back!*
+
+For this tutorial, we'll be adding an `inventory` action to the `characters` endpoint, showing all objects being _worn_ and _carried_ by a character. The first thing you should do is review the [REST API](https://www.evennia.com/docs/1.0-dev/Components/Web-API.html) documentation page. It's not very long, but it covers how to turn the API on and which parts of Django you should be familiar with.
+
+Once you've read that and visited `/api/characters/`, it's time to get started!
+
+## Creating your own viewset
```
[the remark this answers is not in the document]
Griatch: Linking to the django docs on views may also be a good idea, for extra reading.

docs/source/Howtos/Extending-the-REST-API.md
```diff
@@ -0,0 +1,287 @@
...
+		"""
+		Serializes only non-worn objects in the target's inventory.
+		"""
+		carried = [ob for ob in obj.contents if not obj.db.worn]
+		return SimpleObjectDBSerializer(carried, many=True).data
+```
+
+For this, we're assuming that whether an object is being worn or not is stored in the `worn` db attribute, but this can easily be done differently to match how your game's data is structured: `obj.tags.has('equipped')` for example. And you can easily divide up the returned data into different inventory sections by changing or adding the properties, fields, and methods attached.
+
+Just remember: `worn = serializers.SerializerMethodField()` is how the API knows to use `get_worn`, and `Meta.fields` is the list of fields that will actually make it into the final JSON.
```
Griatch, line 172: This is a good repeat, making sure people are not lost! :+1:

Griatch: You could also try formatting this as a Note: 

```
> Just remember: ...
> the list of ...
```

docs/source/Howtos/Extending-the-REST-API.md
```diff
@@ -0,0 +1,287 @@
...
+## Customizing API permissions
+
+Evennia comes with its own custom API permissions class, connecting the API permissions to the in-game permission hierarchy and locks system. Since we're trying to access the object's data now, we need to pass the `has_object_permission` check as well as the general permission check - and that permission class hardcodes the actions into the object permission checks.
+
+So, since we've added a new action - `inventory` - to our characters endpoint, we need to use our own custom permissions on our characters endpoint. Create one more module file: `<mygame>/web/api/permissions.py`
+
+Like with the previous classes, we'll be inheriting from the original and extending it to take advantage of all the work Evennia already does for us.
+
+```python
+from evennia.web.api.permissions import EvenniaPermission
```
Griatch, line 231: Even though it's mentioned on the lines above, I find it's a good idea is to repeat just what file this code should go in. Many just browse the code and don't read the main text unless they have to.

```` 
```python 
# in mygame/web/api/permissions.py

from evennia.web.api.permissions import EvenniaPermission
#...
````

File: docs/source/Howtos/Extending-the-REST-API.md [hunk 5]
```diff
@@ -0,0 +1,287 @@
...
+	"""
+	permission_classes = [CharacterPermission]
+	queryset = DefaultCharacter.objects.all_family()
+
+	@action(detail=True, methods=["get"])
+	def inventory(self, request, pk=None):
+		obj = self.get_object()
+		return Response( InventorySerializer(obj).data, status=status.HTTP_200_OK )
+```
+
```
Griatch, line 284: A suggestion is to have a final `## Conclusions` section at the end. This can summarize what you have learned, suggest further things to try and even link to some other tutorial(s) that could be good to try next.

Griatch (commented):
This is a good and useful effort I think! People will find this useful, no doubt. 
I know this is marked as a draft, but since you write that you want feedback, I have given some high-level inline suggestions.

Griatch:
@InspectorCaracal Ping :)

Griatch:
@InspectorCaracal Ping.

Griatch (approved):
Thanks for the updates! Nice work! I'll merge it and tweak it a little to work better with the doc generator. :+1:

Issue #2706, opened 2022-03-31
[BUG - Develop] Global Scripts loading doesn't catch parse errors


InspectorCaracal:
#### Describe the bug
When there's an error with loading a global scripts module, e.g. it imports the `Room` class but that class has a syntax error, the loading error isn't caught and the traceback points to [this line](https://github.com/evennia/evennia/blob/251a70275bbd98a3e157cbb4c025597a4bb24ac9/evennia/utils/containers.py#L135) with a KeyError

#### To Reproduce
1. Create a simple global script which imports a class such as `Room` from your game dir, and add it to `settings.py`
2. Introduce a syntax error to your `Room` class, such as an extra space in an indent.
3. Reload the server.

#### Expected behavior
It'd be more useful if there was an error thrown which indicated that the script had failed being loaded (and, ideally, why), as the current error invariably leads people to think there is a problem with the global scripts dict in `settings.py`.

#### Develop-branch commit
f093c8bcb

#### Additional context
I feel like it should be catching that kind of an error in [`class_from_module`](https://github.com/evennia/evennia/blob/f093c8bcb1321a44e33003cf91cf9f565e028de7/evennia/utils/utils.py#L1524) or it should at least be throwing an exception in [`load_data`](https://github.com/evennia/evennia/blob/f093c8bcb1321a44e33003cf91cf9f565e028de7/evennia/utils/containers.py#L202) but neither seem to be doing so. Instead, they just silently don't load the script.

Henddher:
Hi @InspectorCaracal

Thanks for the steps to reproduce!

This is the stack trace I'm getting. I think it matches your description.

```shell
...evennia_mygame$ evennia start
Traceback (most recent call last):
  File "/home/henddher/Documents/github/evennia_venv/bin/evennia", line 10, in <module>
    main()
  File "/home/henddher/Documents/github/evennia/evennia/server/evennia_launcher.py", line 2273, in main
    init_game_directory(CURRENT_DIR, check_db=True)
  File "/home/henddher/Documents/github/evennia/evennia/server/evennia_launcher.py", line 1761, in init_game_directory
    check_database()
  File "/home/henddher/Documents/github/evennia/evennia/server/evennia_launcher.py", line 1462, in check_database
    evennia._init()
  File "/home/henddher/Documents/github/evennia/evennia/__init__.py", line 363, in _init
    GLOBAL_SCRIPTS.start()
[...]

PR #2882, opened 2022-10-01
fix: Global Scripts loading now catches parse errors

Henddher:
See https://github.com/evennia/evennia/issues/2706#issuecomment-1264514914 and https://github.com/evennia/evennia/issues/2706#issuecomment-1264527145

#2706

### DIFF
--- a/evennia/utils/containers.py
+++ b/evennia/utils/containers.py
@@ -106,6 +106,7 @@ class GlobalScriptContainer(Container):
         callables from settings but a custom dict of tuples.
 
     """
+    __BASE_SCRIPT_TYPECLASS = class_from_module(settings.BASE_SCRIPT_TYPECLASS)
 
     def __init__(self):
         """
@@ -205,7 +206,9 @@ class GlobalScriptContainer(Container):
             for key, data in list(self.loaded_data.items()):
                 try:
                     typeclass = data.get("typeclass", settings.BASE_SCRIPT_TYPECLASS)
-                    self.typeclass_storage[key] = class_from_module(typeclass)
+                    script_typeclass = class_from_module(typeclass)
+                    assert issubclass(script_typeclass, self.__BASE_SCRIPT_TYPECLASS)
+                    self.typeclass_storage[key] = script_typeclass
                 except Exception:
                     logger.log_trace(
                         f"GlobalScriptContainer could not start import global script {key}. "
--- a/evennia/utils/containers.py
+++ b/evennia/utils/containers.py
@@ -202,14 +202,17 @@ class GlobalScriptContainer(Container):
         """
         if self.typeclass_storage is None:
             self.typeclass_storage = {}
-            for key, data in self.loaded_data.items():
+            for key, data in list(self.loaded_data.items()):
                 try:
                     typeclass = data.get("typeclass", settings.BASE_SCRIPT_TYPECLASS)
                     self.typeclass_storage[key] = class_from_module(typeclass)
                 except Exception:
-                    logger.log_trace(
-                        f"GlobalScriptContainer could not start import global script {key}."
+                    logger.log_err(
+                        f"GlobalScriptContainer could not start import global script {key}. "
+                        "It will be removed (skipped)."
                     )
+                    # Let's remove this key/value. We want to let other scripts load.
+                    self.loaded_data.pop(key)
 
     def get(self, key, default=None):
         """
--- a/evennia/utils/containers.py
+++ b/evennia/utils/containers.py
@@ -207,7 +207,7 @@ class GlobalScriptContainer(Container):
                     typeclass = data.get("typeclass", settings.BASE_SCRIPT_TYPECLASS)
                     self.typeclass_storage[key] = class_from_module(typeclass)
                 except Exception:
-                    logger.log_err(
+                    logger.log_trace(
                         f"GlobalScriptContainer could not start import global script {key}. "
                         "It will be removed (skipped)."
                     )
--- a/evennia/utils/tests/test_containers.py
+++ b/evennia/utils/tests/test_containers.py
@@ -2,6 +2,7 @@ import unittest
 
 from evennia.utils import containers
 from django.conf import settings
+from django.test import override_settings
 from evennia.utils.utils import class_from_module
 
 _BASE_SCRIPT_TYPECLASS = class_from_module(settings.BASE_SCRIPT_TYPECLASS)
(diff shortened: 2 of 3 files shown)


evennia/utils/containers.py
```diff
@@ -202,14 +202,17 @@ def load_data(self):
         """
         if self.typeclass_storage is None:
             self.typeclass_storage = {}
-            for key, data in self.loaded_data.items():
+            for key, data in list(self.loaded_data.items()):
```
Henddher, line 205: to "snapshot" the contents of the `dict` because below I may remove `key`.

Henddher:
@InspectorCaracal, @Griatch

evennia/utils/containers.py
```diff
@@ -202,14 +202,17 @@ def load_data(self):
...
                 try:
                     typeclass = data.get("typeclass", settings.BASE_SCRIPT_TYPECLASS)
                     self.typeclass_storage[key] = class_from_module(typeclass)
                 except Exception:
-                    logger.log_trace(
-                        f"GlobalScriptContainer could not start import global script {key}."
+                    logger.log_err(
```
Griatch, line 210: It sounds like this should still use `log_trace` in order to properly log the exception that it hits to get to this point

Griatch (changes requested):
Thanks for the PR1 I saw the feedback in the issue, but I think this is a good solution. While there's a point that missing a global script may lead to other problems, loading it (or in this case skipping to load it) is in line with Evennia's general policy of at least trying to get things going, but to log the errors. 

One _could_ consider the Typeclass approach of temporarily loading the `EmptyScript` on the global script container, but the reason we fall back `DefaultObject` for typeclasses is that a typeclass is required for the server to boot at all. There is (at least not by default) no requirement for global scripts to load for the server itself to boot. So I'm fine with simply skipping the load like this. 

I have a comment below, and we do need unit tests for this. I'd suggest using `unittest.mock` to test the `containers.GlobalScriptContainer.load_data`.

Henddher:
Thank you @Griatch 

(I addressed your `log_trace` comment)

As I mentioned in the discussion, I am not married to either approach (hard-fail vs forbid-errors). However, I think @InspectorCaracal's "hard fail" view is worth accommodating.

I think that by adding a *sensible default* (*hard-fail* or *forbid-errors*) at the `GENERAL_SCRIPTS.start(settings.FAILED_SCRIPT_LOAD_SETTING)` would be fairly simple.
 
Then, at the script level, an `"allow_load_failure": bool` param would allow any script to reverse that default.
 
Both of these combined will, IMHO, offer both camps of server maintainers the best of both worlds.
I truly think this is good quality of life change and the complexity (fairly low) is worth implementing.

I'd advocate for a configurable option.

Thoughts?

I could even push another PR with that change so we can decide which one to merge. It'll not be much of a hassle, I think. We're in #Hacktoberfest anyways 🎉🎉🎉

RE tests, I'll add them.

evennia/utils/containers.py
Henddher: Reverted it. 
My bad, I thought trace was a lower than DEBUG log level instead of a "traceback/exception" log level.

Griatch:
I'd say, keep this solution. As long as it properly logs the error, it's reasonably close to the way typeclasses are handled. A setting just adds a lot of extra complexity.

Henddher:
Sounds good.
I'll add a couple of tests.

Henddher:
Tests added. No need to mock anything :smile:

selberhad:
Hi @Henddher, what do you think about changing the title of this PR to something more descriptive? E.g.
```
fix: Global Scripts loading now catches parse errors
```
Cheers!

evennia/utils/containers.py
```diff
@@ -106,6 +106,7 @@ class GlobalScriptContainer(Container):
         callables from settings but a custom dict of tuples.
 
     """
+    __BASE_SCRIPT_TYPECLASS = class_from_module(settings.BASE_SCRIPT_TYPECLASS)
```
Henddher, line 109: Added this to load the BASE_SCRIPT_TYPECLASS class so we can `assert` (during `load_data()`) that the specified script class *is* a subclass.

evennia/utils/tests/test_containers.py
```diff
@@ -0,0 +1,86 @@
+import unittest
+
+from evennia.utils import containers
+from django.conf import settings
+from evennia.utils.utils import class_from_module
+
+_BASE_SCRIPT_TYPECLASS = class_from_module(settings.BASE_SCRIPT_TYPECLASS)
```
Griatch, line 7: The problem with this is that it will change with, well, settings, if this is run from `mygame`. I'd suggest
- Since this is in core, just import and use `DefaultScript` here
- or - use the [override_settings](https://docs.djangoproject.com/en/4.1/topics/testing/tools/#django.test.override_settings) to lock the settings per-test.

evennia/utils/containers.py
Griatch: Please don't resolve your own conversations. Presumably they are for the reviewer to see.

evennia/utils/tests/test_containers.py
```diff
@@ -0,0 +1,86 @@
...
+  def objects(self):
+    from evennia import module_that_doesnt_exist
+
+class TestGlobalScriptContainer(unittest.TestCase):
+
+  def setUp(self):
+    settings.GLOBAL_SCRIPTS = {}
+
+  def patch_global_scripts(self, adict):
+    settings.GLOBAL_SCRIPTS.update(adict)
```
Griatch, line 28: Since this is a helper, mark it 'private', by naming it `_patch_global_script`. Makes things clearer. 
You could also opt to use `mock.patch` here to set set `GLOBAL_SCRIPTS` data only for this test.

evennia/utils/tests/test_containers.py
```diff
@@ -0,0 +1,86 @@
...
+class WorseScript(_BASE_SCRIPT_TYPECLASS):
+  """objects will fail upon call"""
+  @property
+  def objects(self):
+    from evennia import module_that_doesnt_exist
+
+class TestGlobalScriptContainer(unittest.TestCase):
+
+  def setUp(self):
+    settings.GLOBAL_SCRIPTS = {}
```
Griatch, line 25: This should not be changed on the `settings` object directly. This could affect other test-runs. Instead you use `override_settings` or `mock.patch` to replace this in-place for this test only. 

If you use an approach like this, you should restore the settings afterwards, by saving their real value and then putting them back in a `tearDown` method (but this could cause issues when running tests in parallel).

Griatch (changes requested):
Getting there! Some comments on the unit tests that need addressing.

evennia/utils/tests/test_containers.py
Henddher: Thanks @Griatch! I'll go with `override_settings`

evennia/utils/tests/test_containers.py
Henddher: By using `override_settings`, `setUp()` and `_patch_global_scripts(...)` became useless so I removed them both.

evennia/utils/tests/test_containers.py
Griatch: (Btw, please let the original author close conversations as needed, otherwise they need to open it to figure out what it was all about.)

Griatch (approved):
Thanks for the changes! Looking good now. :)

Henddher:
Thanks!

PR #2887, opened 2022-10-02
Webclient copy fix

VexiaOnline:

Fix/address #6211. Add the ability to copy text from webclient panels using Ctrl + C, (or Cmd + c on Mac.)

Requested by #6211 , I also thought it would be a good idea to have the ability to copy text without having focus shifted to an input panel automatically.

### DIFF
--- a/evennia/web/static/webclient/js/plugins/default_in.js
+++ b/evennia/web/static/webclient/js/plugins/default_in.js
@@ -17,6 +17,14 @@ let defaultInPlugin = (function () {
             inputfield = $("#inputfield:focus");
         }
 
+        // Allows you to copy from panels.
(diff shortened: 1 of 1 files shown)


Griatch (changes requested):
Thanks for the PR, but this was done against the wrong branch (master) and thus pulls in over a thousand unrelated files. You need to direct this towards the `develop` branch instead. You can try to do so via the github UI https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request.

VexiaOnline:
Oops. I thought I had switched it. It should be the proper one now.

Griatch (approved):
Thanks for the PR!

PR #2888, opened 2022-10-02
Use self.name instead of self.key to make better use of get_numberered_name

VexiaOnline:

Simply changes `self.key` to `self.name` so that if `get_numbered_name` isn't overridden in the call, it will use the name property instead of key, this brings thing in line with other areas, so colorized object names can persist in `get_numbered_name`. 

I believe this is the only place that uses .key instead of .name, and I wanted the consistency of using .name so that other areas didn't need to be overridden to display object name colorization etc.


Griatch (changes requested):
This seems to have been made off a version of Evennia that is 8 months old? Maybe it was initially done based of `master` branch? Either way, all these files have changed in develop branch and now collide with what you have in this commit (see the github alert, you can also see what happens if you try to merge your branch back into a copy of `develop`). 

You need to push a clean PR, so you need to resolve the conflicts and re-push (you can update your PR in place). Alternatively you can redo your changes on a fresh copy of `develop` branch to make sure you are extending that branch and nothing else.

Issue #2883, opened 2022-10-01
[BUG - Develop] False positives on URL detection


InspectorCaracal:
#### Describe the bug
If someone e.g. says "Awww." in-game, the pattern `www.X` gets picked up as a web address by the text2html parser.

#### To Reproduce
1. Open up the webclient for an active game.
2. `emote www.t`
3. See error

#### Expected behavior
The URL detection for the text2html parser needs to be reworked to avoid false positives, or possibly removed entirely. I'm of the opinion to remove it entirely. If I take on fixing this, which is likely, that's the approach I'll take.

#### Develop-branch commit
3992f37da

selberhad:
I looked into this, IMHO the regular expression used to pick up URLs would benefit from being a bit more restrictive.

That said, there would be some design decisions to make. For example, currently typing wwww.example.com will turn into w[www.example.com](www.example.com), which seems pretty weird to me, unless it's just trying to be forgiving of common typos.

I think the ideal user experience would result from requiring a boundary character `\b` before the URL in the regex, but that would break [this test](https://github.com/evennia/evennia/blob/master/evennia/utils/tests/test_tagparsing.py#L307-L311).

Personally I think this is a nice feature to have, as long as it errs on the side of not hyperlinking non-links.

InspectorCaracal:
I'm of the general opinion that things like identifying sections of message strings as URLs should all be done client-side, which in this case would mean reimplementing it in the webclient. If we *do* want URL detection to be done server-side, then I think it should be done consistently for all protocols - but as I said, I think it's the kind of feature that should be done by the client.

selberhad:
I don't have a strong opinion here, but why URLs specifically? You could just send pure ANSI over the websocket and have all of the ANSI to HTML conversion live client-side.

Even if moved client-side, there would still be all of the same decisions to make about permissiveness. E.g. right now it thinks `http!!!!!example.2` is a URL.

aMiss-aWry:
One reason to do it server side could be parsing URLs into clickable links via MXP for clients that support it.

InspectorCaracal:
> I don't have a strong opinion here, but why URLs specifically? You could just send pure ANSI over the websocket and have all of the ANSI to HTML conversion live client-side.

[...]

Issue #1974, opened 2019-10-25
[BUG] Various inconsistencies in evennia/utils/text2html


Henddher:
[text2html.py](evennia/blob/evennia/utils/text2html.py) has various inconsistencies. They are demonstrated by [skipped tests](evennia/blob/evennia/utils/tests/test_text2html.py).

- When an ansi string contains a HILITE effect with a subsequent UNHILITE effect, the UNHILITE effect char seq is removed but the effect is applied until the end of the string. https://github.com/evennia/evennia/pull/1969#discussion_r338683863
- When an ansi string contains an effect followed by a NORMAL effect, the effect terminates at the correct location but the NORMAL effect char sequence is not removed from the string. https://github.com/evennia/evennia/pull/1969#discussion_r338905658, https://github.com/evennia/evennia/pull/1969#discussion_r338906034
- `convert_urls` doesn't perform URL encoding correctly.

GulliblePsychologist:
On top of that, if an attribute contains a URL then:

1. If the attribute is too big, the URL gets cut off but the link is still converted by `convert_urls` to... something.

2. Even if the attribute fits, `utils.crop` adds breaks to the string. It's still picked up `convert_urls` which preserves the breaks in the "href" attribute of the "A" tag. Finally, the newlines are converted to BR elements. The result is a broken link that leads to "&lt;br" and half the text of the object you're examining is converted to a link.

whitenoiseoss:
I believe a better `re_url` may be in order to not cut URLs off. Here is one I've been testing on several cases successfully: `((https?|ftp)://)?(-\.)?([^\s/?\.#-]+\.?)+(/[^\s]*)?$`

InspectorCaracal:
The current development branch `develop` has an entirely rewritten `text2html` parser that solves many, if not all, of these problems.

selberhad:
Cool, I'm happy to put together a PR if it's desired. @Griatch, thoughts?

Griatch:
No, we are not spending time on backporting to `master` at this time, except it's something really breaking. The `develop` branch will merge into `master` when 1.0 is releasing. 

If we can no longer reproduce the issues reported here in `develop` branch, we'll mark this is as `Implemented` so that it will close once it merges.

PR #2895, opened 2022-10-03
fix: improve convert_urls

selberhad:
- Adds a boundary character `\b` before URLs to cut down on false positives.
- Prefixes urls without a protocol with `//` to prevent browsers opening them as relative links on the webclient's domain. (Maybe this should just be `https://` rather than using the protocol-relative `//`? Opinions welcome!)
- Updates tests for urls without protocols, removes leading-w test, adds new test for boundary character requirement

- Links of the form "www.example.com" currently open relative to the webclient URL, which is not expected behavior.
- Cuts down on false positives by requiring a non-word character before a url, so `www.example` will turn into a link on its own, but it won't highlight it in `Awww.example is cute`.

Relates to #2883, #1974


Griatch (changes requested):
This needs to be redirected against `develop` branch. You can do this change in the github UI.

selberhad:
@Griatch Looks like there are a bunch of conflicts when I do it through the github UI, and I've always found rebasing on the command line to be rather confusing. I can just close this one and submit a new PR if that sounds good?

Griatch:
@selberhad No problem :+1:

PR #2897, opened 2022-10-03
fix: improve convert_urls

selberhad:
- Adds a boundary character `\b` before URLs to cut down on false positives.
- Prefixes urls without a protocol with `//` to prevent browsers opening them as relative links on the webclient's domain. (Maybe this should just be `https://` rather than using the protocol-relative `//`? Opinions welcome!)
- Prevent urls with no protocols that are too short or long to be valid URLs according to the RFC from being converted to links
- Updates tests for urls without protocols, removes leading-w test, adds new test for boundary character requirement

- Links of the form "www.example.com" currently open relative to the webclient URL, which is not expected behavior.
- Cuts down on false positives by requiring a non-word character before a url, so `www.example` will turn into a link on its own, but it won't highlight it in `Awww.example is cute`.

Relates to #2883, #1974

### DIFF
--- a/evennia/utils/tests/test_tagparsing.py
+++ b/evennia/utils/tests/test_tagparsing.py
@@ -250,13 +250,13 @@ class TestTextToHTMLparser(TestCase):
     def test_url_scheme_ftp(self):
         self.assertEqual(
             self.parser.convert_urls("ftp.example.com"),
-            '<a href="ftp.example.com" target="_blank">ftp.example.com</a>',
+            '<a href="//ftp.example.com" target="_blank">ftp.example.com</a>',
         )
 
     def test_url_scheme_www(self):
         self.assertEqual(
             self.parser.convert_urls("www.example.com"),
-            '<a href="www.example.com" target="_blank">www.example.com</a>',
+            '<a href="//www.example.com" target="_blank">www.example.com</a>',
         )
 
     def test_url_scheme_ftpproto(self):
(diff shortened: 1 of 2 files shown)


Griatch (changes requested):
This looks like a good change. I'm not really sure about the `//` prefix. I'd think a full `http://` or `https://` prefix looks more solid ... but I don't have a super-strong opinion about it, so I'm open for debate. :)

selberhad:
Since we are essentially guessing what the protocol should be when the user hasn't provided one, there's no definitive best choice. Some points to consider:
1. Most (but not all) webservers will automatically upgrade `http` to `https` if they support secure connections
2. Conversely, webservers that only support `http` will time out if issued an `https` request
3. The vast majority of websites now support TLS (https) thanks to [LetsEncrypt](https://letsencrypt.org/)

My inclination would be to use `http`, but cybersecurity professionals would disagree 😁 . If a web server supports both but doesn't automatically upgrade connections, all modern browsers will at least indicate that the connection is 'Not Secure' for `http` requests, which seems better than just having the request time out for `https` requests if the server doesn't support TLS. That said, due to point 3, 99.9% of websites will work with `https` so either choice would probably be fine.

selberhad:
Hmm, not too sure what's going on with the merge here... will try to fix it up.

Griatch (approved):
Some good points there, and I agree - a `http` prefix seems like it would break the least :) Thanks for the fixes!

PR #2898, opened 2022-10-03
Use `self.name` instead of `self.key` as fallback in `DefaultObject.get_numbered_name`

VexiaOnline:

Simply changes self.key to self.name so that if get_numbered_name isn't overridden in the call, it will use the name property instead of key, this brings thing in line with other areas, so colorized object names can persist in get_numbered_name.

I believe this is the only place that uses .key instead of .name, and I wanted the consistency of using .name so that other areas didn't need to be overridden to display object name colorization etc.

Not sure why it said before that my develop was behind main one, hopefully this time things are in sync?

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -1249,7 +1249,7 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
 
         """
         plural_category = "plural_key"
-        key = kwargs.get("key", self.key)
+        key = kwargs.get("key", self.name)
         key = ansi.ANSIString(key)  # this is needed to allow inflection of colored names
         try:
             plural = _INFLECT.plural(key, count)


evennia/objects/objects.py
```diff
@@ -1249,7 +1249,7 @@ def get_numbered_name(self, count, looker, **kwargs):
 
         """
         plural_category = "plural_key"
-        key = kwargs.get("key", self.key)
+        key = kwargs.get("key", self.name)
```
Griatch, line 1252: This change needs to be mentioned on line 1240 too - explaining that if `key` is not given, the `.name` property will be used instead.

Griatch (changes requested):

Griatch:
@VexiaOnline Ping, since I know you may not be keeping an eye on this :)

VexiaOnline:
I haven not forgotten about this,  just on a plane to glorious Sweden, I will look and make the changes when I get settled in and my laptop set up. :)

Griatch:
@VexiaOnline Ping, and probably the last chance to have this count for Hacktoberfest :)

Griatch:
@VexiaOnline I merged this and added the requested doc string myself instead :)

PR #2903, opened 2022-10-05
Refactor `pronouns.py`

### LINKED ISSUE #2683: [BUG - Develop] $pron incorrectly maps "plural" pronouns
#### Describe the bug
When using the `$pron` inline func from funcparser, many of the pronouns return the singular "neutral" pronouns instead of the "plural" pronouns.

#### To Reproduce
1. Create a second character with the `.gender` attribute of "plural"
3. Pass the parser a string `"$pron(you) $pron(you,object) $pron(your) $pron(yours) $pron(yourself) - $pron(I) $pron(me) $pron(my) $pron(mine) $pron(myself)"` so that you are the receiver and your new character is the caller
4. View the results

#### Expected behavior
The output should be: `they them their theirs themselves - they them their theirs themselves`

The actual output is: `they them their theirs itself  - it it its theirs themselves`

#### Develop-branch commit
473f2e379

InspectorCaracal:
Refactors the module for better readability/ease of future updates, and updates the pronoun data to be consistent across the neutral/plural 'gender' categories.

The original dictionary structure had a lot of redundancy in defining valid mappings, where it started with each possible pronoun and stepped down through all possible outcomes to a final list of pronouns (which results in the same sets of pronouns being entered multiple times)

While the current result still has some redundancy, I streamlined it by breaking it into two different variables:
- `PRONOUN_TABLE` defines the valid viewpoint, part of speech, and gender for each pronoun
- `PRONOUN_MAPPING` gives the target pronoun for each viewpoint, part of speech, and gender

This makes it imo easier to understand what the validity restrictions are, as well as modify them.

The function itself is still pretty similar; the main difference is it separates the differentiation from stepping through the mappings since they're now two separate dictionaries.

Resolves #2683

### DIFF
--- a/evennia/utils/verb_conjugation/pronouns.py
+++ b/evennia/utils/verb_conjugation/pronouns.py
@@ -407,8 +407,8 @@ def pronoun_to_viewpoints(
 
     # step down into the mapping
     viewpoint_map = PRONOUN_MAPPING[viewpoint]
-    pronouns = viewpoint_map.get(pronoun_type) or viewpoint_map.get(DEFAULT_PRONOUN_TYPE)
-    mapped_pronoun = pronouns.get(gender) or pronouns.get(DEFAULT_GENDER)
+    pronouns = viewpoint_map.get(pronoun_type, viewpoint_map[DEFAULT_PRONOUN_TYPE])
+    mapped_pronoun = pronouns.get(gender, pronouns[DEFAULT_GENDER])
     
     # keep the same capitalization as the original
     if pronoun != "I":
--- a/evennia/utils/verb_conjugation/pronouns.py
+++ b/evennia/utils/verb_conjugation/pronouns.py
@@ -383,10 +383,6 @@ def pronoun_to_viewpoints(
 
     # check if pronoun maps to multiple options and differentiate
     # but don't allow invalid differentiators
-    if is_iter(source_gender):
-        gender = gender if gender in source_gender else source_gender[0]
-    else:
-        gender = source_gender
     if is_iter(source_type):
         pronoun_type = pronoun_type if pronoun_type in source_type else source_type[0]
     else:
@@ -396,6 +392,18 @@ def pronoun_to_viewpoints(
         viewpoint = viewpoint if viewpoint in target_viewpoint else target_viewpoint[0]
     else:
         viewpoint = target_viewpoint
+
+    # special handling for the royal "we"
+    if is_iter(source_gender):
+      gender_opts = list(source_gender)
+    else:
+      gender_opts = [source_gender]
+    if viewpoint == "1st person":
+        # make sure plural is always an option when converting to 1st person
+        # it doesn't matter if it's in the list twice, so don't bother checking
+        gender_opts.append("plural")
+    # if the gender is still not in the extended options, fall back to source pronoun's default
+    gender = gender if gender in gender_opts else gender_opts[0]
 
     # step down into the mapping
     viewpoint_map = PRONOUN_MAPPING[viewpoint]
--- a/evennia/utils/verb_conjugation/pronouns.py
+++ b/evennia/utils/verb_conjugation/pronouns.py
@@ -10,21 +10,20 @@ the 3rd person case and back. In some cases, the mapping is not unique; it is as
 differentiate between the options in some other way.
 
 
-====================    =======    ========    ==========    ==========    ===========
-viewpoint/pronouns      Subject    Object      Possessive    Possessive    Reflexive
-                        Pronoun    Pronoun     Adjective     Pronoun       Pronoun
-====================    =======    ========    ==========    ==========    ===========
-1st person                I          me          my           mine         myself
-1st person plural         we         us          our          ours         ourselves
-2nd person                you        you         your         yours        yourself
-2nd person plural         you        you         your         yours        yourselves
+====================  =======  ========  ==========  ==========  ===========
+viewpoint/pronouns    Subject  Object    Possessive  Possessive  Reflexive
+                      Pronoun  Pronoun   Adjective   Pronoun     Pronoun
+====================  =======  ========  ==========  ==========  ===========
+1st person              I        me        my         mine       myself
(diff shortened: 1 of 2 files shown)


InspectorCaracal:
Didn't check to see if the neutral/plural pronouns were in the tests before pushing, will get that updated.

InspectorCaracal:
All right, I just pushed a commit that will fix most of the issues, but I'm mulling over the "royal we" before it'll pass everything.

Griatch (commented):
It certainly looks different :grin: 
It's quite hard to review this big a change for functionality though; it would be useful if you could maybe add some line comments (in the PR on github) on why certain changes are done and the thinking for doing the change.

InspectorCaracal:
@Griatch I'm not really sure how to effectively add line comments to this PR, since as you said it's a very big change (and the git diff breaks things up in ways that make it harder to track), so I updated the PR itself with more explanation.

Griatch (changes requested):
It's hard to review the functionality of this in detail, and there are no comments in the PR to motivate the changes. But since this refactor is done in response to #2683, and this is apparently not covered by existing unit tests (since these pass), I do think we need additional tests to cover what is being fixed here.

InspectorCaracal:
@Griatch Good point! I'll create a separate PR for the testing, since that's an important thing that I don't think should be dependent on my refactor, and then rebase this one onto it to compare the tests.

InspectorCaracal:
As a brief note, if a full refactor is too much, I can close this and make another one that revises the data in the existing module (and hope I don't miss anything else). The refactor isn't strictly required to fix the bug, it just made it a lot easier for me to cross-check that everything is consistent when doing so.

evennia/utils/verb_conjugation/pronouns.py
```diff
@@ -350,44 +381,42 @@ def pronoun_to_viewpoints(
...
+    viewpoint_map = PRONOUN_MAPPING[viewpoint]
+    
+    if pronoun_type in viewpoint_map:
+        pronouns = viewpoint_map[pronoun_type]
+    else:
+        pronouns = viewpoint_map[DEFAULT_PRONOUN_TYPE]
+    if gender in pronouns:
+        mapped_pronoun = pronouns[gender]
+    else:
+        mapped_pronoun = pronouns[DEFAULT_GENDER]
```
Griatch, line 418: It's clearer and more succinct to write these as

```python
pronouns = viewpoint_map.get(pronoun_type, DEFAULT_PRONOUN_TYPE)
mapped_pronoun = pronouns.get(gender, DEFAULT_GENDER)
```

Griatch (changes requested):
Only a minor code suggestion, otherwise I think this does make of a nice refactor. It's just a bit fiddly to get through and review this type of code in general :) 

The separate unit test PR can be applied after this is in place. Don't want to merge failing tests.

evennia/utils/verb_conjugation/pronouns.py
InspectorCaracal: Oh, good catch!

InspectorCaracal: Would this be better than the original?
```py
    viewpoint_map = PRONOUN_MAPPING[viewpoint]

    pronouns = viewpoint_map.get(pronoun_type)
    if not pronouns:
        pronouns = viewpoint_map[DEFAULT_PRONOUN_TYPE]
    
    mapped_pronoun = pronouns.get(gender)
    if not mapped_pronoun:
        mapped_pronoun = pronouns[DEFAULT_GENDER]
```

InspectorCaracal: Ah, never mind that, I figured out how to trim it down further

Griatch: Oh, you are right, I was too hasty. :) Yeah, it'd need to be `pronouns.get(gender, pronouns[DEFAULT_GENDER])`

evennia/utils/verb_conjugation/pronouns.py
```diff
@@ -407,15 +407,8 @@ def pronoun_to_viewpoints(
...
-    if pronoun_type in viewpoint_map:
-        pronouns = viewpoint_map[pronoun_type]
-    else:
-        pronouns = viewpoint_map[DEFAULT_PRONOUN_TYPE]
-    if gender in pronouns:
-        mapped_pronoun = pronouns[gender]
-    else:
-        mapped_pronoun = pronouns[DEFAULT_GENDER]
+    pronouns = viewpoint_map.get(pronoun_type) or viewpoint_map.get(DEFAULT_PRONOUN_TYPE)
+    mapped_pronoun = pronouns.get(gender) or pronouns.get(DEFAULT_GENDER)
```
Griatch, line 411: Sorry for giving the wrong suggested fix before, but the way to write this would be 
```python
pronouns = viewpoint_map.get(pronoun_type, viewpoint_map[DEFAULT_PRONOUN_TYPE])
mapped_pronoun = pronouns.get(gender, pronouns[DEFAULT_GENDER])
```

Griatch (changes requested):
Since we started to correct that line, let's wrap it up before merging :)

evennia/utils/verb_conjugation/pronouns.py
InspectorCaracal: Oh! Okay, I'll change that.

For my own information, what makes the syntax `mapped_pronoun = pronouns.get(gender, pronouns[DEFAULT_GENDER])` better than `mapped_pronoun = pronouns.get(gender) or pronouns[DEFAULT_GENDER]` ?

Griatch: Both work. Generally it's considered more 'Pythonic' to make use of the in-built Python primitives to their fullest. So since `dict.get(value, default)` is a thing, one should make sure to use that functionality explicitly rather than mimic the functionality by separately checking if `get(value)` returns `None` and then use the default value as a 'home brew' solution.

Griatch (approved):

Issue #2695, opened 2022-03-23
[BUG - Develop] xyzgrid spawn doesn't call at_object_creation


InspectorCaracal:
#### Describe the bug
When creating rooms in the game using the xyzgrid module and a custom typeclass, the `at_object_creation` hook is never called.

#### To Reproduce
1. Create custom room and exit classes that inherit from the XYZGrid room and exits
2. Add a simple `at_object_creation` hook to each class with something easily identifiable, e.g. setting a new attribute to `True`
3. Create a basic xyzgrid map module using those room and exit classes
4. Add the map module and spawn using the xyzgrid launcher commands.
5. Go into the game and examine any spawned room or exit. Everything looks fine _except that_ the new attribute is not there.

#### Develop-branch commit
473f2e379

#### Additional context
I tested with these bare-bones classes, but on the more complicated classes where I discovered it, all of the other hooks I overloaded worked correctly.
```python
# Room
class CustomXYZRoom(XYZRoom):
    def at_object_creation(self):
        self.db.checkmark = True

# Exit
class CustomXYZExit(XYZExit):
    def at_object_creation(self):
        self.db.checkmark = True
```
My rooms and exits using those classes look like this, suggesting the hook never ran:
```
Name/key: A major street (#13)
Typeclass: CustomXYZRoom (typeclasses.rooms.CustomXYZRoom)
Home: Limbo (#2)
Locks:   call:true(); control:id(13) or perm(Admin);  delete:id(13) or perm(Admin);
 drop:holds();  edit:id(13) or perm(Admin); examine:perm(Builder);
 get:false(); puppet:false(); tell:perm(Admin); view:all()
Tags:   1[room_y_coordinate], 5[room_x_coordinate], main_streets[zone],
 prototype-4ee42f8[from_prototype], the city[room_z_coordinate]
Persistent Attributes:
  desc=A wide cobbled street.
Exits: west(#84), east(#85), north(#86), southeast(#87)
Characters: admin(#1)

Name/key: east (#85)
Aliases: e
Typeclass: CustomXYZExit (typeclasses.exits.CustomXYZExit)
Location: A major street (#13)
Home: Limbo (#2)
Destination: A street in Spiritgate (#14)
Locks:   call:true(); control:id(85) or perm(Admin);  delete:id(85) or perm(Admin);
 drop:holds();  edit:id(85) or perm(Admin); examine:perm(Builder);
 get:false(); puppet:false(); tell:perm(Admin); traverse:all(); view:all()
[...]

PR #2904, opened 2022-10-06
fix: xyzgrid spawn doesn't call at_object_creation

Henddher:
- Fix installation instructions
- Use specified typeclass in prototype (if any) to instantiate xyz rooms and exits.

#2695

### DIFF
--- a/evennia/contrib/grid/xyzgrid/README.md
+++ b/evennia/contrib/grid/xyzgrid/README.md
@@ -53,28 +53,31 @@ Exits: northeast and east
         (then go back to your mygame/ folder)
 
    This will install all optional requirements of Evennia.
-2. Import and add the `evennia.contrib.commands.XYZGridCmdSet` to the
+2. Import and [add] the `evennia.contrib.grid.xyzgrid.commands.XYZGridCmdSet` to the
    `CharacterCmdset` cmdset in `mygame/commands.default_cmds.py`. Reload
    the server. This makes the `map`, `goto/path` and the modified `teleport` and
    `open` commands available in-game.
+
+[add]: docs/source/Command-Sets.md#defining-command-sets
+
 3. Edit `mygame/server/conf/settings.py` and add
 
-       EXTRA_LAUNCHER_COMMANDS['xyzgrid'] = 'evennia.contrib.launchcmd.xyzcommand'
-
-   and
-
-       PROTOTYPE_MODULES += [’evennia.contrib.grid.xyzgrid.prototypes’]
+       EXTRA_LAUNCHER_COMMANDS['xyzgrid'] = 'evennia.contrib.grid.xyzgrid.launchcmd.xyzcommand'
+       PROTOTYPE_MODULES += ['evennia.contrib.grid.xyzgrid.prototypes']
 
    This will add the new ability to enter `evennia xyzgrid <option>` on the
    command line.  It will also make the `xyz_room` and `xyz_exit` prototypes
    available for use as prototype-parents when spawning the grid.
+
 4. Run `evennia xyzgrid help` for available options.
+
 5. (Optional): By default, the xyzgrid will only spawn module-based
-   [prototypes](../Components/Prototypes.md). This is an optimization and usually makes sense
+   [prototypes]. This is an optimization and usually makes sense
    since the grid is entirely defined outside the game anyway. If you want to
    also make use of in-game (db-) created prototypes, add
    `XYZGRID_USE_DB_PROTOTYPES = True` to settings.
 
+[prototypes]: ../Components/Prototypes.md
 
 ## Overview
 
--- a/evennia/contrib/grid/xyzgrid/tests.py
+++ b/evennia/contrib/grid/xyzgrid/tests.py
@@ -7,7 +7,7 @@ from unittest import mock
 
 from random import randint
 from parameterized import parameterized
-from django.test import TestCase, override_settings
+from django.test import TestCase
 from evennia.utils.test_resources import BaseEvenniaTest
 from . import xymap, xyzgrid, xymap_legend, xyzroom
 
@@ -1477,13 +1477,6 @@ class TestCallbacks(BaseEvenniaTest):
         super().tearDown()
         self.grid.delete()
 
-    # @override_settings(
-    #   XYZEXIT_PROTOTYPE_OVERRIDE={
-    #     "typeclass": "evennia.contrib.grid.xyzgrid.tests.TestXyzExit",
-    #   },
-    #   XYZROOM_PROTOTYPE_OVERRIDE={
-    #     "typeclass": "evennia.contrib.grid.xyzgrid.tests.TestXyzRoom",
-    #   })
     def test_typeclassed_xyzroom_and_xyzexit_with_at_object_creation_are_called(self):
         map_data = dict(MAP_DATA)
         for prototype_key, prototype_value in map_data["prototypes"].items():
--- a/evennia/contrib/grid/xyzgrid/tests.py
+++ b/evennia/contrib/grid/xyzgrid/tests.py
@@ -1522,4 +1522,4 @@ class TestCallbacks(BaseEvenniaTest):
 
         # Two rooms and 2 exits.
         self.assertEqual(mock_room_callbacks.at_object_creation.mock_calls, [mock.call(), mock.call()])
-        # self.assertEqual(mock_exit_callbacks.at_object_creation.mock_calls, [mock.call(), mock.call()])
+        self.assertEqual(mock_exit_callbacks.at_object_creation.mock_calls, [mock.call(), mock.call()])
--- a/evennia/contrib/grid/xyzgrid/tests.py
+++ b/evennia/contrib/grid/xyzgrid/tests.py
@@ -1457,37 +1457,13 @@ MAP_DATA = {
   }
 }
 
-# @override_settings(
-#   XYZEXIT_PROTOTYPE_OVERRIDE= {
-#     "typeclass": "evennia.contrib.grid.xyzgrid.tests.TestXyzExit",
-#   },
-#   XYZROOM_PROTOTYPE_OVERRIDE= {
-#     "typeclass": "evennia.contrib.grid.xyzgrid.tests.TestXyzRoom",
-#   }
-# )
 class TestCallbacks(BaseEvenniaTest):
-    # @override_settings(
-    #   XYZEXIT_PROTOTYPE_OVERRIDE={
-    #     "typeclass": "evennia.contrib.grid.xyzgrid.tests.TestXyzExit",
-    #   },
-    #   XYZROOM_PROTOTYPE_OVERRIDE={
-    #     "typeclass": "evennia.contrib.grid.xyzgrid.tests.TestXyzRoom",
-    #   })
     def setUp(self):
         super().setUp()
         mock_room_callbacks.reset_mock()
         mock_exit_callbacks.reset_mock()
         
     def setup_grid(self, map_data):
-        # from evennia.prototypes import prototypes as protlib
-        # for prototype_key in ('xyz_room', 'xyz_exit', 'xyzroom', 'xyzexit'):
-        #   proto = protlib.search_prototype(
-        #       prototype_key,
-        #       # require_single=True,
-        #       # no_db=True
-        #   )
-        #   print(prototype_key, 'found?', proto)
-
         self.grid, err = xyzgrid.XYZGrid.create("testgrid")
 
         def _log(msg):
@@ -1515,11 +1491,10 @@ class TestCallbacks(BaseEvenniaTest):
                 prototype_value["typeclass"] = "evennia.contrib.grid.xyzgrid.tests.TestXyzRoom"
             if len(prototype_key) == 3:
                 prototype_value["typeclass"] = "evennia.contrib.grid.xyzgrid.tests.TestXyzExit"
-        print(map_data)
         self.setup_grid(map_data)
 
         self.grid.spawn()
 
-        # Two rooms and 2 exits.
+        # Two rooms and 2 exits, Each one should have gotten one `at_object_creation` callback.
         self.assertEqual(mock_room_callbacks.at_object_creation.mock_calls, [mock.call(), mock.call()])
         self.assertEqual(mock_exit_callbacks.at_object_creation.mock_calls, [mock.call(), mock.call()])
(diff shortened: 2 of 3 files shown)


Henddher:
Hi @Griatch - I need help

I'm trying to root cause #2695. My first step is to write a simple test that creates a map using my `TestXyzRoom` and `TestXyzExit` (subclasses of `XYZRoom` and `XYZExit` respectively). Each one of these classes implements the `at_object_creation` callback.

However, I am running into a wall:

Although my test uses `@override_settings` for patching `XYZEXIT_PROTOTYPE_OVERRIDE` and `XYZROOM_PROTOTYPE_OVERRIDE`, they don't seem to *stick*. See 👈👈 below. I query for the prototypes by names but nothing comes back.

... and because of that, `MapNode.spawn(...)` ends up using the default `typeclass` in `xyzgrid/prototypes.py` (`room_prototype` and `exit_prototype`). So, my test xyzroom/xyzexit subclasses *never* get used. See 👈 missing `typeclass`.

I noticed something interesting but I think it's "by design"? There are `NodeTypeclass` and `ExitTypeclass` globals in `xymap_legent.py`. These get initialized on their first use by directly importing `XYZRoom` and `XYZExit` from `xroom.py`. I presume that's ok because after a MapNode and MapLink are built, their prototypes get overriden by `spawner.batch_update_objects_with_prototype(...)`. But again, I think that's "ok" (but I am not sure).

This is the output of test. 

```text
evennia_mygame$ evennia test evennia.contrib.grid.xyzgrid.tests.TestCallbacks

    TESTING: Using Evennia's default settings file (evennia.settings_default).
    (use 'evennia test --settings settings.py .' to run only your custom game tests)
    
Found 1 test(s).
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
xyz_room found? [] 👈👈
xyz_exit found? [] 👈👈
xyzroom found? [] 👈👈
xyzexit found? [] 👈👈
(Re)loading grid ...
Loading map 'map1'...
Loaded and linked 1 map(s).
spawning/updating nodes for Z='map1' ...
  spawning room at xyz=(0, 0, 'map1')👈
  spawning room at xyz=(1, 0, 'map1')👈
spawning/updating links for Z='map1' ...
  spawning/updating exit xyz=(0, 0, 'map1'), direction=east👈
  spawning/updating exit xyz=(1, 0, 'map1'), direction=west👈
(Re)loading grid ...
Loaded and linked 0 map(s).
.
Ran 1 test in 1.926s

OK
Destroying test database for alias 'default'...

``` 

Why cannot I `@override_settings` with my typeclasses?
Do I have to force loading `xyzgrid/prototypes.py` in some special way?

---
side note:
I also asked in an older discussion if it's possible to specify the `typeclass` in the map definition `dict` somehow - it seems like it's not possible? I saw one can use `legend` (which contains `prototype`) but I have not tried that route yet.

Henddher:
FYI - the last commit has my WIP test in case you want to try it out on your end

InspectorCaracal:
@Henddher As I mentioned in the discussion, you can add a "typeclass" key to your prototypes.

```py
             ("*", "*"): {
               "key": "room",
               "desc": "A room.",
               "prototype_parent": "xyz_room",
               "typeclass": "evennia.contrib.grid.xyzgrid.tests.TestXyzRoom",
             },
```

Henddher:
Thank you @InspectorCaracal!

I'll give that a try while I wait for a hint about my "broken `@override_settings` scenario.

Henddher:
After following @InspectorCaracal suggestion (thank you, @InspectorCaracal!), I managed to specify custom `typeclass` before `spawn`ing the map.
I verified `spawner.batch_update_objects_with_prototype(...)` overrode the default typeclass with mine (as expected).

ATM, I am suspecting that both globals, `NodeTypeclass` and `ExitTypeclass`, are closely tied to the issue at hand. 

If they are not set (which I verified on their first pass), they get assigned with the default `XYZRoom` and `XYZExit` from `xyzgrid/xyzroom.py` (line 307). Then, both these are used to create each room and exit (line 325). Hence, any callback overwritten by my `TestXyzRoom` and `TestXyzExit` would never be called. 

This is the interesting code (for the room case):

https://github.com/evennia/evennia/blob/b0f24f9975e14fddeabef06a19b5b61575957b1c/evennia/contrib/grid/xyzgrid/xymap_legend.py#L305-L328

In summary, (1) I think the two globals are to blame, and (2) it would be convenient to use `@override_settings` in this case although that's no longer blocking me.

If anyone has any suggestions, please let me know.

InspectorCaracal:
@Henddher Actually, based on the code you pointed at, I think the problem is it's not using the prototype to spawn the object at all.

Specifically looking at lines 320 to 325, the code *claims* to spawn a node of typeclass `tclass` but it only ever actually creates a node of NodeTypeclass and then applies attribute changes later, which would of course result in the referenced bug: the desired typeclass' `at_object_creation` is never run.

I suspect it was designed that way to ensure all nodes inherit from the base typeclass, but I think it would be a better solution to check that with `inherits_from` after using `spawner.spawn` to create the node initially.

Henddher:
Hi @InspectorCaracal, I think we agree on the culprit. However, I need to carefully understand how `Class.create(...)` works and how it deals with `prototypes`.

I tried the following but `DefaultRoom.create(...)` raises and error. I did not step through it because it's obviously failing due to `kwargs` being a poorly form `prototype` `dict`.

```diff
diff --git a/evennia/contrib/grid/xyzgrid/xymap_legend.py b/evennia/contrib/grid/xyzgrid/xymap_legend.py
index 4dfd05a72..1b64e6224 100644
--- a/evennia/contrib/grid/xyzgrid/xymap_legend.py
+++ b/evennia/contrib/grid/xyzgrid/xymap_legend.py
@@ -22,8 +22,10 @@ from collections import defaultdict
 
 from django.core import exceptions as django_exceptions
 from evennia.prototypes import spawner
+from evennia.utils.utils import class_from_module
 
 from .utils import MAPSCAN, REVERSE_DIRECTIONS, MapParserError, BIGVAL
+from .xyzroom import XYZRoom
 
 NodeTypeclass = None
 ExitTypeclass = None
@@ -318,12 +320,17 @@ class MapNode:
             nodeobj = NodeTypeclass.objects.get_xyz(xyz=xyz)
         except django_exceptions.ObjectDoesNotExist:
             # create a new entity with proper coordinates etc
-            tclass = self.prototype["typeclass"]
+            tclass = self.prototype.pop("typeclass", "evennia.contrib.grid.xyzgrid.xyzroom.XYZRoom")
+            typeclass = class_from_module(tclass)
+            assert issubclass(typeclass, XYZRoom), f"'{tclass}' must be subclass of `{XYZRoom}'"
+            self.prototype["typeclass"] = typeclass
             tclass = (
                 f" ({tclass})" if tclass != "evennia.contrib.grid.xyzgrid.xyzroom.XYZRoom" else ""
             )
-            self.log(f"  spawning room at xyz={xyz}{tclass}")
-            nodeobj, err = NodeTypeclass.create(self.prototype.get("key", "An empty room"), xyz=xyz)
+            self.log(f"  spawning room at xyz={xyz} ({tclass})")
+            if "key" not in self.prototype:
+                self.prototype["key"] = "An empty room"
+            nodeobj, err = typeclass.create(xyz=xyz, **self.prototype)
             if err:
                 raise RuntimeError(err)
         else:
diff --git a/evennia/contrib/grid/xyzgrid/xyzroom.py b/evennia/contrib/grid/xyzgrid/xyzroom.py
index 581ca2688..b824b32ed 100644
--- a/evennia/contrib/grid/xyzgrid/xyzroom.py
+++ b/evennia/contrib/grid/xyzgrid/xyzroom.py
@@ -354,8 +354,10 @@ class XYZRoom(DefaultRoom):
             (str(y), MAP_Y_TAG_CATEGORY),
             (str(z), MAP_Z_TAG_CATEGORY),
         )
+        if "tags" in kwargs:
+            kwargs["tags"] += tags
 
-        return DefaultRoom.create(key, account=account, tags=tags, typeclass=cls, **kwargs)
+        return DefaultRoom.create(key, account=account, **kwargs)
 
     def get_display_name(self, looker, **kwargs):
         """

```

This is the output

```
Creating test database for alias 'default'...

    TESTING: Using Evennia's default settings file (evennia.settings_default).
    (use 'evennia test --settings settings.py .' to run only your custom game tests)
    
Found 1 test(s).
System check identified no issues (0 silenced).
E
ERROR: test_dummy (evennia.contrib.grid.xyzgrid.tests.TestCallbacks)
  File "/home/henddher/Documents/github/evennia/evennia/contrib/grid/xyzgrid/xymap_legend.py", line 320, in spawn
    nodeobj = NodeTypeclass.objects.get_xyz(xyz=xyz)
  File "/home/henddher/Documents/github/evennia/evennia/contrib/grid/xyzgrid/xyzroom.py", line 108, in get_xyz
    raise self.model.DoesNotExist(inp)
evennia.contrib.grid.xyzgrid.xyzroom.XYZRoom.DoesNotExist: Query: xyz=(0,0,map1), 

During handling of the above exception, another exception occurred:

  File "/home/henddher/Documents/github/evennia/evennia/contrib/grid/xyzgrid/tests.py", line 1515, in test_dummy
    self.grid.spawn()
  File "/home/henddher/Documents/github/evennia/evennia/contrib/grid/xyzgrid/xyzgrid.py", line 271, in spawn
    xymap.spawn_nodes(xy=(x, y))
  File "/home/henddher/Documents/github/evennia/evennia/contrib/grid/xyzgrid/xymap.py", line 690, in spawn_nodes
    node.spawn()
  File "/home/henddher/Documents/github/evennia/evennia/contrib/grid/xyzgrid/xymap_legend.py", line 335, in spawn
    raise RuntimeError(err)
RuntimeError: ["An error occurred while creating this 'room' object."]

Ran 1 test in 2.173s

FAILED (errors=1)
Destroying test database for alias 'default'...
{'map': '\n\n    + 0 1\n\n    0 #-#\n\n    + 0 1\n\n  ', 'zcoord': 'map1', 'prototypes': {('*', '*'): {'key': 'room', 'desc': 'A room.', 'prototype_parent': 'xyz_room', 'typeclass': 'evennia.contrib.grid.xyzgrid.tests.TestXyzRoom'}, ('*', '*', '*'): {'desc': 'A passage.', 'prototype_parent': 'xyz_exit', 'typeclass': 'evennia.contrib.grid.xyzgrid.tests.TestXyzExit'}}, 'options': {'map_visual_range': 1, 'map_mode': 'scan'}}
(Re)loading grid ...
Loading map 'map1'...
Loaded and linked 1 map(s).
spawning/updating nodes for Z='map1' ...
  spawning room at xyz=(0, 0, 'map1') ( (evennia.contrib.grid.xyzgrid.tests.TestXyzRoom))
(Re)loading grid ...
Loaded and linked 0 map(s).

```

Hopefully @Griatch will shed some light.

InspectorCaracal:
It's worth noting that when the contrib runs `NodeTypeclass.create`, it does not pass it the prototype dict. Instead, it takes the key and description from the prototype. This is because the typeclassed `create` function has nothing to do with the prototype system.

When creating something from a prototype dict, you need to use the `spawn` method in `evennia.prototypes.spawner`

Instead, it's creating an object of `NodeTypeclass` and then changing the attributes and typeclass after creation.

Griatch:
The problem with using `prototype.spawn` directly here is that `XYZRoom.create` is overridden to understand XYZ coordinates, so we need to first create the room with this, and then apply the prototype _afterwards_ using `batch_update_objects_with_prototype`. I missed that the latter does not fire `at_object_creation`, only the initial `.create` does. 

The issue is that `NodeTypeclass` is used for `.create`, not the intended typeclass. So, you are on the right track here, @Henddher, you should just keep what was there, but use your found `typeclass` instead. No need to worry about the rest of the prototype, it's applied later. For clarity I think we must make sure to verify this typeclass actually set the XYZ coordinates correctly

```python
from .xyzroom import MAP_X_TAG_CATEGORY

class MapNode:
    # ...
    def spawn(self):
         # ... 
         typeclass = class_from_module(tclass) 
         nodeobj, err = typeclass,create(self.prototype.get("key", "An empty room"), xyz=xyz)
         if err:
             raise RuntimeError(err)
 
        if not nodeobj.tags.get('x', MAP_X_TAG_CATEGORY):
             # no coordinate tag added to the room 
             raise RuntimeError(
                 f"MapNode at xyz={xyz}, with typeclass '{tclass}' could not spawn - "
                 "XYZ tags were not set correctly. Check that the typeclass inherits from XYZRoom."
            )
            nodeobj.delete()

        # ...
```

Henddher:
I tried what you suggested @Griatch but `DefaultRoom.create()` failed and produced an error instead (`""`) 
I tried a couple of different combinations (specifying `typeclass`, `key`, etc.) but none succeeded.

I think what @InspectorCaracal suggested is actually more reasonable: use `spawn()` instead. 

I will push PRs (with different experiments) so we can carefully compare.

Henddher:
Applied the same principle (Typeclass with fallback) for exits, and added a test.

evennia/contrib/grid/xyzgrid/tests.py
```diff
@@ -1415,3 +1416,85 @@ def test_build(self):
...
+        self.grid.log = _log
+
+        self.map_data = map_data
+        self.grid.add_maps(map_data)
+
+    def tearDown(self):
+        super().tearDown()
+        self.grid.delete()
+
+    # @override_settings(
```
Henddher, line 1480: I could not get this work in this case. I can remove it but it would be good to know why.

evennia/contrib/grid/xyzgrid/README.md
```diff
@@ -53,28 +53,31 @@ Exits: northeast and east
...
+       PROTOTYPE_MODULES += ['evennia.contrib.grid.xyzgrid.prototypes']
 
    This will add the new ability to enter `evennia xyzgrid <option>` on the
    command line.  It will also make the `xyz_room` and `xyz_exit` prototypes
    available for use as prototype-parents when spawning the grid.
+
 4. Run `evennia xyzgrid help` for available options.
+
 5. (Optional): By default, the xyzgrid will only spawn module-based
-   [prototypes](../Components/Prototypes.md). This is an optimization and usually makes sense
```
Henddher, line 73: Personally, I find this more readable. Although this syntax is valid markdown, I don't know if doc-generation will handle it correctly.

Griatch (changes requested):
Thanks for the PR! I have some comments. :)

Henddher:
Hi @Griatch, for whenever you get a chance?

Griatch (approved):
Should do it. Thanks for the PR!

Issue #2216, opened 2020-10-08
[Feature Request] Select a text field tab but focus isn't on the textbox


dorianturba:
#### My problem
When I select a textfield tab, I have to click a second time in the textbox.

Two tabs:
![1](https://user-images.githubusercontent.com/9381120/95471552-5916d880-0982-11eb-830c-120b8107efa3.png)
Select tab 2:
![2](https://user-images.githubusercontent.com/9381120/95471572-5f0cb980-0982-11eb-9a8c-1ab3f1730ed6.png)
Another click for the textbox:
![3](https://user-images.githubusercontent.com/9381120/95471601-6764f480-0982-11eb-970d-95a744790f10.png)

#### What I want
When I click on a Tab, I want to focus the textbot. Like this, I just have to click one time and start writing.

egrinberg:
Hey, I would like to work on this. Please assign me.

Griatch:
@egrinberg We don't do explicit assignment. Just take it away!

PR #2905, opened 2022-10-06
fix: multiple inputs now retain focus correctly

### LINKED ISSUE #2608: Web Client: Input frame focus is not retained if multiple input frames are open
If you have multiple input frames open on the default web client, inputting a command will unfocus the input frame, requiring you to click into it every time.

1. Start up an Evennia server and log in using the localhost web client
2. Open an extra input frame
3. Input any command
4. Observe input frame focus being lost

The currently-used input frame should stay focused until the user clicks out of it, regardless of command use.

####Screenshot####
Input frames should look like this to repro the bug.
image

selberhad:
- Fixes focus issues when using multiple input tabs
- Cleans up some duplicated event handlers via `once` helper

Fixes #2608, #2216

~~For discussion: in recent times, the trend has been for front-end developers to move away from using Lodash. So why add it as a dependency here? Evennia's webclient is using fairly old-school technologies, like jQuery and Bootstrap. Even [this blog post](https://thejs.dev/jmitchell/its-time-to-let-go-of-lodash-nqc) encouraging developers to let go of lodash acknowledges that "In the world of Bootstrap, jQuery, underscore, prototype and a huge number of other tools, lodash provided a consistent and clear interface. Classes, functional programming, and all the goodness and richness that we take for granted was years away." Therefore, I propose Lodash be added as a dependency until Evennia's webclient is re-written using the latest-and-greatest frontend Javascript technologies. Thoughts?~~
- Removed lodash dependency and pulled `once` helper in

### DIFF
--- a/evennia/web/static/webclient/js/plugins/default_in.js
+++ b/evennia/web/static/webclient/js/plugins/default_in.js
@@ -47,7 +47,7 @@ let defaultInPlugin = (function () {
 
                     // enter key by itself should toggle focus
                     if( inputfield.length < 1 ) {
-                        inputfield = $(".inputfield:last");
+                        inputfield = $(".inputfield.focused");
                         inputfield.focus();
                         if( inputfield.length < 1 ) { // non-goldenlayout backwards compatibility
                             $("#inputfield").focus();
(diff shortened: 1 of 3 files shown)


Griatch (commented):
Thanks for the changes. When originally creating the webclient, I went with a baseline basically because the frontend world changes so quickly. :sweat_smile: No matter what we pick, it will be old news. 
Overall, keeping the code as 'straight forward' as possible has its benefits for teaching. But that said, I don't have a strong opinion on using Lodash or not. @friarzen has been working on the webclient more lately, may want to chip in if there's an opinion on it.

selberhad:
Yeah, it is quite dizzying keeping up with the latest and greatest front-end technologies, and seems to grow ever more complex. 

While I think Lodash is a nice "missing standard library" for frontend Javascript (and something every frontend developer should know about, given its popularity), I can also just pull the `once` function that I depend on in this change into the file where it is used, and drop the full CDN dependency.

Griatch:
@selberhad I think that sounds like a better option for now, in line with the webclient approach we've used so far. 
We could potentially discuss modernizing the webclient overall (wether by introducing more helper libs or a general overhaul) as a separate feature/request project instead (not something for next Evennia version, but worth to discuss for the future).

selberhad:
> @selberhad I think that sounds like a better option for now, in line with the webclient approach we've used so far. We could potentially discuss modernizing the webclient overall (whether by introducing more helper libs or a general overhaul) as a separate feature-request/project instead.

Sounds good, I removed lodash and put the helper function inline.

Griatch (approved):
Thanks for the changes!

PR #2907, opened 2022-10-06
Git Contrib: Integrating Git status, branch, checkout, and pull functionality into Evennia.

aMiss-aWry:
This module integrates a stripped-down version of git within the game, allowing developers to view their git status, change branches, and pull updated code of both their local game and evennia core. 

It's a handy utility and helps keep the development process moving smoothly. Also, I want a t-shirt. 

The more complex operations still need to take place in the command line but the contrib could be fleshed out at a later date to include commit/push functionality - merge conflicts will still have to be amended externally, of course.


evennia/contrib/utils/git/git_integration.py
```diff
@@ -0,0 +1,171 @@
...
+        git evennia status
+        git evennia branch
+        git evennia checkout <branch>
+        git evennia pull
+        
+    If there are any conflicts encountered, the command will abort. The command will reload your game after pulling new code automatically, but for changes involving persistent scripts etc, you may need to manually restart.
+    """
+
+    key = "@git"
+    aliases = ["@git evennia"]
```
Griatch, line 30: No need to hardcode the `@` prefix. You'll be able to use `@git` anyway if you prefer.

evennia/contrib/utils/git/git_integration.py
```diff
@@ -0,0 +1,171 @@
...
+            self.action = "status"
+            self.args = ""
+
+        err_msgs = ["|rInvalid Git Repository|n:",
+            "The {repo_type} repository is not recognized as a git directory.",
+            "In order to initialize it as a git directory, you will need to access your terminal and run the following commands from within your directory:",
+            "    git init",
+            "    git remote add origin {remote_link}"]
+
+        if self.cmdstring == "git evennia":
```
Griatch, line 55: This will never be True; `self.cmdstring` will only ever be `"git"` only. I think you may be looking for 

    if self.raw_string.startswith("git evennia"): ...

But I'd say it's better to properly parse `self.args` instead and do 

    if self.args.startswith("evennia ")

evennia/contrib/utils/git/tests.py
```diff
@@ -0,0 +1,75 @@
...
+from evennia.utils.utils import list_to_string
+
+import git
+import mock
+import datetime
+
+class TestCmdGit(CmdGit):
+    pass
+
+class TestGit(EvenniaTest):
```
Griatch, line 19: The tests seem to distinctly lack the testing of `git evennia` ;)

evennia/contrib/utils/git/README.md
```diff
@@ -0,0 +1,68 @@
+# In-game Git Integration
```
Griatch, line 1: Please rename the folder from `utils/git` to `utils/git_integration/`. Using `git` as a folder name makes it looks like it's, well, something to do with the real git and not a contrib.

Griatch (changes requested):
A very compelling idea! This was a fun contrib :grin: Just needs a little more polish. :wink: 

I have given some inline feedback, but I have some general points to address too: 
- The use of `git status evennia` does not seem to work. This is because you check for `git evennia` in order to figure out to use the `evennia` repo, no room for `status` in-between. 
- There is no reason to *require* your mygame to be a git repo. You will just not be able to use that functionality on mygame if it's not - you should still be able to work with upstream Evennia just fine. I'd consider reworking to docs and command to make sure to gracefully handle both situations - being able to fetch the latest Evennia is useful on its own! 
- Please add unit tests that cover the different use cases. This would have caught the issues with the evennia repo.

aMiss-aWry:
Hi,

That was actually an error in the README.md. I'm fixing it now. It's 'git evennia' + command (so git evennia status) not 'git status evennia'. And it works as 'git evennia' is an alias for the command, thus 'git evennia' is the cmdstring being parsed for those commands. I think it makes more sense as well as being more easily parsed since (to me) 'git branch evennia <branch>' feels more confusing stylewise than 'git evennia branch <branch>'

Everything else I'm fixing, let me know if you feel like the 'git evennia <command> should be changed to 'git <command> evennia' though!

Griatch:
@aMiss-aWry Just FYI, the `self.cmdstring` is not `git evennia`, it's `git`. And even fixing this, I couldn't get `git evennia status` to work, I just get a return of `You can only git status, git branch, git checkout, or git pull.` for that case.

aMiss-aWry:
@Griatch No I mean 
    aliases = ["@git evennia"]
should make it that 'git evennia' is the cmdstring, no? (even removing the @) At least, that seems to be how it's working for me.
image

Griatch:
You are hard-coding @, which should means that the command will require it. `git` would then _not_ be the same as `@git` - this is deliberately so that you can mix the two commands if you really wanted to. 

You are right about aliases though; it does seem like it should end up in `cmdname`. 

However, using `@git evennia status` or `git evennia status` with your raw code (no changes to cmdname) does not work for me - I consistently get the `Invalid Git Repository` error - which indicates it's pointing to the current mygame dir (with no git init) and not to the evennia dir.

aMiss-aWry:
> You are hard-coding @, which should means that the command will require it. `git` would then _not_ be the same as `@git` - this is deliberately so that you can mix the two commands if you really wanted to.
> 
> You are right about aliases though; it does seem like it should end up in `cmdname`.
> 
> However, using `@git evennia status` or `git evennia status` with your raw code (no changes to cmdname) does not work for me - I consistently get the `Invalid Git Repository` error - which indicates it's pointing to the current mygame dir (with no git init) and not to the evennia dir.

Oh! Yes, that would do it, as previously the mygame dir was required to be a git repo. I fixed that in the recent set, it will now only complain if you are using the associated command (so 'git' requires mygame to be a git repo, 'git evennia' requires evennia to be).

Griatch (changes requested):
Thanks for the changes, this seems like a better solution :+1: 

The unit tests fail on CI though, see https://github.com/evennia/evennia/actions/runs/3203415344/jobs/5233482908. The git situation in a CI build can be different to what you see locally; if it works locally it may be that you need to make do with mocking the git library's returns and test that your own code works if it returns the expected things (we are not testing `gitpython` after all).

PR #2914, opened 2022-10-08
Fix `TagHandler.has` to return False for no-match

### LINKED ISSUE #2913: [BUG - Develop] `obj.tags.has` no longer returns `False` when tag is not present
#### Describe the bug
When checking for the presence of a tag, the `has` method no longer returns `False` for tags that are not present on the object. It only returns `True` matches or nothing.

#### To Reproduce
1. Open the `py` console within a clean game.
2. `self.tags.add("sitting")` and `self.tags.add("hungry")`
3. `self.tags.has("sitting")` will return `True`
4. `self.tags.has("kneeling")` will return an empty list
5. `self.tags.has(["sitting", "kneeling", "hungry"])` will return `[True, True]`

#### Expected behavior
- `self.tags.has("kneeling")` should return `False`
- `self.tags.has(["sitting", "kneeling", "hungry"])` should return `[True, False, True]`

#### Develop-branch commit
b0f24f997

#### Additional context
This worked correctly several months ago, but I'm not sure how recently the bug was introduced.

InspectorCaracal:
This fix assumes that a search for a specific tag key with category will only ever return one tag, which to my knowledge is consistent with the rest of the code and intended functionality. As such, it appends a boolean for each tag+category combination representing whether a match was found, rather than extending the list with any found tags.

If that assumption is incorrect, then this'll need a different solution.

Bug fixing

Closes #2913

### DIFF
--- a/evennia/typeclasses/tags.py
+++ b/evennia/typeclasses/tags.py
@@ -419,7 +419,7 @@ class TagHandler(object):
         if key:
             for tag_str in make_iter(key):
                 tag_str = tag_str.strip().lower()
-                ret.extend(bool(tag) for tag in self._getcache(tag_str, category))
+                ret.append(bool(self._getcache(tag_str, category)))
         elif category:
             ret.extend(bool(tag) for tag in self._getcache(category=category))
         else:


Griatch (approved):
Hm, I see what you mean; what this doesn't catch is that `_getache` returns a list, so this is like running `all()` on that return. That said, I don't think you'd ever end up in a situation where that matters, so this seems to be a good fix. :+1:

PR #2918, opened 2022-10-08
Increase test coverage for funcparse.py, finding bugs in the process

audiodude:
Increased test coverage. Based on looking at coveralls coverage for a modified coveralls report based on a local build.

Increase test coverage, uncover bugs, learn more about Evennia.

Will fix bugs found in follow up PR.

### DIFF
--- a/evennia/utils/tests/test_funcparser.py
+++ b/evennia/utils/tests/test_funcparser.py
@@ -292,6 +292,12 @@ class TestFuncParser(TestCase):
         ret = self.parser.parse(string, strip=True)
         self.assertEqual("Test  and  things", ret)
 
+    @unittest.skip("broken due to https://github.com/evennia/evennia/issues/2927")
+    def test_parse_whitespace_preserved(self):
+        string = "The answer is $add(1, x)"
+        ret = self.parser.parse(string)
+        self.assertEqual("The answer is $add(1, x)", ret)
+
     def test_parse_escape(self):
         """
         Test the parser's escape functionality.
--- a/evennia/utils/tests/test_funcparser.py
+++ b/evennia/utils/tests/test_funcparser.py
@@ -262,12 +262,12 @@ class TestFuncParser(TestCase):
         _MAX_NESTING value, which means you can nest 4 values with a value of
         2, as demonstrated by this test.
         """
-        string = "$add(1, $add(1, $add(1, $add(1, $toint(42)))))"
+        string = "$add(1, $add(1, $add(1, $toint(42))))"
         ret = self.parser.parse(string)
 
         # TODO: Does this return value actually make sense?
         # It removed the spaces from the calls.
-        self.assertEqual("$add(1,$add(1,$add(1,$add(1,$toint(42)))))", ret)
+        self.assertEqual("$add(1,$add(1,$add(1,$toint(42))))", ret)
 
         with self.assertRaises(funcparser.ParsingError):
             self.parser.parse(string, raise_errors=True)
--- a/evennia/utils/tests/test_funcparser.py
+++ b/evennia/utils/tests/test_funcparser.py
@@ -79,7 +79,7 @@ def _lsum_callable(*args, **kwargs):
 
 
 def _raises_callable(*args, **kwargs):
-    raise funcparser.ParsingError("Test exception raised by test callable")
+    raise RuntimeError("Test exception raised by test callable")
 
 
 _test_callables = {
@@ -280,7 +280,7 @@ class TestFuncParser(TestCase):
         # It completed the first function call.
         self.assertEqual("test 2 $raise()", ret)
 
-        with self.assertRaises(funcparser.ParsingError):
+        with self.assertRaises(RuntimeError):
             self.parser.parse(string, raise_errors=True)
 
     def test_parse_strip(self):
--- a/evennia/utils/tests/test_funcparser.py
+++ b/evennia/utils/tests/test_funcparser.py
@@ -6,6 +6,7 @@ Test the funcparser module.
 
 import time
 from ast import literal_eval
+import unittest
 from unittest.mock import MagicMock, patch
 
 from django.test import TestCase, override_settings
@@ -77,6 +78,10 @@ def _lsum_callable(*args, **kwargs):
     return ""
 
 
+def _raises_callable(*args, **kwargs):
+    raise funcparser.ParsingError("Test exception raised by test callable")
+
+
 _test_callables = {
     "foo": _test_callable,
     "bar": _test_callable,
@@ -89,6 +94,7 @@ _test_callables = {
     "add": _add_callable,
     "lit": _lit_callable,
     "sum": _lsum_callable,
+    "raise": _raises_callable,
 }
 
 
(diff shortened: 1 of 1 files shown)


evennia/utils/tests/test_funcparser.py
```diff
@@ -216,12 +238,48 @@ def test_parse(self, string, expected):
...
-        string = "Test malformed This is $dummy(a, b) and $bar("
+        with self.assertRaises(funcparser.ParsingError):
+            self.parser.parse(unparseable, raise_errors=True)
+
+    @patch("evennia.utils.funcparser._MAX_NESTING", 2)
+    def test_parse_max_nesting(self):
+        """
+        Make sure it is an error if the max nesting value is reached.
+
+        TODO: Does this make sense? When it sees the first function, len(callstack)
```
Griatch, line 260: Hm, that doesn't seem right, no. It should fail at 3 already. 
To be picky, this test does not show that you can nest 4 though, only that you can't nest 5 :)

evennia/utils/tests/test_funcparser.py
```diff
@@ -216,12 +238,48 @@ def test_parse(self, string, expected):
...
+        is 0. It doesn't raise until the stack length is greater than the
+        _MAX_NESTING value, which means you can nest 4 values with a value of
+        2, as demonstrated by this test.
+        """
+        string = "$add(1, $add(1, $add(1, $add(1, $toint(42)))))"
+        ret = self.parser.parse(string)
+
+        # TODO: Does this return value actually make sense?
+        # It removed the spaces from the calls.
+        self.assertEqual("$add(1,$add(1,$add(1,$add(1,$toint(42)))))", ret)
```
Griatch, line 270: The entire `$(...)` will be replaced in-place by whatever is returned by that outermost `$()`. So whitespace inside arguments are skipped; that is why it compresses the failed output. I agree it would look better to keep the original whitespace when letting the parsing fail like this though.

evennia/utils/tests/test_funcparser.py
```diff
@@ -216,12 +238,48 @@ def test_parse(self, string, expected):
...
+        with self.assertRaises(funcparser.ParsingError):
+            self.parser.parse(string, raise_errors=True)
+
+    def test_parse_underlying_exception(self):
+        string = "test $add(1, 1) $raise()"
+        ret = self.parser.parse(string)
+
+        # TODO: Does this return value actually make sense?
+        # It completed the first function call.
+        self.assertEqual("test 2 $raise()", ret)
```
Griatch, line 281: My guess would be that `$raise` is not a recognized funcparser func, so it's not parsed. If so, that's what you are catching in your `assertRaises` below.

Griatch (commented):
Always good with more unit tests! I think two of your comments identify real bugs
- MAX_NESTING seems to allow too generous nesting
- Compressing the whitespace when 'not parsing' a failed string is not looking good

I suspect the failure to parse `$raise` is an issue in your test setup, but I've not investigated further - something to look into!

evennia/utils/tests/test_funcparser.py
audiodude: Actually, I added `_raises_callable` above on line 81, and registered it with the test parser on line 97, so this is WAI.

evennia/utils/tests/test_funcparser.py
audiodude: I think I noticed this whitespace collapsing when returning any function that had multiple arguments that failed parsing, with `raise_errors=False`, but I don't have test which demonstrates it.

evennia/utils/tests/test_funcparser.py
audiodude: Fixed it, now it nests 4 and that's the minimum required to get the test to pass.

audiodude:
Should we file bugs for `_MAX_NESTING` and whitespace collapsing?

Griatch:
@audiodude Yes, these do sound like bugs to report. If you have a chance to do it please do (since you also have examples to show), otherwise I'll do it a little later. :+1:

Griatch:
Since these tests (in some cases) found and shows the wrong behavior, I'll wait with merging them for now; maybe we can fix those issues and update this before merging. I've set the `hacktoberfest-accepted` label for now, so this still counts for Hacktoberfest credits. :)

audiodude:
> Since these tests (in some cases) found and shows the wrong behavior, I'll wait with merging them for now; maybe we can fix those issues and update this before merging. I've set the `hacktoberfest-accepted` label for now, so this still counts for Hacktoberfest credits. :)

Sounds good, I've got branches in my forked repo that fix most of the skipped tests. As commented on the bugs I filed, I thought it was better to commit a failing test and leave the code-under-test alone for now, so that it could be more easily demonstrated and fixed separately.

Note that the tests aren't actually failing, they're passing CI, they're just marked as skipped with a reference to the Github issue.

The only one that I don't have a fix for is $choice(), because I'm still confused about the parsing behavior of `$choice(red, green)` versus `$choice('red', 'green')` versus `$choice([red, green])` versus `$choice(['red', 'green'])`

evennia/utils/tests/test_funcparser.py
audiodude: Updated the `_raises_callable` to throw `RuntimeException`, to make it more clear that we're testing the exception thrown by the underlying function. Also will make the test more resilient.

audiodude:
> @audiodude Yes, these do sound like bugs to report. If you have a chance to do it please do (since you also have examples to show), otherwise I'll do it a little later. 👍

I don't have time to report right now, but I can do it later as well if you like.

audiodude:
Okay I've reported the additional issues, #2926 and #2927

> I'll wait with merging them for now; maybe we can fix those issues and update this before merging

Could we possibly discuss this further?

1. Having now actually tried the method where I "fix the issues first", it is much more work and less clean to fix the issue without the benefit of the failing test.
1. Assuming the effort has already gone in to write a failing test, it is a simple process to un-`@skip` it, write the fix and verify that it works.
1. This is also demonstrated by the fact that the failing test is reported as the "Steps to reproduce" in the given bugs. Checking in the failing test means that someone who might not be the one who found the issue can easily fix and verify.
1. The failing test also just serves to remind me what part of the code I'm working on and what needs to be fixed, without it being checked in I have to refer to a separate PR on Github to see this information.
1. In fact, if I'm not going to be able to check in `@skip`'d failing tests, it discourages me from writing them in the first place. Then it's possible for bugs to slip through the cracks. It is now extra work to go back and fix this PR to un-skip the tests, whereas if it had been submitted with them skipped, they would have been un-skipped as part of the process of fixing the bugs. 

I'd like to continue working on unit tests for Evennia, but I'd prefer the "submit skipped tests for bad behavior" approach that I've outlined here, if possible.

Thanks!

Griatch:
@audiodude I agree that for debugging, writing a test first to find issues can be practical. It's called test-driven development and is definitely a valid approach. I regularly do this when fixing issues myself.
- The problem is that in general, for an OSS development, most PRs are fire-and-forget: once it is merged, it's up to the the maintainers to manage the code. So merging code that is not yet fully functioning (such has having skipped (thus dead) tests) is normally avoided. As an OSS maintainer, I can't expect you personally to implement the fixes of those things you found (although you did, which is awesome!), so keeping the PR open until all its test actually pass correct behavior makes sense.
- That said, there is nothing that blocks you from making use of these tests for each issue. When fixing something, the approach would be to break out the relevant unit test from this PR and use it to test your fix-code. Those tests can then merge along with the fix, proving your solution and keeping fix + test together in one PR/commit. The remainder of the unit tests of this PR would then merge normally. It's not much difference for you compared to merging this PR first and activating the skipped test.
- (It's in general a perfectly fine thing to include a unit test code with a reported bug issue - this is in fact an excellent approach, because it demonstrates the issue, helps others reproduce it, and also gives code that can be merged to demonstrate the solution of the issue - when it has been fixed). 

That said, this is a special case, you made unit tests in bulk and found issues that way - and seem to be looking into them too, which is much appreciated. So I won't make you jump through the hoops of breaking out tests from this PR. So I'll merge it now, for convenience. :)

audiodude:
Thanks for your consideration of my points, I'll keep all that in mind for the future.

Issue #2915, opened 2022-10-08
[BUG - Develop] Evennia can no longer create global scripts from settings.py


InspectorCaracal:
#### Describe the bug
As of merging #2882 any attempt to create a new game with global scripts defined in settings does not create the script. It simply fails the assertion check and moves on without attempting to create them.

#### To Reproduce
1. Set up a new freshly-made game with `GLOBAL_SCRIPTS` defined in the settings. (I included what I used at the bottom.)
2. Attempt to start the game.

#### Expected behavior
It's supposed to automatically create the script if it doesn't exist.

#### Develop-branch commit
b0f24f997

#### Additional context
```py
# added to typeclasses/scripts.py
from evennia.utils import logger

class TestScript(DefaultScript):
    key = "global_test_script"

    def at_repeat(self):
        logger.log_msg("This just prints a nice message.")
```

```py
# added to server/conf/settings.py
GLOBAL_SCRIPTS = {
    "global_test_script": {
        "typeclass": "typeclasses.scripts.TestScript",
        "persistent": True,
        "interval": 60,
        "desc": "A simple script to test creation.",
    },
}
```

```
2022-10-08 04:42:58 [!!] Traceback (most recent call last):
2022-10-08 04:42:58 [!!]   File "./evgames/evennia/evennia/utils/containers.py", line 214, in load_data
2022-10-08 04:42:58 [!!]     assert issubclass(script_typeclass, _BASE_SCRIPT_TYPECLASS)
2022-10-08 04:42:58 [!!] AssertionError
2022-10-08 04:42:58 [!!] GlobalScriptContainer could not start import global script global_test_script. It will be removed (skipped).
```

Henddher:
Yeap! that's due to #2882. I'll take a look.

Henddher:
Hi @InspectorCaracal

This is the problem:

At `evennia start`, we load and "cache" `settings.BASE_SCRIPT_TYPECLASS`. However, in your example, `TestScript` does **not** derives from `typeclasses.scripts.Script` so the assertion triggers and the script gets skipped. Because your `TestScript` is subclassing `DefaultScript` which is **not** what `settings.BASE_SCRIPT_TYPECLASS` expects.

```
> /home/henddher/Documents/github/evennia/evennia/utils/containers.py(207)load_data()
-> _BASE_SCRIPT_TYPECLASS = class_from_module(settings.BASE_SCRIPT_TYPECLASS)
[...]

PR #2919, opened 2022-10-08
#2915 Add test for (questionably) badly subclassed script.

Henddher:
New test to verify that a good script is always a subclass of `settings.BASE_SCRIPT_TYPECLASS`. If a script derives from `DefaultScript` and `DefaultScript` is NOT the `BASE_SCRIPT_TYPECLASS` is considered a badly subclassesed script.

#2915

### DIFF
--- a/evennia/utils/tests/test_containers.py
+++ b/evennia/utils/tests/test_containers.py
@@ -4,6 +4,7 @@ from evennia.utils import containers
 from django.conf import settings
 from django.test import override_settings
 from evennia.utils.utils import class_from_module
+from evennia.scripts.scripts import DefaultScript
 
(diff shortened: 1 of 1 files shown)


Henddher:
Hi @InspectorCaracal, @Griatch, please take a look?

Assuming this is WAI, documenting this "nuance" explicitly would help

InspectorCaracal:
As I mentioned in the issue [over here](https://github.com/evennia/evennia/issues/2915#issuecomment-1272412922), I don't think the assertion is necessary.

I also believe that breaking when inheriting from `DefaultScript` rather than your game's `Script` class would itself be a bug. The `contrib` modules generally inherit from the generic Default typeclasses and using them directly should not break your game.

Griatch:
No, this is not something we want to enforce. Closing it, since it will not be added.

PR #2920, opened 2022-10-09
#2915 Remove assert that verifies that a Script is a subclass of DefaultScript during global script discovering

Henddher:
Remove assert that verifies that a Script is a subclass of DefaultScript during global script discovering

#2915 
Previous PR #2882

### DIFF
--- a/evennia/utils/tests/test_containers.py
+++ b/evennia/utils/tests/test_containers.py
@@ -4,6 +4,7 @@ from evennia.utils import containers
 from django.conf import settings
 from django.test import override_settings
 from evennia.utils.utils import class_from_module
+from evennia.scripts.scripts import DefaultScript
 
 _BASE_SCRIPT_TYPECLASS = class_from_module(settings.BASE_SCRIPT_TYPECLASS)
 
@@ -19,6 +20,10 @@ class WorseScript(_BASE_SCRIPT_TYPECLASS):
   @property
   def objects(self):
     from evennia import module_that_doesnt_exist
+
+class QuestionableScript(DefaultScript):
+  """Does NOT derive from settings.BASE_SCRIPT_TYPECLASS so it would fail."""
+  pass
 
 class TestGlobalScriptContainer(unittest.TestCase):
 
@@ -64,6 +69,15 @@ class TestGlobalScriptContainer(unittest.TestCase):
 
   @override_settings(GLOBAL_SCRIPTS={'script_name': {'typeclass': 'evennia.utils.tests.test_containers.BadScript'}})
   def test_start_with_bad_typeclassed_script_skips_it(self):
+    gsc = containers.GlobalScriptContainer()
+
+    gsc.start()
+
+    self.assertEqual(len(gsc.typeclass_storage), 0)
+    self.assertNotIn('script_name', gsc.typeclass_storage)
+
+  @override_settings(GLOBAL_SCRIPTS={'script_name': {'typeclass': 'evennia.utils.tests.test_containers.QuestionableScript'}})
+  def test_start_with_questionably_subclassed_script_skips_it(self):
     gsc = containers.GlobalScriptContainer()
 
     gsc.start()
--- a/evennia/utils/tests/test_containers.py
+++ b/evennia/utils/tests/test_containers.py
@@ -78,6 +78,9 @@ class TestGlobalScriptContainer(unittest.TestCase):
 
   @override_settings(GLOBAL_SCRIPTS={'script_name': {'typeclass': 'evennia.utils.tests.test_containers.QuestionableScript'}})
   def test_start_with_questionably_subclassed_script_skips_it(self):
+    # Make sure _BASE_SCRIPT_TYPECLASS is NOT DefaultScript but it's a subclass
+    assert issubclass(_BASE_SCRIPT_TYPECLASS, DefaultScript) and _BASE_SCRIPT_TYPECLASS != DefaultScript
+
     gsc = containers.GlobalScriptContainer()
 
     gsc.start()
--- a/evennia/utils/tests/test_containers.py
+++ b/evennia/utils/tests/test_containers.py
@@ -11,19 +11,33 @@ _BASE_SCRIPT_TYPECLASS = class_from_module(settings.BASE_SCRIPT_TYPECLASS)
 class GoodScript(_BASE_SCRIPT_TYPECLASS):
   pass
 
+
 class BadScript:
-  """Not subclass of _BASE_SCRIPT_TYPECLASS,"""
+  """Not even a subclass of DefaultBaseScript."""
   pass
+
 
 class WorseScript(_BASE_SCRIPT_TYPECLASS):
-  """objects will fail upon call"""
+  """
+  A really bad contrived script.
+
+  It derives from BASE_SCRIPT_CLASS but its objects.filter() will fail
+  during GlobalScriptContainer.start().
+  """
+
   @property
   def objects(self):
-    from evennia import module_that_doesnt_exist
+    objs = object()
+    def _filter(*args, **kwargs):
+      from evennia import module_that_doesnt_exist
+    objs.filter = _filter
+    return objs
+
 
 class QuestionableScript(DefaultScript):
-  """Does NOT derive from settings.BASE_SCRIPT_TYPECLASS so it would fail."""
+  """Does NOT derive from settings.BASE_SCRIPT_TYPECLASS."""
   pass
+
 
 class TestGlobalScriptContainer(unittest.TestCase):
 
@@ -33,12 +47,22 @@ class TestGlobalScriptContainer(unittest.TestCase):
     self.assertEqual(len(gsc.loaded_data), 0)
 
   @override_settings(GLOBAL_SCRIPTS={'script_name': {}})
-  def test_init_with_typeclassless_script(self):
+  def test_init_with_typeclassless_script_accepts_it_asis(self):
 
     gsc = containers.GlobalScriptContainer()
 
     self.assertEqual(len(gsc.loaded_data), 1)
     self.assertIn('script_name', gsc.loaded_data)
+    self.assertNotIn("typeclass", gsc.loaded_data["script_name"])
+
+  @override_settings(GLOBAL_SCRIPTS={'script_name': {'typeclass': 'evennia.utils.tests.test_containers.NonexistentScript'}})
+  def test_init_with_nonexistent_script_accepts_it_asis(self):
+    gsc = containers.GlobalScriptContainer()
+
+    self.assertEqual(len(gsc.loaded_data), 1)
+    self.assertIn('script_name', gsc.loaded_data)
+    self.assertEqual(gsc.loaded_data["script_name"].get("typeclass"),
+        "evennia.utils.tests.test_containers.NonexistentScript")
 
   def test_start_with_no_scripts(self):
     gsc = containers.GlobalScriptContainer()
@@ -68,13 +92,15 @@ class TestGlobalScriptContainer(unittest.TestCase):
     self.assertEqual(gsc.typeclass_storage['script_name'], GoodScript)
 
   @override_settings(GLOBAL_SCRIPTS={'script_name': {'typeclass': 'evennia.utils.tests.test_containers.BadScript'}})
-  def test_start_with_bad_typeclassed_script_skips_it(self):
+  def test_start_with_bad_typeclassed_script_raises(self):
     gsc = containers.GlobalScriptContainer()
 
-    gsc.start()
+    with self.assertRaises(Exception) as cm:
+        gsc.start()
+        self.fail("An exception was expected but it didn't occur.")
 
-    self.assertEqual(len(gsc.typeclass_storage), 0)
-    self.assertNotIn('script_name', gsc.typeclass_storage)
+    self.assertEqual(len(gsc.typeclass_storage), 1)
+    self.assertIn('script_name', gsc.typeclass_storage)
 
   @override_settings(GLOBAL_SCRIPTS={'script_name': {'typeclass': 'evennia.utils.tests.test_containers.QuestionableScript'}})
   def test_start_with_questionably_subclassed_script_skips_it(self):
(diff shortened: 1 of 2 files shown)


Henddher:
Hi @Griatch and @InspectorCaracal

I went ahead and removed the `assert`. However, what I discovered is disappointing, IMHO. 

When one provides bad scripts (`BadScript` and `WorseScript` in the tests), they are both taken *as-is* by `GlobalScriptContainer` during `load_data()` (what I will call "global script discovering"). However, when `GlobalScriptContainer` proceeds to `start()`, it first adds them to `typeclass_storage` and then tries to start them. Of course, given that these scripts are badly constructed, `GlobalScriptContainer.start()` raises and fails.
I recall during the discussion of #2882, @InspectorCaracal noted and countered skipping scripts. It was decided that we would *skip* them instead of hard failing the start of the server (https://github.com/evennia/evennia/pull/2882#pullrequestreview-1127610817).

Now, we are back to failing to start the server and `GlobalScriptContainer` (albeit being thrown-away because server dies), it's getting itself into a bad state. Moreover, if `GlobalScriptContainer` went to commit the ill-conceived discovered global scripts to the database (because of an implementation detail and/or side-effect), the bad scripts will remain in the database potentially causing havoc in every future start.

I will add inline review comments to illustrate these issues.

Now that I see the end result of removing the `assert`, I think it might cause more damage than good. I'd say that adding/accepting bad scripts may compromise integrity and add uncertainty in the long run. IMHO.

If however, that were your preference, I would strongly advocate to clearly define responsibilities to "global script discovering" and `GlobalScriptContainer.start()` to prevent any inconsistency in the future. Documenting all this "quirks" would be very important. Note that a `Nonexistent` script produces different results (it's discovered but doesn't cause server to fail).

I would advocate we keep the assert and maybe relax it to `DefaultScript` instead of removing it. With it, global script discovering is guarding `GlobalScriptContainer.start()` from hard fails.

evennia/utils/tests/test_containers.py
```diff
@@ -28,12 +47,22 @@ def test_init_with_no_scripts(self):
...
 
   @override_settings(GLOBAL_SCRIPTS={'script_name': {}})
-  def test_init_with_typeclassless_script(self):
+  def test_init_with_typeclassless_script_accepts_it_asis(self):
 
     gsc = containers.GlobalScriptContainer()
 
     self.assertEqual(len(gsc.loaded_data), 1)
     self.assertIn('script_name', gsc.loaded_data)
+    self.assertNotIn("typeclass", gsc.loaded_data["script_name"])
```
Henddher, line 56: Maybe it should have defaulted to `settings.BASE_SCRIPT_TYPECLASS`? Surely, there is no functionality to fall back to anything.

evennia/utils/tests/test_containers.py
```diff
@@ -28,12 +47,22 @@ def test_init_with_no_scripts(self):
...
+
+  @override_settings(GLOBAL_SCRIPTS={'script_name': {'typeclass': 'evennia.utils.tests.test_containers.NonexistentScript'}})
+  def test_init_with_nonexistent_script_accepts_it_asis(self):
+    gsc = containers.GlobalScriptContainer()
+
+    self.assertEqual(len(gsc.loaded_data), 1)
+    self.assertIn('script_name', gsc.loaded_data)
+    self.assertEqual(gsc.loaded_data["script_name"].get("typeclass"),
+        "evennia.utils.tests.test_containers.NonexistentScript")
```
Henddher, line 65: Accepted a `typeclass` that does NOT exists.

evennia/utils/tests/test_containers.py
```diff
@@ -63,19 +92,45 @@ def test_start_with_typeclassed_script_loads_it(self):
...
-  def test_start_with_bad_typeclassed_script_skips_it(self):
+  def test_start_with_bad_typeclassed_script_raises(self):
+    gsc = containers.GlobalScriptContainer()
+
+    with self.assertRaises(Exception) as cm:
+        gsc.start()
+        self.fail("An exception was expected but it didn't occur.")
+
+    self.assertEqual(len(gsc.typeclass_storage), 1)
+    self.assertIn('script_name', gsc.typeclass_storage)
```
Henddher, line 103: So `start()` raised but `gcs.typeclass_storage` has it as a "loadable" script. IMHO, this is a bad quirk.

evennia/utils/tests/test_containers.py
```diff
@@ -63,19 +92,45 @@ def test_start_with_typeclassed_script_loads_it(self):
...
+    # Make sure _BASE_SCRIPT_TYPECLASS is NOT DefaultScript but it's a subclass
+    assert issubclass(_BASE_SCRIPT_TYPECLASS, DefaultScript) and _BASE_SCRIPT_TYPECLASS != DefaultScript
+
     gsc = containers.GlobalScriptContainer()
 
     gsc.start()
 
-    self.assertEqual(len(gsc.typeclass_storage), 0)
-    self.assertNotIn('script_name', gsc.typeclass_storage)
+    self.assertEqual(len(gsc.typeclass_storage), 1)
```
Henddher, line 114: From not being a "loadable" script to a "loadable". If we had relaxed the `assert` instead, this new expectation will hold (a script subclassing `DefaultScript` is "loadable").

evennia/utils/tests/test_containers.py
```diff
@@ -63,19 +92,45 @@ def test_start_with_typeclassed_script_loads_it(self):
...
+  def test_start_with_worse_typeclassed_script_raises(self):
+    gsc = containers.GlobalScriptContainer()
+
+    with self.assertRaises(Exception) as cm:
+        gsc.start()
+        self.fail("An exception was expected but it didn't occur.")
+
+    self.assertEqual(len(gsc.typeclass_storage), 1)
+    self.assertIn("script_name", gsc.typeclass_storage)
+    self.assertEqual(gsc.typeclass_storage["script_name"], WorseScript)
```
Henddher, line 128: `WorseScript` is "loadable" after `start()` raised? Bad quirk, IMHO.

evennia/utils/tests/test_containers.py
```diff
@@ -63,19 +92,45 @@ def test_start_with_typeclassed_script_loads_it(self):
...
 
-  @override_settings(GLOBAL_SCRIPTS={'script_name': {'typeclass': 'evennia.utils.tests.test_containers.WorstScript'}})
-  def test_start_with_worst_typeclassed_script_skips_it(self):
+  @override_settings(GLOBAL_SCRIPTS={'script_name': {'typeclass': 'evennia.utils.tests.test_containers.NonexistentScript'}})
+  def test_start_with_nonexistent_script_skips_it(self):
     gsc = containers.GlobalScriptContainer()
 
     gsc.start()
 
     self.assertEqual(len(gsc.typeclass_storage), 0)
```
Henddher, line 136: `NonexistentScript` is  "loadable" but `start()` did not fail? Another quirk with a different outcome when compared with `BadScript` or `WorseScript`, IMHO.

Griatch:
@Henddher Removing the assert is the correct path - because it limits the flexibility of the developer. Users not using the right parent is not the core issue here - a Script could have any number of more likely issues when defining the script - including a syntax error on the module level that means it can't load at all. 

Rather, the point here is that we need to gracefully handle a script that tracebacks on start for _whatever_ reason - ideally, it should not take down the server but just be skipped, with a clear error being logged.

Henddher:
Thanks @Griatch,

I fundamentally agree with your point: "log an error when a script fails to load."

However, removing the `assert` is actually **failing** to achieve that. Allow me to elaborate and hopefully demonstrate that's the case:

As noted in [2706], the original behavior *was misleading*. Indicated that something was wrong with `GLOBAL_SCRIPTS` instead of something wrong with the script itself. Here is that error to compare it side-by-side with the new error.

```
...evennia_mygame$ evennia start
  File "/home/henddher/Documents/github/evennia_venv/bin/evennia", line 10, in <module>
    main()
  File "/home/henddher/Documents/github/evennia/evennia/server/evennia_launcher.py", line 2273, in main
    init_game_directory(CURRENT_DIR, check_db=True)
  File "/home/henddher/Documents/github/evennia/evennia/server/evennia_launcher.py", line 1761, in init_game_directory
    check_database()
  File "/home/henddher/Documents/github/evennia/evennia/server/evennia_launcher.py", line 1462, in check_database
    evennia._init()
  File "/home/henddher/Documents/github/evennia/evennia/__init__.py", line 363, in _init
    GLOBAL_SCRIPTS.start()
  File "/home/henddher/Documents/github/evennia/evennia/utils/containers.py", line 188, in start
    self._load_script(key)
  File "/home/henddher/Documents/github/evennia/evennia/utils/containers.py", line 136, in _load_script
    typeclass = self.typeclass_storage[key]
KeyError: 'key'
```

With the new implementation (without `assert`), this is what happens for `BadScript`, `WorseScript`, and `NonexistentScript`.

```
# BadScript
.../mygame $ evennia start

  File "/home/henddher/Documents/github/evennia_venv/bin/evennia", line 10, in <module>
    main()
  File "/home/henddher/Documents/github/evennia/evennia/server/evennia_launcher.py", line 2273, in main
    init_game_directory(CURRENT_DIR, check_db=True)
  File "/home/henddher/Documents/github/evennia/evennia/server/evennia_launcher.py", line 1761, in init_game_directory
    check_database()
  File "/home/henddher/Documents/github/evennia/evennia/server/evennia_launcher.py", line 1462, in check_database
    evennia._init()
  File "/home/henddher/Documents/github/evennia/evennia/__init__.py", line 363, in _init
    GLOBAL_SCRIPTS.start()
  File "/home/henddher/Documents/github/evennia/evennia/utils/containers.py", line 188, in start
    self._load_script(key)
  File "/home/henddher/Documents/github/evennia/evennia/utils/containers.py", line 137, in _load_script
    script = typeclass.objects.filter(
AttributeError: type object 'BadScript' has no attribute 'objects'
```
One can see that the traceback still shows `GLOBAL_SCRIPTS.start()`. It says that `BadScript` is missing `objects`. Not much of difference between `AttributeError` and `KeyError`. Is there? "Yes! the script name is there!".

Well, let's look at the other two:

For `WorseScript` and `NonexistentScript`, there is **no** traceback whatsoever and the server starts successfully. There is no indication that neither of these scripts was wrong. I find very hard to believe that anyone would say that this is an acceptable behavior.

Just consider the `NonexistentScript` case: that class does **not** exist in the path specified in `GLOBAL_SCRIPTS`.  There is **no** `NonexistentScript` class anywhere and the server starts just fine? This is **not** achieving any "Log error when script fails to load".

In the contrived case of `WorseScript` (which does have `@property objects`), the server took the script as-is and it did **not** "log error" either.

Neither of these 2 scripts *failed* to load ... they were just fine.

IMHO, the fact that the server is taking **bad** scripts "as-is" and reporting nothing back during start makes no sense. On the other hand, with the `assert`, skipping a known to be bad script during start up is substantially more helpful to script developers. It's gone from "traceback random error" to "script XYZ is wrong and was skipped".

To me, the way I think of `GlobalScriptsContainer` is that it's have 2 main responsibilities: (a) discovering global scripts defined in `GLOBAL_SCRIPTS` and (b) `start()`ing and committing them to DB (maybe). 

With that said, I would picture the implementation as follows:

1. When first instantiated, `GlobalScriptsContainer` "discovers" all scripts declared in `GLOBAL_SCRIPTS`. 
2. For each declared script, it verifies that it **exists** and that it **derives from** `DefaultScript` (or `BASE_SCRIPT_TYPECLASS`). Any script that does **not** pass verification is **rejected** and **skipped** permanently. A rejected script gets logged as an error. 
3. At last, when it `start(...)`, it starts all scripts and commits them to storage (maybe). 
4. At runtime, bad scripts that passed through are bound to fail.

I think this clearly defines what the responsibilities and expectations are. It makes the system to obey a given set of well defined rules. 

I truly recommend that you checkout the branch and carefully analyze each of the test cases I highlighted. See if there is a deterministic behavior to document. IMHO, without clear responsibilities and expectations, there cannot be any deterministic behavior but just some "random" quirks in the implementation.

[2706]: https://github.com/evennia/evennia/issues/2706#issuecomment-1264511086

### Note that this PR removes the `assert` and passes all tests (old and new) so it is effectively fixing #2915.

InspectorCaracal:
I believe the main source of the complications here are the decision to change the previous behavior by making the game "load successfully" even when there are broken, un-importable modules in the game code. I don't believe this outcome is necessary nor desired.

The problem in the original issue was never that the game would break, but that the *actual problem* was being suppressed. The actual error being the failure to import the module. After having reviewed the changes and all of your comments further, it does not actually solve that problem and the issue should not have been closed.

InspectorCaracal:
I've identified and tested a much smaller patch at [line 208](https://github.com/evennia/evennia/blob/394bbe2b97872a56b57fccd2af118d1ef95599b3/evennia/utils/containers.py#L208) that would solve both the original issue and the new bug by simply adding a fallback and removing the try/except block in the relevant code entirely:

```py
            self.typeclass_storage = {}
            for key, data in self.loaded_data.items():
                typeclass = data.get("typeclass", settings.BASE_SCRIPT_TYPECLASS)
                self.typeclass_storage[key] = class_from_module(typeclass, fallback=settings.BASE_SCRIPT_TYPECLASS)
```

This will cause the code to traceback when `class_from_module` fails with an exception, pointing at exactly the error that needs to be corrected, which is the desired behavior described in #2706

Henddher:
Great @InspectorCaracal!

Did you try that with all the test cases I had added -  including the `Nonexistent` case I newly added in this PR?
I think you should push a PR with your fix.

If so, I will withdraw this PR. Your solution is sound. Even though, discovered scripts that are later rejected remain in `_loaded_data`, they don't get added to `typeclass_storage`.

InspectorCaracal:
> Great @InspectorCaracal!
> 
> Did you try that with all the test cases I had added - including the `Nonexistent` case I newly added in this PR? I think you should push a PR with your fix.
> 
> If so, I will withdraw this PR. Your solution is sound. Even though, discovered scripts that are later rejected remain in `_loaded_data`, they don't get added to `typeclass_storage`.

Good call! I just did - it passed all except `test_start_with_nonexistent_script_skips_it` which is expected since the behavior for `class_from_module` is to use the fallback if the module doesn't exist, so there was a script loaded.

As for `loaded_data`, I don't expect that will be an issue, since either `class_from_module` will load *a* script and it'll be present in both, or it'll fail and halt server start-up and there won't be anything loaded any more at all.

(I didn't want to go ahead and just do a new PR in case it would step on your toes, I'm still pretty new to community project etiquette. 😅 But I will go ahead and do that now.)

Henddher:
Nice! @InspectorCaracal 
About etiquette, I don't know either :sweat_smile: I think this was a good collaboration and the system will be in better shape. That's a win for all :smile: 

FYI - Two of the tests I added had typos - I realized that and fixed them in this PR. 
I think `WorseScript` was spelled incorrectly so it was effectively testing a script that doesn't exist. And, I think there was another in `BadScript` (two `d`s). Make sure those are fixed and all of them pass! (that caught me by surprise).

If you want, I will wait until your PR is merged before pushing those myself in a new PR; OR, you can "lift" them from this PR and add them to yours. The former will let me have one more contribution towards #Hacktoberfest (that's a gain for me :wink: ). At the end, we should make sure those tests get added, either way works. Just let me know?

Henddher:
Closing because #2922 offers a cleaner solution.

PR #2922, opened 2022-10-09
Allow global scripts container to raise import errors

### LINKED ISSUE #2706: [BUG - Develop] Global Scripts loading doesn't catch parse errors

### LINKED ISSUE #2915: [BUG - Develop] Evennia can no longer create global scripts from settings.py

InspectorCaracal:
Removes the try/except block when loading global scripts, allowing the actual import traceback/error through to devs. Also amends the `class_from_module` loading to use the base script typeclass as a fallback.

Fully resolves #2706, closes #2915

Also h/t to @Henddher's work pinpointing, discussing and testing the issue in #2882 and #2920


evennia/utils/containers.py
```diff
@@ -208,18 +208,8 @@ def load_data(self):
         if self.typeclass_storage is None:
```
Henddher, line 208: Remember to remove 205-206 and 22. We no longer need `_BASE_SCRIPT_TYPECLASS` I added.

evennia/utils/containers.py
```diff
@@ -208,18 +208,8 @@ def load_data(self):
...
-                    assert issubclass(script_typeclass, _BASE_SCRIPT_TYPECLASS)
-                    self.typeclass_storage[key] = script_typeclass
-                except Exception:
-                    logger.log_trace(
-                        f"GlobalScriptContainer could not start import global script {key}. "
-                        "It will be removed (skipped)."
-                    )
-                    # Let's remove this key/value. We want to let other scripts load.
-                    self.loaded_data.pop(key)
+                typeclass = data.get("typeclass", settings.BASE_SCRIPT_TYPECLASS)
```
Henddher, line 211: This is so much cleaner now!!
:+1: :+1:

Henddher:
Oh, and you need to fix two tests (that's why the checks failed):

- test_start_with_bad_typeclassed_script_skips_it
- test_start_with_worst_typeclassed_script_skips_it

evennia/utils/containers.py
InspectorCaracal: Good catch, thanks

InspectorCaracal:
> Oh, and you need to fix two tests (that's why the checks failed):
> 
>     * test_start_with_bad_typeclassed_script_skips_it
> 
>     * test_start_with_worst_typeclassed_script_skips_it

Whoops, thanks for catching, I was doing my testing against the file you'd revised instead of the one in the branch and forgot.

I rewrote them a little to better represent the changes in expected error conditions, too. 👍

evennia/utils/tests/test_containers.py
```diff
@@ -4,78 +4,71 @@
...
 
-_BASE_SCRIPT_TYPECLASS = class_from_module(settings.BASE_SCRIPT_TYPECLASS)
+_BASE_TYPECLASS = class_from_module(settings.BASE_SCRIPT_TYPECLASS)
 
-class GoodScript(_BASE_SCRIPT_TYPECLASS):
-  pass
+class GoodScript(DefaultScript):
+    pass
 
-class BadScript:
```
Henddher, line 13: I would suggest to add two tests for `BadScript`. It's probably closer to a real error than `BrokenScript` is. `BrokenScript` is very contrived!!
(I added to make sure it would fail during `GlobalScriptContainers.start()`

Apart from that, I think these are all good tests!!

InspectorCaracal: What would it test for?

InspectorCaracal: Also: the original issue was specifically for when there's an import error on the script class, so `BrokenScript` might be contrived but it's a good way to make a successful test case for it that doesn't prevent the test module from loading.

Henddher: At first, I thought one for `gcs = GlobalScriptContainer()` ("init") and the other for `gcs.start()`.

1. If I am not mistaken, `init` will succeed, because the class exists. It will place it in `_loaded_data`. This would be testing internal implementation - which it's not a good thing to test. Better not to add this test actually. However, I think that will be the end result.
 
2. If I am not mistaken, `start()` will fail with `AttributeError` because `BadScript` does not have `objects` property. So this test will raise as expected. I think this is expected behavior because a bad test should not let the server start.

evennia/utils/tests/test_containers.py
```diff
@@ -11,64 +11,76 @@
...
 
-		gsc.start()
+    @override_settings(GLOBAL_SCRIPTS={'script_name': {'typeclass': 'evennia.utils.tests.test_containers.InvalidScript'}})
+    def test_start_with_invalid_script(self):
+        """Script class doesn't implement required start methods"""
+        gsc = containers.GlobalScriptContainer()
 
-		self.assertEqual(len(gsc.typeclass_storage), 1)
-		self.assertIn('script_name', gsc.typeclass_storage)
-		self.assertEqual(gsc.typeclass_storage['script_name'], GoodScript)
```
Henddher: Maybe after `assertRaises`, it would be good to assert that `typeclass_storage` does not have `InvalidScript`. Likewise for the others that raise.

InspectorCaracal: I admit I'm fairly new to unit testing, but I don't understand what the purpose would be for checking the specific details of a failed state, particularly one that would halt execution of the entire process.

That is, if it raised the exception, there shouldn't *be* a `typeclass_storage` anymore to contain or not contain the script. It's a bit like checking which tabs your browser has open after it's crashed and isn't running.

InspectorCaracal:
I agree, removing all safety nets would be a bad idea. That's why I altered it to provide a fallback to `class_from_module` so that it would only throw an exception if something is drastically wrong.

This does cause the server itself to fail on more cases than in # but it will fail in fewer cases than previously, and it will provide useful error messages on the situations where it *does* fail.

I believe that the handling of module loading in `class_from_module` when provided with a fallback typeclass is enough of a safety net.

InspectorCaracal:
Since it's relevant here, some additional clarification on the original issue this all is supposed to be fixing:

The situation that was occurring in the _original_ issue caused the server to fail, so failing in the same case will not increase failures at all. The issue was not that it failed, however, but that it had an ambiguous error. Now, it will fail in the same situation - not any additional situations - and it will fail with the actual error traceback rather than a misleading one.

The other part of the issue is that the  aforementioned situation does not result from a bug in the script's module at all. It's caused specifically when the script's module itself will load fine, but it has a *dependency* that is broken. Throwing an error saying that the global script can't load is better than an ambiguous "key error" but is still masking the issue, which is a completely unloadable and broken *different* module.

As such, this solution does not cause your server to fail in more situations than it had before, merely the same situations with an actual tracebacks. In fact, it should fail in fewer cases, since it now provides a fallback. The cases presented in the unit tests should demonstrate that by raising an exception in on `load_data` *only* in the one specific contrived-to-fail example which represents the original issue.

Griatch (changes requested):
I think I buy the argument; we'll see how it works in practice. :+1:  I have only some feedback on testing the raised exceptions a little more thoroughly, since these are after all one of the points of this PR. :)

Griatch:
@InspectorCaracal Ping!

Griatch:
@InspectorCaracal Ping again. :)

InspectorCaracal:
Sorry Griatch! I've been travelling for the past two weeks, I'll look over this within the next few days for sure

Griatch:
@InspectorCaracal No worries, just pinging in case you are looking to have this count for Hacktoberfest :)

evennia/utils/tests/test_containers.py
```diff
@@ -4,78 +4,83 @@
...
-    gsc = containers.GlobalScriptContainer()
+    @override_settings(GLOBAL_SCRIPTS={'script_name': {'typeclass': 'evennia.utils.tests.test_containers.InvalidScript'}})
+    def test_start_with_invalid_script(self):
+        """Script class doesn't implement required start methods"""
+        gsc = containers.GlobalScriptContainer()
 
-    gsc.start()
+        with self.assertRaises(Exception) as cm:
+                gsc.start()
+                self.fail("An exception was expected but it didn't occur.")
```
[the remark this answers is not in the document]
InspectorCaracal: Very good catch; it turns out that the exception tests were _not_ raising the exceptions I thought they would be, so they weren't properly testing the expected cases. The BrokenScript case is proving a little tricky to test but I'll get it and push an update.

InspectorCaracal:
I noticed there's a conflict now:
```py
<<<<<<< global-scripts-raise
=======
        if self.loaded_data:
            # we don't always load this, it collides with doc generation
            global _BASE_SCRIPT_TYPECLASS
            if not _BASE_SCRIPT_TYPECLASS:
                _BASE_SCRIPT_TYPECLASS = class_from_module(settings.BASE_SCRIPT_TYPECLASS)

>>>>>>> develop
```

Currently the fallback in class_from_module in the lines following that conflict uses `settings.BASE_SCRIPT_TYPECLASS` directly - should it use `_BASE_SCRIPT_TYPECLASS` as defined here instead?

Griatch (approved):
Thanks for the fixes, looks ok to me now :)

PR #2923, opened 2022-10-09
Expand pronoun tests

InspectorCaracal:
Adds some additional tests/parameters to hopefully better catch errors without going so far as testing every possible option.

This PR is being submitted with the expectation that tests will fail, as it will now catch pre-existing issues.

Ultimately, bug fixing.

See #2683 and discussion in #2903

### DIFF
--- a/evennia/utils/tests/test_funcparser.py
+++ b/evennia/utils/tests/test_funcparser.py
@@ -477,7 +477,7 @@ class TestDefaultCallables(TestCase):
             ("male", "Char1 smiles at himself"),
             ("female", "Char1 smiles at herself"),
             ("neutral", "Char1 smiles at itself"),
-            ("plural", "Char1 smiles at itself"),
+            ("plural", "Char1 smiles at themselves"),
         ]
     )
     def test_pronoun_gender(self, gender, expected):
--- a/evennia/utils/verb_conjugation/pronouns.py
+++ b/evennia/utils/verb_conjugation/pronouns.py
@@ -59,7 +59,10 @@ PRONOUN_MAPPING = {
             "neutral": "mine",
             "plural": "ours",
         },
-        "reflexive pronoun": {"neutral": "myself", "plural": "ourselves"},
+        "reflexive pronoun": {
+            "neutral": "myself",
+            "plural": "ourselves"
+        }
     },
     "2nd person": {
         "subject pronoun": {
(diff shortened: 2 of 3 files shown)


Griatch:
@InspectorCaracal Hm, I guess the relevant PR for fixing these merged, so these should theoretically now all pass. If you could update the branch against the latest develop, we could see that it works!

InspectorCaracal:
...Well, in a perplexing (because I'd already run these tests on my other PR???) twist, several of the tests are still failing. I'll look into it and add any necessary fixes into this PR.

InspectorCaracal:
Hm, mild complication.

It looks like I hadn't implemented the `"we" -> "he"` direction since the original hadn't either, but I added a test for it here because it doesn't make sense to accommodate the royal we in only one direction. However, my quick attempt to patch in that functionality results in a different test failing instead, so it looks like it'll be a more complex fix.

This brings me to a different, but related, thing that I've been mulling over. While I understand the "validation" aspect of limiting options to disambiguations, it does present some problems when attempting to incorporate special cases such as the _royal we._ Would it be worth removing that aspect, instead using the given pronoun's attributes as the defaults and overriding them with *whatever* options the end user provided? It would certainly simplify things, although I can see the arguments against it.

Griatch:
> This brings me to a different, but related, thing that I've been mulling over. While I understand the "validation" aspect of limiting options to disambiguations, it does present some problems when attempting to incorporate special cases such as the royal we. Would it be worth removing that aspect, instead using the given pronoun's attributes as the defaults and overriding them with whatever options the end user provided? It would certainly simplify things, although I can see the arguments against it.

I'm afraid I don't understand what you mean by this. :sweat_smile: Could you elaborate?

InspectorCaracal:
Sure!

Right now by design it checks if a target combination of options is valid for the input pronoun. e.g. `$pron(he, 1 p)` would return "he" and "I" if there wasn't a special check overriding that because plural is invalid for "he", and `$pron(they, pp)` would return "they" and "you" because possessive pronoun is invalid for "they".

My suggestion is removing that validation and the special-case handling, so that the end user gets whatever target they specify. `$pron(he, 1 p)` would get "he" and "we", while `$pron(they, pp)` would get "they" and "yours".

The clear argument against it is that it would allow people to create nonsensical combinations.

I tend to err towards more permissiveness, personally, so I think it'd be okay, but it's a big difference in design.

InspectorCaracal:
> @InspectorCaracal I see what you mean about the more permissive approach. I suppose I don't mind that too much in this case; if people don't know their grammar it's not really our concern I guess. So sure, I suppose we can be more permissive. We need to properly document this behavior though, so it's clear it's not a bug.

Okay, good to know! I'll need to chew on how to approach this a bit more before tackling it, but I've been thinking about it on and off while I've been out.

I think the sweet spot in permissiveness would be somewhere in between only allowing PoV changes and allowing any changes. Currently I'm leaning towards taking PoV and gender changes but not part-of-speech changes, so I'll give that a shot first and see how it feels.

InspectorCaracal:
Okay, I think this is done, assuming nothing catches me by surprise *this* time....

I wound up ultimately making very minimal changes re: user inputs. Provided gender options now override, since technically all genders are valid, but otherwise it works the same and uses provided options only to differentiate viewpoint and PoS. This allows for freely using the colloquial plural and royal we, e.g.:
> he->we
> we->he
> yourself->themselves
> they->I

I also changed the function declaration very slightly, using None instead of the default variables for the kwargs to make it simpler to use the input pronoun's own attributes as the defaults if nothing is provided in the function call itself. The usage should be identical.

InspectorCaracal:
...welp there's a funcparser test that got added since I started this which is failing, I need to fix that too.

Griatch (approved):
Thanks for addressing this and also adding more tests!

PR #2929, opened 2022-10-10
Fix access control check in $search() function

### LINKED ISSUE #2917: [BUG - Develop] Funcparser `$search()` doesn't check access appropriately
#### Describe the bug
This line in funcparser:

https://github.com/evennia/evennia/blob/develop/evennia/utils/funcparser.py#L1165

This seems to be intended to check if `caller` can access `target`. But it seems to always pass, possibly because `access` is meant to be called with two arguments and this invocation is slurping the `"control"` access level as an unused args?

#### To Reproduce
Failing test:

```
    def test_search_no_access(self):
        string = "Go to $search(Room)"
        with self.assertRaises(funcparser.ParsingError):
            self.parser.parse(string, caller=self.char2, return_list=True, raise_errors=True)
```

Failure:
```
FAIL: test_search_no_access (evennia.utils.tests.test_funcparser.TestCallableSearch)
  File "/home/tmoney/code/evennia/evennia/utils/tests/test_funcparser.py", line 708, in test_search_no_access
    self.parser.parse(string, caller=self.char2, return_list=True, raise_errors=True)
AssertionError: ParsingError not raised

```

#### Expected behavior
Test passes because `self.char2` doesn't have a "control" access on any rooms.

#### Develop-branch commit

#### Additional context
Again, a small bug that should be easy to fix, but I'd rather submit a skipped failing test and fix it after review.

audiodude:
Changes the access control check of the `$search()` callable in funcparser to actually perform the check. Previously, the call was being made with three arguments, which I believe made it so the check was being performed against `target` as the lock and the `access` parameter was being slurped into `*args` and ignored entirely.

Fixes bug.

Fixes #2917

### DIFF
--- a/evennia/utils/funcparser.py
+++ b/evennia/utils/funcparser.py
@@ -1162,7 +1162,7 @@ def funcparser_callable_search(*args, caller=None, access="control", **kwargs):
         )
 
     for target in targets:
-        if not target.access(caller, target, access):
+        if not target.access(caller, access):
             raise ParsingError("$search Cannot add found entity - access failure.")
 
     return list(targets) if return_list else targets[0]
--- a/evennia/utils/tests/test_funcparser.py
+++ b/evennia/utils/tests/test_funcparser.py
@@ -528,6 +528,7 @@ class TestCallableSearch(test_resources.BaseEvenniaTest):
         """
         string = "$search(TestAccount, type=account)"
         expected = self.account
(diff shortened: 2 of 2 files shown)


audiodude:
Note, this might break existing 1.0 clients who didn't realize they were using a search without access.

Griatch (approved):

Griatch:
Nice catch, thanks!

PR #2930, opened 2022-10-10
Remove f-string with unbound variable in funcparser

### LINKED ISSUE #2916: [BUG - Develop] `$search` in funcparser throws wrong exception when no results found
#### Describe the bug
The `$search()` func should throw funcparser.ParsingError when no results found. Instead it throws NameError because the variable used in the f string is not found.

#### To Reproduce
Test:
```
    def test_search_not_found(self):
        string = "$search(foo)"
        with self.assertRaises(funcparser.ParsingError):
            self.parser.parse(string, caller=self.char1, return_str=False, raise_errors=True)

        ret = self.parser.parse(string, caller=self.char1, return_str=False, raise_errors=False)
        self.assertEqual("$search(foo)", ret)

        ret = self.parser.parse(string, caller=self.char1, return_list=True, raise_errors=False)
        self.assertEqual([], ret)
```

Result:
```
ERROR: test_search_not_found (evennia.utils.tests.test_funcparser.TestCallableSearch)
  File "/home/tmoney/code/evennia/evennia/utils/tests/test_funcparser.py", line 688, in test_search_not_found
    self.parser.parse(string, caller=self.char1, return_str=False, raise_errors=True)
  File "/home/tmoney/code/evennia/evennia/utils/funcparser.py", line 540, in parse
    exec_return = self.execute(
  File "/home/tmoney/code/evennia/evennia/utils/funcparser.py", line 259, in execute
    ret = func(*args, **kwargs)
  File "/home/tmoney/code/evennia/evennia/utils/funcparser.py", line 1156, in funcparser_callable_search
    raise ParsingError(f"$search: Query '{query}' gave no matches.")
NameError: name 'query' is not defined

```

https://github.com/evennia/evennia/blob/develop/evennia/utils/funcparser.py#L1160

#### Expected behavior
The proper ParsingError is thrown.

#### Develop-branch commit

#### Additional context
This is a pretty simple bug to fix, but I'm trying to write tests to increase test coverage without modifying any code-under-test, since that confuses things. I will add the failing test with a 'skip' to demonstrate the issue, and it can be unskipped when it is fixed.

audiodude:
The `$search()` bultin function of funcparser uses an f-string in an error message, but the local variable is not found. This bug was found when adding tests in #2918. It seems like the error string is from an older version of the code where the variable `query` was defined. Fix is to simply change the error message.

Fixes bug

Fixes #2916

### DIFF
--- a/evennia/utils/funcparser.py
+++ b/evennia/utils/funcparser.py
@@ -1153,7 +1153,7 @@ def funcparser_callable_search(*args, caller=None, access="control", **kwargs):
     if not targets:
         if return_list:
             return []
-        raise ParsingError(f"$search: Query '{query}' gave no matches.")
+        raise ParsingError(f"$search: Query '{args[0]}' gave no matches.")
 
     if len(targets) > 1 and not return_list:
         raise ParsingError(


Griatch (changes requested):
When the function was  changed to use `*args` directly in the `search_object` call, the `query` parameter was no longer extracted from `args` into its own variable, which led to this error. But `query` is actually `args[0]` and this should be echoed in the error string.

audiodude:
> When the function was changed to use `*args` directly in the `search_object` call, the `query` parameter was no longer extracted from `args` into its own variable, which led to this error. But `query` is actually `args[0]` and this should be echoed in the error string.

Done.

PR #2931, opened 2022-10-13
Remove typeclass_search. It's not used anywhere.

Henddher:

### DIFF
--- a/evennia/typeclasses/managers.py
+++ b/evennia/typeclasses/managers.py
@@ -535,7 +535,7 @@ class TypedObjectManager(idmapper.manager.SharedMemoryManager):
         stats = self.get_typeclass_totals().order_by("typeclass")
         return {x.get("typeclass"): x.get("count") for x in stats}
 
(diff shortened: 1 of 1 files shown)


Griatch (approved):
In a library like Evennia, it's hard to tell what people are actually using. Not all of these tools need to be used by us - they are provided as convenience tools for developers to use in their own code. 

That said, I think the functionality of this has been superceeded by a plain `Typeclass.objects.all()`, so we could probably remove it ... :thinking:

aMiss-aWry:
Being able to search by typeclass string without needing a search term is handy, which makes it slightly different from Typeclass.objects.all() (which requires importing said typeclass?) - I think this is a nice utility function. Although in that sense it seems likely it belongs in utils/search.py instead.

Griatch:
@aMiss-aWry That's a good point. I will close this @Henddher, let's link to `typeclass_search` from `utils.search` instead, to make it more visible.

Griatch:
I made a new ticket for this (as seen above)

PR #2941, opened 2022-10-14
Fix and catch edge cases in `iter_to_str`

InspectorCaracal:
The wrong output from the mismatch of parameters fixed in command suggestions highlighted some *other* problems with `iter_to_str` that weren't caught by the current unit tests.

This adds a couple additional asserts to the test and corrects the issues they would catch.

Bug fixing

#2940 (the "previous" example highlights the spacing issues being fixed here)


Griatch (approved):
This could probably have merged with the other PR. :) But more tests is always good either way.

PR #2945, opened 2022-10-20
Add matching_exit() to DefaultExit

aMiss-aWry:
The addition of the matching_exit() function. It takes as_list as a parameter to return the matching exit(s) as a list or single Exit. "Matching exits" being the 'return' exit, i.e. if you go 'north' to go somewhere and 'south' to return, **north**.matching_exit() would give you **south**.

Useful little utility for exits.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -3133,3 +3133,25 @@ class DefaultExit(DefaultObject):
 
         """
         traversing_object.msg(_("You cannot go there."))
+
+    def matching_exit(self, as_list=False):
+        """
+        Get the exits that pair with this one in its destination room
+        (i.e. returns to its location)
+
+        Args:
+            as_list (bool): Whether to return available results as a
+                            list or single matching exit.
+
+        Returns:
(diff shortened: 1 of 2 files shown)


Griatch (changes requested):
This is a nice idea! 

I have some points of contention: 

- I think `get_return_exit` may be a clearer name, more consistent with other methods.
- I think `as_list` should be `return_all` instead (see below) 
- Looping over `.exits` is a pretty inefficient way to implement this; for a core method we can afford to be more optimized :) 
- It would be good with a unit test for this.

This would be the more generic and likely much faster approach: 

```python
from evennia.objects.models import ObjectDB 
# ...

# Using ObjectDB means it doesn't matter what typeclass of 
# exit we use - only that it has a `destination`. 
query = ObjectDB.objects.filter(
    db_location=self.destination, db_destination=self.location)

if return_all:  # more suitable kwarg name?
    return query  # more generally useful to return 
                  # a queryset rather than a list here.                         
return query.first()
```

Griatch (approved):
Thanks for addressing this and adding the test! Looks good now :+1:

PR #2950, opened 2022-10-25
Update Contrib-Custom-Gametime.md

homeofpoe:
Fixes typo in header


Griatch (approved):
Thanks, but these kind of similar typo changes should really be made together as one PR. So this would normally not count towards Hacktoberfest. But since you are addressing a lot of other stuff next to this as well, I'll approve it. :)

homeofpoe:
Thanks!  Wouldn't expect it to count. I just learned this morning how to do
multiple changes at once, so this should all be tidier in the future!

PR #2951, opened 2022-10-25
Update admin.py

homeofpoe:
Extend and fix fstring usage.

### DIFF
--- a/evennia/commands/default/admin.py
+++ b/evennia/commands/default/admin.py
@@ -80,7 +80,7 @@ class CmdBoot(COMMAND_DEFAULT_CLASS):
                 return
             pobj = pobj[0]
             if not pobj.access(caller, "boot"):
-                caller.msg(f"You don't have the permission to boot {pobj.key,}.")
+                caller.msg(f"You don't have the permission to boot {pobj.key}.")
                 return
             # we have a bootable object with a connected user
             matches = SESSIONS.sessions_from_account(pobj)
@@ -97,7 +97,7 @@ class CmdBoot(COMMAND_DEFAULT_CLASS):
         if "quiet" not in self.switches:
             feedback = f"You have been disconnected by {caller.name}.\n"
             if reason:
-                feedback += "f\nReason given: {reason}"
+                feedback += f"\nReason given: {reason}"
 
         for session in boot_list:
             session.msg(feedback)
@@ -462,7 +462,7 @@ class CmdPerm(COMMAND_DEFAULT_CLASS):
                 caller.msg("You are not allowed to examine this object.")
                 return
 
-            string = f"Permissions on |{obj.key}s|n: "
+            string = f"Permissions on |{obj.key}|n: "
             if not obj.permissions.all():
                 string += "<None>"
             else:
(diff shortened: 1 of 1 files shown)


evennia/commands/default/admin.py
```diff
@@ -76,12 +76,11 @@ def func(self):
...
             pobj = search.account_search(args)
             if not pobj:
-                caller.msg("Account %s was not found." % args)
+                caller.msg(f"Account {args} was not found.")
                 return
             pobj = pobj[0]
             if not pobj.access(caller, "boot"):
-                string = "You don't have the permission to boot %s." % (pobj.key,)
-                caller.msg(string)
+                caller.msg(f"You don't have the permission to boot {pobj.key,}.")
```
Griatch, line 83: This looks like it will traceback. You need to remove the comma from `{obj.key,}`.

evennia/commands/default/admin.py
```diff
@@ -96,18 +95,17 @@ def func(self):
 
         feedback = None
         if "quiet" not in self.switches:
-            feedback = "You have been disconnected by %s.\n" % caller.name
+            feedback = f"You have been disconnected by {caller.name}.\n"
             if reason:
-                feedback += "\nReason given: %s" % reason
+                feedback += "f\nReason given: {reason}"
```
Griatch, line 100: The `f` is inside the string, that's incorrect.

evennia/commands/default/admin.py
```diff
@@ -464,7 +462,7 @@ def func(self):
                 caller.msg("You are not allowed to examine this object.")
                 return
 
-            string = "Permissions on |w%s|n: " % obj.key
+            string = f"Permissions on |{obj.key}s|n: "
```
Griatch, line 465: An errant `s` following `{obj.key}` should be removed.

evennia/commands/default/admin.py
```diff
@@ -483,8 +481,7 @@ def func(self):
         locktype = "edit" if accountmode else "control"
         if not obj.access(caller, locktype):
             caller.msg(
-                "You are not allowed to edit this %s's permissions."
-                % ("account" if accountmode else "object")
+                f"You are not allowed to edit this {'account' if accountmode else 'object'}'s permissions."
```
Griatch, line 484: While this works, it may be better to use `.format()` here, or break out the code. It's generally a bit harder to read full executable code inside a string like this. I would suggest this instead: 

```python
    accountstr = 'account' if accountmode else 'object'
    caller.msg(f"You are not allowed to edit this {accountstr}'s permissions")
```

Griatch (changes requested):
Thanks for fixing this! I have some comments and suggestions before this can merge.

homeofpoe (commented):
Feedback incorporated - thaaaank you!!!!!

evennia/commands/default/admin.py
```diff
@@ -482,10 +480,8 @@ def func(self):
         # we supplied an argument on the form obj = perm
         locktype = "edit" if accountmode else "control"
         if not obj.access(caller, locktype):
-            caller.msg(
-                "You are not allowed to edit this %s's permissions."
-                % ("account" if accountmode else "object")
-            )
+            accountstr = 'account' if accountmode else 'object'
+            caller.msg(f"You are not allowed to edit this {accountmode}'s permissions.")
```
Griatch, line 484: `accountmode -> accountstr`

Griatch (changes requested):
Almost there :)

homeofpoe (commented):
Phew...

Griatch (approved):

PR #2959, opened 2022-10-26
Fix msg_contents funcparsing

### LINKED ISSUE #2909: [BUG - Develop] inline funcparsers sent to `msg_contents` parses the same for everyone
#### Describe the bug
When passing an inline func for the new funcparser, it parses the string from the perspective of the first object in the list of contents and then uses that same parsed message to all of the successive messages.

#### To Reproduce
1. Add the test command class I provided under "Additional context"
2. Create two characters and have them each execute the command.
3. The string will print with the correct *actor* but from the perspective of whichever character was created or entered the room first, to both characters.

#### Expected behavior
I believe it's supposed to parse each time so the viewer sees the correct names relative to themself.

#### Develop-branch commit
3992f37da

#### Additional context
```python
class TestFuncparser(Command):
    """Usage: test <object>"""
    key = "test"

    def func(self):
        caller = self.caller
        location = caller.location

        target = caller.search(self.args.strip())
        if not target:
            return

        message = f"$You() point at $you({target.name})."

        location.msg_contents(message, from_obj=caller, mapping={"you": caller, target.name: target})
```

InspectorCaracal:
Changes the assignment of the parsed message in `msg_contents` to avoid overwriting the input string

Bug fixing

closes #2909

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -763,7 +763,7 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
         for receiver in contents:
 
             # actor-stance replacements
-            inmessage = _MSG_CONTENTS_PARSER.parse(
+            outmessage = _MSG_CONTENTS_PARSER.parse(
                 inmessage,
                 raise_errors=raise_funcparse_errors,
                 return_string=True,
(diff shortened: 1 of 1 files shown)


Griatch (approved):
Thanks!

Issue #2850, opened 2022-09-10
[BUG] Initial character is not `Developer` perm


InspectorCaracal:
#### Describe the bug
When creating a fresh game, Evennia will create a SU account and a character automatically. [The documentation](https://www.evennia.com/docs/latest/Building-Quickstart.html#stepping-down-from-godhood) suggests that this character is supposed to be Developer level, which would make sense. However, it is Player level.

This is true for both 0.9.5 and 1.0-beta

If it is actually intentional that the SU-account character is Player instead of Developer (which I would *not* expect and which has always puzzled me) then the code doesn't need to be fixed, but the documentation needs to be updated.

Griatch:
The superuser account character is indeed intended to be Player level; this is not a bug. The reason is this makes the `quell` functionality max-useful for the superuser - they can quickly become a regular user for testing. The doc is outdated.
I guess one could argue either way on this though.

InspectorCaracal:
Personally I think it makes sense either way, so if it's intended to be Player-level I'd be happy to edit the original post and have this switched to a Documentation issue.

Griatch:
Hm. Reading the docs on this, it does probably make more sense to change the Superuser Character to be `Developer` permission out of the gate. As exemplified in that Builder tutorial, it's nice to quell to be able to test locks properly but also be able to build stuff in that mode.

PR #2961, opened 2022-10-27
Update initial_setup.py

homeofpoe:
Change superuser character permissions per #2850

Per the discussion in #2850, changed the initial superuser character level permissions to Developer.


Griatch (approved):
Thanks for the change!

PR #2963, opened 2022-10-30
chore: convert try/except on TypeError to ValueError

### LINKED ISSUE #2911: [BUG - Develop] Calls to int() wrapped in `TypeError`, should be `ValueError`, in funcparser.py
#### Describe the bug
There are a bunch of calls in the funcparser default callables that are set up to do int conversion/parsing, ie `int(some_var)`. Many if not all of these are wrapped in a try/except block to anticipate the value not being parseable. However, there is a repeated mistake where the caught exception is `TypeError` when it should be `ValueError`:

```
Python 3.9.14 (main, Sep 30 2022, 20:25:24)
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> int('foo')
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: 'foo'
```

See  the following:

https://github.com/evennia/evennia/blob/develop/evennia/utils/funcparser.py#L685
https://github.com/evennia/evennia/blob/develop/evennia/utils/funcparser.py#L917
https://github.com/evennia/evennia/blob/develop/evennia/utils/funcparser.py#L949

etc

#### To Reproduce
N/A

#### Expected behavior
Catch ValueError if the logic in the catch is expected to run. Otherwise, this may be no big deal because the FuncParser itself handles exceptions and the calls to *convert_types would raise the exception before it got to this line if the widths aren't strings. So in that case, remove the exception handling altogether.

#### Develop-branch commit

#### Additional context
N/A

mikejgray:
Correction to TypeError exception catching when converting a string to an integer - ValueError instead of TypeError.

Hacktoberfest! And I love games, with MUDs having a special place in my heart. :)

Closes #2911

### DIFF
--- a/evennia/utils/funcparser.py
+++ b/evennia/utils/funcparser.py
@@ -659,6 +659,8 @@ def funcparser_callable_toint(*args, **kwargs):
     inp = funcparser_callable_eval(*args, **kwargs)
     try:
         return int(inp)
+    except TypeError:
+        return inp
     except ValueError:
         return inp
 
--- a/evennia/utils/funcparser.py
+++ b/evennia/utils/funcparser.py
@@ -657,10 +657,9 @@ def funcparser_callable_eval(*args, **kwargs):
 def funcparser_callable_toint(*args, **kwargs):
     """Usage: $toint(43.0) -> 43"""
     inp = funcparser_callable_eval(*args, **kwargs)
-    try:
-        return int(inp)
-    except TypeError:
-        return inp
+    if isinstance(inp, str) and inp.isnumeric():
+        inp = int(inp)
+    return inp
 
 
 def funcparser_callable_int2str(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


Griatch (changes requested):
Thanks! I do think it's better to keep the `try...except` here but to expand the cases caught though: 
```python
try: 
    inp = int(int) 
except IndexError:   # or except (IndexError, ValueError) where applicable
    return imp 
```
There are a few reasons for this: 
- From a performance perspective, it comes down to what we expect to happen the most. This code will run regularly as part of text processing. If we expect users will _mostly_ use correct inputs, then a `try...except` will produce pretty much no overhead at all. If we, conversely, think that users will _mostly_ pass invalid inputs, then an `if isinstance` is more efficient. In this case I think people will prefer to write correct inputs. 
- The Python mantra is 'leap before you look', and while that doesn't always work, I think it does apply here.

mikejgray:
Understood, I'll make updates and get those over to you when I get a chance.

Griatch (approved):
Thanks for the fixes!

mikejgray:
Hi @Griatch, it appears that Hacktoberfest doesn't recognize this as valid because it doesn't go to trunk. Would you mind applying a `hacktoberfest-accepted` label to this PR?

Griatch:
Will do!

PR #2965, opened 2022-10-30
2926  bugfix nesting off by one

### LINKED ISSUE #2926: [BUG - Develop] Funcparser `_MAX_NESTING` isn't applied properly
#### Describe the bug
The process for checking `_MAX_NESTING` in funcparser.py is too permissive. When it is set to 2 it allows nesting up to 4 before it fails

#### To Reproduce
```
    @patch("evennia.utils.funcparser._MAX_NESTING", 2)
    def test_parse_max_nesting(self):
        """
        Make sure it is an error if the max nesting value is reached.
        """
        string = "$add(1, $add(1, $toint(42)))"
        ret = self.parser.parse(string)
        self.assertEqual("$add(1,$add(1,$toint(42)))", ret)
```

This test currently fails

#### Expected behavior
The value of `_MAX_NESTING` should be applied properly and the test passes. It looks like currently it takes N+2 calls to trigger the nesting failure, since `len(callstack)` is 0 when encountering the first function call. An off by one error.

#### Develop-branch commit

#### Additional context
This was discovered when writing tests in #2918

mikejgray:
Fixes off-by-one error for max nesting. Also updates .gitignore and Makefile for improved development experience and ran a `make fmt`.

Hacktoberfest! And I love games, with MUDs having a special place in my heart. :)

Closes #2926

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,11 @@ docs/build
 
 # Obsidian 
 .obsidian
+
+
+# Virtual environments
+.venv/
+.env/
+
+# Testing folder
+.test_game_dir
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ TESTS ?= evennia
 default:
 	@echo " Usage: "
 	@echo "  make install - install evennia (recommended to activate virtualenv first)"
+	@echo "  make installd - install evennia with requirements_extra (recommended to activate virtualenv first)"
 	@echo "  make fmt/format - run the black autoformatter on the source code"
 	@echo "  make lint - run black in --check mode"
 	@echo "  make test - run evennia test suite with all default values."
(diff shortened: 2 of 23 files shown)


Griatch:
Thanks for the PR! This does not appear to solve the issue though, so I can't merge it: 

- This does not fix it; if you write additional unit tests, you'll find it still allows 4 nestings when MAX_NESTING is 2. 
- Running format on a PR like this is not a good idea - it makes it hard to discern the small changes you made from auto-formatting lines of little value. 
- Some nice little other changes though :+1: 

Since this is also lacking testing, I'll need to reject this PR; I'll cherry pick the changes instead. That said, I still appreciate your work here, so make sure you don't lose Hacktoberfest credit, I've marked this as `hacktoberfest_accepted`, so it should still count.

Issue #2383, opened 2021-04-16
[Feature Request] RPSystem Contrib Matching Options


alephate:
#### Is your feature request related to a problem? Please describe.

I'd like to be able to, in some cases, have unmatched tags in an emote return a generic output. Right now what happens is that it fails to send the emote and delivers an error - "There is no match for <tag>."

#### Describe the solution you'd like

I want to be able to set an option on the functions that in some situations you can tag someone in an emote, and if they're not there, you will see a set sdesc (such as "something") instead of theirs. 

#### Describe alternatives you've considered

In the discord chat we discussed making a dummy object class that would have the necessary features to stand in for a "something". 

#### Additional context

Here is an example of what I mean, using send_emote to send movement messages (I implemented a solution in my own version, but not very tidily).

![image](https://user-images.githubusercontent.com/72240438/115067454-f1a48d00-9ebe-11eb-870b-b726d205cd17.png)

Griatch:
The explicitness here is of course deliberate and goes with Evennia's general approach of not trying to guess if you make a mistake; in a system where recogs matter, asking for clarification is often better than to produce nonsense/generics. 

That said, I suppose one could aim to make this a kwarg on the sdesc replacer function.

alephate:
We had a whole discussion at the time and you asked me to make this feature request. I only made the request because you seemed to think it would be nice, but I have kept fiddling with things and I'm happy with my own fixes. :)

Griatch:
@alephate Yes, I know. I was just clarified why this is this way (same as was clarified in the chat at the time). I do still think this could be useful (but maybe phrased a bit too curtly above), but it would need to be something optional in the RPSystem settings.

PR #2968, opened 2022-11-01
Add optional fallback sdesc to rpsystem emote

InspectorCaracal:
Adds an additional kwarg to `parse_sdescs_and_recogs` (which can be passed through `send_emote`) as an optional generic fallback string. When provided to `send_emote`, any references that have zero matches will instead be replaced with that string.

(This should be easily backportable to v0.9.5 for anyone who wants it, as it doesn't touch the new search algorithm or the updated sdesc system.)

Improving contribs

Implements #2383


Griatch (approved):
This looks like a clean take on that issue, and also adds more flexibility to the contrib. And unit tests give confidence too. Nice!

PR #2974, opened 2022-11-05
Add `get_display_name` to ServerSession class

### LINKED ISSUE #2839: [BUG - Develop] Traceback on multimatching session commands
#### Describe the bug
I've discovered an unusual bug regarding these lines:
https://github.com/evennia/evennia/blob/8b9e1da9413202de69e5467a98e652fe7d58e1de/evennia/commands/command.py#L531-L532

It seems as though it's intending to check if the command is attached to an object and only call the display name hook if it is. However, the `obj` property isn't always an Object or Character - it will _also_ return the Account or Session that the command is attached to, if it's an account- or session-level command.

This becomes a problem if you multimatch on a command input which includes a session-level command match because, unlike Objects, Characters and Accounts, the various Session types have no `get_display_name` method. Which then results in:
```
File "./evennia/evennia/commands/command.py", line 532, in get_extra_info
    return " (%s)" % self.obj.get_display_name(caller).strip()
AttributeError: 'ServerSession' object has no attribute 'get_display_name'
```

#### Possible Solutions
I can see a few different ways to handle this, and I'm unsure which would be best.

1. Change `get_extra_info` on commands to validate the existence of `get_display_name` on the attached `obj` property before calling it
2. Give sessions a `get_display_name` hook
3. Change the `obj` property on commands so it only returns in-game objects (this is probably the worst option)

#### Develop-branch commit
8b9e1da94

InspectorCaracal:
Adds a very basic `get_display_name` method to the Server Session class for command compatibility

bug fixing

resolves #2839

### DIFF
--- a/evennia/server/serversession.py
+++ b/evennia/server/serversession.py
@@ -179,6 +179,15 @@ class ServerSession(_BASE_SESSION_CLASS):
             return self.puppet if self.puppet else self.account
         return None
 
+    def get_display_name(self, **kwargs):
+        """
+        Get display name for this session.
+        
+        Returns:
+            name (str): A formatted string representing the session.
+        """
+        return f"{self.id} ({self.protocol_key})"
+    
     def log(self, message, channel=True):
         """
         Emits session info to the appropriate outputs and info channels.


Griatch:
Do'h, I just fixed and pushed this, without thinking to look if there was a PR for it. :facepalm: Sorry! :( 
My solution takes a little different approach though, so I'll close this now.

PR #2984, opened 2022-11-14
Update menu_login.py

Antrare:
Updating installation instructions to be in line with new requirements.


Griatch (approved):
The I didn't add this second line is because the connection screen is not a part of the `menu_login.py` file; it's actually a separate file and the connection screen used is unrelated to what menu is used. The full instructions combining the two are instead in the README.md for this contrib. 

That said, it's likely that one wants to change the two, and having the description available already here is probably more user-friendly. So I'll add it, thanks. :)

PR #2985, opened 2022-11-19
Move Windows install note

InspectorCaracal:
Since the `python -m evennia` command is required to make the evennia command work in Windows, it'll be easier for the end-user if the note is in the same place as confirming that `evennia -h` works - that way, if/when it doesn't, they'll know exactly what to do.


InspectorCaracal:
Looks like we crossed threads a bit there and you moved it! 👍

PR #2988, opened 2022-11-21
Fix character creator tests

InspectorCaracal:
Updates and corrects broken tests for the character creator contrib


Griatch (approved):
@InspectorCaracal Thanks for fixing! I saw this just before bedtime yesterday, so glad it could be addressed until I woke up :+1:

PR #2990, opened 2022-11-22
Update wilderness contrib for 1.0

InspectorCaracal:
The wilderness contrib is currently a bit buggy in 1.0, with ndb properties not being reapplied on reload, hooks not being called properly on `@teleport`ing out, and probably other things relating to the causes of those issues.

This fixes those hooks and updates/cleans up a few other things, e.g. using (the relatively new) AttributeProperty instead of `@property` wrappers.

I also added a new property, `preserve_items`, to allow for recycling rooms only when they're _actually_ empty rather than just when there are no players present. By default it behaves exactly the same as before, recycling any rooms that have no players.

preparing for 1.0


InspectorCaracal:
There, actually done.

Griatch (approved):
LGTM. :)

Issue #2991, opened 2022-11-22
[BUG - Develop] Extra or missing space before permissions in quell/unquell output


chiizujin:
#### Describe the bug
- When using quell/unquell as a superuser, there is an extra space before "(superuser)".
- When using quell in OOC mode as a normal player, there is no space before "(player)"

#### To Reproduce
1. use "quell" or "unquell" as a superuser.
2. See the extra space before "(superuser)"

and:

1. Create a normal player account.
2. Use the "ooc" command.
3. Use quell.
4. See the missing space before "(player)"

#### Additional context
To fix this, commands/default/account.py needs changing as below:

Change:
`account.is_superuser and " (superuser)" or "(%s)" % ", ".join(account.permissions.all())`
to:
`account.is_superuser and "(superuser)" or "(%s)" % ", ".join(account.permissions.all())`

Change:
`self.msg(f"Quelling Account permissions{permstr}. Use unquell to get them back.")`
to:
`self.msg(f"Quelling Account permissions {permstr}. Use unquell to get them back.")`

There is also a minor thing in the same function that doesn't actually cause an issue. The line:

`if self.cmdstring in ("unquell", "unquell"):`

checks "unquell" twice.

PR #2993, opened 2022-11-22
Fix extra/missing spaces in quell output (#2991)

chiizujin:

Fixes extra and missing spaces in output of quell command as detailed in #2991.

#### Other info

Does not address the other minor thing mentioned relating to the duplication in:
`if self.cmdstring in ("unquell", "unquell")`


Griatch (approved):
Thanks!

PR #3000, opened 2022-11-26
Add Typing issue form

dvoraen:
Adds an [issue form](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) for typing concerns!

To help organize future issue generation.  I feel that any typing concerns generated from my work should be sectioned off for easier tracking.

This is a rough draft concocted in my head that will likely get updated in the future; there isn't a Preview option for issue forms that I could find so you can see the actual form layout.  It just previews the YML text. :)  I'm wondering if I'm going about this the wrong way, and that I have to add this to master to see any results, but I didn't want to "launch" something that was incomplete yet, if that makes sense, so I opted for develop.

Also I booboo'd when I submitted the change on my end.  I, uh, was experimenting with Github features I haven't used before (adding files directly from Github) and for some reason chose not to do it in its own branch as part of committing it.


dvoraen:
Okay, I have discovered that if my .yml file has no errors, the issue form will show up when you open it on Github. :)  I have also discovered that I won't know it has errors until I commit it.  Isn't that lovely? :)

Griatch (commented):
Thanks for the PR!
- Since all our issue templates are written in Markdown, it would be best to keep with that tradition. Both work on github. Markdown templates are not only more human-readable than yaml, using the same style consistently also makes it easier to maintain. 
- That said, I think it's premature to add a separate issue template for typing at this time - it would include reports from the entirety of the library after all. Once we have typing stubs at all, I think that any issues with them can be covered by regular bug reports for the foreseeable future. 
- In conclusion I don't think this should merge, sorry!

PR #3001, opened 2022-11-26
Add discord chat integration

InspectorCaracal:
This adds a built-in Discord chat integration, similar in intent to the IRC and Grapevine integrations. On the Discord side, it requires the end user to sign up for one Discord bot application and manage the permissions and server access. On the Evennia side, all that's required are enabling it in settings and adding the bot token, then using the `discord2chan` command to link and unlink channels

Since there is no twisted library for Discord, I had the pleasure of writing the handling of Discord's API directly into the portal-side protocol, so I kept it to the bare minimum required to enable a chat relay. However, I've set up the evennia-side bot account and the custom protocol so that any courageous souls who want to build a more elaborate integration should be able to do a lot just by extending the bot account without messing with the protocol.

This is a highly requested feature!

### DIFF
--- a/docs/source/Setup/Channels-to-Discord.md
+++ b/docs/source/Setup/Channels-to-Discord.md
@@ -27,12 +27,12 @@ DISCORD_ENABLED = True
 ```
 
 Start/reload Evennia and log in as a privileged user. You should now have a new
-command available: `@discord2chan`. Enter `help discord2chan` for an explanation
+command available: `discord2chan`. Enter `help discord2chan` for an explanation
 of its options.
 
 Adding a new channel link is done with the following command:
 
-     @discord2chan <evennia_channel> = <discord_channel_id>
+     discord2chan <evennia_channel> = <discord_channel_id>
 
 The `evennia_channel` argument must be the name of an existing Evennia channel,
 and `discord_channel_id` is the full numeric ID of the Discord channel.
@@ -122,12 +122,12 @@ DISCORD_ENABLED = True
 Start or reload your game to apply the changed settings, then log in as an account
 with at least `Developer` permissions and initialize the bot account on Evennia:
 
-    @discord2chan/name <your bot name>
+    discord2chan/name <your bot name>
 
 The name you assign it can be anything; it will show up in the `who` list for your
 game and your game's channels, but is otherwise unused.
 
-Lastly, confirm that it's fully enabled by entering `@discord2chan` on its own.
+Lastly, confirm that it's fully enabled by entering `discord2chan` on its own.
 You should receive a message that there are no active connections to Discord.
 
 ### Connecting an Evennia channel to a Discord channel
@@ -140,7 +140,7 @@ then your channel ID is `12345678901234567890`
 
 Link the two channels with the following command:
 
-    @discord2chan <evennia channel> = <discord channel id>
+    discord2chan <evennia channel> = <discord channel id>
 
 The two channels should now relay to each other. Confirm this works by posting a
 message on the evennia channel, and another on the Discord channel - they should
--- a/docs/source/Setup/Channels-to-Discord.md
+++ b/docs/source/Setup/Channels-to-Discord.md
@@ -120,15 +120,9 @@ DISCORD_ENABLED = True
 ```
 
 Start or reload your game to apply the changed settings, then log in as an account
-with at least `Developer` permissions and initialize the bot account on Evennia:
-
-    discord2chan/name <your bot name>
-
-The name you assign it can be anything; it will show up in the `who` list for your
-game and your game's channels, but is otherwise unused.
-
-Lastly, confirm that it's fully enabled by entering `discord2chan` on its own.
-You should receive a message that there are no active connections to Discord.
+with at least `Developer` permissions and initialize the bot account on Evennia with
+the `discord2chan` command. You should receive a message that the bot was created, and
+that there are no active connections to Discord.
 
 ### Connecting an Evennia channel to a Discord channel
 
--- a/docs/source/Setup/Channels-to-Discord.md
+++ b/docs/source/Setup/Channels-to-Discord.md
@@ -1,0 +1,169 @@
+# Connect Evennia channels to Discord
+
+[Discord](https://discord.com) is a popular chat service, especially for game
+communities. If you have a discord server for your game, you can connect it
+to your in-game channels to communicate between in-game and out.
+
+## Configuring Discord
+
+The first thing you'll need is to set up a Discord bot to connect to your game.
+Go to the [bot applications](https://discord.com/developers/applications) page page and make a new application. You'll need the
+"MESSAGE CONTENT" toggle flipped On, and to add your bot token to your settings.
+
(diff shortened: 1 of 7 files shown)


InspectorCaracal:
While skimming the new updated docs, I've found that there are "setting up IRC/Grapevine/etc." pages; I'll write up an equivalent set for Discord as well and include it in this PR.

Griatch:
On a quick glance, this looks pretty complete and in line with the other channel bots. Thanks for spending the time to figure out how to do this! 
(I won't give detailed feedback at this time since this is still marked "Draft").

InspectorCaracal:
Just noticed I missed three "retry" to-do items in the protocol; I will try to get those done and pushed tonight but if not, they'll be in within the next 24h for certain. Otherwise this should be ready to go.

docs/source/Setup/Channels-to-Discord.md
```diff
@@ -0,0 +1,179 @@
...
+
+```python
+# file: server/conf/settings.py
+DISCORD_ENABLED = True
+```
+
+Start or reload your game to apply the changed settings, then log in as an account
+with at least `Developer` permissions and initialize the bot account on Evennia:
+
+    @discord2chan/name <your bot name>
```
Griatch, line 125: This breaks with the style of other bots; where the botname is hard-coded (such as grapewinehaus) or set directly in the definition, such as 

    irc2chan evennia_channel = <network_info> <botname>

Not _that_ big a deal, but it seems like it would be better to not invent yet another initialization style ...

evennia/accounts/bots.py
```diff
@@ -42,16 +40,17 @@ def at_script_creation(self):
...
+    def at_server_start(self):
+        self.at_start()
 
     def at_start(self):
         """
         Kick bot into gear.
 
         """
-        if not self.db.started:
+        if not len(self.account.sessions.all()):
```
Griatch, line 52: No need for `len()` here, just `if not self.account.sessions.all():` is enough.

evennia/accounts/bots.py
```diff
@@ -68,21 +67,6 @@ def at_repeat(self):
...
-        """
-        self.db.started = True
-
-    def at_server_shutdown(self):
-        """
-        Make sure we are shutdown.
-
-        """
-        self.db.started = False
-
```
Griatch, line 85: The purpose of these checks were to make sure you didn't re-initialize the bot even if the script was refreshed multiple times. With this change, it seems like this could start over and over and redo the intialization as long as there are no sessions connected to the server?

evennia/accounts/bots.py
```diff
@@ -576,3 +559,186 @@ def execute_cmd(
...
+            # don't relay our own relayed messages
+            return
+        if channel_list := self.db.channels:
+            # get all the discord channels connected to this evennia channel
+            channel_name = channel.name
+            for dc_chan in [dcid for evchan, dcid in channel_list if evchan == channel_name]:
+                # send outputfunc channel(msg, discord channel)
+                super().msg(channel=(strip_ansi(message.strip()), dc_chan))
+
+    def direct_msg(self, message, sender, **kwargs):
```
Griatch, line 656: Mark what the `**kwargs` are for too; in this case they are passed in into the relay. Are they unused? Are there any particular keywords we know of that can be used?

evennia/accounts/bots.py
```diff
@@ -576,3 +559,186 @@ def execute_cmd(
...
+        Called when the Discord bot receives a direct message on Discord.
+
+        Args:
+            message (str)  - Incoming text from Discord.
+            sender (tuple) - The Discord info for the sender in the form (id, nickname)
+
+        """
+        pass
+
+    def relay_to_channel(
```
Griatch, line 667: `**kwargs` are not used by default, but they should still be in the docstring, describing just that - that they are unused by default.

Griatch (changes requested):
This is very nice work. There is some concern that the bot protocol implementation itself is not tested (it could be done with mocking of the connection, for example). But the other bots are not covered by that level of unit tests either, so I won't require that. 

I have some comments/points to address, but otherwise this looks great!

evennia/accounts/bots.py
InspectorCaracal: That's true, yeah. I was having some issues with `.db.started` not being set back to False when I had to stop the server, and looking at the various bots it seemed as though the main purpose of initializing the bot (that is, running `start()`) is to connect the sessions, so that seemed to me to be an equivalent check. e.g. if it's not connected, start it; if it's connected, it's already been started. I could very well have misinterpreted how that was working, though.

docs/source/Setup/Channels-to-Discord.md
InspectorCaracal: That's true! I originally had this in so you could customize how the relay is displayed, but since then I figured out a better solution to just take the bot name out of the evenni-side relay entirely. I don't mind taking it out for consistency; I'll make sure to remove the respective tests as well.

evennia/accounts/bots.py
Griatch: Hm. Let's try this simplification and see how it goes; it _should_ be okay, I'm just always a wary of changing functionality like this in a parent class affecting many subclasses not under test.

Griatch (approved):
Thanks for the fixes! There is still mentions of the `/name` flag in the documentation, but I can fix that. :)

docs/source/Setup/Channels-to-Discord.md
```diff
@@ -0,0 +1,179 @@
...
+DISCORD_ENABLED = True
+```
+
+Start or reload your game to apply the changed settings, then log in as an account
+with at least `Developer` permissions and initialize the bot account on Evennia:
+
+    discord2chan/name <your bot name>
+
+The name you assign it can be anything; it will show up in the `who` list for your
+game and your game's channels, but is otherwise unused.
```
Griatch, line 128: This `/name` was removed

Griatch (changes requested):
Actually - I realize the unit tests are failing, and it's down to the discord integration, so that needs to be fixed. 
So might as well fix the docs too :)

InspectorCaracal:
This is what I get for assuming my changes are fine and not running tests first. 😅 But at least the tests are doing their job. Fixing this now, hopefully won't take long.

evennia/accounts/bots.py
InspectorCaracal: Yeah, I debated it for a while, but determined that if it wasn't working when I was using it the same way as the other bots, it probably wouldn't work for them either.

I'd like to add unit tests for all of the chat relay bot accounts across the board over the next month or two, but that'll take me a while so I wanted to get this one working and in first. If there are any issues with this revision, I *should* catch it then.

InspectorCaracal:
> This is what I get for assuming my changes are fine and not running tests first. 😅 But at least the tests are doing their job. Fixing this now, hopefully won't take long.

Looks like the problem was just that the hardcoded name I added and the name I used in the test don't match. Rolling this in with the docs fix and pushing. 👍

PR #3006, opened 2022-12-01
Unit testing workflow updates

dvoraen:
This PR updates the unit testing and Docker deployment jobs to be more modular and possibly extensible.

To make the workflows more modularized and potentially extensible.

##### Issues
- I don't know how to test these workflows. :)  My thought was that I would have to branch off my `new_workflows` branch and create a test workflow to manually invoke them, but I wasn't sure if that was an effective solution.
  - The Docker deploy job is untestable by me because I don't have a DockerHub login, for example, plus it specifically looks for a push/PR to `master` and `develop` and that seemed like it would be ripe for trouble if I made artificial commits for testing. :)
- I'm unclear whether you need to use `actions/checkout` for a second job in a workflow (see `jobs.deploy` for what I mean).  GitHub Docs suggests yes, because a job runs in isolation on its own runner, which makes me think that it can't see the file system associated with another job.

##### Changelog
- Database server/service setup has been placed into its own action, and a `.github/actions` folder was created for such code if more become necessary in the future.  This was done mostly so that a lot of steps could be factored out of the primary workflow and the primary workflow becomes "just the facts, ma'am".  I'm being the maid, in other words.
- A timeout limit of 5 minutes was placed on the database setup action in the primary workflow.  If PostgreSQL or MySQL fail to activate, they would force the workflow to run for the entire *six hours* (the default timeout value) before being cancelled by GitHub.  In looking through past workflows on `develop`, typically it would be ready within 10 seconds.
- The coverage test can now be explicitly marked in the `jobs.test.matrix.include` part of the workflow.
  - The unit testing code now checks for the presence of the "is this a coverage test?" flag rather than hardcodes the specific Python version and database type.  I figured this was more extensible and easier to maintain than digging through all the `if` conditions if it should change in the future.  I say extensible because you can just designate version/database combinations as coverage tests in the `jobs.test.matrix.include` at will rather than need to modify conditions elsewhere.
- The build-and-push steps for deploying `master` and `develop` to the Docker image were put into their own job, which is dependent upon the unit tests job succeeding.  This was more my interpretation of making the workflow more modular.  Deployment is dependent on unit testing, but it's not directly involved with it, if that makes sense.  It does its own thing, so I stuck it into its own thing/job.  It also only triggers if the source repo is evennia/evennia.  This is so that forks don't end up with DockerHub job failures that aren't really relevant to them; forks shouldn't be deploying images to Docker, only the source repo.

### DIFF
--- a/.github/actions/setup-database.yml
+++ b/.github/actions/setup-database.yml
@@ -1,0 +1,89 @@
+# evennia/setup-database
+
+# Use this action in a workflow for when you need to do initialization of database services
+# (such as with PostgreSQL and MySQL) before you initiate unit tests that will employ that
+# database service.
+
+# NOTE: This action was written for use with the core Evennia workflows ONLY.
+
+name: Set up Database
+author: dvoraen
+description: "Activates the database server for the passed in service and ensures it's ready for use."
+
+inputs:
+  database:
+    description: "Database service being initialized."
+    type: string
+    required: true
+
+runs:
+  using: "composite"
+
+  steps:
+    # - name: Set up SQLite
+    #   if: inputs.database == "sqlite3"
+
+    - name: Set up PostgreSQL server
+      if: inputs.database == "postgresql"
+      uses: harmon758/postgresql-action@v1
+      with:
+        postgresql version: "11"
+        postgresql db: "evennia"
+        postgresql user: "evennia"
+        postgresql password: "password"
+
+    - name: Wait for PostgreSQL to activate
+      if: inputs.database == "postgresql" && success()
+      run: |
+        while ! pg_isready -h 127.0.0.1 -q >/dev/null 2>&1
+        do
+          sleep 1
+          echo -n .
+        done
+        echo
+      shell: bash
+
+    - name: Set up MySQL server
+      if: inputs.database == "mysql"
+      uses: mirromutth/mysql-action@v1.1
+      with:
+        host port: 3306
+        # character set server: 'utf8mb4'
+        # collation server: 'utf8mb4_unicode_ci'
+        character set server: "utf8"
+        collation server: "utf8_general_ci"
+        mysql database: "evennia"
+        mysql user: "evennia"
+        mysql password: "password"
+        mysql root password: root_password
+
+    - name: Wait for MySQL to activate
+      if: inputs.database == "mysql" && success()
+      run: |
+        while ! mysqladmin ping -h 127.0.0.1 -u root -proot_password -s >/dev/null 2>&1
+        do
+          sleep 1
+          echo -n .
+        done
+        echo
+      shell: bash
+
+    - name: Set up MySQL Privileges
+      if: inputs.database == "mysql" && success()
+      run: |
+        cat <<EOF | mysql -u root -proot_password -h 127.0.0.1 mysql
+          create user 'evennia'@'%' identified by 'password';
+          grant all on \`evennia%\`.* to 'evennia'@'%';
+          grant process on *.* to 'evennia'@'%';
+          flush privileges
+        EOF
+      shell: bash
+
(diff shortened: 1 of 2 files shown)


.github/actions/setup-database.yml
```diff
@@ -0,0 +1,89 @@
...
+    description: "Database service being initialized."
+    type: string
+    required: true
+
+runs:
+  using: "composite"
+
+  steps:
+    # - name: Set up SQLite
+    #   if: inputs.database == "sqlite3"
```
Griatch, line 24: Left over debugs?

.github/workflows/github_action_test_suite.yml
```diff
@@ -5,188 +5,137 @@ name: test-suite
...
 
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10', '3.11']
-        TESTING_DB: ['sqlite3', 'postgresql', 'mysql']
+        python-version: ["3.9", "3.10", "3.11"]
+        TESTING_DB: ["sqlite3", "postgresql", "mysql"]
+        include:
+          - python-version: "3.10"
```
Griatch, line 25: What is this extra `python-version` for?

Griatch (commented):
This is looking quite clean. I don't see any obvious errors, but it's hard to test without actually running it. I just have some minor points and a question I'd like to understand the answer to.

.github/actions/setup-database.yml
dvoraen: I left that there as a reminder in case SQLite requires any additional configuration in the future.  I can remove it if you'd rather it not exist as a ghost.

.github/workflows/github_action_test_suite.yml
dvoraen: TL;DR: It's to specify which run will run coverage testing.  When the matrix strategy currently being employed is `python-version: 3.10` and `TESTING_DB: sqlite3`, the variable `coverage-test` will be "included" and set to `true`.

Longer version: you can "include" additions to your matrix configuration variables without forcing the addition to be a part of every single run.  When you use `include` in your matrix, you're telling the job runner that you want to add extra and/or specific configurations to your run.  In this case, I want to flag a specific run (3.10/sqlite3 as you designated) as the test which runs `coverage` and submits the results.  You did this previously by hardcoding it in the step's if condition: `if: ${{ matrix.python-version == "3.10" && matrix.TESTING_DB == "sqlite3" }}`.  I took out the hardcoded conditions and swapped them for a variable designating a run as a coverage-tested run, which is true only when we include it.

In layman's terms, what I did was say, add `coverage-test` (accessible as `${{ matrix.coverage-test }}`) with a value of true, and the run I'd like to add it to has `python-version` as "3.10" and `TESTING_DB` "sqlite3".  Since 3.10 and sqlite3 exist as part of the existing strategy, it simply inserts `coverage-test` into that run rather than create a new run with that variable, thus avoiding redundancy.  The documentation doesn't explicitly say anything about "non-existent" variables when outside that specific run, but I refer you to the below link which references the variable outside that specific run without any apparent issue.

Compare with this example from GitHub Docs: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-expanding-configurations

To adapt the wording from the official documentation:

> When the job for the `python-version` value of `3.10` and `TESTING_DB` value of `sqlite3` runs, an additional variable called `coverage-test` with the value of `true` will be included in the job.

Griatch: Thanks for the explanation, was not familiar with this syntax :+1:

Griatch (approved):
Thanks for the improvements! It does look much cleaner this way. 
It's hard to see offhand what can be wrong without running it, so let's do that :)

PR #3008, opened 2022-12-01
Workflow string fixes for GitHub

dvoraen:
Fixing strings that I unintentionally broke.

I was bad and I'm trying to be good.

I didn't know that GitHub expressions required single-quotes for strings and that double-quotes were forbidden. 💦


Griatch (approved):
I didn't think about that either :) I was just about to fix it, but this is faster, thanks!

PR #3010, opened 2022-12-01
Action fix

dvoraen:
Attempting to fix action from not importing correctly.  I didn't read the composite and same-repository directions correctly.

Because I can't read directions so I'm trying to be better.

I'm trying to make sure things work, bear with me. :(


dvoraen:
Okay, I am a lot more confident that things work now.

Griatch:
That works better now :) 

For some reason the Coverage reports 0% matches now though. But I don't think that's due to your changes, I see the commit before your github-workflow fix also returned 0% :/

PR #3016, opened 2022-12-03
Adding coverage settings to TOML and workflow

dvoraen:
This PR adds `[tool.coverage.run]` to `pyproject.toml` based on the workflow settings used for running `coverage`.  ~~It also includes a tentative fix to allow coverage testing to run in parallel.~~  (Removed due to `coverage` producing warnings I can't seem to resolve yet.)

A unit testing run now has a max timeout of 35 minutes (for each run).  This is for the PostgreSQL issue where sometimes it will stall until the full six hours elapse.  That's not good.

Also put in environment variables to allow prettier command lines for the actual test commands.  Remember, I'm neurotic.

To consolidate the settings into `pyproject.toml` which is one of its purposes.

The massive diff in the `pyproject.toml` is a side-effect of the TOML extension I'm using for VSCode which automatically formats it to have smaller indentations, spaces between `=` and such.

### DIFF
--- a/.github/workflows/github_action_test_suite.yml
+++ b/.github/workflows/github_action_test_suite.yml
@@ -26,6 +26,12 @@ jobs:
             TESTING_DB: "sqlite3"
             coverage-test: true
 
+    timeout-minutes: 35
+
+    env:
+      UNIT_TEST_SETTINGS: "--settings=settings --keepdb --parallel 4 --timing"
+      COVERAGE_TEST_SETTINGS: "--settings=settings --timing"
+
     steps:
       - uses: actions/checkout@v3
 
@@ -67,12 +73,7 @@ jobs:
         if: ${{ ! matrix.coverage-test }}
         working-directory: testing_mygame
         run: |
-          evennia test \
-            --settings=settings \
-            --keepdb \
-            --parallel 4 \
-            --timing \
-            evennia
+          evennia test ${{ env.UNIT_TEST_SETTINGS }} evennia
 
       # OBS - it's important to not run the coverage tests with --parallel, it messes up the coverage
       # calculation!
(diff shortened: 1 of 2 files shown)


dvoraen:
Okay, now that the rebasing merge fun is over... grumble.

I cannot figure out how to address the module-not-measured warnings x4 for 'evennia' that keep cropping up.

.github/workflows/github_action_test_suite.yml
```diff
@@ -66,27 +68,19 @@ jobs:
       - name: Run test suite
         if: ${{ ! matrix.coverage-test }}
         working-directory: testing_mygame
+        env:
+          UNIT_TEST_SETTINGS: "--settings=settings --keepdb --parallel 4 --timing evennia"
```
Griatch, line 72: The `evennia` part here is not really part of the settings/options of the test run. It would make more sense to just have the options here, so you can do `evennia test ${{ env.UNIT_TEST_SETTINGS}} evennia` later.

.github/workflows/github_action_test_suite.yml
```diff
@@ -66,27 +68,19 @@ jobs:
...
         run: |
-          coverage run \
-            --source=evennia \
-            --omit=*/migrations/*,*/urls.py,*/test*.py,*.sh,*.txt,*.md,*.pyc,*.service \
-            ../bin/unix/evennia test \
-              --settings=settings \
-              --timing \
-              evennia
+          coverage run --rcfile=../pyproject.toml
+          coverage combine
```
Griatch, line 83: While I know this the details are covered in  `[tool.coverage.run]` in `pyproject.toml`, breakiing it out like this makes this file harder to read imo; while the `pyproject.toml` trick is nice, it seems like it would be easier to understand this file if this was also using the `UNIT_TEST_SETTINGS` envvar. 

If keeping it in the pyproject.toml file, we need at least some comment here to explain this.

dvoraen:
Unfortunately, I can't test `--parallel` testing locally.  Windows specifically does not play nicely with it. :(  I can only test changes to the command lines and file structure locally, but once parallel testing comes in I have to switch to non-Windows.

.github/workflows/github_action_test_suite.yml
dvoraen: TL;DR - coverage testing has its own environment variable for the settings for now

I met you halfway here, I think.  I took out `--parallel` from the coverage test for now, because I can't resolve the issue with the `module-not-measured` warning probably until after I figure out how to get WSL working properly on my rig.  I left the general coverage settings in the TOML (the source and omissions) though, which is the "halfway" part of my statement.

Griatch:
Testing with `--concurrency=multiprocessing` locally, I see the same warnings from `coverage` with `module-not-measured`; but the end result after combining is still the same 66% :shrug: Most likely these are benign warnings; it's nevertheless cool to see one can do coverage correctly in a parallel run.

PR #3026, opened 2022-12-10
Change get, drop, give, inv to use get_numbered_name()

chiizujin:

Changed the message output of `get`, `drop`, `give` and `inventory` to use "singular" from `get_numbered_name()`.

This makes the output read a little more nicely. For example:

`Gollum drops ring.` -> `Gollum drops a ring.`
`You pick up ring.` -> `You pick up a ring.`
`You give ring to Gollum.` -> `You give a ring to Gollum.`

The message for a failure of `to_give.move_to()` has also been extended from (for example):

`You could not give ring.`

to:

`You could not give a ring to Sauron.`

##### Inventory

`inventory` output has changed from:

```
 ring  This thing is bad news.
```

to:

```
 a ring  This thing is bad news.
````

##### Other

- One of the lines in `drop` (475) was changed from old-style formatting to an f-string for consistency.
- Unit tests for default commands have been updated.

### DIFF
--- a/evennia/commands/default/general.py
+++ b/evennia/commands/default/general.py
@@ -371,8 +371,9 @@ class CmdInventory(COMMAND_DEFAULT_CLASS):
 
             table = self.styled_table(border="header")
             for item in items:
+                singular, _ = item.get_numbered_name(1, self.caller)
                 table.add_row(
-                    f"|C{item.name}|n",
+                    f"|C{singular}|n",
                     "{}|n".format(utils.crop(raw_ansi(item.db.desc or ""), width=50) or ""),
                 )
             string = f"|wYou are carrying:\n{table}"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: evennia/commands/default/general.py [hunk 1]
```diff
@@ -424,8 +425,9 @@ def func(self):
         if not success:
             caller.msg("This can't be picked up.")
         else:
-            caller.msg(f"You pick up {obj.name}.")
-            caller.location.msg_contents(f"{caller.name} picks up {obj.name}.", exclude=caller)
+            singular, _ = obj.get_numbered_name(1, caller)
+            caller.msg(f"You pick up {singular}.")
+            caller.location.msg_contents(f"{caller.name} picks up {singular}.", exclude=caller)
```
Griatch, line 430: While we are at updating this string, we could update this for the modern way to express this using funcparser embeds:

```python
caller.location.msg_contents(f"$You() $conj(pick) up {singular}.", from_obj=caller)
```
With this, there is no need for a separate `callerm.msg()` on the line above.

evennia/commands/default/general.py
```diff
@@ -472,8 +474,9 @@ def func(self):
         if not success:
             caller.msg("This couldn't be dropped.")
         else:
-            caller.msg("You drop %s." % (obj.name,))
-            caller.location.msg_contents(f"{caller.name} drops {obj.name}.", exclude=caller)
+            singular, _ = obj.get_numbered_name(1, caller)
+            caller.msg(f"You drop {singular}.")
+            caller.location.msg_contents(f"{caller.name} drops {singular}.", exclude=caller)
```
Griatch, line 479: Same here, these two lines can be replaced by 

```python
caller.location.msg_contents(f"$You() $conj(drop) {singular}.", from_obj=caller)
```

Griatch (changes requested):
This is a good idea! While doing this change we should use the funcparser output as well though.

Griatch (approved):
Thanks!

PR #3038, opened 2022-12-20
Fix extendedroom contrib

kenneaal:
Updates documentation in README.md for the extended_room contrib to account for 1.0 changes to contrib structure and default_cmds.

Correcting documentation.

Commit isn't cleaned, but contains only one changed file.


evennia/contrib/grid/extended_room/README.md
```diff
@@ -16,19 +16,23 @@ In more detail, in `mygame/commands/default_cmdsets.py`:
...
 
 ```
 
 Then reload to make the new commands available. Note that they only work
 on rooms with the typeclass `ExtendedRoom`. Create new rooms with the right
-typeclass or use the `typeclass` command to swap existing rooms.
+typeclass or use the `typeclass` command to swap existing rooms. Note that since
+this contrib overrides the `look` command, you will need to add the
+`extended_room.ExtendedRoomCmdSet` to the default character cmdset *after*
+super().at_cmdset_creation(), or it will be overridden by the default look.
```
Griatch, line 35: This is also explicitly noted in the comment in that class. But can't hurt to be extra clear :+1:

Griatch (approved):

PR #3053, opened 2022-12-27
Update extended_room.py

Antrare:
Small QOL improvement to looking at details, tell the room contents (including the player) what they are looking closely at.

Considers the Multiplayer context (showing other players what is going on in the room)

### DIFF
--- a/evennia/contrib/grid/extended_room/extended_room.py
+++ b/evennia/contrib/grid/extended_room/extended_room.py
@@ -359,7 +359,10 @@ class CmdExtendedRoomLook(default_cmds.CmdLook):
                 ):
                     detail = location.return_detail(args)
                     if detail:
-                        # we found a detail instead. Show that.
+                        # we found a detail
+                        # tell all the objects in the room we're looking closely at something
+                        caller.location.msg_contents("$You() $conj(look) closely at the $obj(target).\n", from_obj=caller, mapping={"target": args})
+                        # show the detail to the player
                         caller.msg(detail)
                         return
                 # no detail found. Trigger delayed error messages
--- a/evennia/contrib/grid/extended_room/extended_room.py
+++ b/evennia/contrib/grid/extended_room/extended_room.py
@@ -361,7 +361,7 @@ class CmdExtendedRoomLook(default_cmds.CmdLook):
                     if detail:
                         # we found a detail
                         # tell all the objects in the room we're looking closely at something
-                        caller.location.msg_contents("$You() $conj(look) closely at the $obj(target).\n", from_obj=caller, mapping={"target": args})
+                        caller.location.msg_contents("$You() $conj(look) closely at $obj(target).\n", from_obj=caller, mapping={"target": args})
                         # show the detail to the player
                         caller.msg(detail)
                         return
--- a/evennia/contrib/grid/extended_room/extended_room.py
+++ b/evennia/contrib/grid/extended_room/extended_room.py
@@ -361,7 +361,7 @@ class CmdExtendedRoomLook(default_cmds.CmdLook):
                     if detail:
                         # we found a detail
                         # tell all the objects in the room we're looking closely at something
-                        caller.location.msg_contents("$You() $conj(look) closely at $obj(target).\n", from_obj=caller, mapping={"target": args})
+                        caller.location.msg_contents(f"$You() $conj(look) closely at {args}.\n", from_obj=caller)
                         # show the detail to the player
                         caller.msg(detail)
                         return
--- a/evennia/contrib/grid/extended_room/tests.py
+++ b/evennia/contrib/grid/extended_room/tests.py
@@ -71,6 +71,19 @@ class TestExtendedRoom(BaseEvenniaCommandTest):
         self.call(
             extended_room.CmdExtendedRoomLook(), "nonexistent", "Could not find 'nonexistent'."
         )
+        
+    def test_cmdextendedlook_second_person(self):
+        # char2 is already in the same room. 
+        # replace char2.msg with a Mock; this disables it and will catch what it is called with
+        self.char2.msg = Mock() 
+
+        self.call(
+            extended_room.CmdExtendedRoomLook(), 
+            "testdetail"
+        )
+     
+        # check what char2 saw.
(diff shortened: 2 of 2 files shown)


evennia/contrib/grid/extended_room/extended_room.py
```diff
@@ -359,7 +359,10 @@ def func(self):
                 ):
                     detail = location.return_detail(args)
                     if detail:
-                        # we found a detail instead. Show that.
+                        # we found a detail
+                        # tell all the objects in the room we're looking closely at something
+                        caller.location.msg_contents("$You() $conj(look) closely at the $obj(target).\n", from_obj=caller, mapping={"target": args})
```
Griatch, line 364: I don't understand why there is `$obj(target)` here, with a mapping of `target=args`. For this code to run, `detail` must be non-Null, which only happens if `args` pointed to a detail (an Attribute on the room). So `$obj()` should never find a result to use here - we should be able to inject `{detail}` directly into the string instead. 

Also, are we sure that `the` is always a suitable prefix to the detail. I would suggest removing `the` from the string.

evennia/contrib/grid/extended_room/tests.py
```diff
@@ -63,7 +63,11 @@ def test_cmdextendedlook(self):
             "here",
             "Room(#{})\n{}".format(rid, self.SPRING_DESC),
         )
-        self.call(extended_room.CmdExtendedRoomLook(), "testdetail", self.DETAIL_DESC)
+        self.call(
+            extended_room.CmdExtendedRoomLook(), 
+            "testdetail", 
+            "You look closely at the {}.\n|{}".format("testdetail", self.DETAIL_DESC)
```
Griatch, line 69: Since we are now adding a third-party view to this call, consider verifying by checking what char2 sees (after making sure they are in the same room).

Griatch (changes requested):
A nice idea! I have some comments to address before we can merge.

evennia/contrib/grid/extended_room/extended_room.py
Antrare: Happy to make the first change, that makes a lot of sense, thankyou.

On the 2nd, the 'the' makes sense when looking at all non-named objects I could think of.  
You look closely at the bed.
Antrare looks closely at the bed. 

It doesn't make sense when looking at named details:
You look closely at the Smaug.
Antrare looks closely at the Smaug.

I don't really mind either way, but it seems like 'the' is going to be right more often than not, and named details are far more likely to be a created object.

evennia/contrib/grid/extended_room/tests.py
Antrare: Sorry, I have no idea where to start with this.  This is the first automated test I've ever created, could you point me in the right direction and I'll have a crack at it (even if it's just a similar test that I can try to push in)

evennia/contrib/grid/extended_room/extended_room.py
Griatch: More flexible is, I think, to skip the prefix: 

You look closely at window 
You look closely at Smaug
You look closely at bed 

These all work in the MUD context, especially since you are likely to write "look bed". 

The dev can create the detail as "the bed" if they want that particular grammatical flavor.

Antrare: Done, removed the 'the'.  

After checking, I believe the $obj(target) /etc. is correct.  The {detail} is actually the full detail description, not the 'key' of the detail.

evennia/contrib/grid/extended_room/tests.py
Griatch: Np, since these tests inherit from `BaseEvenniaCommandTest`, there are a few basic entities available. Notably `self.char1` and `self.char2`, both of which start in the same location (found as `self.room`). 

What happens when you do `self.call(cmdname(), "argument", "expected return")` is that `self.char1` is firing the command and then we investigate which `.msg(text="")` was called on it. In this case we want to find out what `msg()` call was called on `char2` (the other person in the room by default). We will use a `mock` for this (see the Evennia documentation on unit testing and Python docs on mocks in general). Here's an example: 

```python 
# in the extended_room testing module 
from mock import Mock 
# ,,, 
class TestExtendedRoom(BaseEvenniaCommandTest): 
    # ...
    def test_looking_at_detail_with_multiple_people(self):
        # we know char2 is already in the same room, otherwise we could do something 
        # like `self.char2.location = self.char1.location` here

        # replace char2.msg with a Mock; this disables it and will catch what it is called with
        self.char2.msg = Mock() 

        # let char1 look at the detail set up for the testing room we are in 
        self.call(extended_room.CmdExtendedRoomLook(), "testdetail", self.DETAIL_DESC) 
     
        # now check what char2 saw.
        self.char2.msg.assert_called_with(text=(f"Char looks closely at testdetail."))  # not complete 

``` 
Run the test from the gamedir with `evennia test evennia.contrib.grid.extended_room.TestExtendedRoom.test_looking_at_detail_with_multiple_people`. The example above will fail because I didn't fill in the `assert_called_with` fully in my example above, but you should easly see what is sends to `msg()` and update your test assert to match.

evennia/contrib/grid/extended_room/extended_room.py
Griatch: Fair, the `{detail}` is incorrect. It should be `{args}`. My main point is that it makes no sense to wrap it in `$obj()`.

evennia/contrib/grid/extended_room/tests.py
Griatch: @Antrare I gave a suggestion on how to add the unit test above; let me know if something more is needed. Once the test is in place, this looks like it should be fine to merge :)

Griatch:
@Antrare Any updates on this? Need more aid with the unit test?

Antrare:
Hoping to have a go this weekend, I have very bursty availability for coding/etc.

evennia/contrib/grid/extended_room/tests.py
Antrare: Added the test, hopefully all good.

Griatch (approved):
Thanks for the additional test, looks fine. :+1:

PR #3059, opened 2023-01-02
Update evform.py

Antrare:
Clarify file path not including .py

Documentation update - removing the .py from the file path & noting that it is not to be added.

Prevent simple issues when following the doco.

### DIFF
--- a/evennia/utils/evform.py
+++ b/evennia/utils/evform.py
@@ -61,8 +61,8 @@ Usage
 ```python
 from evennia import EvForm, EvTable
 
-# create a new form from the template - do not include the .py of the filename!
-form = EvForm("path/to/testform")
+# create a new form from the template - using the python path
+form = EvForm("path.to.testform")
 
 # alteratively, you can supply the template as a dict:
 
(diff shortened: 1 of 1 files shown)


evennia/utils/evform.py
```diff
@@ -61,8 +61,8 @@
 ```python
 from evennia import EvForm, EvTable
 
-# create a new form from the template
-form = EvForm("path/to/testform.py")
+# create a new form from the template - do not include the .py of the filename!
+form = EvForm("path/to/testform")
```
Griatch, line 65: I don't think this correction is, well, correct. 

The module parser _does_ require `.py` if the path is given as a full unix-style path, like exemplified here: `path/to/testform.py` (e.g. `/home/foo/mygame/world/testform.py`) 

However, if using a _python path_ (which is also allowed), you mustn't use the .py suffix, like `path.to.testform` (e.g. `world.testform`). 

I would suggest changing this PR to instead provide the python path (`path.to.testform`) since it's probably more likely to be used than the unix path.

Griatch (changes requested):
I think I see where the confusion may be - I suggest that this could be fixed in a different way.

evennia/utils/evform.py
Antrare: Thanks, I wasn't aware that you could use either method.

Griatch (approved):
Thanks for the changes!

PR #3061, opened 2023-01-03
Add setting to enable/disable new user registration

InspectorCaracal:
Adds a single new setting, `REGISTER_ENABLED` which enables registration. By default it's set to `True` but if set to `False`, it deactivates the new account command and the registration page on the website.

Adding a useful requested feature


evennia/settings_default.py
```diff
@@ -34,6 +34,8 @@
 # Lockdown mode will cut off the game from any external connections
 # and only allow connections from localhost. Requires a cold reboot.
 LOCKDOWN_MODE = False
+# Enables new account registration
```
Griatch, line 37: I would suggest a clearer comment here, something like
```python
# Control new registrations with the default unloggedin-create command. 
# Set to False to lock down your game to newcomers.
```

evennia/settings_default.py
```diff
@@ -34,6 +34,8 @@
...
+REGISTER_ENABLED = True
```
Griatch, line 38: This settings name is a bit ambivalent. I suggest being more explicit, like `NEW_ACCOUNT_REGISTRATION_ENABLED`

Griatch (changes requested):
This is a nice feature! Looks like a clean implementation. I think it would be good with a unit test (using `@override_settings` decorator) to showcase this working for `CmdUnconnectedCreate` command though.

InspectorCaracal:
> This is a nice feature! Looks like a clean implementation. I think it would be good with a unit test (using `@override_settings` decorator) to showcase this working for `CmdUnconnectedCreate` command though.

Oh, good catch! I'll add that today along with the setting name and comments changes.

evennia/settings_default.py
InspectorCaracal: I went with:
```py
# Controls whether new account registration is available.
# Set to False to lock down the registration page and the create account command. 
```

Griatch (approved):
Thanks for the changes!

PR #3075, opened 2023-01-08
Update README.md of TraitHandler contrib to explain undocumented capability for multiple TraitHandlers.

jrsteensen:
Added section at bottom addressing adding additional TraitHandlers. 

Shows example based on first example, but instead using a "stats" TraitHandler, a "skills" TraitHandler and the default "traits" TraitHandler.

Added example at bottom of traits contrib README that expanded first example, but instead breaks the three traits into a stats, skills and traits TraitHandler.

Useful functionality that should be easily accessible to other users from initial read to prevent later refactoring.


Griatch (approved):

Griatch:
Thanks for the addition!

PR #3091, opened 2023-01-22
chore: update Dockerfile to reduce the image size

tdurieux:

Hi there,

I've made a small improvement to the Dockerfile that I think could help optimize the image size.

Summary of the changes:
* I added the `--no-cache-dir` to the `pip` command to disable the package cache.
* I added `--no-cache` flag to `apk add` to ensure that no useless information is kept inside the Docker image.

Impact on the image size:
* Image size before repair: 1.25 GB
* Image size after repair: 1.06 GB
* Difference: 195.49 MB (15.24%)

It reduces the Docker image size, therefore it would make it slightly faster the download Evennia and take less space on the disk.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ FROM python:3.11-alpine
 LABEL maintainer="https://www.evennia.com"
 
 # install compilation environment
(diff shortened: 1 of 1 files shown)


Griatch (approved):
Looks like a nice improvement. Thanks!

PR #3105, opened 2023-02-03
Documentation page for NGINX configuration

InspectorCaracal:
This adds a new documentation page which contains a lightly informative pair of example nginx configurations, including a few weird quirks/gotchas regarding trying to use nginx for telnet.

It is *not* a full guide and isn't intended to be one, since teaching people how to use nginx is outside the scope of evennia docs and the HAProxy guide is better for new server admins.

As someone who's used nginx for a long time, I have a general expectation that any sufficiently mature webserver software will have a sample nginx configuration that I can reference. This fulfills that expectation!


InspectorCaracal:
I think I linked to the HAProxy guide correctly for the doc builder to render it, but please double-check that for me.

Griatch (approved):
This is a good addition! The linking to the HAProxy page is fine. For this to actually show up, this must be linked in from another page though (be part of a doctree structure). I'll add that after I've merged. :)

Issue #123, opened 2014-01-22
Attempt to create duplicate 'Player #1' on first run


Griatch:
_From [braktech](https://code.google.com/u/braktech/) on December 22, 2010 07:57:40_

A) What steps will reproduce the problem? Following http://code.google.com/p/evennia/wiki/GettingStarted :
1. svn checkout http://evennia.googlecode.com/svn/trunk/ evennia
2. python manage.py
3. python manage.py syncdb
4. python evennia.py -i start B) What is the expected output? What do you see instead (tracebacks, error messages)? Expect: Creation of initial objects and server start
Actual:
[user@server game]$ ./evennia.py -i start

Starting Evennia server in interactive mode (stop with keyboard interrupt) ...
Logging to: Standard output.

---

 Server started for the first time. Setting defaults.
 Creating startup screen(s) ...
 Creating objects (Player #1 and Limbo room) ...

---

 Resuming initial setup from step 2.
 Creating objects (Player #1 and Limbo room) ...

Traceback (most recent call last):
<traceback snip>
IntegrityError: column user_id is not unique
<traceback snip>
django.db.utils.IntegrityError: column user_id is not unique

Failed to load application: column user_id is not unique

Full output including full tracebacks (not that helpful as it is Django functions complaining about nonuniqueness) is attached. C) What version/svn revision of Evennia are you using? On what operating system? Give as much information as possible to make it easier for developers to recreate the problem. Evennia Revision: r1266 (latest)
OS: Fedora release 14 (Laughlin)
Kernel: 2.6.35.9-64.fc14.x86_64
Python: ver 2.7 x86_64
Django: ver 1.2.3
Twisted: ver 10.1.0 x86_64 D) Please provide any additional information/suggestions below. The issue occurs when first running scripts on the newly created player object (Player #1).
Making the following changes in src/commands/default/system fixes the problem.  This is reverting changes introduced in r1256 , and I am unaware of any deeper implications.

Line 12: from src.server.server import EVENNIA >> #from src.server.server import EVENNIA
Line 436: EVENNIA.shutdown() >> session.server.shutdown()

According to the pre- r1256 notes, importing the EVENNIA class directly restarts the server.  This in turn forces the initial setup to restart at the 'failed' step and try to recreate that user (which fails due to the no longer unique user_id column).

**Attachment:** [ev_full_trace](http://code.google.com/p/evennia/issues/detail?id=123)

_Original issue: http://code.google.com/p/evennia/issues/detail?id=123_

Issue #3106, opened 2023-02-07
[BUG] Multiple channels with the same name, differed only by case, can exist - and prove difficult to extinguish.


aMiss-aWry:
#### Describe the bug
I (accidentally) created a duplicate channel with a different case as its key (ooc vs OOC) through the server/settings.py. Deleting proved difficult.

#### To Reproduce
1. Create a channel in the game through the usual default fashion, say `channel/create OOC`.
2. Add the below DEFAULT_CHANNELS dict to your settings.py.
3. Restart your game so the updated settings take effect. 
5. Type `channel/delete ooc`
5. Try, in a spontaneous fit of hope, `channel/delete OOC`
6. See error
```
DEFAULT_CHANNELS = [
    {
        "key": "ooc",
        "desc": "Public OOC discussion",
        "locks": "control:perm(Admin);listen:all();send:all()",
        "typeclass": BASE_CHANNEL_TYPECLASS,
    }
]
```
#### Expected behavior
Either channels should be case insensitive and it should be impossible to create a channel of the same name but a different case, or channels should be case sensitive and when passed in as arguments to the channel command the matching-case channel should be returned.

I think the former makes more sense as that's already how it works with the in-game command (you can't `chan/create ooc` and then `chan/create OOC`. The server setting just bypasses it.

#### Environment, Evennia version, OS etc
Evennia version: 1.0-dev (rev 96e486de80)

#### Additional context
Honestly I don't think this is the first time this has happened to me. I think like the last time, I'll just manually delete it through py commands. But I can't be the only person this ever happens to.

InspectorCaracal:
I'm not 100% on the django query subscripts but I think the fix would just be changing this line to `db_key__iexact`
https://github.com/evennia/evennia/blob/0a4e17d06d6c1293f27ddd68f8b4c39ef49a5b25/evennia/server/server.py#L408

aMiss-aWry:
Perfect, thanks! I'll try it out and add a test.

PR #3107, opened 2023-02-07
Prevent recreation of same-named channels

aMiss-aWry:
Changed filtering to be inexact for checking against existing channels in `server.create_default_channels()`
Changed duplicate reference from `connectinfo_chan = settings.CHANNEL_MUDINFO` to point at `settings.CHANNEL_CONNECTINFO` (this was previously causing connect messages to always show up in the mudinfo channel)
Fixed f-string

Bug fix

Fix for https://github.com/evennia/evennia/issues/3106

I would argue that all such filtering should occur within the create_channel method (which is called by both `server.create_default_channels()` and the `CmdChannel` in-game command), but for the sake of minimizing changes I didn't do it.

### DIFF
--- a/evennia/commands/default/comms.py
+++ b/evennia/commands/default/comms.py
@@ -489,9 +489,8 @@ class CmdChannel(COMMAND_DEFAULT_CLASS):
 
     def create_channel(self, name, description, typeclass=None, aliases=None):
         """
-        Create a new channel. Its name must not previously exist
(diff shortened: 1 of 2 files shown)


Griatch (approved):
Looking good. A nice fix. 

(Optional Pro hint - if you link the issue with the phrase `Resolves #123`, merging the PR will automatically close the issue at the same time :wink:  )

PR #3113, opened 2023-02-13
Use filter_family instead of filter

blongden:

Changing the use of filter to filter_family in search_channel

It's possible to create a channel using a custom typeclass, however you cannot search for the channel once it's created using the current functionality. This change allows search_channel to find the channel once it's been created.

### DIFF
--- a/evennia/commands/default/comms.py
+++ b/evennia/commands/default/comms.py
@@ -263,7 +263,6 @@ class CmdChannel(COMMAND_DEFAULT_CLASS):
         caller = self.caller
         # first see if this is a personal alias
         channelname = caller.nicks.get(key=channelname, category="channel") or channelname
-
         # always try the exact match first.
         channels = CHANNEL_DEFAULT_TYPECLASS.objects.channel_search(channelname, exact=True)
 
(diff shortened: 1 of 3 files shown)


blongden:
Ok so this is the wrong fix. It does actually work though in game, however it fails with 'filter_family' not found on class ChannelDBManager in the tests or when using the evennia.search_channel python command. I have no idea why it would work in one scenario (when using the channel <name> <text> syntax to write to the channel) but not the others. Maybe one for discussion.

blongden:
This change appears to work and keeps the tests happy. I can't vouch for it's correctness though so please take a look! :)

blongden:
This is also the wrong fix here. It causes issues with logging in (not captured by tests yet).

```
evennia.comms.models.ChannelDB.MultipleObjectsReturned: get() returned more than one ChannelDB -- it returned 3!
```

Griatch (changes requested):
I would suggest making the unit tests for the case you want to test first (including the login case not covered by unit tests - nice catch). Once those tests are in place you can check solutions. 

If you don't have time to address this, please make a formal ticket, explaining the issue in detail and how to reproduce.

Griatch:
Ping :)

Griatch:
@blongden This has not seen any changes or replies for 3 months, so I'm closing this. Feel free to ping me if you want to keep updating this and I'll be happy to reopen.

Issue #2353, opened 2021-03-23
[BUG] Update the building commands to have less 'common' names


Griatch:
#### Describe the bugs

After the `@` became optional, there are builder commands like `open` and `lock` that regularly collide with commands that users may want to implement themselves (open/locking doors). 

#### To Reproduce
- The `simpledoor` contrib uses an `open` command that collides with the builder `open` command.

#### Expected behavior

While we don't want to go back to `@`-prefixes, we can change the default builder commands to have less 'common' names.

InspectorCaracal:
I'm pretty new to working with Evennia so I don't know the original rationale behind the decision, but the most straightforward solution here would be to reinstate the prefixes into the commands instead of stripping them. That would make it trivial to differentiate between game system commands and IC commands. e.g. `@open` to create Exits, freeing up `open` for IC functionality like opening doors; `@dig` to create Rooms, freeing up `dig` for IC; `@search` and `@find` to search the database for Objects, freeing up `search` and `find` for IC; etc.

fariparedes:
> 
> 
> I'm pretty new to working with Evennia so I don't know the original rationale behind the decision, but the most straightforward solution here would be to reinstate the prefixes into the commands instead of stripping them. That would make it trivial to differentiate between game system commands and IC commands. e.g. `@open` to create Exits, freeing up `open` for IC functionality like opening doors; `@dig` to create Rooms, freeing up `dig` for IC; `@search` and `@find` to search the database for Objects, freeing up `search` and `find` for IC; etc.

I'm inclined to agree (they are traditional for a reason, after all) but apparently these prefixes confused people, which is why they were made optional in the first place.

InspectorCaracal:
> I'm inclined to agree (they are traditional for a reason, after all) but apparently these prefixes confused people, which is why they were made optional in the first place.

Ah, I see.... That sounds like a potential can of worms that I should keep my own opinion out of. >.>

A probably more complex solution I thought of would be if the prefix was still functional, but you could define it in your game settings - in case people want to replace it with a syntax they're more familiar with, like a forward slash. But I don't know how difficult that would be to implement, so it's just a thought.

fariparedes:
> 
> 
> > I'm inclined to agree (they are traditional for a reason, after all) but apparently these prefixes confused people, which is why they were made optional in the first place.
> 
> Ah, I see.... That sounds like a potential can of worms that I should keep my own opinion out of. >.>
> 
[...]

PR #3121, opened 2023-02-25
Change channel command key.

### LINKED ISSUE #3120: [BUG] Channel command nick doesn't match channel command key
#### Describe the bug
Nicks created by the channel command (e.g. the "pub") nick for the Public channel don't use the right command key: they reference `channel` instead of `@channel`. Normally this doesn't matter, but if you set `CMD_IGNORE_PREFIXES = "/+"` in your settings, this mismatch causes channel nicks to break.

```
> pub test
Command 'channel public = test' is not available. Type "help" for help.
```

#### To Reproduce
1. Set CMD_IGNORE_PREFIXES to anything that pays attention to the option @ prefix (like "/+")
2. Run any channel nick-based command.
3. Observe error.

#### Expected behavior
Channel nicks command keys should match the channel command.

Evennia 1.2.0

#### Additional context
The issue looks like it's here: 
https://github.com/evennia/evennia/blob/bee281f82d126482f0acb8e3ea4153cba0ad4779/evennia/comms/comms.py#L69
I think this should be `@channel` instead of `channel`. Then it matches the command key and CMD_IGNORE_PREFIXES shouldn't matter anymore.

owllex:
Change CmdChannel to use `channel` as its command key rather than `@channel`. This fixes an issue with nicks generated at channel subscription.

Fix buggy interaction with CMD_IGNORE_PREFIXES.

Resolves #3120

### DIFF
--- a/evennia/commands/default/comms.py
+++ b/evennia/commands/default/comms.py
@@ -209,8 +209,8 @@ class CmdChannel(COMMAND_DEFAULT_CLASS):
 
     """
 
-    key = "@channel"
-    aliases = ["@chan", "@channels"]
+    key = "channel"
+    aliases = ["chan", "channels"]
     help_category = "Comms"
     # these cmd: lock controls access to the channel command itself
     # the admin: lock controls access to /boot/ban/unban switches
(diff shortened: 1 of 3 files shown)


Griatch (changes requested):
Gah, I knew there was a reason this was changed, because I was sure I originally wrote this as `channel`. 

It was changed to `@channel` as part of https://github.com/evennia/evennia/issues/2353. So we shouldn't change this back. This need to be fixed on the nick-side, like you originally suggested in the ticket. Sorry for misleading you :disappointed:

owllex:
Will re-do this PR.

PR #3127, opened 2023-03-03
Catch RuntimeError in `DefaultObject.move_to`

InspectorCaracal:
At the moment, attempting to use `obj.move_to(somewhere)` when `somewhere` happens to be the same as `obj` will throw a runtime error, which is not caught.

This makes sure that changing an object's location will catch runtime errors as well as exceptions.

Better error handling

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -927,7 +927,7 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
         # Perform move
         try:
             self.location = destination
-        except Exception as err:
+        except (Exception, RuntimeError) as err:
             logerr(errtxt.format(err="location change"), err)
             return False
 


InspectorCaracal:
Looks like the test runner is still having trouble completing some of the test runs, but the PR itself didn't break any tests.

Griatch (changes requested):
Hm, the `RuntimeError` is the correct response to prevent a location loop. We shouldn't add another superfluous check to check what the traceback is already reporting. 

If we want to make an error report here, we should catch the exception instead. Not only will this handle this case, but also any nested assignment (like trying to move and object to an object which is inside another object inside another object inside the moving object etc).

InspectorCaracal:
@Griatch Ahh, okay! I didn't know you could/had to catch runtime errors separately from exceptions, so I figured since it isn't caught by the current try/except it wasn't catchable.

I'll update the PR

Griatch:
A `RuntimeError` is just another exception :)

Griatch (changes requested):
Well, if the original `Exception` was not catching it, then this addition won't either. So most likely this particular `RuntimeError` is raised outside of this try...except statement (that's why it was uncaught). 

So this will not change anything. I would suggest making a unit test to test this case and make sure that what is fixed is actually fixed ;)

InspectorCaracal:
I see.

Since my original fix is unacceptable, I'll close this and make a new one if I figure out the correct solution.

Griatch:
Ok! Maybe it could do with a feature-request ticket as well

PR #3130, opened 2023-03-05
Update README.md

Antrare:
Updating documentation.


Griatch (commented):
Could you explain why you think this change should be done? It should work fine to import from `contrib.rpg.traits` since the `TraitHandler` is imported in `traits/__init__.py` (it's a package).

Antrare:
I didn't realise that, all good I'll close it out.

PR #3135, opened 2023-03-08
Increase code coverage for scripts package

Storsorken:
Increases code coverage for the scripts package by adding more unit tests to the scripts/tests.py file. All new tests pass. 

Added tests target the following classes in the scripts module:

* TickerHandler: +2 tests
* ScriptDBManager: +1 test
* ScriptHandler: +2 tests
* ExtendedLoopingCall: +3 tests
* MonitorHandler: +6 tests

Issue #1458


evennia/scripts/tests.py
```diff
@@ -88,3 +159,132 @@ def test_start__delay(self, MockLoopingCall):
...
+        """ Test the .start method with interval less than zero """
+        with self.assertRaises(ValueError):
+            callback = mock.MagicMock()
+            loopcall = ExtendedLoopingCall(callback)
+            loopcall.start(-1, now=True, start_delay=None, count_start=1)
+
+    def test__call__when_delay(self):
+        """ Test __call__ modifies start_delay and starttime if start_delay was previously set """
+        callback = mock.MagicMock()
+        loopcall = ExtendedLoopingCall(callback)
```
Griatch, line 173: Since the main task here is not mocked, I think this looping call should be explicitly killed at the end of each test; otherwise there is a list of the async component firing outside of the scope of the test, which is never a good idea.

evennia/scripts/tests.py
```diff
@@ -88,3 +159,132 @@ def test_start__delay(self, MockLoopingCall):
...
+    def test_force_repeat(self):
+        """ Test forcing script to run that is scheduled to run in the future """
+        callback = mock.MagicMock()
+        loopcall = ExtendedLoopingCall(callback)
+        loopcall.clock.seconds = mock.MagicMock(return_value=0)
+
+        loopcall.start(20, now=False, start_delay=5, count_start=0)
+        loopcall.force_repeat()
+
+        callback.assert_called_once()
```
Griatch, line 192: Same thing here; make sure to kill the looping async call explicitly.

Griatch (changes requested):
This is nice work! Good to see more tests for these components. I have some comments I'd like to see addressed, but overall this looks very good.

Griatch:
Btw, the test suite fails, probably due to these lingering deferreds mentioned above.

evennia/scripts/tests.py
Storsorken: In this case, the looping call is never put on repeat using .start(...) so I'm guessing it should be fine in this case. It can't be terminated with .stop() as it isn't running.

Storsorken:
Thanks a lot for the feedback! :) 

I've made some small changes now. Sorry for the long delay, unfortunately, I'm quite busy at the moment

PR #3137, opened 2023-03-09
Add a contrib for Containers

InspectorCaracal:
This contrib adds several commands and a typeclass to implement a basic, but extensible, container mechanic.

The cmdset includes extended versions of `look` and `get` to allow looking at or getting objects that are in another object, as well as a new `put` command that allows you to put an object inside a container. These commands reference a new access type, `get_from`, to determine whether an object can be used as a container or not.

The `ContainerObject` typeclass provided has the `get_from` lock set to true by default and also implements a very basic capacity check, along with two new hooks used by the `get` and `put` commands: `at_pre_get_from` and `at_pre_put_in`.

Nearly everyone wanting to build a game with Evennia has wanted containers of some sort. This contrib is intended to fill that need, particularly for those with minimal to no coding experience, while also being useful for those who want to implement more complicated container functionality.

### DIFF
--- a/evennia/contrib/game_systems/containers/containers.py
+++ b/evennia/contrib/game_systems/containers/containers.py
@@ -190,7 +190,9 @@ class CmdContainerGet(CmdGet):
         if caller == obj:
             self.msg("You can't get yourself.")
             return
-        if not obj.access(caller, "get"):
+
+        # check if this object can be gotten
+        if not obj.access(caller, "get") or not obj.at_pre_get(caller):
             if obj.db.get_err_msg:
                 self.msg(obj.db.get_err_msg)
             else:
@@ -199,9 +201,7 @@ class CmdContainerGet(CmdGet):
 
         # calling possible at_pre_get_from hook on location
         if hasattr(location, "at_pre_get_from") and not location.at_pre_get_from(caller, obj):
-            return
-        # calling at_pre_get hook method
-        if not obj.at_pre_get(caller):
+            self.msg("You can't get that.")
             return
 
         success = obj.move_to(caller, quiet=True, move_type="get")
@@ -270,10 +270,12 @@ class CmdPut(CmdDrop):
 
         # Call the object script's at_pre_drop() method.
         if not obj.at_pre_drop(caller):
+            self.msg("You can't put that down.")
             return
 
         # Call the container's possible at_pre_put_in method.
         if hasattr(container, "at_pre_put_in") and not container.at_pre_put_in(caller, obj):
+            self.msg("You can't put that there.")
             return
 
         success = obj.move_to(container, quiet=True, move_type="drop")
--- a/evennia/contrib/game_systems/containers/containers.py
+++ b/evennia/contrib/game_systems/containers/containers.py
@@ -27,8 +27,8 @@ or implement the same locks/hooks in your own typeclasses.
 
 `ContainerObject` implements the following new methods:
 
-    at_pre_get_from(self, getter, target, **kwargs) - called with the pre-get hooks
-    at_pre_put_in(self, putter, target, **kwargs)   - called with the pre-put hooks
+    at_pre_get_from(getter, target, **kwargs) - called with the pre-get hooks
+    at_pre_put_in(putter, target, **kwargs)   - called with the pre-put hooks
 """
 from django.conf import settings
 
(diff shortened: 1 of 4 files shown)


Griatch (changes requested):
This looks like a commonly asked-about contrib indeed. I have some comments, but overall it looks good!

InspectorCaracal:
Thanks for all the catches!

Embarrassingly, I only just realized there's undoubtedly a page in the documentation outlining exactly how to format the docstrings, but I immediately found it and have it bookmarked now for the future.

evennia/contrib/game_systems/containers/containers.py
```diff
@@ -0,0 +1,301 @@
...
+        """
+        This will be called when something attempts to get another object FROM this object,
+        rather than when getting this object itself.
+
+        Args:
+            getter (Object): The actor attempting to take something from this object.
+            target (Object): The thing this object contains that is being removed.
+
+        Returns:
+            boolean: Whether the object `target` should be gotten or not.
```
Griatch, line 75: It would be good to explain what this means; will the get operation be quietly aborted or will an error message be reported automatically - or is that something that should be handled in this method too?

evennia/contrib/game_systems/containers/containers.py
```diff
@@ -0,0 +1,301 @@
...
+    def at_pre_put_in(self, putter, target, **kwargs):
+        """
+        This will be called when something attempts to put another object into this object.
+
+        Args:
+            putter (Object): The actor attempting to put something in this object.
+            target (Object): The thing being put into this object.
+
+        Returns:
+            boolean: Whether the object `target` should be put down or not.
```
Griatch, line 88: Same here; should error messages be handled here or is this expected to be handled in the calling code?

evennia/contrib/game_systems/containers/containers.py
```diff
@@ -0,0 +1,301 @@
...
+        if not obj.access(caller, "get"):
+            if obj.db.get_err_msg:
+                self.msg(obj.db.get_err_msg)
+            else:
+                self.msg("You can't get that.")
+            return
+
+        # calling possible at_pre_get_from hook on location
+        if hasattr(location, "at_pre_get_from") and not location.at_pre_get_from(caller, obj):
+            return
```
Griatch, line 198: This hook has no default error message, so this failure will be completely silent? It seems like this could be confusing.

evennia/contrib/game_systems/containers/tests.py
```diff
@@ -0,0 +1,38 @@
...
+        self.obj1.location = self.container
+        self.call(CmdContainerLook(), "obj in box", "Obj")
+
+    def test_get_and_put(self):
+        # get normally
+        self.call(CmdContainerGet(), "Obj", "You pick up an Obj.")
+        # put in the container
+        self.call(CmdPut(), "obj in box", "You put an Obj in a Box.")
+        # get from the container
+        self.call(CmdContainerGet(), "obj from box", "You get an Obj from a Box.")
```
Griatch, line 38: Please add examples of failing puts and gets too (best to make a separate test for this). This test suite is also missing tests of containers with limited capacity.

Griatch (changes requested):
Thanks for the fixes! I dove more into the code now and have some additional comments. :)

Antrare:
```
        # add the type=look to the outputfunc to make it
        # easy to separate this output in client.
        self.msg(text=(desc, {"type": "look"}), options=None)
```

As an addition, I believe that line 143 should be caller.msg

InspectorCaracal:
@Antrare Commands have a `msg` method attached (which conveniently handles targetting the calling session rather than account/object for you, avoiding multisession strangeness), so `self.msg()` is actualy correct there.

Antrare:
TIL.  That breaks my brain, but it makes sense.  Thanks.

Griatch:
@InspectorCaracal Ping :)

evennia/contrib/game_systems/containers/tests.py
InspectorCaracal: The limited capacity test is in `TestContainer`

evennia/contrib/game_systems/containers/containers.py
InspectorCaracal: I've added `Notes` to this and the other hook matching the extant methods explaining that the action is aborted, but I'm unsure what the correct answer regarding the error messages should be, since the core code is equally unclear on the matter.

evennia/contrib/game_systems/containers/tests.py
Griatch: True, for the explicit `at_pre_put_in` call. It would be good to have the same for the commands too. Also, could be good with a test of looking into a container that is _not_ a container, just to show this works well with existing objects.

Griatch (changes requested):

InspectorCaracal:
As an FYI, this hasn't fallen off my radar; I'm hoping to get to it by this weekend.

InspectorCaracal:
All right! Sorry for the delay. I expanded the tests slightly and added hook-based failure messages to the get/put commands, so this should be set.

Griatch (approved):
Thanks for the changes, looking good now :)

PR #3159, opened 2023-04-04
Replace ANSI color fallback logic

InspectorCaracal:
This reworks the logic for which ANSI colors are chosen as fallbacks when XTERM256 is unavailable. Since Evennia officially supports a minimum Python version of 3.10, this PR also takes advantage of that to use match-case for the logic instead of if-else blocks.

A number of people have expressed dissatisfaction with the ANSI color fallbacks over the past couple of years - most notably with magentas producing blues.

Here is an image of what the current method produces:
image

As you can see, there are *many* colors that are very much not what you'd expect to get from the original xterm color. `405` being bright blue is a strong example, as is `544` being bright red.

Here is an image of what my new method produces:
image

While there are still a number of colors that are significantly different from the originals, this is almost entirely due to the complete lack of an orange in the ANSI palette - the variance from the original XTERM256 palette items is visibly much lower.

### DIFF
--- a/evennia/utils/ansi.py
+++ b/evennia/utils/ansi.py
@@ -332,7 +332,7 @@ class ANSIParser(object):
                 colval = 134 + ord(letter)
 
             # ansi fallback logic expects r,g,b values in [0..5] range
-            gray = (ord(letter) - 97) / 5.0
+            gray = round((ord(letter) - 97) / 5.0)
             red, green, blue = gray, gray, gray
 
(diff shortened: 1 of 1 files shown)


Griatch (approved):
Interesting, It's been a long time since heard any complaints about this, actually - but with the side-by-side comparison I do agree that the changes makes the mapping make more sense in several places. Nice one :+1:

PR #3165, opened 2023-04-10
Adds to Exits documentation

bradleymarques:

Adds two changes to `docs/source/Components/Exits.md`:

1. Adds note that Exits are one-way (this is mentioned elsewhere in the documentation but not here)
2. Adds cross-reference to example of how to create Rooms and Exits that is covered elsewhere in the documentation.

I got tripped up with the above two things, and would have found it useful to have that info on the Exits.md page.

None.


Griatch (approved):
Looks like some good extra info to add.

PR #3171, opened 2023-04-21
Fix Django ID cache not being flushed in EvenniaTestCase

bradleymarques:

`EvenniaTest` is quite heavy-handed and creates a lot of objects you may not need per test. This makes tests run slower than necessary.

However, `EvenniaTestCase` is insufficient for testing because it does not automatically clear the Django ID cache on test teardown. This can lead to strange and flakey test failures.

Fixes some unpredictable test failures when inheriting from `EvenniaTestCase`.

None - was just something I noticed in my own project.

### DIFF
--- a/docs/source/Coding/Unit-Testing.md
+++ b/docs/source/Coding/Unit-Testing.md
@@ -106,7 +106,7 @@ To test this, run
 
 to run the entire test module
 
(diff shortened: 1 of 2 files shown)


Griatch (approved):
Good find, makes sense.

PR #3177, opened 2023-05-05
Improved generation of Player Characters using charcreate

volundmush:
I feel that there are numerous inefficiencies with the way CmdCharCreate duplicates logic found in DefaultAccount.create_character() and how .create_character() lacks logic found in CmdCharCreate...

So, I decided to refactor the way that character slots and player character creation works, moving most of the logic into DefaultAccount.create_character() and adding some new methods onto DefaultAccount to enable per-account handling of character slots, though the existing settings.MAX_NR_CHARACTERS is still the assumed use case. Now different accounts might have different numbers of slots if people want to override those methods.

New hooks have been added to make it easier to customize the creation of player characters by overloading methods on the Account typeclass.

I believe there are improvements which can be made to make it easier for people to customize player character creation which don't require ripping out or replacing CmdCharCreate.

Personally I think virtually all commands should operate more like this. They should contain little direct logic and instead call upon systems with replaceable hook functions which make individual chunks of logic more easily overloadable.

### DIFF
--- a/evennia/accounts/accounts.py
+++ b/evennia/accounts/accounts.py
@@ -273,7 +273,7 @@ class DefaultAccount(AccountDB, metaclass=TypeclassBase):
     objects = AccountManager()
 
     # Used by account.create_character() to choose default typeclass for characters.
-    player_character_typeclass = settings.BASE_CHARACTER_TYPECLASS
+    default_character_typeclass = settings.BASE_CHARACTER_TYPECLASS
 
     # properties
     @lazy_property
@@ -795,18 +795,26 @@ class DefaultAccount(AccountDB, metaclass=TypeclassBase):
 
     def get_character_slots(self) -> typing.Optional[int]:
         """
-        Returns the number of character slots this account has.
+        Returns the number of character slots this account has, or
+        None if there are no limits.
 
         By default, that's settings.MAX_NR_CHARACTERS but this makes it easy to override.
         Maybe for your game, players can be rewarded with more slots, somehow.
 
-        If it returns None, then there are no limits on character slots.
+        Returns:
+            int (optional): The number of character slots this account has, or None
+                if there are no limits.
         """
         return settings.MAX_NR_CHARACTERS
 
     def get_available_character_slots(self) -> typing.Optional[int]:
         """
-        Returns the number of character slots this account has available.
+        Returns the number of character slots this account has available, or None if
+        there are no limits.
+
+        Returns:
+            int (optional): The number of open character slots this account has, or None
+                if there are no limits.
         """
         if (slots := self.get_character_slots()) is None:
             return None
@@ -834,7 +842,7 @@ class DefaultAccount(AccountDB, metaclass=TypeclassBase):
         Args:
             key (str, optional): If not given, use the same name as the account.
             typeclass (str, optional): Typeclass to use for this character. If
-                not given, use settings.BASE_CHARACTER_TYPECLASS.
+                not given, use self.default_character_class.
             permissions (list, optional): If not given, use the account's permissions.
             ip (str, optional): The client IP creating this character. Will fall back to the
                 one stored for the account if not given.
@@ -854,7 +862,7 @@ class DefaultAccount(AccountDB, metaclass=TypeclassBase):
         character_permissions = kwargs.pop("permissions", self.permissions)
 
         # Load the appropriate Character class
-        character_typeclass = kwargs.pop("typeclass", self.player_character_typeclass)
+        character_typeclass = kwargs.pop("typeclass", self.default_character_typeclass)
         Character = class_from_module(character_typeclass)
 
         if "location" not in kwargs:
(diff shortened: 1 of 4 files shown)


evennia/accounts/accounts.py
```diff
@@ -707,6 +711,40 @@ def set_password(self, password, **kwargs):
...
         self.at_password_change()
 
+    def get_character_slots(self) -> typing.Optional[int]:
+        """
+        Returns the number of character slots this account has.
+
+        By default, that's settings.MAX_NR_CHARACTERS but this makes it easy to override.
+        Maybe for your game, players can be rewarded with more slots, somehow.
+
+        If it returns None, then there are no limits on character slots.
```
Griatch, line 721: I know you have added typing, but Evennia follows the google-style standard, which means you should nevertheless include a `Returns:` stanza in your doc strings.

evennia/accounts/accounts.py
```diff
@@ -707,6 +711,40 @@ def set_password(self, password, **kwargs):
...
+        """
+        return settings.MAX_NR_CHARACTERS
+
+    def get_available_character_slots(self) -> typing.Optional[int]:
+        """
+        Returns the number of character slots this account has available.
```
Griatch, line 727: Same here, requires a `Returns:`

evennia/accounts/accounts.py
```diff
@@ -195,6 +196,9 @@ class DefaultAccount(AccountDB, metaclass=TypeclassBase):
 
     objects = AccountManager()
 
+    # Used by account.create_character() to choose default typeclass for characters.
+    player_character_typeclass = settings.BASE_CHARACTER_TYPECLASS
```
Griatch, line 200: best rename this to `default_character_typeclass` to better reflect its purpose.

evennia/accounts/accounts.py
```diff
@@ -870,10 +925,10 @@ def create(cls, *args, **kwargs):
...
             if account and _AUTO_CREATE_CHARACTER_WITH_ACCOUNT:
                 # Auto-create a character to go with this account
-
-                character, errs = account.create_character(
-                    typeclass=kwargs.get("character_typeclass")
-                )
+                call_kwargs = {}
+                if "character_typeclass" in kwargs:
+                    call_kwargs["character_typeclass"] = kwargs["character_typeclass"]
+                character, errs = account.create_character(**call_kwargs)
```
Griatch, line 931: I don't think `account.create_character` makes use of a kwarg `character_typeclass` ... it expects `typeclass`.

evennia/commands/default/account.py
```diff
@@ -149,53 +149,20 @@ def func(self):
...
+        description = self.rhs or "This is a character."
 
-        typeclass = settings.BASE_CHARACTER_TYPECLASS
+        new_character, errors = self.account.create_character(key=key, description=description, ip=self.session.address)
 
-        if ObjectDB.objects.filter(db_typeclass_path=typeclass, db_key__iexact=key):
-            # check if this Character already exists. Note that we are only
-            # searching the base character typeclass here, not any child
-            # classes.
-            self.msg(f"|rA character named '|w{key}|r' already exists.|n")
```
Griatch, line 172: This has been removed, but I don't see this check put back in the `account.create_character` method. Would this not be good to put in `DefaultCharacter.validate_name`?

Griatch (changes requested):
This is a nice refactor I think. :+1: 

- I have some comments and questions below. 
- It would be good to take the opportunity to add unit tests 
   - Tests for the new hook methods
   - Testing that we can't create a character with a name that already exists

evennia/commands/default/account.py
volundmush: This specific part is an issue. I feel that checking for duplicates is a responsibility of .validate_name() for the Character class, but Evennia doesn't distinguish between PC and NPC for its default Character class...

Although there might be games which DO allow multiple player characters to have the same name. That sounds like it would be confusing, but should the server put such enforcement in place without an ability to turn it off?

Griatch: That's a good point. But either way, it should not be removed from the command then.

Griatch (changes requested):
These changes look good, thanks for making the changes. To repeat back to the previous review. though: It would be good to take the opportunity to add unit tests

- Tests for the new hook methods
- Testing that we can't create a character with a name that already exists

Griatch:
@volundmush Ping.

volundmush:
Some of the changes this code makes depends on the changes to the .db._playable_characters pull request, I'm not sure how to proceed until that one's looked at.

Griatch:
@volundmush Ping. The blocking PRs are all reviewed and awaiting further fixes before they can be merged.

Griatch:
@volundmush It doesn't look like the previous feedback was addressed yet :)

volundmush:
Okay I think we're looking good now.

Griatch (changes requested):
One down from the original review, one left. :) 

- Make unit test for these new methods.

volundmush:
Ooookay. I don't feel like this is complete, because the new additions lead to more questions about how naming Objects ought to work...

You want duplication checks, well, I do too. I think that should be a broad feature which can be applied to arbitrary groups of objects, though, not just a little stub in one method. The best way to group objects is Tags, and so I've got a draft of how to do that in the new commit.

Griatch (changes requested):
@volundmush I suggest _not_ adding more to this PR. Namespaces is really a separate feature request that don't belong on this PR. The previous comment was just that you removed the existing duplicate-name check and didn't put it back. 

1. Keep the existing duplicate-name check as before in this PR. If you want to make a new system adding name spaces, put that in a separate PR that can be evaluated independently. We should not bloat this PR with more functionality than it was originally created with.
2. You still haven't added the request unit tests for these changes. Before that this PR is definitely not complete.

Griatch (approved):
Looking good now, thanks for making the changes!

PR #3178, opened 2023-05-06
Improved msg formatting breakup for easier overloading.

volundmush:
This breaks up the msg_contents API on DefaultObject into several smaller pieces which can be overloaded.
Each can be called individually as well for a simpler or different use-case.

There are more scenarios where having that message formatter available conveniently would be super handy, and where you might want to overload it.

I, for instance, would love to overload the new receive_action_msg() so that additional recipients - perhaps people using magic to spy on you, or the occupants of a vehicle looking through a window - can see personalized formatted messages too.


evennia/objects/objects.py
```diff
@@ -672,6 +672,55 @@ def for_contents(self, func, exclude=None, **kwargs):
...
+                else str(obj)
+                for key, obj in mapping.items()
+            }
+        )
+
+        self.msg(text=(outmessage, outkwargs), from_obj=from_obj, **kwargs)
+
+    def send_action_msg(self, message, receiver, mapping=None, raise_funcparse_errors=False, **kwargs):
+        """
+        Convenience wrapper for sending a message as "you" to a receiver.
```
Griatch, line 713: Note that using the `$You()` (actor stance) markup is optional. So better to phrase this something like 'sending a message, supporting optional $You() markup`

evennia/objects/objects.py
```diff
@@ -672,6 +672,55 @@ def for_contents(self, func, exclude=None, **kwargs):
...
+                key: obj.get_display_name(looker=self)
+                if hasattr(obj, "get_display_name")
+                else str(obj)
+                for key, obj in mapping.items()
+            }
+        )
+
+        self.msg(text=(outmessage, outkwargs), from_obj=from_obj, **kwargs)
+
+    def send_action_msg(self, message, receiver, mapping=None, raise_funcparse_errors=False, **kwargs):
```
Griatch, line 711: This is not a useful method name; let's not introduce a new `send_` and `action` nomenclature. Since you can call this in order to send an actor-stance aware message, maybe it's best to note that in the name, like `msg_with_actor_stance`

evennia/objects/objects.py
```diff
@@ -672,6 +672,55 @@ def for_contents(self, func, exclude=None, **kwargs):
         for obj in contents:
             func(obj, **kwargs)
 
+    def receive_action_msg(self, from_obj, message, mapping, raise_funcparse_errors=False, **kwargs):
```
Griatch, line 675: The term `action` doesn't mean much. The point of this is the ability to receive a message that may (optionally) contain actor-stance $You markup. So a better name may be `receive_msg_with_actor_stance`.

evennia/objects/objects.py
```diff
@@ -672,6 +672,55 @@ def for_contents(self, func, exclude=None, **kwargs):
...
+        """
+        Receive and process an action message from another object. Refer to msg_contents for the docs.
```
Griatch, line 677: The term `action message` means nothing in context. Better to describe it as a message that may or may not contain $You markup. 
This method also needs a full doc string to describe the input arguments.

evennia/objects/objects.py
Griatch: This also needs a full docstring, like all methods.

Griatch (changes requested):
I could see this being useful for overloading. 
- See comments in code
- When adding new methods like this, we should take the opportunity to also follow up with unit tests to make sure they work as they should.

Griatch:
Ping.

Griatch:
@volundmush Ping.

Griatch:
@volundmush Ping!

volundmush:
After further thought and a really screwy rebase procedure, I think this one is best saved for some later time.

Issue #3095, opened 2023-01-25
[Feature Request] Rename `basetype_posthook_setup` to `at_post_object_creation` and adjust documentation


owllex:
#### Is your feature request related to a problem? Please describe.
When spawning objects, `basetype_posthook_setup` is the place to handle configuring objects with the spawned attributes present, since that's the only creation hook that occurs after spawned attributes are set. But its name doesn't match the hook naming convention, and as a result it's a little more mysterious than it needs to be. Its docs could be a little more clearer too about this interaction with the spawner.

Real example that spawned this issue (pardon the pun): For Items, I have a handler that stores a number of basic object properties (weight, volume, etc) in a shared dict attribute and then caches that dict locally for efficiency. In `at_object_creation`, I was setting an unrelated attribute (charges) on a Consumable typeclass, a child of Item. The charges value is stored in the same dict as weight/volume. When spawning a new Consumable, the following sequence of actions would occur:

1) Object created,
2) `at_object_creation` called, `charges` set in handler.
3) `spawner.spawn` writes weight/volume/etc attributes directly to the backing attribute dict, bypassing the handler and overwriting what was previously set (the value for charges). It clobbers the entire dict with the one in the prototype.
4) Handler's cached value is now out of sync with the DB, as the DB attribute was modified directly.
5) Desync causes non-obvious behavior later down the road.

Admittedly, none of this is evennia's fault; evennia is probably doing the right thing. But this sequence of events was surprising to me and required careful debugging. We may be able to help the next person who has an issue like this via documentation recommending performing attribute initialization in `basetype_posthook_setup` instead so that they are able to see if the spawner did anything and react accordingly. The state of an object when `at_object_creation` is called could change considerably if spawner is involved compared to if it's not, and I think this should be clearer to implementors.

#### Describe the solution you'd like
1) Rename `basetype_posthook_setup` to `at_post_object_creation`. This name matches the naming convention of other hooks and makes it clearer when it is called and its relation to the other `at_object_creation` hook.
2) Change this method's guidance that says "this should generally not be overloaded unless you are redefining, how a room/exit/object works", as I think the interaction with spawner means that there are a considerable number of cases where you'd want to put initialization of typeclass attributes in the new method instead of `at_object_creation` to avoid the spawner clobbering them (or at least to be able to react).
3) Document the object creation hook call sequence in `at_object_creation` as well for maximum transparency.

#### Describe alternatives you've considered
[...]

PR #3180, opened 2023-05-06
Building Command Hook Improvements

volundmush:
All locations in the default building commands which use create_object now call two additional hooks:
builder.can_build_object(), which can generate arbitrary errors if, for some reason, the builder CANNOT currently build that object,

and new_thing.at_object_construction(builder), which is called shortly after the object is created. As it's aware of who made it, it can check them for things like build modes, information about their current location or whatnot, and this can be used for auto-tagging, quota implementation, any number of things.

This will make building complex game grids easier and more powerful through easier implementation of complex building rules and logical grouping of created objects without needing to completely re-implement perfectly working build commands like Dig.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -203,7 +203,7 @@ class ObjManipCommand(COMMAND_DEFAULT_CLASS):
 
         try:
             type_class = (
-                class_from_module(found_typeclass)
+                class_from_module(found_typeclass, settings.TYPECLASS_PATHS)
                 if isinstance(found_typeclass, str)
                 else found_typeclass
             )
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -579,10 +579,6 @@ class CmdCreate(ObjManipCommand):
     locks = "cmd:perm(create) or perm(Builder)"
     help_category = "Building"
 
-    # lockstring of newly created objects, for easy overloading.
-    # Will be formatted with the {id} of the creating object.
-    new_obj_lockstring = "control:id({id}) or perm(Admin);delete:id({id}) or perm(Admin)"
-
     def func(self):
         """
         Creates the object.
@@ -600,26 +596,26 @@ class CmdCreate(ObjManipCommand):
             string = ""
             name = objdef["name"]
             aliases = objdef["aliases"]
-            typeclass = objdef["option"]
 
-            # create object (if not a valid typeclass, the default
-            # object typeclass will automatically be used)
-            lockstring = self.new_obj_lockstring.format(id=caller.id)
-            if (err := caller.can_build_object()):
-                caller.msg(err)
-                return
-            obj = create.create_object(
-                typeclass,
+            obj_typeclass, errors = caller.get_object_typeclass(obj_type="object", typeclass=objdef["option"])
+            if errors:
+                self.msg(errors)
+            if not obj_typeclass:
+                continue
+
+            obj, errors = obj_typeclass.create(
                 name,
                 caller,
                 home=caller,
                 aliases=aliases,
-                locks=lockstring,
                 report_to=caller,
+                creator=caller
             )
+            if errors:
+                self.msg(errors)
             if not obj:
                 continue
-            obj.at_object_constructed(caller)
+
             if aliases:
                 string = (
                     f"You create a new {obj.typename}: {obj.name} (aliases: {', '.join(aliases)})."
@@ -928,25 +924,27 @@ class CmdDig(ObjManipCommand):
         location = caller.location
 
         # Create the new room
-        typeclass = room["option"]
-        if not typeclass:
-            typeclass = settings.BASE_ROOM_TYPECLASS
+        room_typeclass, errors = caller.get_object_typeclass(obj_type="room", typeclass=room["option"], method="dig")
+        if errors:
+            self.msg("|rError creating room:|n %s" % errors)
+        if not room_typeclass:
+            return
 
         # create room
-        if (err := caller.can_build_object()):
-            caller.msg(err)
-            return
-        new_room = create.create_object(
-            typeclass, room["name"], aliases=room["aliases"], report_to=caller
+        new_room, errors = room_typeclass.create(
+            room["name"], aliases=room["aliases"], report_to=caller, creator=caller, method="dig"
         )
-        lockstring = self.new_room_lockstring.format(id=caller.id)
-        new_room.locks.add(lockstring)
+        if errors:
+            self.msg("|rError creating room:|n %s" % errors)
+        if not new_room:
+            return
+
         alias_string = ""
         if new_room.aliases.all():
             alias_string = " (%s)" % ", ".join(new_room.aliases.all())
-        new_room.at_object_constructed(caller)
+
         room_string = (
-            f"Created room {new_room}({new_room.dbref}){alias_string} of type {typeclass}."
+            f"Created room {new_room}({new_room.dbref}){alias_string} of type {new_room}."
         )
 
         # create exit to room
@@ -962,21 +960,27 @@ class CmdDig(ObjManipCommand):
                 exit_to_string = "\nYou cannot create an exit from a None-location."
             else:
                 # Build the exit to the new room from the current one
-                typeclass = to_exit["option"]
-                if not typeclass:
-                    typeclass = settings.BASE_EXIT_TYPECLASS
-                if (err := caller.can_build_object()):
-                    caller.msg(err)
+                exit_typeclass, errors = caller.get_object_typeclass(obj_type="exit", typeclass=to_exit["option"],
+                                                                     method="dig")
+                if errors:
+                    self.msg("|rError creating exit:|n %s" % errors)
+                if not exit_typeclass:
                     return
-                new_to_exit = create.create_object(
-                    typeclass,
+
+                new_to_exit, errors = exit_typeclass.create(
                     to_exit["name"],
-                    location,
-                    aliases=to_exit["aliases"],
-                    locks=lockstring,
+                    location=location,
                     destination=new_room,
+                    aliases=to_exit["aliases"],
                     report_to=caller,
+                    creator=caller,
+                    method="dig"
                 )
+                if errors:
+                    self.msg("|rError creating exit:|n %s" % errors)
+                if not new_to_exit:
+                    return
+
                 alias_string = ""
                 if new_to_exit.aliases.all():
                     alias_string = " (%s)" % ", ".join(new_to_exit.aliases.all())
@@ -984,7 +988,6 @@ class CmdDig(ObjManipCommand):
                     f"\nCreated Exit from {location.name} to {new_room.name}:"
                     f" {new_to_exit}({new_to_exit.dbref}){alias_string}."
                 )
-                new_to_exit.at_object_constructed(caller)
 
         # Create exit back from new room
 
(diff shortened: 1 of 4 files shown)


evennia/objects/objects.py
```diff
@@ -1606,6 +1606,38 @@ def at_cmdset_get(self, **kwargs):
...
+        they can add to their custom dungeon.
+
+        Returns:
+            can_build (str or None): If self is allowed to build objects.
+                return a string as the error if there is a problem. If
+                this returns True, the build will abort.
+        """
+        pass
+
+    def at_object_constructed(self, builder):
```
Griatch, line 1628: What's with these inconsistent method names :grin: 

I think the comments to [this ticket](https://github.com/evennia/evennia/issues/3095) may be a better approach here - we add a new `at_post_object_creation_hook(*args, **kwargs)` that is called differently depending on how the entity was created (such as by the spawner or by the build command). You could then have the building command pass the `caller` into it as a kwarg to do with as you want.

Griatch (changes requested):
It's a good idea, but I disagree with the need for the first hook. And I think it's better to go with the approach of [this ticket](https://github.com/evennia/evennia/issues/3095) for the second hook - that would make it more generic than just for handling the single building command.

volundmush:
I'm not so sure I agree with your idea about the rename to use at_post_object_creation in regards to that feature request, because, from what I'm reading of that feature request.

The reason being is, owllex is referring to hooks called automatically when they're handled by the spawner or normal creation, for any and all objects, and the inconsistencies caused by that interaction. My concern is purely with objects being created in accordance with "building rules" such as being automatically added to the right zone or deducting from the builder's quota. These are both part of creating an object but are entirely separate concerns.

What this is telling me is that maybe we have too many pipelines for creating objects and the way that the hooks into that pipeline are arranged isn't quite right.

Griatch:
> I'm not so sure I agree with your idea about the rename to use at_post_object_creation in regards to that feature request.
> 
> The reason being is, owllex is referring to hooks called automatically when they're handled by the spawner or normal creation, for any and all objects, and the inconsistencies caused by that interaction. My concern is purely with objects being created in accordance with "building rules" such as being automatically added to the right zone or deducting from the builder's quota. These are both part of creating an object but are entirely separate concerns.
> 
> What this is telling me is that maybe we have too many pipelines for creating objects and the way that the hooks into that pipeline are arranged isn't quite right.

There is one main pipe for creating objects, that is `evennia.utils.create.create_object`. This is ObjectDB-Typeclass agnostic. The `Typeclass.create` methods are wrappers to conveniently create entities of that particular typeclass (which also makes them useful for web-based creation); under the hood they all just call `create_object`, just with different arguments. 

What you are suggesting is to add a separate hook only for creating via a build command, so that'd be yet another 'pipeline' in that case, no? 

I still think having a central post-create hook that can take varying inputs would be the better solution here.

Griatch:
@volundmush Ping.

Griatch:
@volundmush Another one that needs a rebase due to the upstream repo changes.

volundmush:
Sink your teeth into these changes. What do you think about this redesign?

evennia/commands/default/building.py
```diff
@@ -599,22 +595,26 @@ def func(self):
...
-            lockstring = self.new_obj_lockstring.format(id=caller.id)
-            obj = create.create_object(
-                typeclass,
+            obj_typeclass, errors = caller.get_object_typeclass(obj_type="object", typeclass=objdef["option"])
+            if errors:
+                self.msg(errors)
+            if not obj_typeclass:
+                continue
+
+            obj, errors = obj_typeclass.create(
```
Griatch, line 605: Nice! :+1:

evennia/commands/default/building.py
```diff
@@ -599,22 +595,26 @@ def func(self):
...
                 name,
                 caller,
                 home=caller,
                 aliases=aliases,
-                locks=lockstring,
                 report_to=caller,
+                creator=caller
```
Griatch, line 611: It seems suitable to have a method of 'cmd_create` here. There are many reasons to call this creation method.

evennia/commands/default/building.py
```diff
@@ -923,21 +923,27 @@ def func(self):
...
+        if errors:
+            self.msg("|rError creating room:|n %s" % errors)
+        if not room_typeclass:
+            return
 
         # create room
-        new_room = create.create_object(
-            typeclass, room["name"], aliases=room["aliases"], report_to=caller
+        new_room, errors = room_typeclass.create(
+            room["name"], aliases=room["aliases"], report_to=caller, creator=caller, method="dig"
```
Griatch, line 934: Also worth, to call this argument `caller` too, for consistency. Suggest to use "cmd_dig` here to be exclicit.

evennia/commands/default/building.py
```diff
@@ -984,18 +998,25 @@ def func(self):
...
                     back_exit["name"],
-                    new_room,
-                    aliases=back_exit["aliases"],
-                    locks=lockstring,
+                    location=new_room,
                     destination=location,
+                    aliases=back_exit["aliases"],
                     report_to=caller,
+                    creator=caller,
+                    method="dig"
```
Griatch, line 1014: I suggest "cmd_dig" here.

evennia/commands/default/building.py
```diff
@@ -1473,17 +1494,23 @@ def create_exit(self, exit_name, location, destination, exit_aliases=None, typec
...
+            if not exit_typeclass:
+                return
+            exit_obj, errors = exit_typeclass.create(
+                exit_name,
                 location=location,
                 aliases=exit_aliases,
-                locks=lockstring,
                 report_to=caller,
+                creator=caller,
+                method="open"
```
Griatch, line 1508: I suggest "cmd_open" here.

evennia/objects/objects.py
```diff
@@ -311,22 +319,22 @@ def exits(self):
...
+            attribute_name=None,
+            quiet=False,
+            exact=False,
+            candidates=None,
+            use_locks=True,
+            nofound_string=None,
+            multimatch_string=None,
+            use_dbref=None,
+            tags=None,
+            stacked=0,
```
Griatch, line 337: Not sure which auto-formatter you use here, but 8 character indent seems to not be `black` standard.

evennia/objects/objects.py
```diff
@@ -886,7 +894,7 @@ def logerr(string="", err=None):
             # check if source location lets us go
             try:
                 if source_location and not source_location.at_pre_object_leave(
-                    self, destination, **kwargs
+                        self, destination, **kwargs
```
Griatch, line 897: Why 8 character indent here, four should be enough... no big deal though; we'll run `black` on it ... but it's a bit weird since it creates so big diffs.

evennia/objects/objects.py
```diff
@@ -1011,7 +1019,8 @@ def clear_contents(self):
             obj.move_to(home, move_type="teleport")
 
     @classmethod
-    def create(cls, key, account=None, **kwargs):
+    def create(cls, key: str, account: "DefaultAccount" = None, creator: "DefaultObject" = None, method: str = "create",
+               **kwargs):
```
Griatch, line 1023: Naming is hard. 
- To be in line with other uses, I suggest using `caller` instead of `creator` here; we don't use something like a `creater ` anywhere else in Evennia. 
- I'm ambivalent on `method`. I suppose you may be inspired by `GET/POST/etc` methods here? I wonder if `source` may be a more general name, since this can also be called by the web component's views, but not sure :thinking:

evennia/objects/objects.py
```diff
@@ -1597,6 +1609,43 @@ def at_cmdset_get(self, **kwargs):
         """
         pass
 
+    def get_object_typeclass(self, obj_type: str = "object", typeclass: str = None, method: str = "create", **kwargs) -> tuple[
```
Griatch, line 1612: I don't think this makes much sense to have on the typeclass like this. 
- You are accessing a method _on_ a typeclass instance in order to ... get another typeclass instance. While intended for build commands, this method's name reflects nothing of the sort; it's named as a generic method for getting typeclasses. 
- Moreover, you are intending to use this on the caller (which is not clear from this docstring), essentially making this worthless, or at least with no clear usage on other typeclassed entities. 

This really doesn't belong here. It should go in the build command parent; you lose no flexibility this way compared to the current setup since you'd only customize this on the `caller` typeclass anyway.

evennia/objects/objects.py
```diff
@@ -2986,7 +3037,8 @@ def create_exit_cmdset(self, exidbobj):
     # Command hooks
 
     @classmethod
-    def create(cls, key, source, dest, account=None, **kwargs):
+    def create(cls, key: str, location: DefaultRoom = None, destination: DefaultRoom = None, account: "DefaultAccount" = None, creator: DefaultObject = None,
```
Griatch, line 3040: Use `caller` instead of `creator` here.

Good call on changing source/dest to location/destination; that does read better. :+1:

Griatch (changes requested):
Overall, I like these changes. 
- Naming changes
- I don't agree with the location of the `get_object_typeclass` method
- It would be good with some unit tests of these new/changed methods.

evennia/objects/objects.py
volundmush: Putting it in the build command parent makes it far harder to overload without then needing to re-create subclasses of all of the build command children... I am very much attempting to extricate this logic out of the commands and onto something independent of the commands.

Putting it on the object that's doing the building made sense, because that would have access to its location and other relevant information about who's doing the building...

evennia/objects/objects.py
volundmush: I can think of a few cases where the OWNER of an object might not be its creator, but... okay, will use caller, for now.

You're right on the odd naming choice for method, I'm not sure what else to call the darned thing. The hope is to make information about what sort of building is going on available to mechanisms that can make further decisions, and do more checks relevant to a game... I'm not sure how useful it is right this very moment, but I figured it may be useful to SOMEONE so it's better to have it than not...

volundmush:
Looks like there already are tests for the building commands... what specifically do you want tested if not those?

evennia/objects/objects.py
InspectorCaracal: I think `source` probably works well, since the idea of it seems to be "where is this create call coming from", but if it doesn't actually have a specific purpose atm and is more of a "someone might want it someday" *and* you aren't even sure what to call it because the idea is too nebulous for a specific term, then maybe it's best kept in `kwargs`?

evennia/objects/objects.py
Griatch: We have entirely new methods here, like `get_object_typeclass` that has no unit tests on it.

evennia/objects/objects.py
Griatch: I don't really mind having the ability to customize this, but yeah, maybe `source` could be a better name.

evennia/commands/default/building.py
```diff
@@ -164,6 +173,48 @@ def parse(self):
         self.lhs_objattr = obj_attrs[0]
         self.rhs_objattr = obj_attrs[1]
 
+    def get_object_typeclass(
```
Griatch, line 176: Cool, this location makes more sense to me.

Griatch (changes requested):
Thanks for your changes! I think this is getting there. 

1. The Evennia unit tests don't seem to pass. 
2. I'd prefer to have a specific unit test for the new `get_object_typeclass` method, testing it. But I suppose it's correct that the build commands have unit tests already, so this code is covered second-hand through those tests, so I'll not push hard for it, now that the method sits on the command parent.

volundmush:
So I realized that the use of default_typeclasses on the ObjManipCommand was likely bypassing the patched BASE_OBJECT_TYPECLASS. To fix that... well, I didn't. Instead I altered the test to explicitly pick the Typeclass it's gonna use.

One of the tests concerned the BatchProcessor. It was trying to spawn a button and then delete it. Except, it couldn't delete it, because the example batch command was teleporting the button away from self.char1 (which is in self.room1), so the destroy command couldn't find it. I created an altered test batch script that doesn't teleport it and modified the command to destroy/force so it will definitely achieve what it wants.

The final fix was to ensure that class_from_module in get_object_typeclass was targeting settings.TYPECLASS_PATHS properly.

All seems to be well now.

Griatch (approved):
Thanks for the changes, looks clean now.

PR #3184, opened 2023-05-10
Minor refactor for TickerHandler.save()

InspectorCaracal:
This rewrites the .save() method on the TickerHandler to be a little bit easier to read, since referencing the store_key values by index is less than ideal.

When attempting to track down the cause for some ticker issues a couple of us have had, I discovered that the save method is retrieving `store_key[1]` as the interval and using it as such - however, `store_key[1]` is actually the stored method.

This change corrects for that and helps prevent further related issues in one fell swoop.


Griatch (approved):
This does look a lot more readable and cleaner than referencing indices.

PR #3185, opened 2023-05-11
WIP: Moving ANSI codes from the game server to the portal

InspectorCaracal:
This is an active draft/WIP implementing some code changes I've made for my own game into core. The changes are focused on removing ANSIString and preserving the evennia style flags syntax in messaging until the text reaches the actual protocols.

At the moment, cursory manual testing shows that it works correctly, but it's not yet anywhere near a thorough treatment. I also haven't begun implementing any of the required changes in unit testing.

The goal is to convert all of the server-side formatting to be entirely protocol agnostic and *only* use the defined Evennia style codes, such as `|r`, and have those codes converted into protocol-friendly forms from the actual protocol itself.

At the moment, this also includes a few changes in approach to converting colors for the webclient protocol, although it doesn't touch the webclient itself at all. If this PR makes it to the stage of being fully converted and potentially considered for merging, I'll open up discussion on whether to keep those changes along with my justifications for them. (They're in this code because they're in my code and I largely just copy/pasted things.)

### DIFF
--- a/evennia/commands/command.py
+++ b/evennia/commands/command.py
@@ -644,21 +644,21 @@ Command {self} has no defined `func()` - showing on-command variables:
         colornames = ["|%s%s|n" % (column_color, col) for col in args]
 
         h_line_char = kwargs.pop("header_line_char", "~")
-        header_line_char = ANSIString(f"|{border_color}{h_line_char}|n")
+        header_line_char = f"|{border_color}{h_line_char}|n"
         c_char = kwargs.pop("corner_char", "+")
-        corner_char = ANSIString(f"|{border_color}{c_char}|n")
+        corner_char = f"|{border_color}{c_char}|n"
 
         b_left_char = kwargs.pop("border_left_char", "||")
-        border_left_char = ANSIString(f"|{border_color}{b_left_char}|n")
+        border_left_char = f"|{border_color}{b_left_char}|n"
 
         b_right_char = kwargs.pop("border_right_char", "||")
-        border_right_char = ANSIString(f"|{border_color}{b_right_char}|n")
+        border_right_char = f"|{border_color}{b_right_char}|n"
 
         b_bottom_char = kwargs.pop("border_bottom_char", "-")
-        border_bottom_char = ANSIString(f"|{border_color}{b_bottom_char}|n")
+        border_bottom_char = f"|{border_color}{b_bottom_char}|n"
 
         b_top_char = kwargs.pop("border_top_char", "-")
-        border_top_char = ANSIString(f"|{border_color}{b_top_char}|n")
+        border_top_char = f"|{border_color}{b_top_char}|n"
 
         table = EvTable(
             *colornames,
@@ -700,8 +700,8 @@ Command {self} has no defined `func()` - showing on-command variables:
 
         colors = dict()
         colors["border"] = self.account.options.get("border_color")
-        colors["headertext"] = self.account.options.get("%s_text_color" % mode)
-        colors["headerstar"] = self.account.options.get("%s_star_color" % mode)
+        colors["headertext"] = self.account.options.get(f"{mode}_text_color")
+        colors["headerstar"] = self.account.options.get(f"{mode}_star_color")
 
         width = width or self.client_width()
         if edge_character:
@@ -709,20 +709,17 @@ Command {self} has no defined `func()` - showing on-command variables:
 
         if header_text:
             if color_header:
-                header_text = ANSIString(header_text).clean()
-                header_text = ANSIString("|n|%s%s|n" % (colors["headertext"], header_text))
+                header_text = "|n|{}{}|n".format(colors["headertext"], header_text)
             if mode == "header":
-                begin_center = ANSIString(
-                    "|n|%s<|%s* |n" % (colors["border"], colors["headerstar"])
-                )
-                end_center = ANSIString("|n |%s*|%s>|n" % (colors["headerstar"], colors["border"]))
-                center_string = ANSIString(begin_center + header_text + end_center)
+                begin_center = "|n|{}<|{}* |n".format(colors["border"], colors["headerstar"])
+                end_center = "|n |{}*|{}>|n".format(colors["headerstar"], colors["border"])
+                center_string = begin_center + header_text + end_center
             else:
-                center_string = ANSIString("|n |%s%s |n" % (colors["headertext"], header_text))
+                center_string = "|n |{}{} |n".format(colors["headertext"], header_text)
         else:
             center_string = ""
 
-        fill_character = self.account.options.get("%s_fill" % mode)
+        fill_character = self.account.options.get(f"{mode}_fill")
 
         remain_fill = width - len(center_string)
         if remain_fill % 2 == 0:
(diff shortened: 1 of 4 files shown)


Griatch (commented):
First of all, cool that you are working doing this! I know this is a WIP, so just giving some general thoughts;

I think the problem overall here is that we go via ANSI at all. There needs to be a separate parser not called `ANSIParser` but something like `InlineTagParser` that only deals with handling `|-` style tagging (or whatever other formats we will support in the future - the difference is that this would never go via ANSI at all, but just deal with the raw tags. The ANSIParser would only sit by the TelnetProtocol and convert `|-` style to Ansi markup. 

As I've discussed with Volund too; maybe something ANSIString-like should remain under a different name - a string type that deals with `|-` markup but that can also hold other data types (like tables/menus?)  to send through `msg()`.

InspectorCaracal:
> I think the problem overall here is that we go via ANSI at all. There needs to be a separate parser not called `ANSIParser` but something like `InlineTagParser` that only deals with handling `|-` style tagging (or whatever other formats we will support in the future - the difference is that this would never go via ANSI at all, but just deal with the raw tags. The ANSIParser would only sit by the TelnetProtocol and convert `|-` style to Ansi markup.


Yeah! This is exactly my thoughts/goal on it as well. It makes way more sense to have everything be in a consistent editable format on the server and then be converted to what the specific protocol needs right in the protocol.

> As I've discussed with Volund too; maybe something ANSIString-like should remain under a different name - a string type that deals with `|-` markup but that can also hold other data types (like tables/menus?) to send through `msg()`.

That's a really interesting idea... 🤔 So the idea is that the format classes would stay class objects until reaching the protocol? Would this custom string replacement be able to contain an arbitrary collection of strings and tables in one object, or would it just provide an interface for stitching them together?

That also makes me think that converting table objects to a string output should be done in the protocol parser, not as part of the the EvTable class, so that it can be rendered protocol-specifically. That would open up the opportunity of using actual html table tags on the webclient, as well.

InspectorCaracal:
Some idle thoughts on the topic from thinking through the suggestions:

I could implement some kind of foundational `EvString` class which supports all basic string functions like replace and concatenation, storing the string internally as chunks delineated by evennia format tags. `EvString` itself can support basic sizing information, effectively replacing the fundamental data element of `EvCell`, meaning it could juts be a subclass of `EvString` that supports borders. Neither `EvTable` nor `EvString` would implement their own rendering (aside from basic repr strings) but instead would be handled directly by the protocol-level parser.

However... laying that out in text, I'm not fully convinced that having a custom `EvString` class has much of an advantage over simply removing `ANSIString`, since both approaches would still offload the final formatting to the parser. I think personally I prefer maintaining using plain strings, since the strings themselves don't need to know how to be formatted for the client - that is all still specific to the protocol layer.

For example, all of the calculations ANSIString regarding display vs real length and how to break text for wrapping and the like are crucial for formatting text going through telnet, since the simple text stream requires you to have all that formatting done explicitly by adding the right number of characters in the right places. Evennia has quite a lot of useful code for doing this, integrated pretty thoroughly into all of its various formatting tools.

But while all of that **is** extremely useful for telnet, it's also completely unnecessary for the webclient. Instead, it would be better to take advantage of the fact that web browsers are designed to do that visual rendering *for* you, using HTML and CSS to define the desired width and spacing without needing to measure the content at all. So I think I would need to be convinced that some sort of `EvString` object would be more generically useful than just keeping the strings as strings until reaching the protocol parser, and at the moment I'm not.

(I *do* still think having EvTable objects able to pass directly through `msg` all the way to the protocol and having explicit table parsing on the protocol parser would be great; I might give that a shot.)

Griatch:
@InspectorCaracal  
(this is pretty long, so TLDR at the end ;) )

I do agree that if one were able to just have plain old Python strings to pass around and not _have_ to convert to something like an `EvString` (or today `ANSIString`) format, it would be nicer for the programmer. So as long as one would have an expressive-enough formatting language (if using Rich, this would be something like our current `|-` language in combination with `bbcode`), one could just pass these around as strings; that's fine. In the simplest case, one would then just send this to the portal and it would be converted at the protocol level to ANSI for telnet and HTML/CSS for webclient etc. 

We still need something like `ANSIString` (I like `EvString`, why not), but in that case purely as a utlity tool rather than as a standard to pass around. Having a string-like entity that correctly ignores formatting markup for length calculations just makes your code so much easier. EvTable is only one example, there are other cases where you will want to place text at particular distances without wanting to consider colors etc. This utility should just not go via `ANSI` like `ANSIString` does, but deal with raw formatting markers (making it a lot easier to implement, probably). 

In that case, I like the idea of EvTable/EvForm and maybe other structures being possible to be sent directly to `msg()`. I think we need to make a new common class for that kind of sendable entity. So I'd suggest that we change what is being sent between the Portal and Server into a new `OutputFunc` class. 

This is a container that knows how to de/serialize itself, and which has the needed methods that the protocols will need to extract the data to send over the wire. An `EvTable` and `EvForm` would _be instances of this class_ and implement table/form-specific methods. 

So the change to `msg` would be that when you send e.g. `msg(text=(args), {kwargs})`, those args/kwargs could be either raw strings (with formatting markers) _or_ `OutputFunc`entities with more metadata for things like EvTable or EvForm. The portal protocol would look at the args/kwargs coming in and either run a direcet-to-wire converter on the raw strings, or use the extra meta-info from the `OutputFunc` entities to better represent that type of data for that protocol (such as building a HTML table for the webclient). 

I can see two ways to handle `OutputFunc` entities on the protocol level: 
1. The `OutputFunc` class is agnostic, and has methods like `.get_rows` or `.get_columns` for EvTable. Each protocol knows how to deal with and `EvTable` and uses the methods to get the relevant data to send on the wire (e.g. like a HTML table to the webclient or MSDP table in telnet, if supported). This has the advantage of separation of concern - all protocol info is in the portal, none is known on the server side. For customization, one would need two steps: a) Making a new subclass of `OutputFunc` and b) Registering in a new handler for this subclass on the Portal side.
2.  Each `OutputFunc` class provides `to_<protocol>` methods for converting itself to the wire. So `.to_telnet` and `to_webclient` respectively. The advantage is that this is pretty straightforward to override/customize (only one step compared to two above). The drawback is that this stuff is now on the server side, making this less 'clean' when it comes to separation of concerns. 

TLDR: 
- I like the idea of passing raw Python strings (containing formatting markers) around as much as possible. 
- We still need an `ANSIString` analogue (but without ANSI) for convenient string operations with formatting markers.
- I suggest a new `OutputFunc` parent class for `EvTable/EvForm` and similar, for serializing and passing more metadata to help the protocol send more relevant data over the wire. Exactly how to organize this is still up in the air.

InspectorCaracal:
@Griatch

I'm going to have to read that all a few times to digest it, but you make a lot of good points and I like the ideas! I'll think about it some more, as well.

InspectorCaracal:
Two thoughts:

- I think that the replacement of `ANSIString` with a more generic, evennia-flavored, protocol-agnostic `EvString` would be within the scope of this PR's intent, so I'll see if that looks like something I can do.
- It would, to me, make sense for this `EvString` to share the same proposed parent class as `EvTable` and `EvForm`, since all three of them would be custom objects with special formatting information attached that needs to be parsed into something protocol-specific when being sent.

Griatch:
@InspectorCaracal I think something like EvString needs to have as a requisite a separate `|-` syntax parser module that replaces  `ansi.py` and is also extendible in the future (with Rich or similar). This is a pretty big thing but if you want to take this on as part of this PR, feel free. I'm fine with considering this PR blocked until that is in place, otherwise.
Having EvString and container objects for EvString, EvTable, EvForm all share parent is a good idea I think.

InspectorCaracal:
Based on the discussion here, I decided to scrap this branch and make a new one building on the EvString idea, since there's very little overlapping code between the two. So I'm closing this PR and will create a new draft PR on the new branch referencing it.

PR #3186, opened 2023-05-11
Adding client-side flags for supporting TRUECOLOR and extending MTTS...

volundmush:

A few minor changes to the portal-side protocols so that they can detect TRUECOLOR support for the few clients that do support it, as well as the rest of the expanded MTTS options.

Now the protocol_flags are more standardized for color detecting routines, and the Telnet protocol gathers more information from the client. We might not be USING that information just yet, but we're gathering it now.

It is assumed that the WebClient and SSH both support ANSI, XTERM256, and TRUECOLOR.
I am not sure if this is actually true, but it seems like a reasonable assumption.

Not that it will matter in the short term, since ANSIString isn't capable of generating truecolor codes (yet.)

### DIFF
--- a/evennia/server/portal/ssh.py
+++ b/evennia/server/portal/ssh.py
@@ -127,6 +127,10 @@ class SshProtocol(Manhole, _BASE_SESSION_CLASS):
         self.width = width
         self.height = height
 
+        # Set color defaults
+        for color in ("ANSI", "XTERM256", "TRUECOLOR"):
+            self.protocol_flags[color] = True
+
         # initialize the session
         client_address = self.getClientAddress()
         client_address = client_address.host if client_address else None
--- a/evennia/server/portal/ttype.py
+++ b/evennia/server/portal/ttype.py
@@ -19,6 +19,10 @@ SEND = bytes([1])  # b"\x01"
 
 # terminal capabilities and their codes
 MTTS = [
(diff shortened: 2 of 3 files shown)


Griatch (commented):
While cool, what would be the effect server-side of activating this? Just a different flag for activating colors? 

I thought true colors was an actual feature standard (supported by very few clients). This seems like it's supporting it without having anything on the server side to ... support it?

volundmush:
> 

All it does is add more data to session.PROTOCOL_FLAGS noting that the client supports true color. It doesn't affect any current server logic at all.

To clarify, during MTTS negotiation, the client sends an integer which is bit-checked with each bit representing a different feature. Evennia hasn't been checking for the bits of certain features the client could be advertising, so I just added it to the bit map so it will see if the client is advertising support for the feature or not.

It doesn't really change anything the server is doing apart from storing the data to PROTOCOL_FLAGS like everything else.

Griatch (approved):
I suppose it's a bit weird to me that we would be checking for this flag when we are not actually supporting something like `#rrggbb` hex colors on the Evennia side. 

I can picture our users wondering why `TRUECOLOR` is in our session info when we are not supporting `TRUECOLOR`. I suppose if a client reports `TRUECOLOR` and not any other color flag, this still has an effect though, so merging.

PR #3187, opened 2023-05-12
Changed importing of SESSION_HANDLER to avoid circular imports when e…

volundmush:
When trying to extend PortalSessionHandler or ServerSessionHandler it's easy to fall afoul of a circular import issue due to them attempting to instantiate when the module loads by pulling straight from settings. That has been changed so now evennia.SESSION_HANDLER is referred to for the active session handler instead of a weird import to a module variable that might be none.

While attempting to create my own PortalSessionHandler, I ran into a circular import issue related to how I was trying to inherit from the same .py module that wants to import the thing I'm inheriting to...

I had to make extensive changes to how things were reaching the session_handler to make this work. I think the result is cleaner and shouldn't be any less performant, though.

### DIFF
--- a/evennia/__init__.py
+++ b/evennia/__init__.py
@@ -136,7 +136,7 @@ __version__ = _create_version()
 del _create_version
 
 
-def _init():
+def _init(portal_mode=False):
     """
     This function is called automatically by the launcher only after
     Evennia has fully initialized all its models. It sets up the API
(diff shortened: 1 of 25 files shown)


Griatch (approved):
This is a nice refactor, looks much cleaner and since it's referencing the sessionhandler indirectly rather than in the top of the module I agree it should be less prone to circular import problems. Nice job!

PR #3188, opened 2023-05-12
Py adjustment

volundmush:
Just adding to the options dictionary for outgoing Py output so that it can be specially highlighted by an eventual Rich usage.

We eventually want to move to Rich. I want to use Rich now in my Project, and this change breaks nothing.


Griatch (approved):
Should be fine, and this has some nice overall cleanups too. It _is_ a bit iffy to have an unused keyword in the output though, so I'll add some more documentation to that after merge.

PR #3192, opened 2023-05-13
BugFix: Restore dynamic class kwarg assignment to goldenlayout clients

friarzen:

In the default_output.js there is a feature to allow a passed-in kwarg to define what html class to assign to the resulting div.
This feature was not included in the processing that Goldenlayout performs.

This Pull Request allows Goldenlayout to use the same 'cls' kwarg to customize each message with a different html style.


Griatch (approved):
Thanks for the fix. I wonder if there should be a documentation update as well?

PR #3197, opened 2023-05-23
Alter Global Scripts handling

volundmush:
Currently, Global Scripts are started at the end of evennia._init(), which is called by both the portal and the server. The GlobalScriptContainer is also having trouble distinguishing "scripts defined in its settings.py section" from "ALL Scripts which have no db_obj attached" and tries to start them all...

Not only that, but the scripts are starting before at_server_init() and all startup hooks are called, so they might fire off before the game has finished loading critical resources.

Even if it's running on the portal.

To combat this, I moved the GLOBAL_SCRIPTS.start() to the very end of the server startup process, after calling startup hooks.

The above is definitely not something we want happening.

### DIFF
--- a/evennia/__init__.py
+++ b/evennia/__init__.py
@@ -106,6 +106,9 @@ MONITOR_HANDLER = None
 GLOBAL_SCRIPTS = None
 OPTION_CLASSES = None
 
+# variables
+PORTAL_MODE = False
+
 
 def _create_version():
     """
@@ -160,6 +163,8 @@ def _init(portal_mode=False):
     global EvMenu, EvTable, EvForm, EvMore, EvEditor
     global ANSIString, FuncParser
     global AttributeProperty, TagProperty, TagCategoryProperty
+    global PORTAL_MODE
+    PORTAL_MODE = portal_mode
 
     # Parent typeclasses
     # utilities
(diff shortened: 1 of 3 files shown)


Griatch (changes requested):
Overall, I think this looks fine. The unit tests don't pass though.

Griatch:
@volundmush Fixing the unit tests should be what remains for this PR.

Griatch:
@volundmush Ping. This and the other PRs you made often require rather small additions/unit test fixes.

volundmush:
It looks like blongden got to this first - sorry I lost my momentum there. This PR has been superseded.

Griatch:
@volundmush Not sure about that ... Blongden made a fix to make sure global scripts does not start up in the portal; I think this PR covers a different case.

blongden:
> @volundmush Not sure about that ... Blongden made a fix to make sure global scripts does not start up in the portal; I think this PR covers a different case.

Agreed. This incorporates a nicer fix to the same issue, moving the scripts start into the server rather than keeping it in init.

Griatch:
@volundmush Is this ready to re-review?

Griatch (approved):
LGTM! And unit tests pass now too :+1:

PR #3203, opened 2023-06-07
Remove obsolete Windows install note

InspectorCaracal:
`twisted-iocpsupport` now has an updated pip package supporting python 3.11, so the build tools are no longer necessary. As such, I've removed the note about the SDK requirement.


Griatch:
@InspectorCaracal Hm, https://pypi.org/project/twisted-iocpsupport/ only lists 3.10 in its metadata. I see there is a commit that mentions 3.11 support, but it's not clear this actually resolves all 3.11 issues. Where do you see that 3.11 is now fully supported?

InspectorCaracal:
I didn't read any official support statement per se, aside from a closed issue on twisted indicating that the latest package of iocpsupport works with 3.11 and pointing at the relevant tracking there. However, someone new in discord confirmed it works now by installing on Windows with 3.11 without needing build tools.

Griatch (approved):
Let's clean it up, I also saw the successful install mentioned in discord. If people are still having trouble we can always extend the troubleshooting docs again.

Issue #3239, opened 2023-07-22
[BUG] Default Channels from settings.py do not Auto-Sub


jaborsh:
#### Describe the bug
The default channels provided by `settings.py` do not appear to properly function. They are created at reload, but not auto-subbed for subsequent created users.

#### To Reproduce
1. Define new channel in `settings.py`
2. Reload
3. Create new character
4. `channel/all` should yield something like the picture attached.
![image](https://github.com/evennia/evennia/assets/73198594/6b670e17-758f-4f40-b2b9-d7f4640a2069)

#### Expected behavior
Newly created characters (or all characters, really) should auto-sub to default channels as is stated by the Evennia Docs.

#### Environment, Evennia version, OS etc
Version 2.1.0

#### Additional context
Might have something to do with MULTISESSION settings (baseless accusation; haven't actually looked). If it's relevant, I'm using the default: one account to one character.

jaborsh:
Fixed by #3246

PR #3246, opened 2023-07-31
Bug Fix: Autosub to all channels instead of just first

jaborsh:
This PR resolves issue #3239. Original account implementation did not auto-sub accounts to default channels, only the very first one.

Documentation and intuition would lead one to believe that all characters automatically sub to the channels listed in DEFAULT_CHANNELS

Imports were not changed. Linter made those changes.


Griatch:
Thanks for the PR! This is not quite working as it should though - this will add new users to _all_ channels they have listen-access to, which could be a huge number (considering players can also create channels). Only the channels in the DEFAULT_CHANNELS settings should be auto-joined. This code change also needs to be done in more than one place. 

So I'm closing this and will fix this in a slightly different way. But thanks a lot for putting in the effort to help!  :)

PR #3252, opened 2023-08-07
Point to default settings on documentation site instead of evennia install

InspectorCaracal:
Changes the auto-generated `settings.py` docstring so it has a link to the Default Settings page of the documentation site.

The current settings file is generated with a line pointing to the location of the actual parent settings file on the computer you created the gamedir on. Since the recommended install method now is `pip` and it's discouraged to go digging around in the core library files, and since we have such a nice documentation website, that approach seems obsolete.


Griatch (approved):
I see your point; I suppose this looks kinda strange on a pip install :thinking:

PR #3253, opened 2023-08-08
Rework ANSIString to more generic EvString

InspectorCaracal:
This will replace the telnet-focused `ANSIString` with an equivalent `EvString` built around Evennia markup.

As of this initial PR, the `EvString` class parses around Evennia markup rather than ANSI codes, and has methods for converting its data to the separate protocols (at the moment `ansi` and `html`). Those methods are then used in their respective protocol modules to convert the markup on sending.

The further goal is to integrate that approach and the `EvString` class into the other evennia formatting classes such as `EvTable`, but I haven't yet reviewed them. The tests have also not been updated yet.

The goal is to convert all of the server-side formatting to be entirely protocol agnostic and only use the defined Evennia style codes, such as `|r`, and have those codes converted into protocol-friendly forms from the actual protocol itself.

See #3185 for the discussion.


InspectorCaracal:
A question for this: should converting an `EvString` or `EvStringContainer` (such as `EvTable`) to a `str` object render it *with* the markup, or *without* the markup?

InspectorCaracal:
Some other questions/issues that arose, for reference by me/others.

One EvString renders as equal to another if there is "invisible" markup at the ends - e.g. `EvString("|relectric |c")` is identified as equal to `EvString("|relectric ")`. This shouldn't be the case - while they may render the same on their own, when adding another string like `test` to it would result in `|relectric |ctest` versus `|relectric test` which is clearly not equal. I haven't identified why it is yet to make sure I fix it properly.

Also, while working on updating forms, I discovered that EvString slices behave very unpredictably... At minimum, when taking a slice such as `mystr[9:]` it's including all of the codes from the first segment of the string that's not being included. The code affixing doesn't seem to occur for codes in the remaining string when doing something like `mystr[:9]` - only when taking a tail slice (although also probably prefixing a center slice).

evennia/utils/evstring.py
```diff
@@ -0,0 +1,1231 @@
+
+# ------------------------------------------------------------
+#
+# EvString - ANSI-aware string class
```
Griatch, line 4: I suppose this is aware of evennia-style |-formatting rather than ANSI per-se, so this docstring is a bit misleading.

evennia/utils/evstring.py
```diff
@@ -0,0 +1,1231 @@
...
+        if _difference <= 0:
+            return self
+        return func(self, width, fillchar, _difference)
+
+    return wrapped
+
+# 
+def _to_evstring(obj):
+    """
+    convert to ANSIString.
```
Griatch, line 221: No longer an ANSIString ;)

Griatch (commented):
This is a great start of this project. I wonder if this kind of change is something one should merge directly into main or if one need to have this in a separate develop branch for a while; either way doing this change would mean a major version change (e.g. to Evennia 3.0.0). 
The original implementation of ANSIString (which, as far as I understand was retained here with different internal storage) was itself based on the state of the Python string library in Python 2.x. So one should probably review what has happened in `string` since then and possibly refactor/clean up/modernize ANSISTring (EvString) as needed; possibly some of the weirdness comes from this (There is no concept of a separate unicode string class anymore for example).

InspectorCaracal:
This still has a lot of work left - there's a *lot* of code to update and tests to work through - but since it came up recently and I was working on it more today, I figured I'd offer some reassurance that I haven't forgotten about it by committing and pushing my current progress.

ChrisLR:
> Some other questions/issues that arose, for reference by me/others.
> 
> One EvString renders as equal to another if there is "invisible" markup at the ends - e.g. `EvString("|relectric |c")` is identified as equal to `EvString("|relectric ")`. This shouldn't be the case - while they may render the same on their own, when adding another string like `test` to it would result in `|relectric |ctest` versus `|relectric test` which is clearly not equal. I haven't identified why it is yet to make sure I fix it properly.
> 
> Also, while working on updating forms, I discovered that EvString slices behave very unpredictably... At minimum, when taking a slice such as `mystr[9:]` it's including all of the codes from the first segment of the string that's not being included. The code affixing doesn't seem to occur for codes in the remaining string when doing something like `mystr[:9]` - only when taking a tail slice (although also probably prefixing a center slice). Based on some strange outputs I'd noticed in formatting codes previously with ANSIString, I think this has been happening all along but the change in parsing has made it more of a problem.
> 
> There's also a question of what should happening with leading and trailing color codes. Currently it's designed to preserve trailing color codes when slicing, but I think that they should only be preserved on stripping and the slice should occur before the code, since they're essentially switches, not open/close tags.
> 
> i.e. `|relectric |cboogaloo|n` should slice so that [:9] is `|relectric ` and [9:] is `|cboogaloo|n`, rather than what currently seems intended, which is `|relectric |c` and `|cboogaloo|n`

Comparing two EvStrings should include comparing the markup, if someone wants to compare without the markup they can cast to string first.

I agree for the slicing.
Similarly, len() should not include the markup.
Anything that wants to include the EvString's markup externally should need to access it explicitly.

InspectorCaracal:
> > A question for this: should converting an `EvString` or `EvStringContainer` (such as `EvTable`) to a `str` object render it _with_ the markup, or _without_ the markup?
> 
> Without the markup. EvString is the one handling the markup, accessing the str is accessing the internal value. repr could include the markup but converting EvString to html/others is not EvString's job.
> 
> That said, serializing the EvString should not simply serialize a str, it will need to be customized.

Sorry, I don't follow; what does converting an EvString to a normal string have to do with html?

At the moment, I convert it with the markup. e.g. `str(EvString("some |rred|n text"))` would be `"some |rred|n text"`. I figured that it makes the most sense to flatten back to the original string than to lose information.

ChrisLR:
>Sorry, I don't follow; what does converting an EvString to a normal string have to do with html?

Your EvString is used to store additional information, like hey, make this text red.
It does not decide how it is going to be shown to the user (in which format).
Somewhere down the line you will have code that will convert an EvString into whatever else for displaying, like html.


>At the moment, I convert it with the markup. e.g. str(EvString("some |rred|n text")) would be "some |rred|n text". I figured that it makes the most sense to flatten back to the original string than to lose information.

If you do that, there is very little benefit to not just using a string until it is converted to whatever destination format.
If the code using the string does not know about markup it won't be able to do anything with it, might have to clean it out and could also break the markup... while still needing to be parsed yet again into EvString (or other markup aware classes) for anything to be done with it.

The information will only be useful when something aware of Evennia markups detects it, so why not keep the EvString object all the way?

InspectorCaracal:
@ChrisLR Oh! You're misunderstanding the use case xD this is for things like printing to logs and other situations that just want the actual flat string 

If you take a look at the PR's changes to the telnet and webclient protocols and at the evstring module, you'll see that it already implements sensible protocol-format methods that are called when about to be sent down the wire, which themselves use separate renderer classes built from the previous ANSIParser and Text2Html classes

The intent is that the EvString and its containers, such as tables, will be kept as that object while being passed around internally.

ChrisLR:
@InspectorCaracal my bad then I must've missed something somewhere 😅

Griatch:
@InspectorCaracal This one will also need a rebase as well, due to the changes of the upstream repo to reduce its size.

Griatch:
@InspectorCaracal Is this still in progress?

InspectorCaracal:
@Griatch Yep! I'm just on a particularly tricky part and I've been a lot busier since the school year started.

InspectorCaracal:
Looks like I messed up a few of the conflicts when rebasing 😩 I'll make sure to clean that up before this leaves draft stage

InspectorCaracal:
As I was afraid of after some conversations on Discord last week, I've run into a major setback with my intended plan here.

When these updated objects are passed through AMP from the server to the portal, they're converted to their `str` representations. This is then being re-parsed back to a single EvString on the portal's protocol class before being rendered.

I *can* move ahead with it working this way, but it loses a lot of the pre-parsing efficiency of having the messages packaged into objects, and it prevents any truly HTML-focused rendering in the future due to forcing things like tables to use whitespace alignment rather than styled HTML elements.

As such, I'm not sure how best to proceed.

fariparedes:
Currently on this branch, any message with MXP markup in it will silently fail to send to the user.

evennia/utils/evstring.py
```diff
@@ -0,0 +1,1497 @@
...
+        decode strings, as escapes can only be respected once.
+
+        Internally, EvString can also passes itself precached code/character
+        indexes and clean strings to avoid doing extra work when combining
+        EvStrings.
+
+        """
+        text = args[0]
+        if not isinstance(text, str):
+            text = to_str(text)
```
fariparedes, line 566: `to_str()` strips MXP from EvLinks, unlike EvCode, because it converts them to strings, so MXP will not survive being added to an EvString, e.g. by joining chunks. Perhaps EvLink should have a `__str__` function which returns `self.raw()` to avoid this?

evennia/utils/evstring.py
```diff
@@ -0,0 +1,1497 @@
...
+        """
+        Splits the link text on the first instance of sep. Returns a tuple of length 3 containing a link
+        with the first segment of text, the separator, and an EvString of the tail. If sep is not found,
+        returns a tuple of self followed by two empty strings.
+        """
+        if sep not in self.clean():
+            return (self, '', '')
+        
+        result = self._ev_string.partition(sep)
+        return ( EvLink(result[0], link_value=self._link_string, link_key=self._link_key), result[1], result[2] )
```
fariparedes, line 500: I believe you would want to wrap `EvLink()` around all of these in order to retain the unbroken link functionality, possibly even the separator.

evennia/utils/evstring.py
```diff
@@ -0,0 +1,1497 @@
...
+        # now we check our results
+        if not result:
+            return ('', '', self) if reverse else (self, '', '')
+        
+        else:
+            # we broke the loop at i, so split the code chunks there
+            first = self._code_chunks[:i] + (result[0],)
+            last = (result[2],) + self._code_chunks[i+1:]
+            # create new EvStrings from our partitioned results and return
+            return ( EvString(''.join(first), chunks=first), sep, EvString(''.join(last), chunks=last), )
```
fariparedes, line 963: Despite your own advice in `EvString.join()`'s doc string, you are joining EvStrings with `str.join()` here. ;)

InspectorCaracal:
I've been arguing with myself for a while on this, because this conversion and task is something I feel strongly about getting completed, but I think it's time for me personally to throw in the towel on this PR.

I think that the fundamentals and direction I've taken here are sound, but I don't have the technical knowledge or time necessary to work through the remaining issues, particularly those regarding the way ANSIString is integrated into the more complex formatting classes like EvForm and EvTable.

In other words: I'm forced to admit that this is as far as I can take it.

I'll leave my branch intact in case someone else can and wants to pick up and finish what I've started. The PR itself can be left as a draft or closed.

Griatch:
@InspectorCaracal Ok, no worries! Thanks for your work so far! As you said, let's keep this branch open for now so your work can act as a basis for future development.

PR #3267, opened 2023-09-05
Add recache step to ObjectSessionHandler

InspectorCaracal:
I was looking through all sessions references and spotted this - I'm pretty sure it's supposed to be recaching to sync up before trying to get again, rather than go into infinite recursions.

Bug Prevention :tm:


Griatch:
Nice catch!

PR #3270, opened 2023-09-12
Update MSSP value

InspectorCaracal:
According to https://mudhalla.net/tintin/protocols/mssp/ "Evennia" is a recognized value for FAMILY now, so we no longer need to identify ourselves as "Custom"


Griatch (approved):
How cool. Merging.

PR #3281, opened 2023-09-30
Adds $you() and $your() to the ACTOR_STANCE callables.

volundmush:
Adds $your() and $Your() to ACTOR_STANCE_CALLABLES

People are gonna want more of these.

Nah it's really that simple.

### DIFF
--- a/evennia/utils/funcparser.py
+++ b/evennia/utils/funcparser.py
@@ -1252,6 +1252,76 @@ def funcparser_callable_you_capitalize(
     )
 
 
+def funcparser_callable_your(
+    *args, caller=None, receiver=None, mapping=None, capitalize=False, **kwargs
+):
+    """
+    Usage: $your() or $your(key)
+
+    Replaces with your for the caller of the string, with the display_name +'s
+    of the caller for others.
+
+    Keyword Args:
+        caller (Object): The 'your' in the string. This is used unless another
+            your-key is passed to the callable in combination with `mapping`.
+        receiver (Object): The recipient of the string.
+        mapping (dict, optional): This is a mapping `{key:Object, ...}` and is
+            used to find which object `$you(key)` refers to. If not given, the
+            `caller` kwarg is used.
+        capitalize (bool): Passed by the You helper, to capitalize you.
+
+    Returns:
+        str: The parsed string.
+
+    Raises:
+        ParsingError: If `caller` and `receiver` were not supplied.
+
+    Notes:
+        The kwargs should be passed the to parser directly.
+
+    Examples:
+        This can be used by the say or emote hooks to pass actor stance
+        strings.
+
+        - `$your() pet jumps at $you(tommy).`
+
+        The caller-object will see "Your pet jumps Tommy."
+        Tommy will see "CharName's pet jumps at you."
+        Others will see "CharName's pet jumps at Tommy."
+
+    """
+    if args and mapping:
+        # this would mean a $your(key) form
+        try:
+            caller = mapping.get(args[0])
+        except KeyError:
+            pass
+
+    if not (caller and receiver):
+        raise ParsingError("No caller or receiver supplied to $your callable.")
+
+    capitalize = bool(capitalize)
+    if caller == receiver:
+        return "Your" if capitalize else "your"
+
+    name = caller.get_display_name(looker=receiver) \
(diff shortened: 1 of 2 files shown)


evennia/utils/funcparser.py
```diff
@@ -1252,6 +1252,76 @@ def funcparser_callable_you_capitalize(
...
+        Tommy will see "CharName's pet jumps at you."
+        Others will see "CharName's pet jumps at Tommy."
+
+    """
+    if args and mapping:
+        # this would mean a $your(key) form
+        try:
+            caller = mapping.get(args[0])
+        except KeyError:
+            pass
```
Griatch, line 1298: When would this ever happen? `mapping.get()` returns `None` if unset, and `args[0]` should never fail since we check for `args` above (and would return `IndexError` anyway).

evennia/utils/funcparser.py
```diff
@@ -1252,6 +1252,76 @@ def funcparser_callable_you_capitalize(
...
+        raise ParsingError("No caller or receiver supplied to $your callable.")
+
+    capitalize = bool(capitalize)
+    if caller == receiver:
+        return "Your" if capitalize else "your"
+
+    name = caller.get_display_name(looker=receiver) \
+        if hasattr(caller, "get_display_name") else str(caller)
+
+    return name + "'s"
```
Griatch, line 1310: I was taught that if the name ends with `s`, it would be just a single apostrophe, like `Charles'` . But looking it up, it appears it's allowed to also use `'s` here, like `Charles's`. English is weird.

Griatch (changes requested):
Looks good! 
- I have a comment to address.
- A unit test is needed; can be added in `evennia/utils/tests/test_funcparser.py`.

Griatch:
@volundmush Ping, this just needs some minor changes, and to be rebased against the upstream since upstream repo was changed to make it a smaller download.

evennia/utils/funcparser.py
volundmush: That isn't true, actually. .get() will raise a KeyError when given a single argument and will return a default value if given a second. For this to return none if args[0] doesn't exist, it would need to be `.get(args[0], None)`

I copied this exactly from the normal $you(). Do you wanna rewrite both?

evennia/utils/funcparser.py
volundmush: I may not be using the exactly correct grammatical terms, but `Charles'` would be a multiple possessive for a word ending in s while `Charles's` is singular possessive or... something like that.

The intent of this function is to show...

"Your ability level has increased!" to Charles and
"Charles's ability level has increased!" to everyone watching.

So it should be correct as-is.

Multiple possessive would be...

"Charles' abilities have increased in level!" and that's not possible to rectify using funcparser.

InspectorCaracal: `Charles'` is the correct singular possessive firm of `Charles`

volundmush: O_o
If you say so...
That's not how I was taught in school.

evennia/utils/funcparser.py
Griatch: Let's do that. I could do it, but a unit test need to be added to this PR anyway, so if you could fix this as well, that'd be great!

Griatch (changes requested):
I have a comment, and this is also still missing a unit test. Good otherwise!

volundmush:
There we go.

evennia/utils/tests/test_funcparser.py
```diff
@@ -437,20 +437,25 @@ def test_callable(self, string, expected):
             ("$You() $conj(smile) at $You(char1).", "You smile at You.", "Char1 smiles at Char1."),
             ("$You() $conj(smile) at $You(char2).", "You smile at Char2.", "Char1 smiles at You."),
             (
-                "$You(char2) $conj(smile) at $you(char1).",
-                "Char2 smile at you.",
-                "You smiles at Char1.",
+                    "$You(char2) $conj(smile) at $you(char1).",
```
Griatch, line 440: You really need to fix your auto-formatter, dude. We don't use 8 character indents in Evennia (nor in PEP8). These PRs become unnecessarily spammy this way.

Griatch (approved):
Looking good! 
(but please take a look at your auto-formater ...)

Issue #2247, opened 2020-11-01
[BUG] Quirk in saving webclient layout


Griatch:
#### Describe the bug

It's visibly possible to circumvent the block against saving the `default` webclient layout. It doesn't actually work, but it's confusing.

#### To Reproduce
In webclient, open options: 

1.  Start to edit the `default` layout name by adding more letters to its name, like `default1234232`
2.  The `Close Options and Save` button becomes active
3.  Without saving, erase the extra letters, down to `default`
4.  The `Close Options and Save` buttons _remains_ active and you can now save your modified layout as 'default'

The save does not _actually_ happen (so this is correct), as loading default returns the 'real' default layout.

#### Expected behavior

The `default` button should remain greyed out whenever it's about to save something named `default`. While the above behavior is apparently benign in the sense that the `default` layout is not *actually* saved, it nevertheless confuses the user. 

#### Environment, Evennia version, OS etc
Evennia master branch.

Griatch:
Afaik this is still an issue.

count-infinity:
#3282  is merged and closes this issue I believe.  Currently unable to reproduce.

PR #3282, opened 2023-10-01
added listener + logic to Webclient Save Layout button -  #2247

johnnyvoruz:
#2247 

**Brief overview of PR changes/additions**

- Introduced a jQuery listener on the #layoutName input field.
- Automatically disables the "Save" button (class .savelayout) if the input is exactly "default".
- Re-enables the "Save" button for other input values.
- This change ensures that users don't have the perception that they can overwrite the default layout.

**Motivation for adding to Evennia**

- I'm still learning the engine, but excited to contribute to the web site of things!

**Other info (issues closed, discussion etc)**

- This change was inspired by an observed behavior where users would often mistakenly try to save a new layout as "default".
- By disabling the button for this specific input, we add a subtle hint to the user about the importance of this default name and guide them towards choosing a different name.

### DIFF
--- a/evennia/web/static/webclient/js/plugins/goldenlayout.js
+++ b/evennia/web/static/webclient/js/plugins/goldenlayout.js
@@ -4,8 +4,6 @@
  *
  */
 
-
-
 let goldenlayout = (function () {
 
     var myLayout; // The actively used GoldenLayout API object.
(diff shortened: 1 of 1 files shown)


evennia/web/static/webclient/js/plugins/goldenlayout.js
```diff
@@ -584,12 +587,33 @@ let goldenlayout = (function () {
     }
 
 
-    //
+    // Listener for realtime changes to the layout name input field.
+    // If the layout name is "default", the save button is disabled
+    // to prevent the perception of overwriting the default layout.
+    $(document).on("input", "#layoutName", function () {
+        console.log("Input changed to:", $(this).val());
```
Griatch, line 594: This seems like a leftover debug message (no need to log every change like this, right?)

Griatch (changes requested):
Nice! Good to get this fixed. Just a minor tweak needed before merge.

evennia/web/static/webclient/js/plugins/goldenlayout.js
johnnyvoruz: Bingo, thanks for catching that @Griatch - that was just for debugging! Thanks for your eagle eyes!

Griatch (approved):
Looking good!

PR #3296, opened 2023-10-22
docs(Adding-Commands): :bug: fix for lesson 8.1.2

iLPdev:
if we're already importing `CmdEcho` manually, we should just be able to remove the reference to command in the `self.add` and end it with `()`

existing tutorial code throws runtime error


iLPdev:
this is deprecated by issue #3297

PR #3301, opened 2023-10-26
Configurable Django Admin

volundmush:
Currently, the EvenniaAdminSite's get_all_apps method filters out any app which doesn't exist in its app_order listing.

This replaces that logic to allow settings.py to provide priority ordered listing to a selection of apps, and also to provide a list of apps which can be excluded from the listing.

Now it is possible to use the Django Admin properly for any added models.

The Django Admin is currently filtering out access to models from various Django plugins/libraries like Django Wiki and Django Helpdesk, as well as any models I might create in my projects.

An extendable, more generic approach is far better.

The Django Admin site doesn't nicely categorize the displays of where models come from. I wish it displayed in categories like "Here are Evennia models", "Here are wiki models", "here are helpdesk models"...

### DIFF
--- a/evennia/settings_default.py
+++ b/evennia/settings_default.py
@@ -1033,6 +1033,24 @@ MIDDLEWARE = [
     "evennia.web.utils.middleware.SharedLoginMiddleware",
 ]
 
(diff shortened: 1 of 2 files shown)


Griatch (approved):
This looks good!

PR #3304, opened 2023-10-27
docs(Sittable-Object): :memo: add terminal punctuation to do_stand code

iLPdev:
missing a period in both code example blocks for do_stand

progress, not perfection

should I make and close issues for small changes like this? it seems excessive but I will if that's the preference


Griatch (approved):
Thanks! No, there's no need to make a formal issue for a small thing like this if you are fixing it right away anyway. Issues are there to collect and track progress on an issue and to make sure it's not forgotten if it's not fixed right away. :)

PR #3306, opened 2023-10-27
Fix Clothing contrib ignoring clothing type limits

InspectorCaracal:
Corrects a typo in `CmdWear` that caused the command to check the wrong attribute on clothing.

Also adds a test on clothing limits to prevent future regressions.

Bug fixing


Griatch (approved):
Thanks! Also thanks for the unit test, that's great. I thought I saw an issue for this to close, but now I don't see it.

PR #3309, opened 2023-10-28
docs(Sittable-Object): :bug: call `adjective` from `self.db`

iLPdev:
call from db in all instances

progress, not perfection


Griatch (changes requested):
This is incorrect. You can see on line 157 that `adjective` is a local variable, created like this: 

```py 
adjective = self.db.adjective or "on"
```
In other words, we have a fallback of "on", if the adjective is not set explicitly in the database.
So this will not merge.

iLPdev:
this is confusing because, without the commit, i was getting the fallback incorrectly.

iLPdev:
by the end of the lesson, the fallback wasn't working, so I'm pretty sure the problem was my code/thinking

PR #3313, opened 2023-10-29
Improve OptionHandler.set()

volundmush:
Simply alters the output type of OptionHandler.set() so that it returns the BaseOption that was matched, allowing access to its methods and properties like .key and .display() by other systems.

I'm implementing a lot of various things which use OptionHandler, and having access to what was .set() will help with output formatting considerably.


Griatch (approved):
Makes sense. Also note to self, that this is a backwards incompatible change.

PR #3319, opened 2023-10-31
Refactoring Server Startup to be more idiomatic for Twisted.

volundmush:
A massive cleanup of server.py and portal.py twisted application startups that moves much of the logic into a Twisted Service.

Cleaning this up to be more modular and upgradeable in the future. It's currently a mess.

The tests do not currently pass... I cannot figure out how to fix them.


volundmush:
It looks like one of the reasons that some tests for AMP are failing is because I moved the startup sequence for AMP to be "started by Twisted when it starts services." IE: the privilegedStartService() hooks on the new EvenniaServerService and EvenniaPortalService classes. Since those aren't being run, the AMP service isn't starting, isn't connecting, and so isn't setting amp_protocol on the service object.

Griatch:
@volundmush If the AMP is no longer automatically started by the test runner, then it may be possible to boot strap for the sake of unit testing?

Griatch (approved):
This is a really nice cleanup I think. Good that you got the tests working too. Kudos!

PR #3322, opened 2023-11-04
Fixing output of BaseOption.display to always be str

volundmush:
ensure that BaseOption.display() always returns a string.

Turns out this wasn't actually minding its own API; sometimes it would return a number or some other value.

Not everything that wants to use it liked that...


Griatch (approved):
Thanks for the fix!

PR #3324, opened 2023-11-06
SIGNAL_ACCOUNT_POST_LOGIN_FAIL is now properly used. Added a CUSTOM_S…

volundmush:
the SIGNAL_ACCOUNT_POST_LOGIN_FAIL signal wasn't being called anywhere, so that's been fixed.

A SIGNALS_CUSTOM object has been added to signals.py using a defaultdict to make for an easy event handler.

Although I regret pushing for adding these signals all over the place (the custom variant using the defaultdict is a much more elegant approach), they ought to do what they say they do, so... fixing this oversight.

### DIFF
--- a/evennia/server/signals.py
+++ b/evennia/server/signals.py
@@ -93,6 +93,11 @@ SIGNAL_CHANNEL_POST_CREATE = Signal()
 SIGNAL_EXIT_TRAVERSED = Signal()
 
 # Used as a generic event emitter. Use to make your own signals easily in one place!
+# To use it, import SIGNALS_CUSTOM and use it like a dictionary of Signal objects.
+# Example:
+# from evennia.utils.signals import SIGNALS_CUSTOM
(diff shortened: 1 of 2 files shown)


evennia/server/signals.py
```diff
@@ -91,6 +92,9 @@ def my_callback(sender, **kwargs):
 # Called just after at_traverse hook.
 SIGNAL_EXIT_TRAVERSED = Signal()
 
+# Used as a generic event emitter. Use to make your own signals easily in one place!
+SIGNALS_CUSTOM: dict[str, Signal] = defaultdict(Signal)
+
```
Griatch, line 97: This could do with some example of how to use this.

Griatch (changes requested):
The new `SIGNALS_CUSTOM` needs example usage description so people can actually make use of it without having to go to Django docs.

Griatch (approved):
LGTM. Will probably need to update the signal docs at some point too.

PR #3326, opened 2023-11-09
Breakup DefaultObject.msg() and add an at_post_msg_receive() hook

volundmush:
Pretty much exactly what it says on the tin.

This refactors DefaultObject.msg(), breaking it up to rely on a series of helper methods and adding a new hook that's called at the end of a successful message send.

There is currently no way to access the finalized kwargs at the end of the .msg() call, which I wish to be able to nab for purposes of logging. Since it's a pain in the ass to reimplement/overload .msg() in general I figured out breaking it up at the same time would be a good idea.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -620,9 +620,9 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
         self._msg_helper_text_format(text, kwargs)
 
         # try send hooks
-        self._msg_helper_from_obj(text=text, from_obj=from_obj, **kwargs)
+        self._msg_helper_from_obj(from_obj=from_obj, **kwargs)
 
-        if not self._msg_helper_receive(text=text, from_obj=from_obj, **kwargs):
+        if not self._msg_helper_receive(from_obj=from_obj, **kwargs):
             return
 
         # relay to session(s)
@@ -671,7 +671,7 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
                     text = repr(text)
             kwargs["text"] = text
 
-    def _msg_helper_from_obj(self, text=None, from_obj=None, **kwargs):
+    def _msg_helper_from_obj(self, from_obj=None, **kwargs):
         """
         Helper method for .msg() that handles calling at_msg_send on the from_obj.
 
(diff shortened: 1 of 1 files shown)


evennia/objects/objects.py
```diff
@@ -612,25 +612,56 @@ def msg(self, text=None, from_obj=None, session=None, options=None, **kwargs):
...
+        """
+        Overloadable hook which receives the kwargs that exist at the tail end of self.msg()'s processing.
+
+        This might be used for logging and similar purposes.
+
+        Kwargs:
+            from_obj (DefaultObject or list[DefaultObject]): The objects that sent the message.
+            **kwargs: The kwargs from the end of message, using the Evennia outputfunc format.
+        """
+        pass
```
Griatch, line 643: If this is the only method intended to be overridden, it would be more visible to move it to the end of this block, after all the private methods.

evennia/objects/objects.py
```diff
@@ -612,25 +612,56 @@ def msg(self, text=None, from_obj=None, session=None, options=None, **kwargs):
...
-        try:
-            if not self.at_msg_receive(text=text, from_obj=from_obj, **kwargs):
-                # if at_msg_receive returns false, we abort message to this object
-                return
-        except Exception:
-            logger.log_trace()
+        self._msg_helper_text_format(text, kwargs)
+
+        # try send hooks
+        self._msg_helper_from_obj(from_obj=from_obj, **kwargs)
```
Griatch, line 623: While I can see the point of these being private from a code perspective, using a private `_` prefix also suggests that this is a private method that is not intended to be customized by the developer.

Griatch (changes requested):
I don't mind the `msg` method being broken up into smaller parts; I think this could be useful for devs! I have a few questions/concerns about this though before merge:  

1. Majority of these break-out methods are "private" (`_` prefix), and while this makes sense from a purely Pythonic perspective, Evennia offers and API, and in an API, private methods suggests that the API is not to be overridden by the user (because it's private it could be changed without warning upstream). Since the reason to break up `msg` would ostensibly be to make it easier to override, I'd think these sub-methods should not be private. 
2. Following the style used everywhere in Evennia, the sub-methods should appear above the method using them in the class. So please move `msg` to the end, after its helper methods.

Griatch:
@volundmush Ping.

volundmush:
This PR is rendered completely useless by #3356 so I'm gonna retire it. Thanks for keeping up with this though.

PR #3334, opened 2023-11-21
Adjusted checks for Object/Account to prevent conflicts.

volundmush:
Tweaked a few uses of hasattr and made a few other minor adjustments where Evennia tries to detect the "account" property to determine if it's looking at an Account or an Object.

I did this by altering hasattr(target, "account") to hasattr(target, "has_account") since DefaultObject has that method and Accounts do not.

Although Evennia itself doesn't strictly need this, one of the projects I'm trying to integrate adds models that establish a OneToOneField to the Django User (settings.AUTH_USER_MODEL, etc). Since this means that Accounts have a .account, it sends Evennia's code off on a wild goose chase and breaks things.

The library in question is Pinax's django-user-accounts, which adds quite an array of features. You might wanna check it out.

Everything appears to be working.

### DIFF
--- a/evennia/comms/models.py
+++ b/evennia/comms/models.py
@@ -646,8 +646,6 @@ class SubscriptionHandler:
             from django.core.exceptions import ObjectDoesNotExist
 
             try:
-                if hasattr(obj, "account") and obj.account:
-                    obj = obj.account
                 if not obj.is_connected:
                     continue
             except ObjectDoesNotExist:
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -8,6 +8,7 @@ from django.conf import settings
 from django.core.paginator import Paginator
 from django.db.models import Max, Min, Q
 
+import evennia
 from evennia import InterruptCommand
 from evennia.commands.cmdhandler import get_and_merge_cmdsets
 from evennia.locks.lockhandler import LockException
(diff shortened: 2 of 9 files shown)


Griatch (changes requested):
I see the argument for not relying on `.account` for the sake of third-party libs. I'm not a fan of checking for the `has_account` method like this though. It smells pretty magical and unintuitive. 

Honestly the original check for `account` is not great either. :disappointed: 

While duck typing is all good and well, I wonder if it would not be best to simply be explicit, and use `utils.inherits_from(DefaultAccount)` instead here?

volundmush:
Explicit it is.

evennia/comms/models.py
```diff
@@ -646,8 +646,6 @@ def online(self):
             from django.core.exceptions import ObjectDoesNotExist
 
             try:
-                if hasattr(obj, "account") and obj.account:
-                    obj = obj.account
```
Griatch, line 650: This got lost?

Griatch (changes requested):
Looks like one of these got lost on the way, otherwise looks ok

evennia/comms/models.py
volundmush: That was deliberate.
Both Accounts and Objects have an is_connected property.

Griatch: Fair enough. 
(by the way - you can add comments to your own PR lines - explaining your reasoning for the reviewer to avoid unnecessary questions).

Issue #3336, opened 2023-11-22
[BUG] Dockerfile's ONBUILD command fails at --chown=evennia


poisonbl:
#### Describe the bug
Dockerfile's ONBUILD command fails at `--chown=evennia` because the user hasn't been created.

#### To Reproduce
Create a Dockerfile consisting of:

```Dockerfile
FROM evennia/evennia:latest

ENTRYPOINT evennia start -l
```

Run `docker build .` in the folder with it.

```
$ docker build .
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/

<download spam>
Sending build context to Docker daemon  5.153MB
Step 1/2 : FROM evennia/evennia:latest
# Executing 1 build trigger
unable to convert uid/gid chown string to host mapping: can't find uid for user evennia: no such user: evennia
```

#### Expected behavior
The evennia image build should either preemptively create the user, or not incorrectly expect it to be their for a derivative build.

#### Additional context
Appears to've been introduced potentially as far back as Dec 2, 2022 in 86e2722

Simplest fix would likely be just to change

```Dockerfile
ONBUILD COPY --chown=evennia . /usr/src/game
```
to
```Dockerfile
ONBUILD COPY . /usr/src/game
```

PR #3341, opened 2023-11-24
Fixes evennia/evennia#3336 - Dockerfile's ONBUILD command fails at --chown=evennia

poisonbl:
Addressing evennia/evennia#3336

Taking the assumption that the user creation was commented out for a reason and following suit by fixing this to work without it.


Griatch (approved):

Griatch:
There was an issue with the `evennia` user for the general case as far as I can recall; left it in for those that wanted to use it, but not as a default.

PR #3342, opened 2023-11-24
Add Command.cmdset_source

volundmush:
This simply modifies CmdSet.add() to stamp each newly added command with cmd.cmdset_source = self

So, Commands now know EXACTLY which CmdSet they came from.

This allows a Command to reference properties, methods, state, etc, on their CmdSet, which can be useful in a number of ways.

See, for instance, creating a kind of pseudo-menu which is aware of all of the commands available to the 'menu'.

This is such a tiny - but useful - change, it doesn't appear to cause any issues.

### DIFF
--- a/evennia/commands/cmdset.py
+++ b/evennia/commands/cmdset.py
@@ -541,6 +541,8 @@ class CmdSet(object, metaclass=_CmdSetMeta):
         system_commands = self.system_commands
 
         for cmd in cmds:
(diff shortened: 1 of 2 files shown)


Griatch (approved):
I dislike the use of cmdsets for data storage. But I like the notion of storing the source-cmdset on the cmd for debug purposes. This may be useful in combination with expanding the examine command to show where each available cmd is actually coming from.

PR #3344, opened 2023-11-25
Overhaul IP Matching

volundmush:
Extracted and streamlined the logic for extracting the 'source IP' from a Django Request into a new utility function, ip_from_request

Created a match_ip function which can test if the given IP matches a second IP OR a subnet in CIDR format (uses Python3's standard library, hooray for the ipaddress module...)

Adds and integrates an OriginIpMiddleware into Evennia's middleware which will use this on every incoming request and sets the result to request.origin_ip so that it's only ever used once.

Modified webclient_ajax to use the new standard.

Added relevant tests.

When my friend wanted to docker-ify Evennia and put it behind several layers of caching/reverse proxies/hardening, etc, UPSTREAM_IPs proved to be wildly inadequate to the task.

In order to compensate, he coded this monstrosity:

```py
UPSTREAM_IPS = ['127.0.0.1']

for a in range(16,32):
    for b in range(254):
        for c in range(254):
            UPSTREAM_IPS.append(f"172.{a}.{b}.{c}")
```

to this I say, HECK NAW, that's crazy and it drives me up a wall every time I see it.
How about we use some proper IP matching libraries here... Python3 has 'em, so let's use 'em.

Everything appears to be backwards compatible.


Griatch (approved):
This looks very nice. :+1:

PR #3346, opened 2023-11-25
Streamlined cmdhandler to support more extension.

volundmush:
Reworked some of the cmdhandler guts to grow more agnostic of the KINDS of objects it can be executing commands from, without breaking existing conventions.

The current cmdhandler code is very messy and not very pythonic; it's also difficult to extend in ways that challenge the internal architecture of Evennia itself. These alterations are a major step in divorcing the process of merging cmdsets from the triad of (session, account, object). It is now possible to add MORE "command-providing objects".

evennia.commands.default.tests.TestCmdTasks.test_new_task_waiting_input keeps failing and I have no idea why.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -206,9 +206,9 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
 
     # Determines which order command sets begin to be assembled from.
     # Objects are usually third.
-    cmd_order = 100
-    cmd_order_error = 100
-    cmd_type = "object"
+    cmdset_provider_order = 100
+    cmdset_provider_error_order = 100
+    cmdset_provider_type = "object"
 
     # Used for sorting / filtering in inventories / room contents.
     _content_types = ("object",)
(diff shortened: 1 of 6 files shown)


evennia/objects/objects.py
```diff
@@ -205,6 +205,12 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
 
     """
 
+    # Determines which order command sets begin to be assembled from.
+    # Objects are usually third.
+    cmd_order = 100
+    cmd_order_error = 100
+    cmd_type = "object"
+
```
Griatch, line 213: I feel like sneaking this into a PR that ostensibly is for refactoring/cleanup of the cmdhandler is a bit iffy :wink:  This is a HUGE change to the cmdset merge process. CmdSets already have merge priority and if you want commands from certain cmdsets to merge ahead of others, you change the merge priority of the cmdset. If you want an object's commands to override others, you increase the prio of that cmdset. It's not something you set on the object level.

Unless I'm missing something here, this should be removed.

Griatch (changes requested):
I like most of the cleanup done here, it does make for less repetition :+1: . I also have some concerns/required changes: 

- Renamings 
- The cmd merge prio changes don't belong in this PR, and is probably a no-go overall. 
- The test suite was failing from other tests previously, rebasing this PR against the latest Evennia should make for a better show of things (working on that).

evennia/objects/objects.py
volundmush: This mechanism is used for sorting the order in which the refined cmdhandler uses to do the merging. like, it first checks session, then account, object. Since the dictionary generated by `get_cmdset_providers` has no concept of order.

Griatch (approved):
Nice refactor overall! 
I'm not a fan of sneaking in a new priority mechanism in a refactor like this. But it does make sense as a static thing making the merge more deterministic; so will accept this for the PR too.

PR #3347, opened 2023-11-25
Improved lock-setting logic

volundmush:
Where relevant, typeclass objects now have a new classmethod: generate_default_locks(), which is used to generate lockstrings for .create()

The lockhandler now has better logic for handling empty strings.

The enhancements to the lock system I'm adding to Athanor exposed that Evennia doesn't know what to do if the lockhandler.add() is fed an empty string, and that .create() will explode if .lockstring is empty since an empty string can't be .format()'d...

This change makes it more flexible, while remaining backwards compatible.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -207,10 +207,6 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
     # Used for sorting / filtering in inventories / room contents.
     _content_types = ("object",)
 
-    # lockstring of newly created objects, for easy overloading.
-    # Will be formatted with the appropriate attributes.
-    lockstring = "control:id({account_id}) or perm(Admin);delete:id({account_id}) or perm(Admin)"
-
     objects = ObjectManager()
 
     # populated by `return_appearance`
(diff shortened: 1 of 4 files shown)


evennia/objects/objects.py
```diff
@@ -1011,6 +1011,26 @@ def clear_contents(self):
...
+        Args:
+            account (Account): Account to attribute this object to.
+            caller (DefaultObject): The object which is creating this one.
+            **kwargs: Arbitrary input.
+
+        Returns:
+            lockstring (str): A lockstring to use for this object.
+        """
+        if cls.lockstring:
+            account_id = account.id if account else -1
```
Griatch, line 1030: This convention of setting id to `-1` is new. It seems like this could create some really hard-to-find bugs, if objects start being silently created with locks of `id(-1)` ... I would say we should error out in this situation; clearly something is wrong with the object creation or setup. The default should always return something reasonable out of the box.

Griatch (changes requested):
Having a default lockstring getter is fine. I have some comments on the naming and defaults chosen. I wonder if one could add some more unit testing to cover these new methods too?

Griatch (changes requested):
This is looking good. But since we have new `get_default_locks` functionality, with new logic inside it, we need to also make sure they are covered by unit tests and are sure to generate expected locks/errors.

Griatch (approved):
LGTM, thanks for adding the unit tests.

PR #3353, opened 2023-12-02
Add a unique ID to each webclient instance

### LINKED ISSUE #3327: [BUG] Inconsistent session tracking between websocket and AJAX webclient
#### Describe the bug
When connecting to evennia with websockets, each tab establishes its own websocket and counts as a separate session for the server.

However, by default, a browser uses the same _website_ session across all tabs. The AJAX method references this same one session for all requests, so all tabs count as the _same_ session for the server and will override each other, even in multisession modes.

#### Additional context
I don't think either one of these is necessarily wrong or correct, but the fact that they behave differently is less than ideal. That said, I think it would probably be simpler to make the websocket behave like the AJAX method than vice versa.

InspectorCaracal:
Generates a low-security unique ID on loading a new webclient page, so that multiple pages associated with a single browser session properly count as their own play sessions.

Resolves #3327

Quick note: while there is a more cryptographically sound UUID generator available in modern browsers via the `crypto` library, its usage is restricted to secure contexts. Since Evennia doesn't require running a server with HTTPS and individual tabs on a single browser don't actually need to be guaranteed unique on a large scale, I opted for a quicker and more widely functional generator function.

### DIFF
--- a/evennia/server/portal/webclient.py
+++ b/evennia/server/portal/webclient.py
@@ -68,7 +68,7 @@ class WebSocketClient(WebSocketServerProtocol, _BASE_SESSION_CLASS):
 
         """
         try:
-            # client will connect with wsurl?csessid&browserid
+            # client will connect with wsurl?csessid&page_id&browserid
             webarg = self.http_request_uri.split("?", 1)[1]
         except IndexError:
             # this may happen for custom webclients not caring for the
@@ -86,7 +86,7 @@ class WebSocketClient(WebSocketServerProtocol, _BASE_SESSION_CLASS):
         if len(cargs) == 1:
             self.browserstr = str(cargs[0])
         elif len(cargs) == 2:
-            self.cuid = str(cargs[0])
+            self.page_id = str(cargs[0])
             self.browserstr = str(cargs[1])
 
         if self.csessid:
--- a/evennia/server/portal/webclient.py
+++ b/evennia/server/portal/webclient.py
@@ -82,9 +82,12 @@ class WebSocketClient(WebSocketServerProtocol, _BASE_SESSION_CLASS):
             logger.log_trace(str(self))
             return None
 
-        self.csessid, *browserstr = webarg.split("&", 1)
-        if browserstr:
-            self.browserstr = str(browserstr[0])
+        self.csessid, *cargs = webarg.split("&", 2)
+        if len(cargs) == 1:
+            self.browserstr = str(cargs[0])
+        elif len(cargs) == 2:
+            self.cuid = str(cargs[0])
+            self.browserstr = str(cargs[1])
 
         if self.csessid:
             return _CLIENT_SESSIONS(session_key=self.csessid)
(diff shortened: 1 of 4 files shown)


evennia/server/portal/webclient_ajax.py
```diff
@@ -171,19 +183,19 @@ def lineSend(self, csessid, data):
             dataentries.append(jsonify(data))
             self.databuffer[csessid] = dataentries
 
-    def client_disconnect(self, csessid):
+    def client_disconnect(self, cusessid):
```
Griatch, line 186: The docstring still mentions `csessid`. It should reflect that cusessid is a combination of the session id and browser id.

evennia/server/portal/webclient.py
```diff
@@ -82,9 +82,12 @@ def get_client_session(self):
...
             return None
 
-        self.csessid, *browserstr = webarg.split("&", 1)
-        if browserstr:
-            self.browserstr = str(browserstr[0])
+        self.csessid, *cargs = webarg.split("&", 2)
+        if len(cargs) == 1:
+            self.browserstr = str(cargs[0])
+        elif len(cargs) == 2:
+            self.cuid = str(cargs[0])
```
Griatch, line 89: Hm. We now hae cuid, cusessid, csessid, browserstr,  ... it's starting to be confusing to track what's what here. While csessid is probably something we need to keep for backwards compatibility, maybe it would be worthwhile to start renaming local variables to something more useful, like `client_session_id`, `webclient_session_id` etc?

Griatch (changes requested):
Good overall; I have some comments as to readability of this as we start to add more ids.

evennia/server/portal/webclient.py
InspectorCaracal: Sure! I'd be happy to rename them. `csessid` is tracking the browser's login session and needs to be kept separate for authentication, but I can alias. `cuid` is the individual client page's ID and I can name absolutely anything. What would you prefer for each?

Griatch: @InspectorCaracal I don't think it would hurt to be explicit here; so something like `webclient_page_id` instead of `cuid` and so on - something that actually tells you what the id points to.

Griatch:
@InspectorCaracal If this is to go into Evennia 3.0, it would be good to wrap this one up. :)

InspectorCaracal:
I'll get it done in the next few hours. 👍

InspectorCaracal:
FYI, the PR here says it has merge conflicts, but when I look at the "Resolve conflicts" page, I don't see any...? I'm a bit loath to push the "mark as resolved" button in case I'm just bad at reading the UI, but it *seems* fine, if you could just double-check it for me.

Griatch:
@InspectorCaracal Just to be sure, it could be an idea to rebase your branch against latest upstream and re-push.

InspectorCaracal:
All right, the confusion was rooted in the AjaxWebClient class having been moved from the top of the module to the bottom, so I rebased, shifted it, and patched up all the conflicts.

Issue #3354, opened 2023-12-03
[Feature Request] Rework the message path concept, decoupling from ANSI


Griatch:
#### Is your feature request related to a problem? Please describe.

Today, while Evennia's message path is generic, it's heavily tied to ANSI markup: Text strings are handled with ANSI codes, and conversion to e.g. HTML also go via ANSI-to-HTML conversion. Furthermore, there is no generalized way to handle more context output types, like menus or tables/forms. It's all going to the Portal as ANSI-marked strings and sent over the wire that way, despite certain clients potentially being able to do more advanced things with them. An example is that the webclient could theoretically render an EvTable using HTML `<table>` tags. 

#### Describe the solution you'd like

We want to decouple Evennia's string handling from ANSI. ANSI is a telnet-specific standard that is irrelevant for other client types. We want to keep the Server as protocol-agnostic as possible. After discussion with @volundmush, this is the suggestion: 

- Use Evennia's internal markup (`|r` markup) for as long as possible. @InspectorCaracal's `EvString` PR would be required here, to move away from ANSI-dependence.
- Use some other additional library (Such as `rich`) for more expressive markup.
- When sending data from the Server to the Portal, data elements are converted into *Sendables*. A Sendable is a container that can contain any pickle-able python data. A Sendable could also wrap multiple entities, if they are intended to be rendered together (such as a set of tables going to a specific webclient pane). Finally, the Sendable can and will contain metadata related to the send, such as command-id. 
- The Sendable (or a list of them) are (depending on MULTISESSION_MODE) routed to different sessions, pickled (they must always be possible to pickle) and sent to the Portal. On the Portal side, they are in turn routed to all client protocols that should be receiving this data. 
- The client protocols contain customizable/extendable handlers for processing different Sendables in to the right form for that protocol. For example the telnet protocol would undertstand to convert table Sendables to MSDP TABLE, and/or just convert to ANSI concersion. Meanwhile the webclient protocol would understand other Sendable primitives, converting them to HTML elements as needed.  
- The protocol then sends this 'rendered' data to the client. 

### Discussion 

- This means changing the outgoing message path; including how `msg()` works (or introduce some lower-level send mechanism). This is a major backwards-incompatible change. 
- What about the incoming data? Should we allow custom machinery for converting advanced client concepts into Evennia-primitives as well? Initially, keeping the current `(funcname, (args), {kwargs})` may be enough. 
[...]

PR #3356, opened 2023-12-03
Rework Outgoing MessagePath

volundmush:
Added settings.SENDABLE_CLASS_MODULES for loading up (replaceable) Sendable classes from modules using callables_from_module. They're cached to evennia.SENDABLES (a dictionary)

Added `evennia.server.sendables` with some example Sendables.

Adds `.send()` method to Characters, Sessions, Accounts, Commands which handles the new outgoing path. This method takes a list of `Sendable` objects and a metadata dict to associate with the Sendables. Everything involved must be pickleable for this to work.

Altered the `.msg()` methods on Character, Sessions, Accounts, and Commands to convert their outgoing messages into a list of Sendables (the EvString placeholder and OOBFunc Sendable). These two ultimately are rendered exactly as existing Evennia messages are, for backwards compatibility.

Added a feature to AMP so it can actually send the Sendables. This process is actually a bit more efficient than the old way, since it can multi-cast the message to multiple sessids with a single AMP transmission.

Created a new PortalSession base class and made it customizable via class_from_module. It has methods to handle the Sendables.

Well, just gotta read over issue #3354 to get the picture.

Well I haven't exactly rewritten all of the documents and some more documentation is probably needed.


evennia/server/serversession.py
```diff
@@ -297,10 +294,28 @@ def msg(self, text=None, **kwargs):
...
-            self.data_out(text=text, **kwargs)
-        else:
-            self.data_out(**kwargs)
+            if not (isinstance(text, str) or isinstance(text, tuple)):
+                # sanitize text before sending across the wire
+                try:
+                    text = to_str(text)
+                except Exception:
+                    text = repr(text)
+            kwargs["text"] = text
```
Griatch, line 305: What if the data comes in as, say, a list rather than a tuple? There is a reason the original `clean_senddata` is so long; it handles a lot of edge cases in order to allow for pretty generous input formats (because it's easy for devs to miss this).

evennia/server/serversession.py
```diff
@@ -382,57 +397,6 @@ def at_cmdset_get(self, **kwargs):
...
-    # @ndb.deleter
-    def ndb_del(self):
-        """
-        Stop accidental deletion.
-
-        """
-        raise Exception("Cannot delete the ndb object!")
-
-    ndb = property(ndb_get, ndb_set, ndb_del)
-    db = property(ndb_get, ndb_set, ndb_del)
```
Griatch, line 434: Why are these removed?

evennia/utils/ravensgleaning.py
```diff
@@ -0,0 +1,441 @@
+"""
+This is a Python re-implementation of the ANSI Parser "RavensGleaning" written in JavaScript.
+
+It converts raw ANSI strings to HTML spans.
+
+# TODO: Add proper accreditation to the original Author, @PresJPolk
```
Griatch, line 6: We need proper licensing and accreditaion, yes. Also licensing, if any, needs to be made clear.

evennia/utils/ravensgleaning.py
```diff
@@ -0,0 +1,441 @@
+"""
```
Griatch, line 1: The name of this module means nothing for Evennia users. We should reference the original name and accredit properly in the docstring, but the module should be renamed to something more usable, such as `ansi2html.py` (so as to not collide with the existing `text2html.py`. 

Also, as far as I understand, this is more complete/produces better results than the old `text2html.py` implementation? Are there unit tests to prove this?

Griatch (changes requested):
I must say that this looks quite good overall. I appreciate your work on this. I have made many comments below; apart from minor things, like naming conventions and a few questions, we need extensive unit tests for this: 
- Unit tests for the new default sendable logics on the server side, including serialization/deserialization tests. 
- Unit tests for portal, making sure sendables output expected results (mocking out any AMP/wire calls, enough to test what is generated)
- Full unit test suite for the new `ansi-to-html` (`ravensgleanin.py`) module. 
- Also, legacy unit tests are not passing as it is; that needs fixing too; to make it  clear how backwards-compatible this is with the legacy message path. 
- Performance. We are still serializing to tuples across the AMP connection, so that should be fine - but is there any impact to serializing to/from a Sendable on each side? Would be good to have some comparisons or discussions around this.

Also, as a note (mainly to myself); if this merges; it will mean rewriting all documentation for the message path and how to extend these parts of Evennia.

evennia/utils/utils.py
```diff
@@ -3018,3 +3018,32 @@ def ip_from_request(request, exclude=None) -> str:
 
     logger.log_warn("ip_from_request: No valid IP address found in request. Using remote_addr.")
     return remote_addr
+
+
+def split_oob(data) -> ["any", dict]:
+    if isinstance(data, (tuple, list)) and len(data) == 2:
```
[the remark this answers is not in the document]
volundmush: I'm skeptical if server/utils.py is going to be any more informative than utils/utils.py when you get down to it...

volundmush:
Had to rebase due to merge conflicts.

evennia/server/serversession.py
volundmush: The madness of `clean_sendata` makes me want to cry. I'm not sure what to say here: I'd like to imagine that developers are reading how things work and not doing things they're not supposed to do. It's one thing to add some niceties like calling to_str or repr() on things which aren't strings, quite another to recurse arbitrary data structures to hell and back when people should have just sent the right to begin with.

I could quote the Zen of Python? Hmm.

Explicit is better than implicit.
Simple is better than complex.
Readability counts.
Special cases aren't special enough to break the rules.
Errors should never pass silently.
Unless explicitly silenced.
In the face of ambiguity, refuse the temptation to guess.

Hmmm. Maybe it would be better to start raising exceptions if bad data is fed into methods like this. But I'm trying to avoid breaking working APIs...

evennia/server/serversession.py
volundmush: I'm not sure why that happened. I think I was meaning to create some kind of mixin to deduplicate some code and I got distracted.

evennia/utils/ravensgleaning.py
volundmush: It is as close to perfect as you can get. Evennia's own implementation cannot handle arbitrary; it can only handle the output produced by ANSIString, which it does by pre-defining classes. RavensGleaning defines spans with inline styles specific to the ANSI state. So it's more accurate since it doesn't need predefined CSS classes.

evennia/settings_default.py
```diff
@@ -1167,7 +1171,10 @@
 # to tons of classes. You generally shouldn't need to touch protocols.
 BASE_SESSION_CLASS = "evennia.server.session.Session"
 
-# Telnet Protocol inherits from whatever above BASE_SESSION_CLASS is specified.
+# The base Portal Session class, which by default inherits from BASE_SESSION_CLASS.
+BASE_PORTAL_SESSION_CLASS = "evennia.server.portal.portalsession.PortalSession"
```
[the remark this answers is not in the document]
volundmush: No, because there is only ONE Server Session class for ALL Protocol Types... that wouldn't make any sense.

SERVER:
BASE_SESSION_CLASS -> SERVER_SESSION_CLASS

PORTAL:
BASE_SESSION_CLASS -> BASE_PORTAL_SESSION_CLASS -> <specific_protocol_class>

volundmush:
Geeze this has come a long way.

Griatch:
Thanks for your work on this, @volundmush! 
This will take a bit longer to review and and I think it needs to sit a bit in main before it goes into a release. So I'll review and hopefully merge this after I get Evennia 3.0 out.

volundmush:
I am starting to have second thoughts about the specific implementation of this. We should pow-wow and go over the details because yikes this is not something I want to screw up.

Griatch:
@volundmush I know you had doubts about this implementation, but I still think this Sendables concept is better than passing JSON around :/

Griatch:
@volundmush So where are we on this one?

PR #3358, opened 2023-12-06
Reset portal's `server_restart_mode` on server start

InspectorCaracal:
Currently, AMP sets an attribute of `server_restart_mode` on the portal to the restart mode when the server is being stopped - however, it never resets it on reconnect.

This just sets it to `None` on server connect regardless of mode.

Making that flag actually potentially useful. 👍


Griatch (approved):
Thanks!

PR #3365, opened 2023-12-12
Add a reference prefix setting to rpsystem

InspectorCaracal:
Changes the prefix marker to be settings-defined, with '/' as a fallback instead of hardcoded.

I also embedded the prefix into the docstring for CmdEmote to avoid the command's help text being confusing; not sure if this breaks any rules, though.

Convenience!


Griatch (changes requested):
Cool idea. You cannot define a class docstring as a dynamic f-string (or `.format`ed string) like this though. The `CmdEmote.__doc__` property of the class comes out empty.

What you need to do is to edit `CmdEmote.get_help` and in it do: 

```python 

class CmdEmote(RPCommand): 
     """ 
     ... 
    Example: 
        emote {prefix}me looks around 
        ...
     """

     def get_help(self, caller, cmdset): 
        return self.__doc__.format(prefix=_PREFIX)
```

InspectorCaracal:
I always forget `get_help` exists!

Griatch (approved):
LGTM!

PR #3366, opened 2023-12-13
Updating Component contrib - allow arbitrarily long subclass chains of Component

scyfris:

This change allows a subclass of Component to itself be subclassed.  This

would come in handy if a game wanted to make a custom base component
class that contains common functionality and further subclass those into
more specific components.  The base component class would inherit from
Component, and the subclasses would inherit from that base component
class.  Unfortunately this wasn't possible since get_component_class
(used internally by the component contrib) only considered one layer of
subclasses of Component and not a possible inheritance chain.

This change fixes that, and allows you to subclass Component subclasses.
Example is the following:
```
class MyGameBaseComponent(Component):
  # common definitions, data, etc

class MyGameComponentA(MyGameBaseComponent):
  # specific functionality for MyGameComponentA

class MyGameComponentA_inherit(MyGameBaseComponentA):
  # specific functionality for MyGameComponentA_inherit
```
Etc...

In my personal project I want a base Component class with some common stuff for my game, and wanted to derive all Components from that Base class.  This is similar to creating a game-specific BaseObject class which in turn you can subclass different Object types

### DIFF
--- a/evennia/contrib/base_systems/components/__init__.py
+++ b/evennia/contrib/base_systems/components/__init__.py
@@ -15,9 +15,16 @@ from evennia.contrib.base_systems.components.holder import (
     ComponentProperty,
 )
 
+# Recursively check inheritance chain of the input class.
+def all_component_subclasses(cls):
+    return set(cls.__subclasses__()).union(
+        [s for c in cls.__subclasses__() for s in all_component_subclasses(c)])
+
 
 def get_component_class(component_name):
-    subclasses = Component.__subclasses__()
+    
+    subclasses = all_component_subclasses(Component)
+
     component_class = next((sc for sc in subclasses if sc.name == component_name), None)
     if component_class is None:
         message = (
--- a/evennia/contrib/base_systems/components/dbfield.py
+++ b/evennia/contrib/base_systems/components/dbfield.py
@@ -76,10 +76,18 @@ class TagField:
         It is called with the component class and the name of the field.
         """
         self._category_key = f"{owner.name}::{name}"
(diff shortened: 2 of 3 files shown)


scyfris:
Looking at this a bit more - I _think_ the behaviour where TagField on a base class is not automatically added to a class when adding the component is incorrect (that is the behaviour of my current fix).
Instead I think that when adding a component that inherits from another component, the TagField of the parent components should _also_ be added to the typeclass that is getting the component attached - The "category" will be <base class>::<default val> anyways, so it will still be unique even if the subclass overrides the TagField.

ChrisLR:
I currently have some code refactoring that will allow this already, using a metaclass.
This PR will conflict with those changes so I suggest holding off on this for now.

ChrisLR:
Here, if you want to take a look

https://github.com/ChrisLR/evennia/pull/2/files#diff-33f8c9d4af17fd52b4aff00ca10ba1251f29df8d934ced36e9f5416da32edb62R10

It also contains some unrelated changes, I will make a proper clean PR soonish

scyfris:
@ChrisLR - Thanks! I'll hold off on my change then.  I also found some other issues when doing further testing on my change- such as TagField and DBField/NDBField aren't getting properly inherited, but I couldn't quite figure it out.

I'm glad to hear this is on your radar and I'll take a look at your change and close this pull request out for now.

ChrisLR:
@scyfris the PR is opened, still in draft because I need to update the readme and add tests but thanks for reminding me the changes are needed :smile:

PR #3367, opened 2023-12-14
[Components] Refactor Component registering with cherry-picked additions

ChrisLR:
**WIP**
This refactors Components to remove limitations and add some new functionality

- Registering a component is now done via Metaclass, which allows components to inherit other components
- Components may now specify a slot name, which is used to determine the attribute to access it from.
  This would let us have two components for the same feature, using a different name but the same slot name.
  Example: CharacterHealth, ItemHealth, they would both have a distinct name but use the slot health and both
  would be accessed using object.health or object.cmp.health
- ComponentHolders (Characters/Objects Etc) will also handle a bit better pulling components from inherited classes, 
  such as Mixins
- Components will by default use `__slots__` to be a bit lighter, as they are not intended to store data on themselves.
- DBFields will now auto-create correctly when they are added, but are autocreate=False by default until set, to preserve behavior

A good improvement to Components Contrib

### DIFF
--- a/evennia/contrib/base_systems/components/tests.py
+++ b/evennia/contrib/base_systems/components/tests.py
@@ -20,11 +20,23 @@ class ComponentTestA(Component):
     my_list = DBField(default=[], autocreate=True)
 
 
+class ShadowedComponentTestA(ComponentTestA):
+    name = "shadowed_test_a"
+    slot = 'ic_a'
+
+
 class InheritedComponentTestA(ComponentTestA):
     name = "inherited_test_a"
     slot = 'ic_a'
 
     my_other_int = DBField(default=2)
+
+
+class ReplacementComponentTestA(InheritedComponentTestA):
+    name = "replacement_inherited_test_a"
+    slot = "ic_a"
+
+    replacement_field = DBField(default=6)
 
 
 class ComponentTestB(Component):
@@ -50,11 +62,18 @@ class ComponentTestD(Component):
     mixed_in = DBField(default=8)
 
 
+class ShadowedCharacterMixin:
+    ic_a = ComponentProperty("shadowed_test_a")
+
+
 class CharacterMixinWithComponents:
+    ic_a = ComponentProperty("inherited_test_a", my_other_int=33)
     test_d = ComponentProperty('test_d')
 
 
-class CharacterWithComponents(ComponentHolderMixin, CharacterMixinWithComponents, DefaultCharacter):
+class CharacterWithComponents(
+    ComponentHolderMixin, ShadowedCharacterMixin, CharacterMixinWithComponents, DefaultCharacter
+):
     test_a = ComponentProperty("test_a")
     test_b = ComponentProperty("test_b", my_int=3, my_list=[1, 2, 3])
     ic_a = ComponentProperty("inherited_test_a", my_other_int=4)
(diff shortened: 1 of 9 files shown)


Griatch:
@ChrisLR Just leaving a comment (this is a draft so I won't review the code in detail): 
Since people is already using this contrib, refactoring needs to be done with backwards incompatibility in mind. And if backwards-compatibility is not 100% possible, the changes needs to be clearly documented. Normally, your old unit tests should still pass, and new tests added for any new functionalty. :)

ChrisLR:
@Griatch all changes here should be compatible so I suspect there's one or a couple bugs hiding in there, I will try to clean and finish this PR in the coming days/weeks.

ChrisLR:
I think the last thing I need to do here is the documentation

Griatch:
Ping.

ChrisLR:
This should be good for a first pass :)

evennia/contrib/base_systems/components/tests.py
```diff
@@ -50,73 +68,80 @@ class TestComponents(EvenniaTest):
...
-        assert self.char1.test_a.my_int == 1
-        assert self.char1.test_a.my_list == []
+        self.assertEquals(self.char1.test_a.my_int, 1)
+        self.assertEquals(self.char1.test_a.my_list, [])
 
     def test_character_assigns_default_provided_values(self):
-        assert self.char1.test_b.my_int == 3
-        assert self.char1.test_b.my_list == [1, 2, 3]
+        self.assertEquals(self.char1.test_b.my_int, 3)
+        self.assertEquals(self.char1.test_b.my_list, [1, 2, 3])
```
Griatch, line 92: `my_list` is configured with autocreate=True, but to make sure this is actually created in the db, one need to make sure this is actually available from the AttributeHandler as well.

Another classic test here is to make sure to check what happens if you do `self.char1.test_b.my_list.append(4)` and make sure behavior is as expected.

Griatch (changes requested):
Overall, I think these changes looks fine. It's hard to fully predict if it's backwards-compatible or not but it looks like the existing unit tests passes so that's good. 
One thing I'm wondering about (and why I marked this as 'request changes' is that I don't see additional unit tests several of the new features you are announcing with this refactor: 

- Ability to use the same name with different slots
- Components with associated cmdsets
- DbFields with autocreate=True (see cdommend)

I could be missing some of these being tested as part of other tests though, point it out in that case :)

Griatch:
@ChrisLR Ping.

ChrisLR:
@Griatch Sorry for the delay, I encountered a few additional bugs I've changed on my end and still testing.
I will remove the part about cmdsets as its easy enough for anyone to implement and introduces potential bugs.
I will also write a few additional tests as soon as I get some time.

ChrisLR:
@Griatch I have pushed the last fixes I made and put the PR back to draft until I get the tests written

evennia/contrib/base_systems/components/tests.py
ChrisLR: I am adding `test_character_has_autocreated_values ` in my next commit.
For the append test, I think `test_mutables_are_not_shared_when_autocreate` already tests that.

ChrisLR:
Regarding your comment "Ability to use the same name with different slots"
Its the opposite, a component is required to have a unique name.
The shared slot lets us have components that expose the same interface but behave differently.

For example, a Player class could have stats that are regular DBFields but a Mob class could have stats that are fixed and defined in code. Both would be accessed by using obj.stats.x.

An existing test was
`test_component_inheritance_assigns_proper_values`

I am adding tests
`test_component_inheritance_properly_overrides_slots`
`test_replacing_class_component_slot_with_runtime_component`

ChrisLR:
I continue to test these changes locally but I think this is ready for a 2nd review :)

Griatch (approved):
Thanks for the changes and explanations! LGTM.

Issue #3374, opened 2023-12-21
[BUG] CmdTunnel requires settings.py has CMD_IGNORE_PREFIXES = "@"


chromancer:
#### Describe the bug
@tunnel under the hood is really just a specific use case of @dig, and so it uses @dig and simply rearranges the arguments and creates reciprocal exits by boxing the compass. However the implementation fails to include the prefix '@' and that is part of the base key of CmdDig, so if CMD_IGNORE_PREFIXES does not minimally include @, the command it formats for you is not recognized.

#### To Reproduce
1. Change settings.py to CMD_IGNORE_PREFIXES = ""
2. Reload the server
3. Try '@tunnel n' or any cardinal direction

#### Expected behavior
@tunnel should correctly call @dig rather than dig

#### Environment, Evennia version, OS etc
Evennia 2.3.0

#### Additional context
Add the prefix @ to line 1107 of evennia/commands/default/building.py

Further considerations: currently even though the keys of most evennia system commands use the prefix @—e.g. @dig, @link, @examine, @tunnel—the docstrings for each function show no prefix, which is strictly wrong and will cause confusion, except that settings.py ships with a setting turned on that automatically ignores all command prefixes, rendering the issue moot. I suggest changing each docstring to use the prefix as default, even if the default settings are to ignore it. Prefixes are desirable because OOC commands need to be disambiguated from potential IC commands, anyway: dig, lock, open, and other common verbs are often used on MUDs to take IC actions.

Griatch:
Closed in https://github.com/evennia/evennia/pull/3375

PR #3375, opened 2023-12-21
Add prefix to CmdTunnel implementation

chromancer:
Adds command prefix for @dig to @tunnel's function

Should fix issue #3374


Griatch (approved):
This is an interesting case/bug. Nice catch.

PR #3377, opened 2023-12-22
Update character_creator contrib

InspectorCaracal:
Properly updates the `character_creator` contrib for the recent refactor of the character creation system, as well as re-applies the feature respecting START_LOCATION, which has apparently been missing/broken for some time.

The unit tests have also been slightly modified/extended to more accurately match and test the contrib's functionality.

Bug fixing, maintenance


Griatch (approved):
Not super-familiar with this contrib, but it looks ok!

PR #3382, opened 2023-12-27
Start global scripts

InspectorCaracal:
In the recent server startup refactor, the line initializing the global scripts appears to have been lost. This just adds it back.

Bug fixing

### DIFF
--- a/evennia/server/service.py
+++ b/evennia/server/service.py
@@ -481,6 +481,9 @@ class EvenniaServerService(MultiService):
         # always call this regardless of start type
         self.at_server_start()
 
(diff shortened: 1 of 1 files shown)


Griatch (commented):
Hm, are these not started from another location now? :thinking: ButI presume this fix comes from seeing global scripts not starting, so let's add it then.

InspectorCaracal:
Yeah, Aleph reported them not starting and I scoured the code looking for where they're started to find the problem and it turned out the call to start had just gotten lost entirely. 👍

PR #3388, opened 2024-01-02
Fix tutorial

gas-public-wooden-clean:
- Align the code in the character generation tutorial documentation (which doesn't work) with the actual code in the referenced contrib (which does).
- Fix typos in documentation.

I got confused when the documentation code didn't work for me.

This is my first PR. Please be more picky than usual so I can learn what's expected.


docs/source/Howtos/Beginner-Tutorial/Part3/Beginner-Tutorial-Chargen.md
```diff
@@ -354,7 +354,12 @@ def start_chargen(caller, session=None):
     # this generates all random components of the character
     tmp_character = TemporaryCharacterSheet()
 
-    EvMenu(caller, menutree, session=session, tmp_character=tmp_character)
+    EvMenu(
+        caller,
+        menutree,
+        session=session,
+        startnode_input=("sgsg", {"tmp_character": tmp_character}),
```
gas-public-wooden-clean, line 361: Passing tmp_character via kwargs would work and would then be accessible as `caller.ndb._evmenu.tmp_character, but node_chargen tries to get it from its kwargs. I'm not actually sure what the "sgsg" is doing here or if it's necessary. I just made it match the actual code in the contrib, which worked in my testing.

Griatch: The `startnode_input` mimics calling the start node as `startnode(caller, "sgsg", tmp_character=tmp_character)`. The string argument is the node-text, normally the input from the previous node, but since it's the start node, I just put a random string there that turned out to be `"sgsg"`. It should have been just `""`, but I might have been unsure if the string needed to be truthy when I wrote the code and forgot to check later :). 

I will merge this fix, but rewrite it later to use just the `tmp_character=...` kwarg in both places since that is much more understandable.

Griatch (approved):
Good fixes, and a nice observation about the tutorial code; I'll rewrite it in both places to use the kwarg directly instead of the `startnode_input` kwarg, it's easier to use and explain.

PR #3389, opened 2024-01-02
Refactor CI for postgres

ChrisLR:
PSQL tests usually fail after 30 minutes, looking at the actions file I saw the pg version is very outdated.

Fixing the CI


ChrisLR:
@Griatch the only change worth keeping so far is using postgres by declaring it in services instead of using another action, it should let us provide a conf to postgres more easily, I did research a few ways to make the tests faster, using one or a few parameters mentioned here

https://www.postgresql.org/docs/current/non-durability.html

However I don't understand why the pg tests would be that slow compared to other databases so I would have to investigate more.

ChrisLR:
Its on the TODO pile, I'll come back to it :sweat_smile:

Griatch:
Ping. :)

Griatch:
@ChrisLR Another ping

ChrisLR:
@Griatch Still experimenting, unsure why having the two pg tests run in parallel makes one of them fail while one after the other works

Griatch:
@ChrisLR Added Python3.12 to the CI matrix, and the postgres issue remains :disappointed:

ChrisLR:
@Griatch using the code from this PR ? :(

Griatch:
Nono, just adding a data point since you were looking at this.

ChrisLR:
@Griatch :relieved: great, so there's a chance this PR is sufficient, I will add 3.12 to it and see

Griatch:
@ChrisLR This is still marked as a Draft. Is this still being worked on?

ChrisLR:
@Griatch sorry about the delay, yes this is still on my plate.

ChrisLR:
No longer worked after resolving conflicts.
It likely didn't solve the root issue.
Closing for now.

PR #3394, opened 2024-01-06
Fixing Replace Legacy Time of Day Markup

### LINKED ISSUE #3340: [BUG] Expanded Room time-of-day tags do not update until post-character entry
#### Describe the bug
When entering an Expanded Room, characters will see an out-of-date description if time has reached a new tag but the room hasn't updated yet.

#### To Reproduce
1. Use Expanded Rooms contrib
2. Create a room with time of day tags.
3. Visit room after time-of-day change.

#### Expected behavior
Rooms should update before a character enters so that the description is not incorrect.

image

jaborsh:
1. Renamed the loop variable in the for loop to tod to avoid shadowing the current_time_of_day.
2. Used a loop for applying the regex substitutions to make the code more concise and easier to maintain. This change fixed the main issue, ensuring that current_time_of_day retains the correct value from self.get_time_of_day() throughout the function.
3. Ensured that the current_time_of_day is correctly used for the regex lookup.

Prior implementation did not correctly update room description on first instance of call, only on subsequent iterations. With this change, entering/looking at a room now displays the proper description instead of one that is potentially incorrect.

Resolves #3340

### DIFF
--- a/evennia/contrib/grid/extended_room/extended_room.py
+++ b/evennia/contrib/grid/extended_room/extended_room.py
@@ -47,6 +47,7 @@ from collections import deque
 
 from django.conf import settings
 from django.db.models import Q
+
 from evennia import (
     CmdSet,
     DefaultRoom,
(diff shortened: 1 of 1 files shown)


Griatch (approved):
Aha, nice fix!

PR #3397, opened 2024-01-09
Fixes incorrect variable name in a doc string

bradleymarques:

Simply fixes a mistake in a variable name in a doc string.

Before: `pre_logout_location`

After: `prelogout_location`

This is how the variable is actually used in the code:

```py
# evennia/objects/objects.py

def at_pre_puppet(self, account, session=None, **kwargs):
       ...
    if self.location is None:
        # Make sure character's location is never None before being puppeted.
        # Return to last location (or home, which should always exist)
        location = self.db.prelogout_location if self.db.prelogout_location else self.home
```          
          
> My goals are beyond your understanding.

Nah, just kidding. I got tripped up with this one, and want to fix so that other people don't make the same mistake :)

No


Griatch (approved):
Honestly, `pre_logout_location` is a better and more consistent name than "prelogout_location" ... but yeah, it's not what's used in code, so this is a good fix.

PR #3398, opened 2024-01-09
Fix rpsystem language syntax to be consistent with documentation

InspectorCaracal:
Corrects the regex for identifying language keys, as well as expands slightly on speaking in languages in the main documentation and adds a unit test case.

Bug fixing


bradleymarques (commented):
Thanks for doing this so quickly @InspectorCaracal :)

Griatch (approved):
Thanks for the fix!

PR #3414, opened 2024-01-19
Fixed get_numbered_name decoloration.

### LINKED ISSUE #3413: [BUG] Color is lost with `get_numbered_name`
#### Describe the bug
Color is lost with `get_numbered_name`

#### To Reproduce
1. Create item. Make name something like `|yChest|n`
2. Put in room
3. See error

#### Expected behavior
Function maintains coloring.

jaborsh:
The logic for `get_numbered_name` has been rewritten so that _INFLECT does not remove colors from singular & pluralized names.

Closes #3413

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Main branch
+
+- [Feature][pull3470]: New `exit_order` kwarg to
+  `DefaultObject.get_display_exits` to easier customize the order in which
+  standard exits are displayed in a room (chiizujin)
+
+[pull3470]: https://github.com/evennia/evennia/pull/3470
+
 ## Evennia 4.1.1
 
 April 6, 2024


jaborsh:
Note: In my fix, I've included regular expressions for @InspectorCaracal's Hex implementation which follows the syntax `|#RRGGBB`, but it's addition here does not diminish the overall integrity of the function.

Griatch:
Thanks for working on this! I have looked over this, and I need to ponder if this is the way to go about it though ... this re-implements |-style parsing in a second place, which breaks DRY principles and feels a little hackish to me ... not sure I have a better alternative atm, but will wait to merge this a bit, nevertheless.

InspectorCaracal:
Not as a solution, but as an alternative that might help spark a better idea: for my games, I've written the following function that I use pretty much everywhere. It's not a fantastic solution, due to its dependence on adding a temporary "marker" character that you don't really have a guarantee won't be included in someone's just normal string, which is why I only use it personally and so haven't bothered making it "nice".

```py
re_colorful = re.compile(r'\|.+?\|n')
re_articles = re.compile(r'^(a|an|and|the)(\s+)', re.IGNORECASE)

def numbered_name(name, count, pair=False, pluralize=True, cap=False, prefix=True):
	name = re_articles.sub('', name)
	colorlist = re_colorful.findall(name)
	colorkeys = {}
	for colorword in colorlist:
		index = strip_ansi(colorword)+'@'
		colorkeys[index] = colorword
		name = name.replace(colorword, index)
	if count == 1 and prefix:
		name = INFLECT.an(name)
	else:
		if not count:
			num = "no"
		elif pair and count == 2 and not name.startswith("pair of"):
			num = "a pair of"
		else:
			num = INFLECT.number_to_words(count)
		if pluralize:
			name = INFLECT.plural_noun(name) or name
		if prefix:
			name = f"{num} {name}"

	if cap:
		name = name[0].upper() + name[1:]

	for index, colorword in colorkeys.items():
		name = name.replace(index, colorword)

	return name
```

PR #3421, opened 2024-02-04
Removes need for hardcoding line breaks into display hooks

InspectorCaracal:
This adds a new utility function, `strip_extra_whitespace`, which compresses down repeating sequences of whitespace, and uses it instead of `.strip()` as the default appearance format method in `DefaultObject.format_appearance`

This change removes the need to have the exit/thing/character items on the same line in the appearance template, which also removes the need to have the line breaks for those items included in their return value.

The current method makes using the appearance template to reorder items more work than it should be, since the get_display methods themselves still need to be overloaded. This change means you can now just change the appearance template itself, as intended.

Additionally, the whitespace function allows you to define your maximum preferred spacing, which is then easily overridden via the `format_appearance` method.

### DIFF
--- a/evennia/utils/utils.py
+++ b/evennia/utils/utils.py
@@ -472,6 +472,31 @@ def iter_to_str(iterable, sep=",", endsep=", and", addquote=False):
 list_to_string = iter_to_str
 iter_to_string = iter_to_str
 
+re_empty = re.compile("\n\s*\n")
+
+def strip_extra_whitespace(text, max_linebreaks=1, max_spacing=2):
+    """
+    Removes extra sequential whitespace in a block of text. This will also remove any trailing
+    whitespace at the end.
+
+    Args:
+        text (str):   A string which may contain excess internal whitespace.
+        
+    Keyword args:
+        max_linebreaks (int):  How many linebreak characters are allowed to occur in a row.
+        max_spacing (int):     How many spaces are allowed to occur in a row.
+
+    """
+    text = text.rstrip()
+    # replaces any non-visible lines that are just whitespace characters with actual empty lines
+    # this allows the blank-line compression to eliminate them if needed
+    text = re_empty.sub("\n\n", text)
+    # replace groups of extra spaces with the maximum number of spaces
+    text = re.sub(f" {{{max_spacing},}}", " "*max_spacing, text)
+    # replace groups of extra newlines with the maximum number of newlines
+    text = re.sub(f"\n{{{max_linebreaks},}}", "\n"*max_linebreaks, text)
+    return text
+    
 
 def wildcard_to_regexp(instring):
     """
--- a/evennia/utils/utils.py
+++ b/evennia/utils/utils.py
@@ -474,14 +474,14 @@ iter_to_string = iter_to_str
 
 re_empty = re.compile("\n\s*\n")
 
-def strip_extra_whitespace(text, max_linebreaks=1, max_spacing=2):
+def compress_whitespace(text, max_linebreaks=1, max_spacing=2):
     """
     Removes extra sequential whitespace in a block of text. This will also remove any trailing
     whitespace at the end.
 
     Args:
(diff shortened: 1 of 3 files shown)


InspectorCaracal:
A note re: the most recent commit: `strip_extra_whitespace` deliberately does not do a full .strip() on the input text in order to support indented first lines, so I had to add the .strip() back on in `format_appearance` since I'd forgotten `appearance_template` begins with a blank newline. There's other ways to account for that, I think, but that still seems like the most logical and extensible place to be doing it.

evennia/utils/utils.py
```diff
@@ -472,6 +472,31 @@ def iter_to_str(iterable, sep=",", endsep=", and", addquote=False):
 list_to_string = iter_to_str
 iter_to_string = iter_to_str
 
+re_empty = re.compile("\n\s*\n")
+
+def strip_extra_whitespace(text, max_linebreaks=1, max_spacing=2):
```
Griatch, line 477: I think the naming of this function is confusing, the `strip` name does not make it clear this is also removing whitespace _inside_ a string. Maybe a name like `limit_whitespace_in_string` would be clearer?

evennia/utils/utils.py
```diff
@@ -472,6 +472,31 @@ def iter_to_str(iterable, sep=",", endsep=", and", addquote=False):
...
+    Removes extra sequential whitespace in a block of text. This will also remove any trailing
+    whitespace at the end.
+
+    Args:
+        text (str):   A string which may contain excess internal whitespace.
+        
+    Keyword args:
+        max_linebreaks (int):  How many linebreak characters are allowed to occur in a row.
+        max_spacing (int):     How many spaces are allowed to occur in a row.
+
```
Griatch, line 488: I think it would be useful with an `Example:` block here: 

```
Example:
::
    limit_whitespace_in_string("""
         A long    string     with whitespaces 


         A line separated      by   many line breaks 
         """, max_linebreaks=2, max_spaces=1)
becomes 
:: 
    """
    A long string with whitespaces
          
    A line separated by many line breaks"""
```

evennia/utils/utils.py
```diff
@@ -472,6 +472,31 @@ def iter_to_str(iterable, sep=",", endsep=", and", addquote=False):
...
+    """
+    text = text.rstrip()
+    # replaces any non-visible lines that are just whitespace characters with actual empty lines
+    # this allows the blank-line compression to eliminate them if needed
+    text = re_empty.sub("\n\n", text)
+    # replace groups of extra spaces with the maximum number of spaces
+    text = re.sub(f" {{{max_spacing},}}", " "*max_spacing, text)
+    # replace groups of extra newlines with the maximum number of newlines
+    text = re.sub(f"\n{{{max_linebreaks},}}", "\n"*max_linebreaks, text)
+    return text
```
Griatch, line 498: This function will ruin right-sided indent. It could be interesting to mention this in the docs (or even add support for retaining justification, only removing whitespace inside the indented text block).

Griatch (changes requested):
A good idea that should make the override easier. Also great with accompanying unit test. My main comment is that this could be even more useful if it retained indents.

evennia/utils/utils.py
InspectorCaracal: Ahaha, I've been trying to figure out why this happens in my game code and somehow never realized it's here! I'll update the function to retain leading indent spaces.

evennia/utils/utils.py
InspectorCaracal: I couldn't figure out how to write an effective and legible example that didn't break the docstring.

Griatch (approved):
I think this looks good now. Thanks for the fixes and extra unit tests!

evennia/utils/utils.py
Griatch: Ok, I'll investigate.

PR #3423, opened 2024-02-09
Fix underlying logic for moving in and out of Wilderness areas

InspectorCaracal:
The current logic for moving into a wilderness room completely falls apart if the room already exists - most notably, the object-leave hooks for any prior locations never get checked. Additionally, it contains a lot of redundant checks.

This fixes the logic and makes it more concise.

Bug fixing.


Griatch (approved):
LGTM

PR #3425, opened 2024-02-11
Override example crafting recipes

InspectorCaracal:
Replaces the example recipes if you've set your own recipe modules, rather than forcing people's games to either include them or go out of their way to hack them back out.

I need to not have these recipes in my game. >.>


InspectorCaracal:
I know this is not exactly kosher but I found another bug in the crafting contrib just now and decided it would be easier for everyone to put the fix in here rather than open another tiny PR, since they're both the same contrib.

Griatch (approved):
Thanks for the fixes. Makes sense. 

Also fine to add the bug-fix since it's a small one (hint: you can click on lines in your own PR and add line comments - very useful for explaining parts of your PR code you think the reviewer may wonder about).

Issue #3432, opened 2024-02-19
[BUG] Plural aliasing fails to match in stacked/quiet searches


InspectorCaracal:
#### Describe the bug
Something in the way Evennia adds plural aliases to objects, or possibly the way it searches for them, is breaking search results and only returning one match, even when `stacked` or `quiet` are set.

#### To Reproduce
Create 3 items named "rock" in your location and then enter the following py lines
```
py self.search('rock', quiet=True)
>>> self.search('rock', quiet=True)
['rock', 'rock', 'rock']
py self.search('rocks', quiet=True)
>>> self.search('rocks', quiet=True)
['rock']
```

#### Expected behavior
If the auto-generated alias "rocks" is expected to be usable at all, it should match exactly the same search results as the regular key "rock".

#### Environment, Evennia version, OS etc
evennia 3.1.1

Griatch:
Looks like the objects `.get_numbered_name(num, looker)` must have been called for the `two rocks` alias to be created, which is what makes `self.search("rocks")` work. 

I suspect it's only called on the first rock normally (the others are stacked with it so they are not treated as individual objects). That's why there's only one return in the `search("rocks")` case. 

I'm reluctant to always create these aliases for all objects on-creation, but unless we device some other, more programmatic way to query for plural forms, it's probably what  needs to happen. :thinking:

InspectorCaracal:
Yeah, I puzzled over how exactly to write up this issue for a while because of exactly that. But I think at the moment the expectation of how it should work is *probably* to add the plural aliases to everything, mostly because of the docstring for `get_numbered_name`:
https://github.com/evennia/evennia/blob/0593144e8dd9daae69bba659c6f3d0a0868745ea/evennia/objects/objects.py#L1447-L1448

Griatch:
An alternative here is to maybe expand the `evennia.utils.strip_partial_matching` which is what is used to do partial matches (like finding out that `sw` works as a shortcut for `sword` as long as it becomes a unique match among the match candidates). Here, `rocks` doesn't register as a valid partial match for `rock` since it's longer. 

But maybe it could be a special case ... then there would be no need for an explicit alias (at least not for this use case). Then again, english plural forms are not always easy to generalize :sweat_smile:

PR #3433, opened 2024-02-19
Add the ability to get/give/drop stacks of things

InspectorCaracal:
This adds a barebones syntax for pulling the number out of a command argument on get, drop, and give - and more significantly, makes the command funcs therein able to handle moving around multiple objects.

Being able to get, give, and drop multiple of an object is an extremely common and fairly fundamental feature that the vast majority of people using Evennia wind up implementing. I consider it's nearly as fundamental as multimatching, honestly. And since the object search itself supports stacking, it seems reasonable that these basic object manipulation commands should also support stacking.

This doesn't work as smoothly as it should, at the moment, due to the bug described in #3432. Since the pluralization search isn't working correctly, for whatever reason, `get 3 rocks` will only ever get one rock. However, fixing that bug is outside of the scope of this PR, and `get 3 rock` works correctly.

Also, the `get` command had its help file locked to Developer, for some reason, so I fixed that in this PR. If for whatever reason this PR is rejected, I'll put in a smaller patch to fix that separately.

### DIFF
--- a/evennia/commands/default/general.py
+++ b/evennia/commands/default/general.py
@@ -379,8 +379,58 @@ class CmdInventory(COMMAND_DEFAULT_CLASS):
             string = f"|wYou are carrying:\n{table}"
         self.msg(text=(string, {"type": "inventory"}))
 
+class NumberedTargetCommand(COMMAND_DEFAULT_CLASS):
+    """
+    A class that parses out an optional number component from the input string. This
+    class is intended to be inherited from to provide additional functionality, rather
+    than used on its own.
 
-class CmdGet(COMMAND_DEFAULT_CLASS):
+    Note that the class's __doc__ string (this text) is used by Evennia to create the
+    automatic help entry for the command, so make sure to document consistently here. 
+    """
+    def parse(self):
+        """
+        This method is called by the cmdhandler once the command name
+        has been identified. It creates a new set of member variables
+        that can be later accessed from self.func() (see below)
+
+        The following variables are available for our use when entering this
+        method (from the command definition, and assigned on the fly by the
+        cmdhandler):
+            self.key - the name of this command ('look')
+            self.aliases - the aliases of this cmd ('l')
+            self.permissions - permission string for this command
+            self.help_category - overall category of command
+
+            self.caller - the object calling this command
+            self.cmdstring - the actual command name used to call this
+                             (this allows you to know which alias was used,
+                             for example)
+            self.args - the raw input; everything following self.cmdstring.
+            self.cmdset - the cmdset from which this command was picked. Not
+                          often used (useful for commands like 'help' or to
+                          list all available commands etc)
+            self.obj - the object on which this command was defined. It is often
+                       the same as self.caller.
+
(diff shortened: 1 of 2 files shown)


InspectorCaracal:
Realized slightly belatedly I should add tests for the stacking syntax, so there you go.

evennia/commands/default/general.py
```diff
@@ -387,48 +387,72 @@ class CmdGet(COMMAND_DEFAULT_CLASS):
...
       get <obj>
 
-    Picks up an object from your location and puts it in
-    your inventory.
+    Picks up an object from your location and puts it in your inventory.
     """
 
     key = "get"
     aliases = "grab"
-    locks = "cmd:all();view:perm(Developer);read:perm(Developer)"
```
Griatch, line 396: I honestly don't remember why this was :thinking: It seems pretty deliberate though :shrug:

evennia/commands/default/general.py
```diff
@@ -387,48 +387,72 @@ class CmdGet(COMMAND_DEFAULT_CLASS):
...
+        super().parse()
+        self.number = 0
+        if self.args:
+            # check for numbering
+            count, *args = self.args.split(maxsplit=1)
+            # we only use the first word as a count if it's a number and
+            # there is more text afterwards
+            if args and count.isdecimal():
+                self.number = int(count)
+                self.args = args[0]
```
Griatch, line 408: This same block is repeated over and over.
- Maybe commands with item targeting like this should all share the same parent class. 
- Another option would be to add this kind of functionality to the `object.search()`, but possibly that makes it harder to customize.

Griatch (changes requested):
I think I agree with you that this kind of stack manipulation makes sense to have in core. If deciding to take that step though, it feels like we could help the dev more than to just make custom additions to a few general commands: 

The `obj.search()` method is used in almost all commands. Even today, with `quiet=True`, this can return a list. It already supports custom query input to separate matches (like `apple-4`), so maybe it could be made to also support `two apple` or `2 apple` in a general form (as you say, `two apples` is another issue, but maybe it's time for a general refactor of the search to be more naturalistic ...). I'm open to discussion. 

Alternatively, even with the per-command fixes, I think we could avoid a lot of code repetition here.

Griatch:
@InspectorCaracal Ping :)

Griatch:
Ping.

InspectorCaracal:
I haven't forgotten about this, but I haven't come to a decision about what to do re: your comments. A more fundamental change to searching and/or command targeting *would* be a better solution, but I don't think I'm the right person to tackle it.

Griatch:
@InspectorCaracal That's fine. We could move a potential search refactor into a feature request instead. I think we could still get away with less repetition here though.

InspectorCaracal:
I've rebased the PR to the current main, pulled out the number parsing portion into a new parent command class, `NumberedTargetCommand`, and also updated the three commands to use the nice new `return_string` keyword in `get_numbered_name`. Once the tests finish, if there's no new issues, it should be good to go :+1:

Griatch (approved):
Looking cleaner!

PR #3434, opened 2024-02-21
Patches to improve "help" usability

InspectorCaracal:
Changes the lunr weightings to prevent command aliases from taking matching precedence to command keys. I'm not sure why having the boost weights so close was causing this effect, but it does.

Also removes the "building" default filehelp entry, since it conflicts with the "building" category and is thus inaccessible and confusing as well as being a not-very-useful help file anyway. The existing "evennia" entry is sufficient of an example as well as being potentially useful.

mostly bug fixing

### DIFF
--- a/evennia/commands/default/help.py
+++ b/evennia/commands/default/help.py
@@ -480,9 +480,9 @@ class CmdHelp(COMMAND_DEFAULT_CLASS):
             # lunr search fields/boosts
             search_fields = [
                 {"field_name": "key", "boost": 10},
-                {"field_name": "aliases", "boost": 9},
-                {"field_name": "no_prefix", "boost": 8},
-                {"field_name": "category", "boost": 7},
+                {"field_name": "aliases", "boost": 7},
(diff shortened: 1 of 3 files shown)


evennia/commands/default/help.py
```diff
@@ -480,9 +480,9 @@ def do_search(self, query, entries, search_fields=None):
             # lunr search fields/boosts
             search_fields = [
                 {"field_name": "key", "boost": 10},
-                {"field_name": "aliases", "boost": 9},
-                {"field_name": "no_prefix", "boost": 8},
-                {"field_name": "category", "boost": 7},
+                {"field_name": "aliases", "boost": 7},
+                {"field_name": "no_prefix", "boost": 6},
+                {"field_name": "category", "boost": 5},
```
Griatch, line 485: Tuning these weights are really more a matter of black magic ...

Griatch (approved):
Fair enough, if tuning the weights is enough to resolve the issues, let's go for it!

PR #3446, opened 2024-03-18
Change comparison for singular vs plural

InspectorCaracal:
The form used for describing the absence of a noun is plural, not singular. e.g. "no apples" rather than "no apple".


Griatch (approved):
Hm, I suppose that's true, didn't know this was a grammatical, general rule though.

PR #3447, opened 2024-03-21
Achievements contrib

InspectorCaracal:
Adds a new contrib which provides a mechanism for tracking and checking game achievements.

### DIFF
--- a/evennia/contrib/game_systems/achievements/README.md
+++ b/evennia/contrib/game_systems/achievements/README.md
@@ -24,9 +24,9 @@ To allow players to check their achievements, you'll also want to add the `achie
 from evennia.contrib.game_systems.achievements.achievements import CmdAchieve
 
 class CharacterCmdSet(default_cmds.CharacterCmdSet):
-	key = "DefaultCharacter"
+    key = "DefaultCharacter"
 
-	def at_cmdset_creation(self):
+    def at_cmdset_creation(self):
         # ...
         self.add(CmdAchieve)
 ```
@@ -191,11 +191,11 @@ A utility function for retrieving a specific achievement's data from the achieve
 from evennia.contrib.game_systems.achievements import get_achievement
 
 def toast(achiever, completed_list):
-	if completed_list:
+    if completed_list:
         # `completed_data` will be a list of dictionaries - unrecognized keys return empty dictionaries
-		completed_data = [get_achievement(key) for key in args]
-		names = [data.get('name') for data in completed]
-		achiever.msg(f"|wAchievement Get!|n {iter_to_str(name for name in names if name)}"))
+        completed_data = [get_achievement(key) for key in args]
+        names = [data.get('name') for data in completed]
+        achiever.msg(f"|wAchievement Get!|n {iter_to_str(name for name in names if name)}"))
 ```
 
 #### `search_achievement`
--- a/evennia/contrib/game_systems/achievements/README.md
+++ b/evennia/contrib/game_systems/achievements/README.md
@@ -1,0 +1,119 @@
+# Achievements
+
+A simple, but reasonably comprehensive, system for tracking achievements. Achievements are defined using ordinary Python dicts, reminiscent of the core prototypes system, and while it's expected you'll use it only on Characters or Accounts, they can be tracked for any typeclassed object.
+
+## Installation
+
+Once you've defined your achievement dicts in a module, assign that module to the `ACHIEVEMENT_CONTRIB_MODULES` setting in your settings.py
+
+Optionally, you can specify what attribute achievement progress is stored in, with the setting `ACHIEVEMENT_CONTRIB_ATTRIBUTE`. By default it's just "achievements".
+
+There is also a command available to let players check their achievements - `from evennia.contrib.game_systems.achievements.achievements import CmdAchieve` and then add `CmdAchieve` to your default Character and/or Account cmdsets.
+
+#### Settings examples
+
+One module:
+
+```python
+# in server/conf/settings.py
+
+ACHIEVEMENT_CONTRIB_MODULES = "world.achievements"
+```
+
+Multiple modules, with a custom-defined attribute:
+
+```python
+# in server/conf/settings.py
+
+ACHIEVEMENT_CONTRIB_MODULES = ["world.crafting.achievements", "world.mobs.achievements"]
+ACHIEVEMENT_CONTRIB_ATTRIBUTE = "achieve_progress"
+```
+
+A custom-defined attribute including category:
+
+```python
+# in server/conf/settings.py
+
+ACHIEVEMENT_CONTRIB_MODULES = "world.achievements"
+ACHIEVEMENT_CONTRIB_ATTRIBUTE = ("achievements", "systems")
+```
+
+## Usage
+
+### Defining your achievements
+
+This achievement system is designed to use ordinary dicts for the achievement data - however, there are certain keys which, if present in the dict, define how the achievement is progressed or completed.
+
+- **name** (str): The searchable name for the achievement. Doesn't need to be unique.
+- **desc** (str): A longer description of the achievement. Common uses for this would be flavor text or hints on how to complete it.
+- **category** (str): The type of actions this achievement tracks. e.g. purchasing 10 apples might have a category of "buy", or killing 10 rats might have a category of "defeat".
+- **tracking** (str or list): The *specific* things this achievement tracks. e.g. the previous example of killing rats might have a `"tracking"` value of `"rat"`. An achievement can also track multiple things: instead of only tracking buying apples, you might want to track apples and pears. For that situation, you can assign it to a list of values to check against: e.g. `["apple", "pear"]`
+- **tracking_type** (str): *Default value if unset: `"sum"`* There are two valid tracking types: "sum" (which is the default) and "separate". `"sum"` will increment a single counter every time any of the tracked items match. `"separate"` will have a counter for each individual item in the tracked items. (This is really only useful when `"tracking"` is a list.)
+- **count** (int): *Default value if unset: 1* The number of tallies this achievement's requirements need to build up in order to complete the achievement. e.g. the previous example of killing rats would have a `"count"` value of `10`. For achievements using the "separate" tracking type, *each* item being tracked must tally up to this number to be completed
+- **prereqs** (str or list): The *keys* of any achievements which must be completed before this achievement can start tracking progress. An achievement's key is the variable name it's assigned to in your achievement module.
+
+You can add any additional keys to your achievement dicts that you want, and they'll be included with all the rest of the achievement data when using the contrib's functions. This could be useful if you want to add extra metadata to your achievements for your own features.
+
+#### Examples
+
+A simple achievement which you can get just for logging in the first time. This achievement has no prerequisites and it only needs to be fulfilled once to complete, so it doesn't need to define most of the keys.
+```python
+# This achievement has the unique key of "FIRST_LOGIN_ACHIEVE"
+FIRST_LOGIN_ACHIEVE = {
+    "name": "Welcome!", # the searchable, player-friendly display name
+    "desc": "We're glad to have you here.", # the longer description
+    "category": "login", # the type of action this tracks
+    "tracking": "first", # the specific login action
+}
+```
+
+An achievement for killing 10 rats, and another for killing 10 *dire* rats which requires the "kill 10 rats" achievement to be completed first.
+```python
+ACHIEVE_TEN_RATS = {
+    "name": "The Usual",
+    "desc": "Why do all these inns have rat problems?",
+    "category": "defeat",
+    "tracking": "rat",
+    "count": 10,
+}
+
+ACHIEVE_DIRE_RATS = {
+    "name": "Once More, But Bigger",
+    "desc": "Somehow, normal rats just aren't enough any more.",
+    "category": "defeat",
+    "tracking": "dire rat",
+    "count": 10,
+    "prereqs": "ACHIEVE_TEN_RATS",
+}
+```
+
+An achievement for buying 5 each of apples, oranges, and pears.
+```python
+FRUIT_BASKET_ACHIEVEMENT = {
+    "name": "A Fan of Fruit", # note, there is no desc here - that's allowed!
+    "category": "buy",
+    "tracking": ("apple", "orange", "pear"),
+    "count": 5,
+    "tracking_type": "separate",
+}
+```
+
+### `track_achievements`
+
+The primary mechanism for using the achievements system is the `track_achievements` function. In any actions or functions in your game's mechanics which you might want to track in an achievement, add a call to `track_achievements` to update the achievement progress for that individual.
+
+For example, you might have a collection achievement for buying 10 apples, and a general `buy` command players could use. In your `buy` command, after the purchase is completed, you could add the following line:
+
+```python
+    track_achievements(self.caller, "buy", obj.name, count=quantity)
+```
+
+In this case, `obj` is the fruit that was just purchased, and `quantity` is the amount they bought.
+
+The `track_achievements` function does also return a value: an iterable of keys for any achievements which were newly completed by that update. You can ignore this value, or you can use it to e.g. send a message to the player with their latest achievements.
+
+### The `achievements` command
+
+The contrib's provided command, `CmdAchieve`, aims to be usable as-is, with multiple switches to filter achievements by various progress statuses and the ability to search by achievement names.
+
+To make it easier to integrate into your own particular game (e.g. accommodating some of that extra achievement data you might have added), the code for formatting a particular achievement's data for display is in `CmdAchieve.format_achievement`, making it easy to overload for your custom display styling without reimplementing the whole command.
--- a/evennia/contrib/game_systems/achievements/README.md
+++ b/evennia/contrib/game_systems/achievements/README.md
@@ -2,63 +2,26 @@
 
 A simple, but reasonably comprehensive, system for tracking achievements. Achievements are defined using ordinary Python dicts, reminiscent of the core prototypes system, and while it's expected you'll use it only on Characters or Accounts, they can be tracked for any typeclassed object.
 
-## Installation
-
-Once you've defined your achievement dicts in a module, assign that module to the `ACHIEVEMENT_CONTRIB_MODULES` setting in your settings.py
-
-Optionally, you can specify what attribute achievement progress is stored in, with the setting `ACHIEVEMENT_CONTRIB_ATTRIBUTE`. By default it's just "achievements".
-
-There is also a command available to let players check their achievements - `from evennia.contrib.game_systems.achievements.achievements import CmdAchieve` and then add `CmdAchieve` to your default Character and/or Account cmdsets.
-
-#### Settings examples
-
-One module:
-
-```python
-# in server/conf/settings.py
-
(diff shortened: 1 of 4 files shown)


evennia/contrib/game_systems/achievements/achievements.py
```diff
@@ -0,0 +1,370 @@
...
+- prereqs (str or list): An optional achievement key or list of keys that must be completed before
+        this achievement is available.
+
+To add achievement tracking, put `track_achievements` in your relevant hooks.
+
+Example:
+
+    def at_use(self, user, **kwargs):
+        # track this use for any achievements about using an object named our name
+        finished_achievements = track_achievements(user, category="use", tracking=self.key)
```
Griatch, line 42: Why not use the actual achievement-dict you defined above as the example when calling this function? This example doesn't make it clear to me how this function interacts with the achievement.

evennia/contrib/game_systems/achievements/README.md
```diff
@@ -0,0 +1,119 @@
...
+    track_achievements(self.caller, "buy", obj.name, count=quantity)
+```
+
+In this case, `obj` is the fruit that was just purchased, and `quantity` is the amount they bought.
+
+The `track_achievements` function does also return a value: an iterable of keys for any achievements which were newly completed by that update. You can ignore this value, or you can use it to e.g. send a message to the player with their latest achievements.
+
+### The `achievements` command
+
+The contrib's provided command, `CmdAchieve`, aims to be usable as-is, with multiple switches to filter achievements by various progress statuses and the ability to search by achievement names.
```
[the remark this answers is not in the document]
Griatch: Why not show what it returns, extending the example flower achievement from above?

evennia/contrib/game_systems/achievements/README.md
```diff
@@ -0,0 +1,119 @@
...
+## Usage
+
+### Defining your achievements
+
+This achievement system is designed to use ordinary dicts for the achievement data - however, there are certain keys which, if present in the dict, define how the achievement is progressed or completed.
+
+- **name** (str): The searchable name for the achievement. Doesn't need to be unique.
+- **desc** (str): A longer description of the achievement. Common uses for this would be flavor text or hints on how to complete it.
+- **category** (str): The type of actions this achievement tracks. e.g. purchasing 10 apples might have a category of "buy", or killing 10 rats might have a category of "defeat".
+- **tracking** (str or list): The *specific* things this achievement tracks. e.g. the previous example of killing rats might have a `"tracking"` value of `"rat"`. An achievement can also track multiple things: instead of only tracking buying apples, you might want to track apples and pears. For that situation, you can assign it to a list of values to check against: e.g. `["apple", "pear"]`
```
Griatch, line 50: Reading this, it's not clear what are the valid values of these 'things' you talk about. What is an 'apple' or 'rat' here? Must they be matched to objects? Are they just lables?

A common achievement doesn't track anything - you just get it when you perform a particular action or fulfills some condition. How is that represented here? Is the 'tracking' field optional?

evennia/contrib/game_systems/achievements/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+For example, you might have a collection achievement for buying 10 apples, and a general `buy` command players could use. In your `buy` command, after the purchase is completed, you could add the following line:
+
+```python
+    track_achievements(self.caller, "buy", obj.name, count=quantity)
+```
+
+In this case, `obj` is the fruit that was just purchased, and `quantity` is the amount they bought.
+
+The `track_achievements` function does also return a value: an iterable of keys for any achievements which were newly completed by that update. You can ignore this value, or you can use it to e.g. send a message to the player with their latest achievements.
```
Griatch, line 113: If I just want to give the player an achievement based on some _external_ condition, how would I go about that? 

Example: Player steps into the dragon's lair for the first time, I independently check this and declare that this means they should get the _Wannabe Dragon Slayer_ achievement.

evennia/contrib/game_systems/achievements/achievements.py
```diff
@@ -0,0 +1,370 @@
...
+
+An example of an achievement dict:
+
+    EXAMPLE_ACHIEVEMENT = {
+        "name": "Some Achievement",
+        "desc": "This is not a real achievement.",
+        "category": "crafting",
+        "tracking": "box",
+        "count": 5,
+        "prereqs": "ANOTHER_ACHIEVEMENT",
```
Griatch, line 16: prereqs are cool :+1:

evennia/contrib/game_systems/achievements/achievements.py
```diff
@@ -0,0 +1,370 @@
...
+
+    Args:
+        achiever (Account or Character):  The entity that's collecting achievement progress.
+
+    Keyword args:
+        category (str or None):  The category of an achievement.
+        tracking (str or None):  The specific item being tracked in the achievement.
+
+    Returns:
+        completed (tuple):  The keys of any achievements that were completed by this update.
```
Griatch, line 94: This actually is nameless; so this should just be `tuple: ...` In older code of Evennia, the return has a name, but this is not standard and messes with autodocs.

evennia/contrib/game_systems/achievements/README.md
```diff
@@ -0,0 +1,119 @@
...
+ACHIEVEMENT_CONTRIB_ATTRIBUTE = ("achievements", "systems")
+```
+
+## Usage
+
+### Defining your achievements
+
+This achievement system is designed to use ordinary dicts for the achievement data - however, there are certain keys which, if present in the dict, define how the achievement is progressed or completed.
+
+- **name** (str): The searchable name for the achievement. Doesn't need to be unique.
```
Griatch, line 47: Is this correct? In the `get_achievement` function, the `key` used to acquire the achievement is the variable name and not this. Since this `name` does _not_ need to be unique, it seems like a poor choice for a 'searchable' key?

Edit: Reading on, I see the `search_achievement` function does search by this field. This probably needs to be explicitly mentioned here, since the variable key is used everywhere else.

evennia/contrib/game_systems/achievements/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+ACHIEVEMENT_CONTRIB_MODULES = "world.achievements"
+```
+
+Multiple modules, with a custom-defined attribute:
+
+```python
+# in server/conf/settings.py
+
+ACHIEVEMENT_CONTRIB_MODULES = ["world.crafting.achievements", "world.mobs.achievements"]
```
Griatch, line 28: If this works like other similar systems in Evennia, same-named achievements defined in latter modules would override those in previous modules. This is not explained here though.

evennia/contrib/game_systems/achievements/achievements.py
```diff
@@ -0,0 +1,370 @@
...
+    """
+    if not (all_achievements := _load_achievements()):
+        # there are no achievements available, there's nothing to do
+        return tuple()
+
+    # split out the achievement attribute info
+    attr_key = _ACHIEVEMENT_ATTR[0]
+    attr_cat = _ACHIEVEMENT_ATTR[1] if len(_ACHIEVEMENT_ATTR) > 1 else None
```
Griatch, line 102: This won't change without a reload. Why not do this split directly once, in the header of this module?

evennia/contrib/game_systems/achievements/achievements.py
```diff
@@ -0,0 +1,370 @@
...
+    Keyword args:
+        category (str or None):  The category of an achievement.
+        tracking (str or None):  The specific item being tracked in the achievement.
+
+    Returns:
+        completed (tuple):  The keys of any achievements that were completed by this update.
+    """
+    if not (all_achievements := _load_achievements()):
+        # there are no achievements available, there's nothing to do
+        return tuple()
```
Griatch, line 98: You do this loading in every function, storing to a global variable ... why not make a singleton handler instead? 

```python
class AchivementHandler: 
    def __init__(self):
        self.achivements = _load_achievements()
    def get(): 
         # .. .
    def all(): 
         # ... 
    def check():
         # ...


ACHIEVEMENTS_HANDLER = AchivementsHandler()
```

evennia/contrib/game_systems/achievements/achievements.py
```diff
@@ -0,0 +1,370 @@
...
+    attr_cat = _ACHIEVEMENT_ATTR[1] if len(_ACHIEVEMENT_ATTR) > 1 else None
+    if progress_data := achiever.attributes.get(attr_key, default={}, category=attr_cat):
+        # detach the data from the db to avoid data corruption and return the data
+        return progress_data.deserialize().get(key, {})
+    else:
+        # just return an empty dict
+        return {}
+
+
+def search_achievement(search_term):
```
Griatch, line 213: So this is where the 'name' search mentioned in the README comes in. :)

evennia/contrib/game_systems/achievements/achievements.py
```diff
@@ -0,0 +1,370 @@
...
+
+        Returns
+            str: The display string to be sent to the caller.
+
+        """
+        template = """\
+|w{name}|n
+{desc}
+{status}
+""".rstrip()
```
Griatch, line 277: Why embed this inside the `format_achievement` method? Put it on the class instead - then if people don't want to change much, they can just do 

```py
class CmdMyAchieve(CmdAchieve): 
    template = """ my custom style """ 
```

Griatch (changes requested):
It's a cool idea! I think this can do with some improvements and clarifications though. 

I also have a few use-cases I'm unsure about, such as how to 'just' add an achievement to a user based on an external condition (something which I've found to be the most common case whenever I've done achievements).

evennia/contrib/game_systems/achievements/README.md
```diff
@@ -0,0 +1,119 @@
...
+There is also a command available to let players check their achievements - `from evennia.contrib.game_systems.achievements.achievements import CmdAchieve` and then add `CmdAchieve` to your default Character and/or Account cmdsets.
+
+#### Settings examples
+
+One module:
+
+```python
+# in server/conf/settings.py
+
+ACHIEVEMENT_CONTRIB_MODULES = "world.achievements"
```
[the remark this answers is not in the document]
InspectorCaracal: Oh, my bad. I was referencing the code for the crafting contrib when writing this up and the code is written to handle a string, so I assumed that was expected to be valid.

evennia/contrib/game_systems/achievements/achievements.py
InspectorCaracal: I had it that way originally, but I realized that most, if not all, people who want to override the template will want to change the strings being *put into* the template as well, making it fairly useless to override just the template. I'll split it out anyway though.

evennia/contrib/game_systems/achievements/achievements.py
InspectorCaracal: If I renamed it to `_get_achievements` instead of `_load_achievements` to make it clearer that it's an access function, would that be acceptable? Or is the issue that it's a function at all?

evennia/contrib/game_systems/achievements/README.md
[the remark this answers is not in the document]
Griatch: I didn't remember it did - that's not so good honestly; it should be explicit, but harder to change after it's already there, alas.

evennia/contrib/game_systems/achievements/achievements.py
Griatch: I think the naming is fine; You can structure your contrib like this if you want; just pointing out that you are essentially treating the module as a 'handler' of sorts, storing achievements into a module scope. It leads to a lot of code repetition (every function needs to load the global scope since they are not otherwise related to each other), which you could avoid with a singleton pattern.

evennia/contrib/game_systems/achievements/achievements.py
Griatch: While you are probably right in practice, I just know that _someone_ will eventually ask why this simple override is not possible down the line :wink: Breaking it out costs nothing, so best to go for max flexibility here.

evennia/contrib/game_systems/achievements/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+An achievement for buying 5 each of apples, oranges, and pears.
+```python
+FRUIT_BASKET_ACHIEVEMENT = {
+    "name": "A Fan of Fruit", # note, there is no desc here - that's allowed!
+    "category": "buy",
+    "tracking": ("apple", "orange", "pear"),
+    "count": 5,
+    "tracking_type": "separate",
+}
```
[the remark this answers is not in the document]
InspectorCaracal: After some thought: I can appreciate the usefulness of this as an approach, but I've personally found it to be more confusing to actually use in practice, personally, and it would add an extra layer of complexity to the install/setup directions, so I've opted not to implement it.

evennia/contrib/game_systems/achievements/README.md
```diff
@@ -0,0 +1,119 @@
...
+- **count** (int): *Default value if unset: 1* The number of tallies this achievement's requirements need to build up in order to complete the achievement. e.g. the previous example of killing rats would have a `"count"` value of `10`. For achievements using the "separate" tracking type, *each* item being tracked must tally up to this number to be completed
+- **prereqs** (str or list): The *keys* of any achievements which must be completed before this achievement can start tracking progress. An achievement's key is the variable name it's assigned to in your achievement module.
+
+You can add any additional keys to your achievement dicts that you want, and they'll be included with all the rest of the achievement data when using the contrib's functions. This could be useful if you want to add extra metadata to your achievements for your own features.
+
+#### Examples
+
+A simple achievement which you can get just for logging in the first time. This achievement has no prerequisites and it only needs to be fulfilled once to complete, so it doesn't need to define most of the keys.
+```python
```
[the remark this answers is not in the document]
Griatch: Still no comment here.

evennia/contrib/game_systems/achievements/README.md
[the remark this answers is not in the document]
Griatch: Ok. I find this style to be _considerably_ more useful in practice, but I guess it's a matter of preference. :smiley:

evennia/contrib/game_systems/achievements/README.md
Griatch: Not sure my question was answered?

evennia/contrib/game_systems/achievements/achievements.py
Griatch: This question is still open to me.

Griatch (changes requested):
Improvements for sure :+1: , I'm not sure I see some of my comments as actually being addressed though. 

----

As a point of PR etiquette, please don't "resolve" review comments on your own. It just makes it harder for the reviewer - since there are several commits making changes harder to track what changed, and it's not clear if a resolved comment means it's closed because you disagreed with the comment or if you agreed and implemented it. Easiest is if you reply to the comment and say if you implemented the change, or why you disagree and won't (like you did in one place). Leave the "resolving" of review comments to the reviewer. :) 

You also did a bunch of other changes to this PR after the last review, without commenting on what these changes were. This makes it hard to know what should be re-reviewed from scratch and what were just refactoring without changing functionality.

InspectorCaracal:
@Griatch Ack! Sorry about that. I will refrain from marking things resolved in the future. 🫡

evennia/contrib/game_systems/achievements/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+ACHIEVEMENT_CONTRIB_MODULES = "world.achievements"
+ACHIEVEMENT_CONTRIB_ATTRIBUTE = ("achievements", "systems")
+```
+
+## Usage
+
+### Defining your achievements
+
+This achievement system is designed to use ordinary dicts for the achievement data - however, there are certain keys which, if present in the dict, define how the achievement is progressed or completed.
```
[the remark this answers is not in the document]
InspectorCaracal: It's in the header paragraph and the first line of the install instructions.

evennia/contrib/game_systems/achievements/README.md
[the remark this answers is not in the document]
InspectorCaracal: I don't understand what comment you want me to add, sorry.

Griatch:
@InspectorCaracal No worries, these things are not obvious in the GH interface and won't come to mind one has reviewed a few hundred PRs like I have ;)

evennia/contrib/game_systems/achievements/achievements.py
InspectorCaracal: The code reviewed here is outdated. The updated docstring resolved this.

evennia/contrib/game_systems/achievements/README.md
InspectorCaracal: The very first example provided is an example of an achievement that "just happens".

evennia/contrib/game_systems/achievements/README.md
InspectorCaracal: It's addressed in the search_achievement section. This section is just supposed to define what the fields are, not how to use them; using them is in the usage section.

evennia/contrib/game_systems/achievements/README.md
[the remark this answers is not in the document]
Griatch: Yes, but it's not mentioned here, where the actual dicts are described, which is confusing to the reader. A line 'These dicts are defined in Python modules you tell Evennia to import" or something would clarify things.

evennia/contrib/game_systems/achievements/README.md
```diff
@@ -0,0 +1,174 @@
...
+    "name": "A Fan of Fruit", # note, there is no desc here - that's allowed!
+    "category": "buy",
+    "tracking": ("apple", "orange", "pear"),
+    "count": 5,
+    "tracking_type": "separate",
+}
+```
+
+
+## Installation
```
Griatch, line 69: I note you moved this down. I can see some reason for this in that the previous sections are introducing the system. But all other contribs put the installation as close to the top as possible, so it's probably better to follow that.

evennia/contrib/game_systems/achievements/README.md
Griatch: Which one? :thinking:

Griatch (changes requested):
Just some nitpicks when it comes to (IMO) a bit confusing descriptions in the readme, but otherwise getting there :+1:

evennia/contrib/game_systems/achievements/README.md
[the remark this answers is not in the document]
Griatch:
```python
# in a module you set from settings.<SETTINGS_NAME>

# This achievement has the unique key of ...
FIRST_LOGIN_ACHIEVE = ...
```

Griatch:
@InspectorCaracal I had missed to answer one of your follow-up questions in the code comments, did so now :)

Griatch:
@InspectorCaracal This is almost done, with just some minor doc comments to fix :)

InspectorCaracal:
I did a major revision of the README and I *think* it covers all the issues now.

InspectorCaracal:
Oops, was just doing another pass to make sure I didn't miss anything and realized I left some tab indents, so I'll be putting in another commit in a second just to fix those.

Griatch (approved):
Sorry for the delay, thanks for fixing the readme and bringing things up-to-date! Approved!

PR #3453, opened 2024-03-29
Fix aliases not being shown for disambiguation

chiizujin:

Fixed missing alias display when utils.at_search_result() asks for disambiguation.

Additionally, objects without aliases were showing '[]' (instead of nothing) due to a misplaced 'if'.

Old:
```
More than one match for 'box' (please narrow target):
 box-1 []
 box-2 []
```
New:
```
More than one match for 'box' (please narrow target):
 box-1 [cube]
 box-2
```

Unit tests for this only exist as part of contrib/rpg and were testing for the presence of '[]' since this is what at_search_result()  always returned. These test have been updated to test for the new behaviour.

When utils.at_search_result() asks for disambiguation it always lists the aliases as '[]' for every matching object, whether or not it has any (non-pluralisation) aliases. This doesn't seem to be the desired behaviour.

### DIFF
--- a/evennia/contrib/rpg/rpsystem/tests.py
+++ b/evennia/contrib/rpg/rpsystem/tests.py
@@ -6,6 +6,7 @@ import time
 
 from anything import Anything
 from evennia import DefaultObject, create_object, default_cmds
+from evennia.commands.default import building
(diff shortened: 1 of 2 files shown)


Griatch (approved):
LGTM! Thanks for the fix!

PR #3454, opened 2024-03-29
Fix traceback in crafting system recipe.msg

InspectorCaracal:
The `msg` method used by crafting recipes has faulty syntax, resulting in the outgoing text kwargs being interpreted as the `from_obj` keyword argument.

Bug fixing

### DIFF
--- a/evennia/contrib/game_systems/crafting/crafting.py
+++ b/evennia/contrib/game_systems/crafting/crafting.py
@@ -237,7 +237,7 @@ class CraftingRecipeBase:
             **kwargs: Any optional properties relevant to this send.
 
         """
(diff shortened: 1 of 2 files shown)


Griatch (approved):
Good catch!

Griatch (changes requested):
Good fix, but I notice that the unit tests actually assume the faulty syntax :sweat: Apparently went all in with the wrong here. So these unit tests need updating too while we are at it.

InspectorCaracal:
I didn't even think to check the unit tests and it really is all through there, isn't it? :sweat_smile: I'll get that done in a bit.

Griatch (approved):
Thanks, fixed all the way down now! :D

PR #3459, opened 2024-03-30
Fix editor echo not displaying correctly for lines with whitespace

chiizujin:

When in normal editing mode (i.e. not in code editing mode) and adding lines with extra whitespace (leading or between words) the input echoing does not correctly reflect what is actually added to the buffer (the extra space is stripped).

This fixes the echo to correctly show what is added.

Correct editor echo display.

#### Other info

**Before (in the editor):**
```
>  This  line   has    whitespace.
01| This line has whitespace.
>:
----------Line Editor [topic test]--------------------------------------------
01|   This  line   has    whitespace.
----------[l:01 w:004 c:0033]------------(:h for help)------------------------
```
**After (in the editor):**
```
>  This  line   has    whitespace.
01|   This  line   has    whitespace.
>:
----------Line Editor [topic test]--------------------------------------------
01|   This  line   has    whitespace.
----------[l:01 w:004 c:0033]------------(:h for help)------------------------
```

### DIFF
--- a/evennia/utils/tests/test_eveditor.py
+++ b/evennia/utils/tests/test_eveditor.py
@@ -168,13 +168,13 @@ class TestEvEditor(BaseEvenniaCommandTest):
             eveditor.CmdLineInput(),
             'First test "line".',
             raw_string='First test "line".',
-            msg='01First test "line" .',
+            msg='01First test "line".',
         )
         self.call(
             eveditor.CmdLineInput(),
             "Second 'line'.",
             raw_string="Second 'line'.",
-            msg="02Second 'line' .",
+            msg="02Second 'line'.",
         )
         self.assertEqual(
             self.char1.ndb._eveditor.get_buffer(), "First test \"line\".\nSecond 'line'."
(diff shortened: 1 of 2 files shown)


evennia/utils/tests/test_eveditor.py
```diff
@@ -168,13 +168,13 @@ def test_eveditor_COLON_UU(self):
...
             raw_string='First test "line".',
-            msg='01First test "line" .',
+            msg='01First test "line".',
         )
         self.call(
             eveditor.CmdLineInput(),
             "Second 'line'.",
             raw_string="Second 'line'.",
-            msg="02Second 'line' .",
+            msg="02Second 'line'.",
```
Griatch, line 177: It seems like a good opportunity to actually add a new test here to verify the echo is indeed as expected?

Griatch (changes requested):
Good catch! I agree with the fix, but I think this is also a good opportunity to add a new test to make sure to test this case and make sure to avoid regressions in the future.

chiizujin:
I've added a unit test for this. I added it the end of the editor tests so that the addition of another line doesn't break the tests that rely on the buffer content.

Griatch (approved):
New test seems fine, thanks for adding it!

PR #3464, opened 2024-03-31
Fix editor's range specification

chiizujin:

The editor's range specification (n:n) returns at most two lines. This has been fixed.

Also, as part of this:
 - An existing unit test has been corrected
 - New unit tests have been added to test various ranges
 - `utils.value_is_integer(value)` has been added which returns `True` if `value` can be type-cast to an integer and `False` otherwise. This is to allow checking for negative integers, which `str.isdigit()` can't do.

Fix editor range specification.

#### Other info

**Test buffer**
```
>:
----------Line Editor [desc]--------------------------------------------------
01| line 1
02| line 2
03| line 3
04| line 4
05| line 5
----------[l:05 w:010 c:0034]------------(:h for help)------------------------
```
**Before**
```
>: 1:5
----------Line Editor [desc]--------------------------------------------------
01| line 1
02| line 2
----------[l:02 w:004 c:0013]------------(:h for help)------------------------
```
**After**
```
>: 1:5
----------Line Editor [desc]--------------------------------------------------
01| line 1
02| line 2
03| line 3
04| line 4
05| line 5
----------[l:05 w:010 c:0034]------------(:h for help)------------------------
```
**Changed unit test**

```
def test_eveditor_COLON_F(self):
    eveditor.EvEditor(self.char1)
    self.call(
        eveditor.CmdEditorGroup(),
        "",
        cmdstring=":",
        msg="Line Editor []\n01\n[l:01 w:000 c:0000](:h for help)",
    )
    self.call(eveditor.CmdLineInput(), "line 1", raw_string="line 1", msg="01line 1")
    self.call(eveditor.CmdEditorGroup(), "1:2", cmdstring=":f", msg="Flood filled lines 1-2.
```

This test expects the output "Flood filled lines 1-2." for a buffer containing only one line. After the fix the above command will return "Flood filled line 1.", which is correct. The test has been updated to look for the new output.

### DIFF
--- a/evennia/utils/utils.py
+++ b/evennia/utils/utils.py
@@ -3077,3 +3077,21 @@ def ip_from_request(request, exclude=None) -> str:
 
     logger.log_warn("ip_from_request: No valid IP address found in request. Using remote_addr.")
     return remote_addr
+
+
+def value_is_integer(value):
+    """
+    Determines if a value can be type-cast to an integer.
+
+    Args:
+        value (any): The value to check.
+
+    Returns:
+        result (bool): Whether it can be type-cast to an integer or not.
+    """
+    try:
+        int(value)
+    except ValueError:
+        return False
+
+    return True
--- a/evennia/utils/eveditor.py
+++ b/evennia/utils/eveditor.py
@@ -305,12 +305,9 @@ class CmdEditorBase(_COMMAND_DEFAULT_CLASS):
         linerange = False
         if arglist and arglist[0].count(":") == 1:
             part1, part2 = arglist[0].split(":")
-            if part1 and part1.isdigit():
-                lstart = min(max(0, int(part1)) - 1, nlines)
-                linerange = True
-            if part2 and part2.isdigit():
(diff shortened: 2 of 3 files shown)


evennia/utils/utils.py
```diff
@@ -3077,3 +3077,21 @@ def ip_from_request(request, exclude=None) -> str:
...
+
+    Returns:
+        result (bool): Whether it can be type-cast to an integer or not.
+    """
+    try:
+        int(value)
+    except ValueError:
+        return False
+
+    return True
```
Griatch, line 3097: This is fine, but in case you were not aware, you can also do `try: ... except: ... else: ...` in Python :)

Griatch (approved):
This is good attention to detail. And looks like a clean fix supported by unit tests. Nice!

evennia/utils/utils.py
chiizujin: I always forget about `else` in `try` blocks.

PR #3465, opened 2024-04-01
Add proper name support to get_numbered_name()

chiizujin:

Objects created with a key beginning with an upper-case letter are now treated by `get_numbered_name()` as being proper-named and will return the key as-is. The default behaviour can be restored per-object by setting the object attribute `proper_named=False`.

See below for examples.

This makes the names of proper-named objects scan better. I think the behaviour will work naturally for most cases, but can be overridden for exceptions.

I will add unit tests if you decide to accept this change.

#### Regarding existing unit tests

The existing unit tests use and object called "Obj", which is proper-named. That tests looking for things like `You pick up an Obj.` have been changed to look for `You pick up Obj.`, etc. While this looks a little odd it is correct for these instances.

A better way would be to rename the objects along the lines of "obj" and "box" (another that caused failures). However, the casing is quite deeply embedded in the test suite for objects and would require many other changed to expected outputs, so I'm reluctant to change it.

As I mentioned above, I will add additional tests to cover differences between, for example, "Obj" and "obj".

#### Examples
```
>create ring
You create a new Object: ring.
>i
You are carrying:
 a ring  You see nothing special.
```
```
>create The One Ring
You create a new Object: The One Ring.
>i
You are carrying:
 The One Ring  You see nothing special.
 a ring        You see nothing special.
```
```
>create Evennia manual
You create a new Object: Evennia manual.
>i
You are carrying:
 Evennia manual  You see nothing special.
 The One Ring    You see nothing special.
 a ring          You see nothing special.
```
```
>set manual/proper_named = False
Created attribute Evennia manual/proper_named [category:None] = False
>i
You are carrying:
 an Evennia manual  You see nothing special.
 The One Ring       You see nothing special.
 a ring             You see nothing special.
```
](https://github.com/evennia/evennia/pull/3465/files)

### DIFF
--- a/evennia/commands/default/tests.py
+++ b/evennia/commands/default/tests.py
@@ -108,13 +108,13 @@ class TestGeneral(BaseEvenniaCommandTest):
         self.call(general.CmdNick(), "/list", "Defined Nicks:")
 
     def test_get_and_drop(self):
-        self.call(general.CmdGet(), "Obj", "You pick up an Obj")
-        self.call(general.CmdDrop(), "Obj", "You drop an Obj")
+        self.call(general.CmdGet(), "Obj", "You pick up Obj")
+        self.call(general.CmdDrop(), "Obj", "You drop Obj")
 
     def test_give(self):
         self.call(general.CmdGive(), "Obj to Char2", "You aren't carrying Obj.")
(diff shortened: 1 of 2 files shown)


chiizujin:
After some discussion on Discord about this, I'm going to rethink my approach.

PR #3466, opened 2024-04-01
Add support to get_numbered_name() for omitting the article

chiizujin:

Adds support for a `kwarg` '`no_article`' to `get_numbered_name()`. When set to `True` and `count` is 1, it will return the key as-is without an article.

See below for usage and examples.

(Thanks to Inspector Caracal for the brainstorming on this one.)

Adds support for making the names of proper-named objects scan better.

### Usage
```
from typeclasses.objects import Object

class ProperNamedObject(Object):
    """
    This is a base class for objects that have a proper name, for example "The One Ring'.

    The effect of using this class is that the object's numbered name will not include the
    indefinite article when there is only one of them in scope.
    """

    def get_numbered_name(self, count, looker, **kwargs):
        return super().get_numbered_name(count, looker, no_article=True, **kwargs)
```

#### Examples
```
>create The One Ring
You create a new Object: The One Ring.
>i
You are carrying:
 a The One Ring  You see nothing special.
```
```
>type the one ring = typeclasses.utils.ProperNamedObject
(...)
>i
You are carrying:
 The One Ring  You see nothing special.
```
```
>create The Other One Ring : typeclasses.utils.ProperNamedObject
You create a new ProperNamedObject: The Other One Ring.
>i
You are carrying:
 The One Ring        You see nothing special.
 The Other One Ring  You see nothing special.
```

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -1483,8 +1483,12 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
                 including the count.
 
         Examples:
-            ::
-                obj.get_numbered_name(3, looker, key="foo") -> ("a foo", "three foos")
+        ::
+            - obj.get_numbered_name(3, looker, key="foo") -> ("a foo", "three foos")
+            - obj.get_numbered_name(1, looker, key="Foobert", return_string=True)
+                  -> "a Foobert"
+            - obj.get_numbered_name(1, looker, key="Foobert", return_string=True, no_article=True)
+                  -> "Foobert"
 
         """
         plural_category = "plural_key"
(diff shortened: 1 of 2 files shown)


evennia/objects/objects.py
```diff
@@ -1476,6 +1476,7 @@ def get_numbered_name(self, count, looker, **kwargs):
                 method is used.
             return_string (bool): If `True`, return only the singular form if count is 0,1 or
                 the plural form otherwise. If `False` (default), return both forms as a tuple.
+            no_article (bool): If `True`, do not return an article if `count` is 1.
```
Griatch, line 1479: With this change, I think it would be suitable to also add an `Example:` block to this doc string to exemplify the use cases. the "The One Ring" example as well as "an Obj" vs just "Obj" are relevant.

Griatch (changes requested):
I don't have any trouble with the implementation, but the PR has a good description which won't appear anywhere in docs unless an (abbreviated) version of it is also added to the doc string for `get_numbered_name`.

chiizujin:
I've added some examples to the doc string and re-aligned the `::` to see if that removes it from the HTML docs.

Griatch (approved):

PR #3489, opened 2024-04-07
Prevents admin from walking out non-existent exits in the wilderness

t34lbytes:

When navigating the wilderness as an Admin you can see exits that are otherwise invalid.  If you try to use one of these invalid exits, you will encounter stack traces and your character will get stuck unable to teleport themselves out.  You have to go ooc, switch characters, teleport them to safety, and switch back.

Suggested this to `evennia-minimud` but it was rejected on the grounds that the issue is with the contrib, so I'm suggesting it here.

Probably not relevant, but since I mention it above, that rejected pull request:

https://github.com/InspectorCaracal/evennia-minimud/pull/10


Griatch (approved):
I think it's better to be friendly here, so this seems ok.

Issue #3468, opened 2024-04-01
[Feature Request] Centralize searching for N number of identical objects


Griatch:
#### Is your feature request related to a problem? Please describe.

In PR https://github.com/evennia/evennia/pull/3433, there, @InspectorCaracal creates a helper class for allowing get/drop/give default commands to manipulate stacks, such as `give 3 apples`. This kind of thing seems like something we should investigate more fundamental support for.

#### Describe the solution you'd like

Add `DefaultObject.search` support for input on the form `3 apples`, and make it clear that this should return 3 instances of objects identified by `apples`. We have support for disambiguation syntax, like `apple-3`, so it ought to be possible to do something similar for fetching stacks of identical items as well. The `evennia.utils.utils.group_objects_by_key_and_desc` could be looked at for detecting stacks.

#### Alternatives

If adding search-support is deemed too compex or heavy, one could also picture adding/extracting helper tools dealing with stack handling into one central place so devs don't need to implement or find it spread out across different default commands.

PR #3493, opened 2024-04-07
Framework for multi-search.

jaborsh:
This PR isn't intended to go live yet (although it won't break anything). This is my proposal on how we rebuild the search function in order to handle queries such as: `get sword`, `get 2 swords`, and `get all` as discussed in #3468.

With this framework, I use regular expression patterns within `parse` for functions `get`, `drop`, `give`, and `put` which would also need to be implemented, but I think this gives an idea for a direction to go. If desired, I can attach `CmdGet` and `CmdDrop`.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -73,10 +73,16 @@ class ObjectSessionHandler:
         self._sessid_cache = list(
             set(int(val) for val in (self.obj.db_sessid or "").split(",") if val)
         )
-        if any(sessid for sessid in self._sessid_cache if sessid not in evennia.SESSION_HANDLER):
+        if any(
+            sessid
+            for sessid in self._sessid_cache
+            if sessid not in evennia.SESSION_HANDLER
+        ):
             # cache is out of sync with sessionhandler! Only retain the ones in the handler.
             self._sessid_cache = [
-                sessid for sessid in self._sessid_cache if sessid in evennia.SESSION_HANDLER
+                sessid
+                for sessid in self._sessid_cache
+                if sessid in evennia.SESSION_HANDLER
             ]
             self.obj.db_sessid = ",".join(str(val) for val in self._sessid_cache)
             self.obj.save(update_fields=["db_sessid"])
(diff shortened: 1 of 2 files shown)


jaborsh:
There's also a series of linter formatting. I can undo that if desired.

evennia/objects/objects.py
```diff
@@ -530,13 +539,17 @@ def handle_search_results(self, searchdata, results, **kwargs):
 
         nofound_string = kwargs.get("nofound_string")
         multimatch_string = kwargs.get("multimatch_string")
+        return_quantity = kwargs.get("return_quantity")
```
jaborsh, line 542: Point of interest.

evennia/utils/utils.py
```diff
@@ -2354,48 +2413,65 @@ def display_len(target):
 # Replace this hook function by changing settings.SEARCH_AT_RESULT.
 
 
+class SearchReturnType(Enum):
```
jaborsh, line 2416: Point of interest.

evennia/objects/objects.py
```diff
@@ -73,10 +73,16 @@ def _recache(self):
         self._sessid_cache = list(
             set(int(val) for val in (self.obj.db_sessid or "").split(",") if val)
         )
-        if any(sessid for sessid in self._sessid_cache if sessid not in evennia.SESSION_HANDLER):
+        if any(
+            sessid
+            for sessid in self._sessid_cache
+            if sessid not in evennia.SESSION_HANDLER
+        ):
```
Griatch, line 80: Please set up your IDE to follow Evennia line width standard (100 characters). That makes the PR smaller (easier to review) and clearer since you only see the important changes.

evennia/utils/utils.py
```diff
@@ -2354,48 +2413,65 @@ def display_len(target):
...
         # no results.
-        error = kwargs.get("nofound_string") or _("Could not find '{query}'.").format(query=query)
+        error = kwargs.get("nofound_string") or _("Could not find '{query}'.").format(
+            query=query
+        )
         matches = None
     elif len(matches) > 1:
+        return_quantity = kwargs.get("return_quantity", 1)
+        return_type = kwargs.get("return_type", SearchReturnType.DEFAULT)
+        if return_type == SearchReturnType.ONE:
```
Griatch, line 2460: This can be a `match ... case` structure.

evennia/utils/utils.py
```diff
@@ -2354,48 +2413,65 @@ def display_len(target):
...
+            return matches[return_quantity - 1]
```
Griatch, line 2461: Would this not traceback if `matches` and `return_quantity` are out of sync?

evennia/utils/utils.py
```diff
@@ -2354,48 +2413,65 @@ def display_len(target):
...
+        elif return_type == SearchReturnType.MULTIPLE:
+            matches = matches[:return_quantity]
+            return matches
+        elif return_type == SearchReturnType.ALL:
```
Griatch, line 2465: Is this not the default?

Griatch (changes requested):
I need to look into this a bit more, have a few questions so far. This also doesn't actually exemplifies how this would be used which could be the next step.


Also, please use the correct line width in your IDE (Evennia uses 100 character line width) to make your PR diff smaller (and you won't have to specify your 'points of interests'). As it is I'll have to reformat your code after merging.

Issue #3248, opened 2023-08-01
[Feature Request] Support for color beyond XTERM256


11Spades:
When I make a room or item, I like to *really* color them up. Unfortunately, XTERM256 occasionally lacks the range required for me to express my idea in full. To remedy this, I'd like to propose support for full RGB color. I've implemented this before on an LP MUD, and know that at least Mudlet can handle it nicely.

Unfortunately, I am not very familiar with the Evennia codebase, but if someone could direct me to the components that handle color code -> ansi code replacement, I'd be happy to take a crack at it myself.

Griatch:
I think this sounds interesting. What you are talking about sounds like an extension to the `|r` and `|345` style syntax, which would be interesting to have. Some questions and points: 

- Could you give some examples for how the syntax would look, for Evennia? I take it this would be a core addition, or are you picturing it as a contrib? 
- Is there a Telnet protocol standard (well, as standard as MUD Telnet extension go) for clients to use - if so, do you have a link to the spec? 
- Webclient support would require updating / adding logic to convert this to CSS on the server level (and in the future on the client level when we get to implement the requirements of [PR 3185](https://github.com/evennia/evennia/pull/3185)).

InspectorCaracal:
I'm not the OP, but for my game, I added full RGB support using hex code syntax, like so:

`|#ff8800this is orange text|n`
![image](https://github.com/evennia/evennia/assets/51038201/48911361-8210-498b-95ff-0da4e08baf13)

Personally I find the syntax works nicely with the existing Evennia syntax and I figure it would be easy to pick up, since virtually everyone is familiar with hex color codes by now.

(I never managed to find an answer to "what's the accepted way of sending RGB via telnet", myself, but hopefully the OP knows.)

11Spades:
>Could you give some examples for how the syntax would look, for Evennia? I take it this would be a core addition, or are you picturing it as a contrib?

The syntax that @InspectorCaracal is using is the syntax I had in mind -- this is most akin to how I implemented it elsewhere.

>Is there a Telnet protocol standard (well, as standard as MUD Telnet extension go) for clients to use - if so, do you have a link to the spec?

I'm unsure as to whether you're referring to a standard for figuring out whether the client supports RGB color (True Color) or a standard for displaying True Color.

[...]

PR #3494, opened 2024-04-07
Hex colors

michaelfaith84:
Added support for truecolor in the portal (xterm truecolor) and the web portal using the style attribute. Syntax used is |#0f0, |#0f0f764, |[#00f, etc.

#3248 Who doesn't like more colors?


evennia/server/portal/ttype.py
```diff
@@ -143,13 +143,23 @@ def will_ttype(self, option):
...
 
+            # use name to identify support for xterm truecolor
+            truecolor = False
+            if (clientname.endswith("-TRUECOLOR") or
+                    clientname in (
+                            "AXMUD",
+                            "TINTIN"
+                    )):
+                truecolor = True
+
```
Griatch, line 158: The [TTYPE/MTTS](https://tintin.mudhalla.net/protocols/mtts/) protocol supports `TRUECOLOR` as a valid parameter. With this check it seems you are assuming `truecolor=False` and ignoring the `TTYPE/MTTS` value sent by the client completely. 

I think you should honor the handshake and only override if there is no such value in the communication, or if you know that the given client sends a faulty value (or don't support the proper handshake, which is what most of these client-exceptions are for).

evennia/utils/ansi.py
```diff
@@ -459,13 +462,18 @@ def parse_ansi(self, string, strip_ansi=False, xterm256=False, mxp=False):
...
+        cachekey = f"{string}-{strip_ansi}-{xterm256}-{mxp}-{truecolor}"
+
         if cachekey in _PARSE_CACHE:
             return _PARSE_CACHE[cachekey]
 
         # pre-convert bright colors to xterm256 color tags
         string = self.brightbg_sub.sub(self.sub_brightbg, string)
 
+        def do_truecolor(part: re.Match, truecolor=truecolor):
+            hex2truecolor = HexColors()
```
Griatch, line 474: Why initialize this every regex query? It seems needlessly expensive. It's a stateless class, it could be initialized once at the top of this module.

evennia/utils/ansi.py
```diff
@@ -516,11 +525,12 @@ def do_xterm256_gbg(part):
 #
 
 
-def parse_ansi(string, strip_ansi=False, parser=ANSI_PARSER, xterm256=False, mxp=False):
+def parse_ansi(string, strip_ansi=False, parser=ANSI_PARSER, xterm256=False, mxp=False, truecolor=False):
     """
     Parses a string, subbing color codes as needed.
 
     Args:
+        truecolor:
```
Griatch, line 533: Missing docstring here.

evennia/utils/hex_colors.py
```diff
@@ -0,0 +1,173 @@
...
+            res = re.search(self._RE_XTERM_TRUECOLOR, bg, re.DOTALL)
+            fg_bg, r, g, b = res.groups()
+            r = hex(int(r))[2:].zfill(2)
+            g = hex(int(g))[2:].zfill(2)
+            b = hex(int(b))[2:].zfill(2)
+            prop += f"background-color: #{r}{g}{b};"
+        prop += f'"'
+        return prop
+
+    def _split_hex_to_bytes(self, tag: str) -> tuple[str, str, str]:
```
Griatch, line 104: Minor style-nitpick: Place the methods being used _before_ the methods using them. This is standard across Evennia. So move these private methods up towards the top of this class.

evennia/utils/tests/test_ansi.py
```diff
@@ -8,7 +8,9 @@
 
 from django.test import TestCase
 
-from evennia.utils.ansi import ANSIString as AN
+from evennia.utils.ansi import ANSIString as AN, ANSIParser
```
Griatch, line 11: This is not ANSI though, it's Truecolor. I think this belongs in its own test module `utils/tests/test_hex_colors.py`.

evennia/utils/tests/test_ansi.py
```diff
@@ -52,3 +54,124 @@ def test_split_with_mixed_strings(self):
         self.assertEqual(split2, split3, "Split 2 and 3 differ")
         self.assertEqual(split1, split2, "Split 1 and 2 differ")
         self.assertEqual(split1, split3, "Split 1 and 3 differ")
+
+# TODO: Better greyscale testing
```
Griatch, line 58: Ok.

evennia/utils/text2html.py
```diff
@@ -333,8 +362,16 @@ def format_styles(self, text):
...
                         classes.append(color_class)
+
                     # define the new style span
-                    prefix = '<span class="{}">'.format(" ".join(classes))
+                    if truecolor_fg == '' and truecolor_bg == '':
+                        prefix = f'<span class="{" ".join(classes)}">'
+                    else:
+                        # Classes can't be used for true color
+                        prefix = (f'<span '
+                                  f'class="{" ".join(classes)}" '
```
Griatch, line 372: You comment that classes cannot be used, yet the `class=` is set here? At the very least the comment is confusing.

Griatch (changes requested):
As far as I can tell (haven't tested this yet), this looks like a cool implementation. 
- I have a few inline comments, check them out
- We don't really have a way to test this actually works. It sounds like the `color ansi|xterm` default command should be extended to able to do `color truecolor` for supported clients and have it show a suitable range of colors (clearly showing all colors is not possible). It is also a good way to show that it gracefully maps to xterm256 (or even to ANSI) properly.

evennia/utils/text2html.py
michaelfaith84: Sorry about that! Classes can't be used for colors because of the verbosity of truecolor BUT other classes can still be applied (for things like blink, etc). I will clarify the comment.

evennia/server/portal/ttype.py
michaelfaith84: Gotcha. I was just following the example set in the case above for xterm256. Thanks for the correction.

evennia/utils/ansi.py
michaelfaith84: 🤦‍♂️ Mistake on my part! Sorry.

michaelfaith84:
> As far as I can tell (haven't tested this yet), this looks like a cool implementation.

Thanks! I was worried because it was a different method of implementation than Evennia was using but it seemed like a good choice given @InspectorCaracal's project in #3253.

>     * We don't really have a way to test this actually works. It sounds like the `color ansi|xterm` default command should be extended to able to do `color truecolor` for supported clients and have it show a suitable range of colors (clearly showing all colors is not possible). It is also a good way to show that it gracefully maps to xterm256 (or even to ANSI) properly.

I hadn't noticed that command. Thanks for pointing it out!

Griatch (approved):
Sorry for the delay, and thanks for the fixes! I think I'll approve this as is now :+1:

Issue #3317, opened 2023-10-31
[BUG] `rate` change calculations for traits in the Traits contrib break on reload


InspectorCaracal:
#### Describe the bug
A game reload will reset the "last updated" value for traits and thus stop all rate changes for all traits on all objects. The rate changes will only begin again when something else changes the current value for that specific trait, completely losing the time in between the reload and that value change.

#### To Reproduce
1. Add the traits handler from the contrib to any typeclass according to the instructions
2. Add a Counter or Gauge trait to an object or character with a non-zero rate
3. Change the current value and verify it is changing
4. `reload` the game server
5. Check the value a few times over the next ten seconds, minute, however long you want - the value will not change
6. Manually change the current value - it will begin changing again.

#### Expected behavior
The traits should actually remember their last change for the rate so that they give the correct updated value, even with a reload in the interim.

#### Environment, Evennia version, OS etc
    Evennia 2.3.0
    Windows 10
    Python 3.10

jaborsh:
The issue does not appear to require a value update, as demonstrated in the screenshot, but the number remains static until the trait is referenced again.

```py
    @lazy_property
    def traits(self):
        return TraitHandler(self)

    @property
    def test(self):
        return self.traits.get("test")

    def at_object_creation(self):
        self.traits.add(
            "test", "Test", trait_type="counter", base=1, min=0, max=100, rate=1
        )
```

![image](https://github.com/evennia/evennia/assets/73198594/de0328be-251c-444c-943c-5aab23140a8a)

PR #3495, opened 2024-04-07
CounterTrait now checks for last_update before defaulting to current time.

jaborsh:
* Obligatory Linter PEP adherence.
* CounterTrait calls `validate_input` after a server reload. Previously, this method would default `._data['last_update']` to the current time making rate freeze until trait referenced. Fix has `validate_input` check for an existing last update before assigning it the default value `time()`. #3317


Griatch (approved):
Nice fix, thanks!

> Note: This seems to be formatted to 78 character line width. Evennia uses 100 character line width (allowed by PEP8; see CODE_STYLE.md). I'll reformat that after merge, but linting like this makes it much harder to see what the actual functional change is, so best avoid doing so in a PR unless the reformat is the point of said PR :) )

Issue #3484, opened 2024-04-03
Add support for client resizing window width after initial connect


MorquinDevlar:
#### Describe the bug
When setting client-screen-width to 80 chars, many commands like "help" or "examine" will not respect this and will display output wider than the char limit set by the client.

Using Mudlet and setting the char-width first to 100 and then to 80, gives me this display:

<img width="818" alt="Screenshot 2024-04-03 at 22 46 44" src="https://github.com/evennia/evennia/assets/94325528/9cc96c82-2241-433a-8dd2-0973acf308de">

#### To Reproduce
Set client width to less than 86 chars, and type "help" or "examine"

#### Expected behavior
I would expect the commands to respect client char width, and not display borders or even text that is too wide for the display.

#### Environment, Evennia version, OS etc
Mudlet, on OSX

jaborsh:
Tried to reproduce. Set Mudlet length to 80 and length for both `CmdExamine` and `CmdHelp` was consistent with expectations. I reproduced your error by manually using `options/save screenwidth = 100`.

We could make an upgrade by having a client-side width adjustment automatically update the character options, but I believe you're seeing the error because of your character's screenwidth option and not because of a bug.

MorquinDevlar:
Thanks for value inputs @jaborsh 

I did some further investigations:

I believe the issue is that the client width is only read once, on login. When you change it in the client and not using the options command in game, the value is not updated, unless you reconnect to the game.

1. You are right in that setting the client width in options, produces the result I was expecting.
2. Setting the client width, in client, and then reconnecting will also set the client width values correctly.
3. The default value in settings is apparently only used if the client does not set anything.

The question is - which settings should have priority?

When setting the client width in client and connecting the game is then set to the client width. If we then change the client width while still connected, the game ignores it, unless we reconnect again. This, to me, seems like faulty behaviour. Either we always use the client width, or never. Having it used only on connect seems to be half a solution.

Either we ignore the client and set the values in game, or we respect the client every time.

A compromise would be to have the command setting the width, read the client width and then warn the player if the game-command is setting the width to something wider than the client is allowing, thus forcing ugly linebreaks.

Griatch:
Thanks for the help reproducing and further discussion, @jaborsh and @MorquinDevlar! 

[...]

PR #3498, opened 2024-04-09
Automatic Terminal resizing

michaelfaith84:
It turns out the protocol_flags were being updated with NAWS updates but weren't being saved. I added an exception in the Telnet dataReceived method to handle them and a flag in the options to ignore them if NAWS isn't supported, the flag is set to false, or a size is manually assigned.

Fun problem to work on that involved exploring the bowels of Evennia.

See #3484.


Griatch (approved):
Awesome, this looks good! A nice feature to support.

PR #3499, opened 2024-04-10
Fix room creation showing room name instead of typeclass path

chiizujin:

When rooms are created in-game with `dig` or `tunnel`, the feedback message shows the room's name/key instead of its typeclass path.

Bug fix.

#### Other info

**Before**
```
>dig cave1
Created room cave1(#3) of type cave1.
```
**After**
```
>dig cave2
Created room cave2(#4) of type typeclasses.rooms.Room.
```


chiizujin:
I'm making the assumption that this is what is intended.

Griatch (approved):
Thanks, but this makes me wonder why this is not caught by a unit test :thinking:

PR #3503, opened 2024-04-12
Change page date/times to display in local time based on style timezone setting

chiizujin:

The `page` command currently shows page date/times in UTC. This changes the date/times to the local time zone based on the setting in `style`.

Show local time for pages.

#### Other info

**UTC:**
```
>page
Your latest pages:
 14:07:52 to Chiizujin:> Test.
 14:07:52 from Chiizujin:< Test.
```
**Local time:**
```
>page
Your latest pages:
 00:07:52 to Chiizujin:> Test.
 00:07:52 from Chiizujin:< Test.
```

### DIFF
--- a/evennia/commands/default/comms.py
+++ b/evennia/commands/default/comms.py
@@ -18,7 +18,7 @@ from evennia.locks.lockhandler import LockException
 from evennia.utils import create, logger, search, utils
 from evennia.utils.evmenu import ask_yes_no
 from evennia.utils.logger import tail_log_file
-from evennia.utils.utils import class_from_module, strip_unsafe_input, utc_to_local
+from evennia.utils.utils import class_from_module, strip_unsafe_input, time_as_timezone
 
 COMMAND_DEFAULT_CLASS = class_from_module(settings.COMMAND_DEFAULT_CLASS)
 CHANNEL_DEFAULT_TYPECLASS = class_from_module(
@@ -1485,7 +1485,7 @@ class CmdPage(COMMAND_DEFAULT_CLASS):
                     sender = f" {sender} " if multi_send else f" {sender}"
 
                 time_zone = self.account.options.get("timezone")
-                date_created = utc_to_local(page.date_created, time_zone)
+                date_created = time_as_timezone(page.date_created, time_zone)
                 listing.append(
                     template.format(
                         date=utils.datetime_format(date_created, time_zone),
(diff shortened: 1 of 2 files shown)


Griatch (changes requested):
Hm, I don't think this is correct. While the default time zone for the server is UTC, it doesn't have to be. This is configured by the `settings.TIME_ZONE` setting. If this is changed, I'm pretty sure the `Msg` objects (`page.date_created`) will be set in that time zone, and then this solution will be wrong. 

What should be done here is instead to convert from the server `TIME_ZONE` (also if it's not UTC) to the player's time zone.

chiizujin:
When looking at the value of `page.date_created`, the messages always seem to have the time in UTC, no matter what the server's time zone setting is at the time of creation. I may be missing some other piece of the puzzle somewhere, but that's why I did a conversion from UTC to the `style` timezone setting.

However, I think the better solution is to convert the time to the local time zone if the two time zones aren't the same, whch is probably how I should have done it in the first place. This way it doesn't matter what the time zones of the server or `page.date_created` are; it gets converted if it's not the requested one. I've committed a change to do this.

Here's an example of the UTC times in the messages (this is with the current behaviour):

```
>py evennia.settings.TIME_ZONE
Australia/Sydney

>page me TEST 1
Account Chiizujin pages: TEST 1
You paged Chiizujin with: 'TEST 1'.

>page
Your latest pages:
 03:22:50 to Chiizujin:> TEST 1
 03:22:50 from Chiizujin:< TEST 1
```
page.date_created:
```
2024-04-28 03:22:50.823627+00:00
2024-04-28 03:22:50.823627+00:00
```

```
>py evennia.settings.TIME_ZONE
US/Central

>page me TEST 2
Account Chiizujin pages: TEST 2
You paged Chiizujin with: 'TEST 2'.

>page
Your latest pages:
 03:24:43 to Chiizujin:> TEST 2
 03:24:43 from Chiizujin:< TEST 2
```
page.date_created:
```
2024-04-28 03:24:43.674000+00:00
2024-04-28 03:24:43.674000+00:00
```

For context, the UTC time according to https://www.timeanddate.com/worldclock/timezone/utc was:
``03:26:03 UTC``

I'm not sure if this is related, but `django.utils.timezone.now()` always returns a UTC time, regardless of the server's time zone setting.

InspectorCaracal:
So, I discovered something interesting while poking at this, and I believe the current timezone settings are actually bugged. :laughing: I'm going to file an issue since it got a bit complicated but I'll reference it here.

chiizujin:
OK, I'll park this and https://github.com/evennia/evennia/pull/3504 until you've go that sorted.

Griatch:
Ok, will see if #3522 can be addressed before returning here.

Griatch:
#3522 was resolved.

Griatch:
@chiizujin Ping. The blocking issue was fixed.

Griatch:
@chiizujin Ping.

PR #3507, opened 2024-04-13
Fix protoype object updates changing cached prototype

### LINKED ISSUE #3505: [BUG] Prototype wizard loses changes to prototype when updating spawned objects
(Resolved by https://github.com/evennia/evennia/pull/3507.)

#### Describe the bug

When updating existing objects from a template, you are asked twice if you want to save changes to the template -- once when you use `sa` to save and again after you update the objects. If you do not answer "yes" at the second prompt, the changes you made to the template are lost after quitting and re-entering the wizard (*example 1*).

This can also result in changes that have already been saved in a template being reverted to the the spawned object's state (*example 2*). In a scenario where you make changes to a template but don't update spawned objects, and then decide later to update them and *don't* do the second save, the template will be reverted.

#### To Reproduce
In the examples, only the relevant parts of the output are shown to keep things brief.

**Example 1 - make changes to a template and update spawned objects**

```
>4
Current key: test
>test2
Current key: test2
```
```
>sa
Do you want to save/overwrite the existing prototype 'test'?
>y
Prototype saved.
Do you want to update 1 object(s) already using this prototype?
>y
- key:    [1] UPDATE: test -> test2
>u
1 objects were updated successfully.
```
At this point you are asked again if you want to save the prototype, even though the wizard has confirmed the save.
```
Do you want to save/overwrite the existing prototype 'test'?
n
```
From the main index:
```
>4
Current key: test2
```
The key *is* changed in the wizard.

After quitting the wizard, the object is correctly updated:
```
>ex test2
Name/key: test2 (#29)
```
Return to the wizard and load the template. They key has reverted to "test":
```
>4
Current key: test
```
**Example 2 - apply changes from a previously-saved template**

In this example, the prototype has had its key changed and has been saved without updating spawned objects. The wizard was quit and then re-entered:
```
>4
Current key: test2
```
```
>sa
Do you want to save/overwrite the existing prototype 'test'?
>y
Prototype saved.
Do you want to update 1 object(s) already using this prototype?
>y
- key:    [1] UPDATE: test -> test2
>u
1 objects were updated successfully.
```
Next we answer "no" because we haven't made any changes, we just saved it again to update spawned objects:
```
Do you want to save/overwrite the existing prototype 'test'?
n
```
From the main index:
```
>4
Current key: test2
```
Leave the wizard, go back in, load template. The key has reverted to that of the spawned object's *old* key (although the spawned object itself has been updated with the new key):
```
>4
Current key: test
```
#### Expected behavior

The first save prompt should be sufficient to save the template, and answering "no" should not cause existing saved changes to be reverted.

It also seems odd to be asked to be saved again -- in fact you can go round in circles indefinitely if you keep answering "yes". Perhaps it should return to the main index after the update is (or isn't) done.

chiizujin:

Fixes behaviour in https://github.com/evennia/evennia/issues/3505.

This was caused by the spawned object being used as a comparison for the update directly updating the dict in the prototype cache if it found one, rather than copy of it. This resulted in the cached prototype  being changed to reflect the spawned object.

Bug fix.

#### Other info

Note that this does *not* change the repeated save request behaviour mentioned in the bug report.

**Before**

Starting with an object spawned from a prototype and then adding a new tag ('tag_on_object') directly to the object:
```
>tag test
Tags on test: 'tag_on_object', 'test' (category: from_prototype)
```
Add a new tag to the prototype:
```
Actions: examine <num> | remove <num>
Back (attrs) | Forward (locks) | Index | Validate prototype | Quit

 1: new_tag          
```
Save and update spawned objects, then say "no" at the second save prompt:
```
Do you want to save/overwrite the existing prototype 'test'?
(...)
>n
```
View tags on prototype, without exiting the wizard:
```
Actions: examine <num> | remove <num>
Back (attrs) | Forward (locks) | Index | Validate prototype | Quit

 1: new_tag          
```
Reload the prototype while still in the wizard. The prototype in the wizard now has the tags that the spawned object had before it was updated:
```
Actions: examine <num> | remove <num>
Back (attrs) | Forward (locks) | Index | Validate prototype | Quit

 1: tag_on_object      
 2: test               
```
Note that the database has the correct tags. so ~quitting the wizard~ restarting the server and reloading the template will result in the correct tags.

The spawned object has the new tag:
```
>tag test
Tags on test: 'new_tag', 'tag_on_object', 'test' (category: from_prototype)
```

**After**

Starting from the same point as the start of the "before" test:
```
>tag test
Tags on test: 'tag_on_object', 'test' (category: from_prototype)
```
Go through the same steps to add a new tag, save, update, and say "no" at the second save prompt. Look at the tags on the prototype, without leaving the wizard:
```
Actions: examine <num> | remove <num>
Back (attrs) | Forward (locks) | Index | Validate prototype | Quit

 1: new_tag     
```
Reload the prototype while staying in the wizard and check the tags:
```
Actions: examine <num> | remove <num>
Back (attrs) | Forward (locks) | Index | Validate prototype | Quit

 1: new_tag          
```
The object has the new tag:
```
tag test
Tags on test: 'new_tag', 'tag_on_object', 'test' (category: from_prototype)
```

### DIFF
--- a/evennia/prototypes/spawner.py
+++ b/evennia/prototypes/spawner.py
@@ -323,7 +323,7 @@ def prototype_from_object(obj):
         prot["prototype_locks"] = "spawn:all();edit:all()"
         prot["prototype_tags"] = []
     else:
-        prot = prot[0]
+        prot = prot[0].copy()
 
     prot["key"] = obj.db_key or hashlib.md5(bytes(str(time.time()), "utf-8")).hexdigest()[:6]
     prot["typeclass"] = obj.db_typeclass_path


Griatch (approved):
Nice investigation and good job finding the fix!

Issue #1529, opened 2017-12-12
Feature Request: Indicate that a person is currently typing in webclient


Griatch:
#### Brief summary of issue / Description of requested feature:

When a person is typing in the webclient, other accounts/characters in the webclient should get a little indicator that they are typing, even before they actually send the text. Who sees the typing-info should be possible to customize - for example one may specify that only Characters in the same room may see that the player of a given Character is currently typing. For more chat-like setups one could even have all server-wide typing  be notified. 

#### Reasons for adding feature:

This was originally suggested by user *Apos* [here](http://musoapbox.net/topic/1959/alternative-formats-to-mu/195?page=10). This might be a very interesting feature for RP-heavy environments as well as chat-like implementations since it immediately gives an indication as to if poses and communication is forthcoming.  

This is also an example of a feature that's not possible to make (reliably) available on telnet clients but which we can reliably implement on the webclient since we control both sides. It would be tricky (but possible) to implement with a traditional telnet client with UI expansion abilities like mudlet. 

#### Expected result of feature

The suggested UI would be a overlay in a corner of the webclient. This would be a vertical list of `<name or colored circle> typing ...` or similar. You can hover over each line to get the full name of who's typing. These pop up and go away as people are actively typing or send their messages. 

#### Extra information, such as Evennia revision/repo/branch, operating system and ideas for how to solve / implement:

This should be implemented by OOB messages. 

1. A player types in the webclient.
2. The webclient sends the `is_typing(True)` OOB message to Evennia. 
3. When a player has not typed anything for a certain amount of time (10 seconds? 20?), it sends the `is_typing(False)` OOB message. 
3. In the `is_typing` inputfunc, Evennia collates these messages from all players in the game. It stores all currently typing entities and remove those who it is told are now not typing. It figures out who should see the information from each player (this should be configurable, such as those in the same room, by a lock etc). No persistent storage should be needed for this. 
4. On a fixed interval the current list of typing entities is dispersed to every eligible player using the `is_typing` outputfunc. This will be something like `target.msg(is_typing=("Griatch", "The tall man", "Anna"))`.
5. The client of each receiving player receives the `is_typing` OOB message from Evennia and use this information to build/update its display. 

#### Caveats: 
This would only work between webclient-using players. Interestingly, this OOB info could in principle be used to update the prompt of a telnet client too, but it's unlikely most (any?) telnet clients could be set up to send equivalent info based on the user typing or not. Worth investigating maybe.

[...]

Issue #2028, opened 2020-01-18
[BUG] Fatal error when using reload command


TehomCD:
#### Describe the bug
Attempting to use the `@reload` command from in-game often produces a traceback of the webserver internal port being in-use that prevents the server from restarting, leaving the game dead.

```
20-01-18 17:21:03-05 [..] Server disconnected from the portal.
20-01-18 17:21:03-05 Main loop terminated.
20-01-18 17:21:03-05 Server Shut Down.
20-01-18 17:21:03-05 Loading /home/tehom/arx/evennia/evennia/server/server.py...
20-01-18 17:21:05-05 Loaded.
20-01-18 17:21:05-05 twistd 19.10.0 (/home/tehom/arx/arxvenv/bin/python 3.8.1) starting up.
20-01-18 17:21:05-05 reactor class: twisted.internet.epollreactor.EPollReactor.
20-01-18 17:21:05-05 Traceback (most recent call last):
20-01-18 17:21:05-05   File "/home/tehom/arx/arxvenv/lib/python3.8/site-packages/twisted/internet/tcp.py", line 1361, in startListening
20-01-18 17:21:05-05     skt.bind(addr)
20-01-18 17:21:05-05 OSError: [Errno 98] Address already in use
20-01-18 17:21:05-05
20-01-18 17:21:05-05 During handling of the above exception, another exception occurred:
20-01-18 17:21:05-05
20-01-18 17:21:05-05 Traceback (most recent call last):
20-01-18 17:21:05-05   File "/home/tehom/arx/arxvenv/bin/twistd", line 11, in <module>
20-01-18 17:21:05-05     load_entry_point('Twisted==19.10.0', 'console_scripts', 'twistd')()
20-01-18 17:21:05-05   File "/home/tehom/arx/arxvenv/lib/python3.8/site-packages/twisted/scripts/twistd.py", line 31, in run
20-01-18 17:21:05-05     app.run(runApp, ServerOptions)
20-01-18 17:21:05-05   File "/home/tehom/arx/arxvenv/lib/python3.8/site-packages/twisted/application/app.py", line 674, in run
20-01-18 17:21:05-05     runApp(config)
20-01-18 17:21:05-05   File "/home/tehom/arx/arxvenv/lib/python3.8/site-packages/twisted/scripts/twistd.py", line 25, in runApp
20-01-18 17:21:05-05     runner.run()
[...]

PR #3510, opened 2024-04-15
Is typing

michaelfaith84:
New JS plugin listens for keydown events and messages the server if they're found. The plugin has several cleanup functions for potential stale data from disconnects. It also only updates when a say command is being used and only counts typing as words, letters, and punctuation (no F-keys, alt, etc).

Server side code isn't quite what @Griatch had in mind. I'll be refactoring it. Currently, it just targets others in the room. Also, the server doesn't keep track of anything, just relays messages. Also, the speaker isn't currently excluded because it made testing easier.

@Griatch asked for it some time ago.

#1529


michaelfaith84:
@Griatch In terms of getting this done in a more efficient manner, it seems like a new script typeclass would be the best idea. I'm thinking of a couple of nattributes as lists. The first list would just contain the sessions that are subscribed. it would be populated when the client requests "Say" aliases. The second would be a list of dicts, maybe something like { "speaker": session, expiry: timestamp }. Am I on the right track?

evennia/server/inputfuncs.py
```diff
@@ -650,6 +651,26 @@ def msdp_send(session, *args, **kwargs):
...
 
+def is_typing_get_aliases(session, *args, **kwargs):
+    session.msg(is_typing={'type': 'aliases', 'payload': CmdSay.aliases})
+
+
+def is_typing_state(session, *args, **kwargs):
+    # audience = ObjectDB.objects.filter(db_typeclass_path="typeclasses.characters.Character",
+    #                                    db_location=session.puppet.location).exclude(db_key=session.puppet.key)
+
+    audience = ObjectDB.objects.filter(db_typeclass_path="typeclasses.characters.Character",
```
Griatch, line 662: Only filtering on Characters is a very limited thing since devs may well change the typeclass they use for characters. Maybe you are rather looking for `DefaultCharacter.filter_family(...)` here?

Griatch (commented):
Cool functionality! This is just a draft so far, so I'll just comment. 
To be honest, I think a room scope for this is not a bad start (ideally one should be able to configure that scope though, so configurability (and ability to deactivate!) would be useful). 

While I laid out some ideas for how to do it in the issue, I'm quite fine with a simpler implementation too. 

You mention that the server only relays and don't save ... and looking at it, I think that's just fine. The less database querying going on here the better. I don't think we need to use a Script (or ServerConfig) to store anything here. What one may want to do is to collate the information and only echo it now and then - and for that one could use an in-memory cache.

michaelfaith84:
I abandoned the notion of trying to use TickerHandler and switched to delay on @InspectorCaracal 's suggestion. TickerHandler seems to persist (and then break everything) even when you set persist to false (or whatever the flag is). I did notice something odd while finishing this up... The server doesn't want to shutdown or restart while people are connected. It seems to partially disconnect users but then hangs. Closing the tabs (or just refreshing them) at that point will then allow you to stop the server.

This still only targets the same room. I have been considering reworking it so the send updates method takes an optional method that would return a queryset to make it more flexible.

evennia/server/inputfuncs.py
```diff
@@ -654,6 +662,119 @@ def msdp_send(session, *args, **kwargs):
...
+    """
+    participants = list(_IS_TYPING_PARTICIPANTS.keys())
+
+    for participant in participants:
+        if _IS_TYPING_PARTICIPANTS[participant]["session"].puppet is not None:
+
+            payload = []
+            # Get potentials
+            potentials = (DefaultCharacter.objects.filter_family(db_location=_IS_TYPING_PARTICIPANTS[participant]["session"].puppet.location)
+                          .exclude(db_key=_IS_TYPING_PARTICIPANTS[participant]["session"].puppet.key))
```
Griatch, line 678: This could be a lot more readable if extracting `_IS_TYPING_PARTICIPANTS[participant]["session"] into a variable earlier. But then again, it seems much of this could be skipped if you track the participant object sessions directly.

evennia/server/inputfuncs.py
```diff
@@ -654,6 +662,119 @@ def msdp_send(session, *args, **kwargs):
...
 
 
+def is_typing_send_update():
+    """
+    Send relevant updates to participants
+
+    """
+    participants = list(_IS_TYPING_PARTICIPANTS.keys())
+
+    for participant in participants:
```
Griatch, line 672: The _IS_TYPING_PARTICIPANT never needs to be sent across the wire as far as I can tell, so why do all this conversion back and forth to strings/keys? Just store a dict of sessions  for quick lookup and adding/removal (maybe a `{session: puppet }` map?)

evennia/server/inputfuncs.py
```diff
@@ -654,6 +662,119 @@ def msdp_send(session, *args, **kwargs):
...
+                    payload.append({
+                        "name": puppet.name,
+                        "state": _IS_TYPING_PARTICIPANTS[str(puppet.sessid)]['state'],
+                    })
+
+            _IS_TYPING_PARTICIPANTS[participant]['session'].msg(is_typing={
+                'type': 'typing',
+                'payload': payload
+            })
+            delay(5, is_typing_send_update)
```
Griatch, line 693: The delays may be what is causing the server issues, but it's not clear why. Maybe it would be worth experimenting with a direct send here (no delay) just to see if that resolves things.

Griatch (changes requested):
Thanks for your work on this. It's a cool feature. I tested locally and found some blocking issues though: 

- With players connected on the webclient, the server cannot reload. You commented seeing this as well. This is of course not acceptable; server must be able to reload in-situ, it's an important feature of Evennia. 
- Trying to chat between two players, just typing random letters on the input line, I did see the "X is typing ..." notifier pop up for the other player - very nice. However, after one or two exchanges the server became completely unresponsive, and any input I added would yield nothing. It feels like there is a bug or issue here with the delays or the queueing. 
- From a code perspective, while the core inputfunc belongs in `server/inputfuncs`, I think there's enough extra code here to worth extracting the bulk of it into a new module under `server, something like `webclient_is_typing`.
- The code itself can probably be cleaned up a bit too; I gave some suggestions inline.
- It's tricky to write unit tests for inputfuncs since they are tied to incoming data, but I think there's enough code here to warrant a unit test; it could be in a new module `test_inputfuncs`. One would not test the entire path, but one could mock the list of participants and make sure each function behaves as expected with mocked inputs/arguments.
- All of these functions need proper doc strings, as per the Evennia CODESTYLE doc.

Griatch:
Interestingly, this PR appears to be the only one that causes me to see https://github.com/evennia/evennia/issues/2028 in action locally; the server locks and won't reboot with players connected to the webclient, and then some 3-4 minutes later I get that `CannotListenError` ... so maybe adding `delays()` to the inputfunc is a venue for debugging that other issue :sweat_drops:

evennia/server/inputfuncs.py
michaelfaith84: That's how it worked originally--no issues. I have rewritten this four times from when I had it working without the server batching the responses. This issue isn't directly with delay(), it's with TICKER_HANDLER. If I so much as looked at it directly, the server would be broken on restart to the point where I'd have to nuke the install and I wasn't saving anything to the DB so nothing should have changed... It's especially interesting in that using delay can prevent the server from restarting. Regardless, nothing is logged anywhere that I could...

I'll roll the whole thing back to just relaying in real time, clean  it up, separate it, and see about writing tests.

Griatch:
@michaelfaith84 Ping.

michaelfaith84:
> @michaelfaith84 Ping.

Sorry--things have been hectic. Also, sorry for the shitty PR. I'll try and find some time to straighten it out this Monday. Thanks for the feedback.

michaelfaith84:
Aaaaand it really bothered me I made that dumpster fire and walked away so here's the stripped down version. If you feel it still warrants tests, please let me know : )

Again, @Griatch , thank you for the feedback and patience.

evennia/server/is_typing.py
```diff
@@ -0,0 +1,71 @@
...
+
+    if not is_typing:
+        return
+
+    session.msg(
+        is_typing={
+            "type": "setup",
+            "payload": {"say_aliases": CmdSay.aliases, "talking_timeout": _IS_TYPING_TIMEOUT},
+        }
+    )
```
Griatch, line 33: I feels unnecessary to just single out `say` and its aliases here.

evennia/server/is_typing.py
```diff
@@ -0,0 +1,71 @@
...
+    is_typing = options.get("ISTYPING", True)
+
+    if not is_typing:
+        return
+
+    state = kwargs.get("state")
+
+    audience = DefaultCharacter.objects.filter_family(db_location=session.puppet.location).exclude(
+        db_key=session.puppet.key
+    )
```
Griatch, line 57: I suggest `session.puppet.location.contents_get(content_type="character")` here (making sure to handle the case where `puppet` has no location) - this is a cached resource updated as characters move in and out of rooms, so avoids the db query here.

evennia/server/is_typing.py
```diff
@@ -0,0 +1,71 @@
+"""
+This module allows users based on a given condition (defaults to same location) to see
+whether applicable users are typing or not. Currently, only the webclient is supported.
+"""
+
```
Griatch, line 5: If the user wants to customize this functionality, they need to replace this module in `settings.INPUT_FUNC_MODULES`, implementing both of `is_typing_setup` and `is_typing_state`. Exactly what is required for this customization should be documented here (and added to documentation as well, but that can come later).

evennia/server/is_typing.py
```diff
@@ -0,0 +1,71 @@
...
+    options = session.protocol_flags
+    is_typing = options.get("ISTYPING", True)
+
+    if not is_typing:
+        return
+
+    session.msg(
+        is_typing={
+            "type": "setup",
+            "payload": {"say_aliases": CmdSay.aliases, "talking_timeout": _IS_TYPING_TIMEOUT},
```
Griatch, line 31: It's a dangerous practice to have core depend on default commands like this; we actively recommend users override default commands with their own alternatives, including `CmdSay`.

Griatch (changes requested):
Thanks for the updates! I think this looks a lot cleaner already. Since this is a new core functionality, this does need some further refinement though, and I think it can in fact be simplified further.

1. This makes the `CmdSay` command special. I see no reason for this. 
    - Not only is the `say` overridable in mygame (which means basing this on the default command implementation is dangerous), there are plenty of commands that would warrant an `is_typing` notification. Even in standard PP, the emote command (especially the rpsystem contrib one) is regularly used instead of say.
    - An added complexity here is that people can define `nicks`, which are personal aliases to commands. They could for example define a nick "foo" that the system maps to `say` or `emote` or whatever. 
    - So, in conclusion, I think the way to actually support this is to ditch the filter and allow for notifiying whenever someone is typing something. The issue is how heavy this becomes, but with location-caching (see comment) and timeouts on the webclient side it may be okay. It would definitely simplify the implementation though.
2. We do need unit tests for the new inputfuncs. No need to test the browser functionality per se (our testing facilities of the webclient is rather limited at this time), but it should be fine to mock the inputs to the inputfuncs to make sure they work reliably. 
3. Before merging this, I think we'll need to run the [dummyrunner](https://www.evennia.com/docs/latest/api/evennia.server.profiling.dummyrunner.html?highlight=dummyrunner) stress tester to make sure this doesn't tank performance with a lot of new cross-talk. If you want to try it, let me know, otherwise I can do it when the earlier points are addressed and this is ready to merge (note to self).

michaelfaith84:
> 1. This makes the `CmdSay` command special. I see no reason for this

Following up on this. I'm happy to complete this PR but I have very limited time. The way you want me to implement this is very inefficient--it flies in the face in what I felt was the spirit of optimization in the original outline. A helper could be added in the spirit of django's get_user_model such as get_social_cmds to return relevant commands (and by extension, their aliases). This would dramatically reduce the number of exchanges between the client and webserver. The least intrusive approach would be a dict in the config for this plugin. A deeper integration (and more convenient) method would be adding "is_social" flag or similar to commands. If you are dead set on the "send a message every single keystroke" we can do that, too. Let me know.

Griatch:
@michaelfaith84 I'm not expecting to track every key stroke, but hard coding CmdSay is not a good idea, since this is code in the core and not obvious to change. A particular developer might for example embed says in their CmdEmote command instead. Or want a typing notification for more commands (because the cost is acceptable to them). 

What we want is for the developer to be able to select which commands actually leads to the typing indicator. One could do this by the code in the portal reading a setting with the commands to report, but I think a flag on the Command class, something like `client_live_report_typing` is probably the most elegant solution.

Griatch:
@michaelfaith84  Ping :)

michaelfaith84:
> @michaelfaith84 Ping :)

Thanks for the ping. I think these changes are more  what you are looking for. Thanks again for your patience.

evennia/server/is_typing.py
```diff
@@ -0,0 +1,104 @@
+import pprint
```
Griatch: This doesn't appear used. Leftover from debugging?

evennia/server/is_typing.py
```diff
@@ -0,0 +1,104 @@
...
+
+"""
+This module allows users based on a given condition (defaults to same location) to see
+whether applicable users are typing or not. Currently, only the webclient is supported.
+"""
+
+# Notification timeout in milliseconds + <=100ms polling interval in the client.
+_IS_TYPING_TIMEOUT = 1000 * 5
```
Griatch: While we are at it, let's make this customizable in settings as well, such as `WEBCLIENT_TYPING_TIMEOUT = 5` and convert that to milliseconds.

evennia/server/inputfuncs.py
```diff
@@ -25,8 +25,9 @@
 
 from django.conf import settings
 
-from evennia.accounts.models import AccountDB
+from evennia import ObjectDB, DefaultCharacter
 from evennia.commands.cmdhandler import cmdhandler
+from evennia.commands.default.general import CmdSay
```
Griatch, line 29: This line should be removed, I think, it's not used after the refactor.

evennia/server/inputfuncs.py
```diff
@@ -25,8 +25,9 @@
 
 from django.conf import settings
 
-from evennia.accounts.models import AccountDB
+from evennia import ObjectDB, DefaultCharacter
```
Griatch: Are these used for anything?

evennia/server/is_typing.py
```diff
@@ -0,0 +1,104 @@
...
+    audience_including_typer = session.puppet.location.contents_get(content_type="character")
+
+    audience = [puppet for puppet in audience_including_typer if puppet.id != session.puppet.id]
+
+    return audience
+
+
+def is_typing_setup(session, *args, **kwargs):
+    """
+    This fetches any aliases for the "say" command and the
```
Griatch: This is not unique to the 'say' command anymore.

evennia/server/is_typing.py
```diff
@@ -0,0 +1,104 @@
...
+        return
+
+    live_report_commands = [
+        cmd for cmd in session.puppet.cmdset.current if hasattr(cmd, "client_live_report_typing")
+    ]
+
+    live_report_keywords = []  # full list
+    for cmd in live_report_commands:
+        live_report_keywords.append(cmd.key)  # actual commands
+        live_report_keywords += cmd.aliases  # aliases
```
Griatch: This will be much faster:

```python

live_report_keywords = [kw for kw in live_report_commands for kw in (cmd.key, *cmd.aliases)]
```

evennia/server/is_typing.py
```diff
@@ -0,0 +1,104 @@
...
+
+    live_report_keywords += [
+        nick.value[2]  # this is the 'nick'
+        for nick in session.puppet.nicks.all()
+        if nick.value[1].split(" ")[0] in live_report_commands  # spaced keywords
+        or nick.value[1][0] in live_report_commands  # handle things like ', ", :
+    ]
```
Griatch, line 57: This seems a bit fragile. The `nick.value` tuple is `(regex, template, pattern, replacement)`, it seems better to operate on the actual replacement `nick.value[3]` (which is the 'real' mapped result, so if you have nick 'dir' for 'look', then the `replacement` will be 'look'):

```python
live_report_keywords += [
    nick.value[2] 
    for nick in session.puppet.nicks.all()
    if nick.value[3].split(" ", 1])[0] in live_report_keys
]
```

evennia/server/is_typing.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    This should return a list of puppets to relay our client live reporting messages.
+    The example returns other puppets in the same room as the reporting session with
+    the typing puppet filtered out.
+
+    Args:
+        session: The player's current session.
+    """
+
+    audience_including_typer = session.puppet.location.contents_get(content_type="character")
```
Griatch, line 23: `session.puppet` can be `None` if the user is in OOC mode, or has not yet started puppeting a character/object. This will raise an `NoneType` error in that case. This needs to guard against this gracefully.

evennia/server/is_typing.py
```diff
@@ -0,0 +1,104 @@
...
+"""
+This module allows users based on a given condition (defaults to same location) to see
+whether applicable users are typing or not. Currently, only the webclient is supported.
+"""
+
+# Notification timeout in milliseconds + <=100ms polling interval in the client.
+_IS_TYPING_TIMEOUT = 1000 * 5
+
+
+def is_typing_get_audience(session, *args, **kwargs):
```
Griatch: It feels like it would be useful to be able to replace this module easily, without modifying server-side code - for example a game may have 'stealth', and you wouldn't want to reveal that a stealthy player is in the room with you typing.

Easiest is probably to keep this as the default, but to load this method via a setting, like `IS_TYPING_GET_AUDIENCE_CALLBACK` pointing to this by default. There are helper tools in evennia/utils.py for importing functions based on settings-provided Python paths.

evennia/web/static/webclient/js/plugins/is_typing.js
```diff
@@ -0,0 +1,265 @@
...
+            } else if (Date.now() + timeout > state.timeout) {
+                stillTyping();
+
+            }
+        // Not talking anymore but state hasn't been updated yet.
+        } else if (state.is_typing) {
+            stoppedTyping();
+        }
+
+        return true
```
Griatch: This means `onKeyUp` will always claim this event exclusively - no other plugin can handle key-up events since the plugin loop exits when getting a `true`. It probably makes sense to return `false` here since the typing indicator should not be blocking other potential events from triggering.

evennia/web/static/webclient/js/plugins/is_typing.js
```diff
@@ -0,0 +1,265 @@
...
+        setup()
+    }
+
+    /**
+     * Sends a typing indicator to the server.
+     *
+     * @param {KeyboardEvent} event - The typing state, e.g., "typing" or "idle".
+     */
+    const onKeyup = function (event) {
+        const regex = new RegExp(`^\W*(${liveReportKeywords.reduce((acc, cur)=> acc + "|" + cur, "").substring(1)})`)
```
Griatch: It feels wasteful to recompile this regex every key press. The keywords don't change all that often, it should be enough to recompile it only then.

evennia/web/static/webclient/js/webclient_gui.js
```diff
@@ -72,6 +72,24 @@ var plugin_handler = (function () {
...
+    var onKeyup = function (event) {
+        // cycle through each plugin's keyup
+        for (let n = 0; n < ordered_plugins.length; n++) {
+            let plugin = ordered_plugins[n];
+            // does this plugin handle keyup events?
+            if ('onKeyup' in plugin) {
+                // yes, does this plugin claim this event exclusively?
+                if (plugin.onKeyup(event)) {
+                    // 'true' claims this event has been handled
+                    return;
```
Griatch, line 85: Why is a is-typing event exclusive?

evennia/web/static/webclient/js/webclient_gui.js
```diff
@@ -72,6 +72,24 @@ var plugin_handler = (function () {
...
+                }
+            }
+        }
+        console.log('NO plugin handled this Keyup');
```
Griatch: Not sure we need this log on every keypress just because no handling plugin was set...

evennia/web/static/webclient/js/plugins/is_typing.js
```diff
@@ -0,0 +1,265 @@
...
+     *
+     */
+    const cleanupTimedOutPlayers = function () {
+        const now = Date.now();
+        const timedOut = []
+
+        state.typing_players.forEach((player, index)=>{
+            if (player.timeout < now) {
+                timedOut.push(index)
+                $(`#istyping-${player}`).remove()
```
Griatch: `player.name` here? `payload.name` is used below.

evennia/server/portal/telnet.py
```diff
@@ -305,8 +306,6 @@ def connectionLost(self, reason):
 
         """
         self.sessionhandler.disconnect(self)
-        if self.nop_keep_alive and self.nop_keep_alive.running:
-            self.toggle_nop_keepalive()
```
Griatch, line 309: Nop_keep_alive is important for legacy clients. Why is this removed? It seems unrelated to this PR.

evennia/web/static/webclient/css/webclient.css
```diff
@@ -346,6 +346,38 @@ div {margin:0px;}
...
 
+#istyping {
+    position: fixed;
+    top: 25px;
+    right: 25px;
+    width: auto;
+    height: auto;
+    display: none;
+    color: white;
+}
```
Griatch, line 357: It seems like this could be placed relative to existing elements, to make sure it doesn't overlap existing UI elements.

Griatch (changes requested):
This looks much better now. I went back over this now that there are changes, and found a bunch of cleanup and refactorings that could be done here, including some bugs that need to be addressed before this can merge.

michaelfaith84:
Sorry about the mess. Let me know how these changes look.

evennia/web/static/webclient/css/webclient.css
michaelfaith84: > to make sure it doesn't overlap existing UI elements.

Eh, I guess if we're worried about this we could make it draggable and persist the offset if you want.

evennia/server/inputfuncs.py
```diff
@@ -25,8 +25,8 @@
 
 from django.conf import settings
 
-from evennia.accounts.models import AccountDB
 from evennia.commands.cmdhandler import cmdhandler
+from evennia.commands.default.general import CmdSay
```
Griatch: This is still here ... we mustn't pull default commands into the core.

evennia/commands/command.py
```diff
@@ -512,6 +512,7 @@ def get_command_info(self):
                 ("self.locks", self.locks),
                 ("self.help_category", self.help_category),
                 ("self.cmdset", self.cmdset),
+                ("self.client_live_report_typing", self.client_live_report_typing),
```
Griatch, line 522: This `self.client_live_report_typing` is not defined in the base `Command` class; it will raise an AttributeError if the command doesn't have a `func()` (unlikely, but possible). Easiest way to fix this is to add `self.client_live_report_typing = False` above (next to `arg_regex` default).

evennia/server/is_typing.py
Griatch: This needs documentation, still (see previous comment).

evennia/server/is_typing.py
```diff
@@ -45,24 +46,22 @@ def is_typing_setup(session, *args, **kwargs):
         cmd for cmd in session.puppet.cmdset.current if hasattr(cmd, "client_live_report_typing")
```
Griatch: I didn't point out every instance of t his in the previous review, but this `session.puppet` can, as said, be `None`, so that needs to be defended against here too.

evennia/server/is_typing.py
```diff
@@ -45,24 +46,22 @@ def is_typing_setup(session, *args, **kwargs):
...
-    live_report_keywords = []  # full list
-    for cmd in live_report_commands:
-        live_report_keywords.append(cmd.key)  # actual commands
-        live_report_keywords += cmd.aliases  # aliases
+    # Commands and aliases
+    live_report_keywords = [kw for kw in live_report_commands for kw in (kw.key, *kw.aliases)]
 
     live_report_keywords += [
         nick.value[2]  # this is the 'nick'
         for nick in session.puppet.nicks.all()
```
Griatch, line 76: Same here - `session.puppet` can (and often is), `None`.

evennia/web/static/webclient/js/plugins/is_typing.js
```diff
@@ -197,6 +197,8 @@ let is_typing = (function (){
                     const {live_report_keywords, typing_timeout } = kwargs.payload
                     timeout = typing_timeout
                     setLiveReportKeywords(live_report_keywords)
+
+                    regex = new RegExp(`^\W*(${liveReportKeywords.reduce((acc, cur)=> acc + "|" + cur, "").substring(1)})`)
```
Griatch: The template literal will have `^\W` collapse to `^W` I think. You need to escape as `^\\W` - it will only work for no-leading space as-is.

evennia/server/is_typing.py
```diff
@@ -85,9 +84,14 @@ def is_typing_state(session, *args, **kwargs):
     if not is_typing:
         return
 
+    audience_getter = class_from_module(
+        settings.WEBCLIENT_TYPING_AUDIENCE_GETTER
+        or "evennia.server.is_typing.is_typing_get_audience_common_location"
+    )
```
Griatch: Why is this run inside every typing event? It should be able to be resolved once at the top of the module and stored in a global variable you then use henceforth (you need to reload the server to change this anyway).

evennia/web/static/webclient/js/plugins/is_typing.js
Griatch: Not part of the PR, but this will probably fire very often, making it a pretty poor throttle - and now that every keypress could lead to a server-call, this may matter (and why an actual dummyrunner stress test would be good to run on this).

Griatch (changes requested):
This is definitely an improvement, some changes remain (please don't "mark as resolved" things I've opened - let me do that once I've verified the things have actually resolved).

evennia/server/is_typing.py
michaelfaith84: It had to be loaded after the method it was pointed to was defined so I couldn't place it at the very top.

michaelfaith84:
I noticed something I hadn't considered before. Is it intentional that a command such as
```
sayhello
You say, "hello"
```
Work? 
```
------------------------------------------------------------------------------
Help for say (aliases: ', ")

speak as your character

Usage:
  say <message>

Talk to those in your current location.
------------------------------------------------------------------------------
```
Seems to suggest that the space should be enforced.

```
------------------------------------------------------------------------------
Help for look (aliases: l, ls)

looks at the room and on details

Usage:
    look <obj>
    look <room detail>
    look *<account>

Observes your location, details at your location or objects
in your vicinity.

Tutorial: This is a child of the default Look command, that also
allows us to look at "details" in the room.  These details are
things to examine and offers some extra description without
actually having to be actual database objects. It uses the
return_detail() hook on TutorialRooms for this.
------------------------------------------------------------------------------
Command 'lookmike' is not available. Maybe you meant "look"?
mike(#1)
This is User #1.
```
Further suggests that the space should be enforced.

I had intended this to *not* send to the server unless they had started entering content, not just a watch word.

Griatch (commented):
@michaelfaith84  Approved and merged with some minor cleanup. Thanks a lot for keeping at this PR and addressing all feedback along the way! 

I added a WEBCLIENT_TYPING_ENABLED setting to turn this off in case performance is affected for certain setups.

As for your last comment, let's see if people complain and we can tweak along the way. :)

PR #3511, opened 2024-04-15
Add support for article=True and format=fmt to $You(r) and $you(r)

chiizujin:

This adds two kwargs to the funcparser functions `$You()`/`$you()` and `$Your()`/`$your()`.

- `article=True`

  This will use `get_numbered_name` instead of `get_display_name` to return, for example, "a mouse" rather than "mouse".

- `format=upper|lower|capital|title`

  This enforces specific casing of the output. Note that this overrides the current behaviour of only capitalising "You/Your", making the upper- and lower-case variations of the functions interchangeable. If no format is specified then the current behaviour is used, with the separate functions working independently.

More options for the output of `$You()`, etc.

#### Examples

**Before:**

Using:
```
attacker.location.msg_contents(
    "$You() $conj(attack) $you(target).",
     from_obj=attacker, mapping={"target": target}
)
```

```
rat attacks you.
You attack rat.
```

**After:**

Using capitalisation formatting. This forces the first noun to always be capitalised, even if it is not "you".:
```
attacker.location.msg_contents(
    "$You(format=capital) $conj(attack) $you(target).",
     from_obj=attacker, mapping={"target": target}
)
```

```
Rat attacks you.
You attack rat.
```
Including a request for the indefinite article:
```
attacker.location.msg_contents(
    "$You(article=True, format=capital) $conj(attack) $you(target, article=True).",
     from_obj=attacker, mapping={"target": target}
)
```

```
A rat attacks you.
You attack a rat.
```
Without any of the new formatting kwargs:
```
attacker.location.msg_contents(
    "$You() $conj(attack) $you(target).",
     from_obj=attacker, mapping={"target": target}
)
```
Output is the same as the current behaviour:
```
rat attacks you.
You attack rat.
```
An example using `$your()`:
```
attacker.location.msg_contents(
    "$You(article=True, format=capital) $conj(strike) $your(target, article=True) leg.",
     from_obj=attacker, mapping={"target": target}
)
```
```
A rat strikes your leg.
You strike a rat's leg.
```

### DIFF
--- a/evennia/utils/funcparser.py
+++ b/evennia/utils/funcparser.py
@@ -70,6 +70,12 @@ _CLIENT_DEFAULT_WIDTH = settings.CLIENT_DEFAULT_WIDTH
 _MAX_NESTING = settings.FUNCPARSER_MAX_NESTING
 _START_CHAR = settings.FUNCPARSER_START_CHAR
 _ESCAPE_CHAR = settings.FUNCPARSER_ESCAPE_CHAR
+_YOU_FORMAT_FUNCS = {
+    "upper": str.upper,
+    "lower": str.lower,
+    "capital": str.capitalize,
+    "title": str.title,
+}
 
 
 @dataclasses.dataclass
@@ -1230,12 +1236,30 @@ def funcparser_callable_you(
 
     capitalize = bool(capitalize)
     if caller == receiver:
-        return "You" if capitalize else "you"
-    return (
-        caller.get_display_name(looker=receiver)
-        if hasattr(caller, "get_display_name")
-        else str(caller)
-    )
+        name = "You" if capitalize else "you"
+    else:
+        if kwargs.get("article"):
+            name = (
+                caller.get_numbered_name(1, looker=receiver, return_string=True)
+                if hasattr(caller, "get_numbered_name")
+                else str(caller)
+            )
+        else:
+            name = (
+                caller.get_display_name(looker=receiver)
+                if hasattr(caller, "get_display_name")
+                else str(caller)
+            )
+
+    # a specified format overrides 'capitalize' and also applies to "you"
+    if (fmt := kwargs.get('format')):
+        try:
+            name = _YOU_FORMAT_FUNCS[fmt](name)
+        except KeyError:
+            callable_name = "$You" if capitalize else "$you"
+            raise ParsingError(f"Unsupported format supplied to {callable_name} callable: {fmt}.")
+
+    return name
 
 
 def funcparser_callable_you_capitalize(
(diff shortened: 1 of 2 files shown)


Griatch (changes requested):
This looks good except for one thing: As per your examples, 

`$You()` would have to be supplied as `$You(format=capital)` for `rat` to become `Rat`. I think this is unintuitive. 
- Since we are now moving to modifying the object's display name anyway, I think it makes sense that `$You()` would return capitalized `Rat` by default, and you'd have to do `$You(format=lower)` to make it return `rat`. Without this, there is no difference between `$You` and `$you`. 
- On the other hand `$you()` makes sense to just keep the native display name unless a format is given (it shouldn't default to lower() I don't think, but I'm open to discussion on that).

chiizujin:
That sounds fair enough. The only thing I'd do differently is to *not* use `capitalize()` or `title()` for the $You/rat combination. Rather, I'd do `s[:1].upper() + s[1:]`

This is because the two functions above both change more than just the first character:

```
>>> s = 'The Knights of Ni'

>>> s.capitalize()
'The knights of ni'
# Not desired as it has lower-cased the rest of the string

>>> s.title()
'The Knights Of Ni'
# Not desired as it has upper-cased the first letter of every word

```
```
>>> s = 'the Knights of Ni'
>>> s[:1].upper() + s[1:]
'The Knights of Ni'
# This seems more acceptable
```
I have already added `"first": lambda s: s[:1].upper() + s[1:]` to `_YOU_FORMAT_FUNCS` in my local version as I have a use case for it, so I could add that here. I'm not sure about calling it `first`, though, as that does't really say what it does. I just thought `upper_first` was a bit long.

InspectorCaracal:
I know this is a bit of a nitpick, but I think it would be better if this change were implemented using a different keyword, such as `case`, rather than `format`. The keyword of `format` suggests that it would take a string matching [Python's formatting syntax](https://docs.python.org/3.8/library/string.html#formatexamples), so I think having it specified instead as e.g. `case` would be more intuitive for the actual feature.

chiizujin:
`case` does sound better, yes. It's closer to what it's doing.

Griatch:
@chiizujin To follow up on the comments, yes, I agree that that `s[:1].upper() + s[1:]` is the better way to do it for that combination. I do think that there is an advantage to calling it `format` over `case` though, since the `format` name allows for possibly expanding this functionality with other non-case related capabilities in the future (not sure what those would be, but maybe color/nocolor or similar?)

Griatch:
@chiizujin Ping.

Griatch:
Author is non-responsive for almost two years, but this is nice functionality, so finishing up this PR while retaining original attribution.

PR #3517, opened 2024-04-23
Don't ignore empty search candidates

InspectorCaracal:
Providing an empty list of search candidates currently results in the unexpected behavior of falling back to the default search candidates rather than finding nothing. This replaces the truthiness check to more explicitly verify that the candidates are not the default kwarg of `None`

Bug fixing


Griatch (approved):
Nice catch!

PR #3521, opened 2024-04-27
Support full capitalization in GMCP commands

InspectorCaracal:
Changes the GMCP command encoding to only run capitalize() on command parts that are not fully uppercase.

A previous conversation on Discord identified the current GMCP encoding's command capitalization as somewhat lacking, based on their own previous use of GMCP in other MUD libraries. Additionally, today a specific use-case arose that is currently impossible: sending a Mudlet package install command, as per [the Mudlet docs](https://wiki.mudlet.org/w/Manual:GMCP_Extensions) - it requires the ability to send a fully capitalized command section, `Client.GUI`.

I'm not extremely familiar with the specifications of the GMCP protocol, so I wrote this under the assumption that mixed-case command parts are not allowed.

### DIFF
--- a/evennia/server/portal/telnet_oob.py
+++ b/evennia/server/portal/telnet_oob.py
@@ -257,11 +257,8 @@ class TelnetOOB:
         if cmdname in EVENNIA_TO_GMCP:
             gmcp_cmdname = EVENNIA_TO_GMCP[cmdname]
         elif "_" in cmdname:
-            if cmdname.istitle():
-                # leave without capitalization
-                gmcp_cmdname = ".".join(word for word in cmdname.split("_"))
-            else:
-                gmcp_cmdname = ".".join(word.capitalize() for word in cmdname.split("_"))
+            # enforce initial capitalization of each command part, leaving fully-capitalized sections intact
+            gmcp_cmdname = ".".join(word.capitalize() if not word.isupper() else word for word in cmdname.split("_"))
         else:
             gmcp_cmdname = "Core.%s" % (cmdname if cmdname.istitle() else cmdname.capitalize())
 


Griatch (approved):
This is not clear from the GMCP specs, hence different clients implement this differently. My interpretation is that GMCP commands are always on the form `Word.Word` and not `Word.WORD` but that may also be because the IRE reference implementation looked like that (but I could be wrong too, it's been a while). Either way, if people find the current implementation hinders them, I don't see a problem with making it more lenient.

MorquinDevlar:
I would still advocate for not enforcing any kind of capitalisation on GMCP output. Let the developer set it like they want.

Take a look at this list from IRE:

_Room.WrongDir
 Sent if the player attempts to move in a non-existant direction using the standard movement commands
 Upon receiving this message, the client can safely assume that the specified direction does not lead anywhere at this time
 Message body is a string with the name if the non-existant exit
 Example: Room.WrongDir "ne"

Room.Players
 Object containing player details, each key is the short name of the player and each value is the full name including titles for the player
 Example: Room.Players [{ "name": "Tecton", "fullname": "Tecton the Terraformer" }, {"name": "Cardan", "fullname": "Cardan, the Curious" }]

Room.AddPlayer
Message body has the same object structure as Room.Players except that it only contains the one player being added to the room.
Example: Room.AddPlayer { "name": "Cardan", "fullname": "(Cardan, the Curious)" }

Room.RemovePlayer
 Message body has the same object structure as Room.Players except that it only contains the one player being removed from the room.
 Example: Room.RemovePlayer "Cardan"_


They have a lot of commands sent with SnakeCase, which would not be compatible with an enforced capitalisation.

InspectorCaracal:
@MorquinDevlar Since this PR has already been merged and closed, you'd be best off creating a feature request or second PR for your desired outcome.

MorquinDevlar:
@InspectorCaracal You are absolutely right, and I actually also meant CamelCase not snake_case. My only excuse was the lateness of the hour, as I was writing. 

I'll push in a PR, with my suggested change. 

I really appreciate you taking the time to make the first change. Makes it easier to put in my own suggestion.

PR #3523, opened 2024-04-28
Polish translation (Moonchasered)

Moonchasered:
Hello!

As in the title - just a Polish translation I made guided by the Evennia docs. I covered most of the .po file (>80%) avoiding those which did not have much sense as those were mostly errors for developers. 

That was my first step in trying to adjust Evennia for a polish language. Later I will have to deal with a declension and other weird stuff in the Polish language so I guess i will just continue working on it as a fork. Or at least I will try to do so. 

Anyway, Evennia is an awesome mud engine and I'm glad I found it and hopefully someone will take advantage of this small PR.


Griatch (approved):
The polish translation is fine, but I'm wondering about the reason for adding `unidecode` to the dependencies.

pyproject.toml
```diff
@@ -78,6 +78,7 @@ dependencies = [
   "Jinja2 < 3.1",
   "tzdata >= 2022.6",
   "pydantic < 2.0",
+  "unidecode <= 1.3.8",
```
Griatch, line 81: There is no reason given for this. Why is this added?

Griatch (changes requested):
While the polish translation is fine, I'm wondering about the added library.

pyproject.toml
Moonchasered: Sorry, my bad, that should go to a different branch of my fork, something for my own project. Fixing it.

Moonchasered:
A mistake, sorry. Supposed to be part of a different branch of my own.  Reverted.

PR #3526, opened 2024-04-29
Update host provider info

InspectorCaracal:
The Akami/Linode service described is their shared computing suite, which are all VPS. Cloud services are also offered by the company but are not that service.


Griatch (approved):
Good catch!

Issue #3532, opened 2024-05-06
[BUG] CmdHelp File Identification Flaw


jaborsh:
#### Describe the bug
When using `CmdHelp`, Evennia searches for appropriate documentation. Usually it successfully chooses the most relevant page given the keyword provided by the user, but it does this even if an exact match is given.

![image](https://github.com/evennia/evennia/assets/73198594/50d7ba86-07d4-4570-b49c-d0997020a19e)

#### To Reproduce
1. Create `CmdInteract` and `CmdInteractions` or any other version of commands so long as one is "contained" within the other (i.e: "interact" is within "interactions")
2. View in `help`.
3. Attempt `help interactions`
4. Receive help page for `CmdInteract`

#### Expected behavior
Evennia should, in my opinion, first determine if an exact match exists before searching by closest association.

#### Environment, Evennia version, OS etc
- Py 3.10
- Ev 4.11
- Win11

jaborsh:
I removed my local changes to `CmdHelp` to verify it isn't developer error. The issue remains the same, just semantically different. The screenshot is from `help interact` but it yields the `interactions` docstring.

![image](https://github.com/evennia/evennia/assets/73198594/a0b508a6-fa85-474f-939b-b2009109d3f1)

PR #3533, opened 2024-05-07
Improve help search matching

chiizujin:

If help entries have similar names, it may be impossible to access one or more of them. This change fixes that, resolving https://github.com/evennia/evennia/issues/3532.

Improvement/bug fix.

#### Other info

This seems to be related to lunr and is further complicated by the fact that commands can be prefixed with "@".

In the case of "test" and "tests", lunr ranks both of these the same, no matter which is searched for:

```
help test
{'ref': 'test', 'score': 44.819, 'match_data': <MatchData "test">}
{'ref': 'tests', 'score': 44.819, 'match_data': <MatchData "test">}

help tests
{'ref': 'test', 'score': 44.819, 'match_data': <MatchData "test">}
{'ref': 'tests', 'score': 44.819, 'match_data': <MatchData "test">}
```

I don't know if this PR is the most elegant fix but I couldn't find any lunr-based solution.

#### Examples

**Before:**

```
>help test
Help for test
(...)
Other topic suggestions:
tests

>help tests
Help for test
(...)
Other topic suggestions:
tests
```
```
>help spawn
Help for spawns
(...)
Other topic suggestions:
spawn

>help spawns
No help found

There is no help topic matching 'spawns'.
... But matches were found within the help texts of the suggestions below.

Other topic suggestions:
spawn  spawns
```
Example from the bug report:
```
>help interactions
Help for interact
(...)
Other topic suggestions:
interactions

>help interact
Help for interact
(...)
Other topic suggestions:
interactions
```

**After:**

```
>help test
Help for test
(...)
Other topic suggestions:
tests

>help tests
Help for tests
(...)
Other topic suggestions:
test
```
```
>help spawn
-------------------------------------------------------------------------------------
Help for spawn (aliases: olc)
(...)
Other topic suggestions:
spawns

>help spawns
-------------------------------------------------------------------------------------
Help for spawns
(...)
Other topic suggestions:
spawn
```
```
>help interact
Help for interact
(...)
Other topic suggestions:
interactions

>help interactions
Help for interactions
(...)
Other topic suggestions:
interact
```


evennia/commands/default/help.py
```diff
@@ -487,16 +487,36 @@ def do_search(self, query, entries, search_fields=None):
                 {"field_name": "tags", "boost": 1},  # tags are not used by default
             ]
         match, suggestions = None, None
+        base_query = query.lstrip("@")
```
Griatch, line 490: This is a good idea, but this is too specific since `@` is not the only possible prefix understood (and ignored) by the command parser. The understood prefixes are given by `settings.CMD_IGNORE_PREFIXES`.

Griatch (changes requested):
This looks like it could work; Command syntax is arbitrary though, so more than just `@` needs to be considered.

evennia/commands/default/help.py
chiizujin: I've added support for `settings.CMD_IGNORE_PREFIXES`.

Griatch (changes requested):
Nice, just needs to handle the empty query as well.

evennia/commands/default/help.py
```diff
@@ -477,6 +477,11 @@ def do_search(self, query, entries, search_fields=None):
             tuple: A tuple (match, suggestions).
 
         """
+        def strip_prefix(query):
+            if query[0] in settings.CMD_IGNORE_PREFIXES:
```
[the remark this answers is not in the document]
chiizujin: I've updated this to handle the empty string.

Griatch (approved):

PR #3541, opened 2024-05-15
Rework Object searching to behave more consistently

InspectorCaracal:
This PR makes modifications to both the DefaultObject manager methods of `get_objs_by_key_and_alias` and `search_object`.

`get_objs_by_key_and_alias` will now properly respect partial matches in its queries, leading to the expected result of an empty search string partial-matching all candidates and additionally running fewer queries.

`search_object` has been moderately streamlined, removing the need for special-case handling of tags or search terms, and applying all filtering keywords consistently to all types of searches rather than unintuitively picking and choosing based on which combination of keywords were provided.

Usability improvements

The undesirable behavior brought up on Discord where searching with tag criteria but no search string is resolved by this PR. It additionally resolves several equally problematic behaviors that currently result when giving the search method an empty search string, such as returning an empty result when people typically expect an empty string to partial-match *all* keys not none, and the unfortunate tag-only search disregarding all other keywords entirely.

### DIFF
--- a/evennia/contrib/game_systems/clothing/tests.py
+++ b/evennia/contrib/game_systems/clothing/tests.py
@@ -85,10 +85,7 @@ class TestClothingCmd(BaseEvenniaCommandTest):
         )
 
         # Test remove command.
-        # NOTE: commenting out due to failing via the search refactor - however, this command 
-        # SHOULD be providing standard wrong-args feedback, like CmdWear.
-        # this will be fixed and the test amended in a separate PR
-        # self.call(clothing.CmdRemove(), "", "Could not find ''.", caller=self.wearer)
+        self.call(clothing.CmdRemove(), "", "Usage: remove <worn clothing object>", caller=self.wearer)
         self.call(
             clothing.CmdRemove(),
             "hat",
--- a/evennia/contrib/game_systems/clothing/tests.py
+++ b/evennia/contrib/game_systems/clothing/tests.py
@@ -85,7 +85,10 @@ class TestClothingCmd(BaseEvenniaCommandTest):
         )
 
         # Test remove command.
-        self.call(clothing.CmdRemove(), "", "Could not find ''.", caller=self.wearer)
+        # NOTE: commenting out due to failing via the search refactor - however, this command 
+        # SHOULD be providing standard wrong-args feedback, like CmdWear.
+        # this will be fixed and the test amended in a separate PR
+        # self.call(clothing.CmdRemove(), "", "Could not find ''.", caller=self.wearer)
         self.call(
             clothing.CmdRemove(),
             "hat",
--- a/evennia/objects/manager.py
+++ b/evennia/objects/manager.py
@@ -281,8 +281,7 @@ class ObjectDBManager(TypedObjectManager):
         Args:
             ostring (str): A search criterion.
             exact (bool, optional): Require exact match of ostring
-                (still case-insensitive). If `False`, will do fuzzy matching
-                using `evennia.utils.utils.string_partial_matching` algorithm.
+                (still case-insensitive). If `False`, will do fuzzy matching with a regex filter.
             candidates (list): Only match among these candidates.
             typeclasses (list): Only match objects with typeclasses having thess path strings.
 
@@ -305,7 +304,7 @@ class ObjectDBManager(TypedObjectManager):
         cand_restriction = candidates is not None and Q(pk__in=candidates_id) or Q()
         type_restriction = typeclasses and Q(db_typeclass_path__in=make_iter(typeclasses)) or Q()
         if exact:
-            # exact match - do direct search
+            # exact matches only
             return (
                 (
                     self.filter(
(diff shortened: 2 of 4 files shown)


InspectorCaracal:
Whoops.

Converting this to a draft because a) there's more failing unit tests, and b) I just realized while working on adding new tests for the two methods I modified that I unwittingly lost the match scoring mechanism. Since that needs to be accounted for, I need to rethink some of my changes to `get_objs_with_key_or_alias`

InspectorCaracal:
On further review of how `string_partial_matching` works, it seems that the "score" is just how many of the input words were matched in the alternative text. Since the score is zeroed when any input words are lacking, this means my single query will in practice return equivalent results since it requires that all input words are partially matched in order. As such, I'm leaving the modifications as I originally wrote them.

The failed unit test is a bit more of a pain, however. It seems as though the clothing contrib, rather than having a proper usage feedback on the `remove` command when used with no args, was set up to verify that it returned a "no object found" error instead. Since caller.search is fuzzy by default, this is obviously failing now that the empty term fuzzy search works "correctly"....

evennia/contrib/game_systems/clothing/tests.py
```diff
@@ -85,7 +85,10 @@ def test_clothingcommands(self):
         )
 
         # Test remove command.
-        self.call(clothing.CmdRemove(), "", "Could not find ''.", caller=self.wearer)
+        # NOTE: commenting out due to failing via the search refactor - however, this command
```
InspectorCaracal, line 88: I've temporarily just commented out this assertion since I intend to make a second PR to correct the oversight in the contrib command, but I'm unsure if this is the correct approach.

I could alternatively write the other PR, wait for it to be merged, and rebase this one onto it.

Or, I could leave this one failing with the assumption I'll correct it. (I'm not a fan of this third option. >.>)

Griatch (commented):
At first glance it looks like it would comprehensive. I'm not entirely convinced that it's obvious that you'd find all candidates by fuzzy-matching with an empty string honestly, but it does simplify things a bit too. Also nice use of iregex. 

I need to run this more to check before merging this.

InspectorCaracal:
> I'm not entirely convinced that it's obvious that you'd find all candidates by fuzzy-matching with an empty string honestly, but it does simplify things a bit too. 

I tried it out in a test game and it seemed like it was working, so let me know if you find cases where it doesn't. 👍

evennia/objects/manager.py
```diff
@@ -486,51 +440,59 @@ def _search_by_tag(query, taglist):
...
-            matches = self.none()
-        elif len(matches) > 1 and match_number is not None:
-            # multiple matches, but a number was given to separate them
-            if 0 <= match_number < len(matches):
+        if match_number is not None:
+            # this indicates someone attempting to get a single match with a match-number
+            if match_number >= len(matches):
+                # the search attempted to target a higher match index than exists in the candidates.
+                # This leads to a no-match.
+                matches = self.none()
```
Griatch, line 486: This first `if` conditional seems like it is superfluous ... both this and the `else:` statement fulfills the same condition (being outside of the range).

Griatch (changes requested):
Testing some more, I think this is a nice change, it also cleans up/shortens the code a lot, that `iregex=` is a nice touch. Apparently I wrote much of this code a whopping 12 years ago! :astonished: 

(By the way, changing the search result in a way that breaks unit tests makes it clear this is a backwards-incompatible change.)  

So far it seems to behave as expected in my testing. I comment on some code below. I did find one piece of strangeness though: 

```python
> py self.search("")
Griatch-1
Limbo-2
rose-3
cactus-5
box-1
box-2
box-3
```
Whereas the box-1..3 are the expected multi-match differentiators, the other added numbers appear for _single_ matches and appear to be the dbrefs rather than the run-numbers. Expectedly, doing e.g. `py self.search("rose-3")` will fail. As far as I can tell, this PR interacts in an unexpected way with the listing display of multi-matches.

InspectorCaracal:
> Whereas the box-1..3 are the expected multi-match differentiators, the other added numbers appear for single matches and appear to be the dbrefs rather than the run-numbers.

Ohh, that's weird and very interesting. I'm looking forward to figuring that one out!

InspectorCaracal:
Interestingly, in a fresh gamedir on this branch of evennia, I had a completely different result:
```
> py self.search('')
>>> self.search('')
More than one match for '' (please narrow target):
 admin-1
 Limbo-2
 box-3
 box-4
 box-5
 ```

This is also technically wrong, but it's more in line with what I would've expected. :thinking:

It seems to also actually be highlighting a flaw in the existing multimatching system's ability to handle partial matches - if I switch to the current evennia main, it behaves like this:
```
>>> self.search('red')
More than one match for 'red' (please narrow target):
 red box-1
 red block-2
>>> self.search('red block-2')
Could not find 'red block-2'.
```

I'll try a few more times to see if I can replicate your weird result, but I'll make a separate bug report for the partial-match multimatching issue.

**Comment by InspectorCaracal** [comment]
@Griatch I officially cannot replicate anything like your result. I've tried multiple different collections of items, quelled and unquelled on the superuser and auto-created character, destroying and creating new objects, and even made a command that does nothing but an empty-string search to try it on a Player-level character, and the results are *always* numbered in simple list order.

```
More than one match for '' (please narrow target):
 box-1
 box-2
 red box-3
 blue box-4
 red block-5
 red block-6
 field-7
 south-8 [s]
 tester-9
 box-10
 ```
 
 So as far as I can tell, that wasn't/isn't an issue with this PR.

evennia/objects/tests.py
```diff
@@ -247,6 +247,51 @@ def test_get_objs_with_key_and_typeclass(self):
         )
         self.assertEqual(list(query), [self.char1])
 
+    def test_get_objs_with_key_or_alias(self):
+        query = ObjectDB.objects.get_objs_with_key_or_alias("Char")
+        self.assertEqual(list(query), [self.char1])
+        self.assertEqual(list(query), [self.char1])
```
Griatch, line 253: This is done twice? A typo or some reason?

evennia/objects/tests.py
```diff
@@ -247,6 +247,51 @@ def test_get_objs_with_key_and_typeclass(self):
...
+        query = ObjectDB.objects.get_objs_with_key_or_alias(
+            "Char", typeclasses="evennia.objects.objects.DefaultObject"
+        )
+        self.assertFalse(query)
```
Griatch, line 257: Should this be `self.assertNone` to be explicit?

evennia/objects/manager.py
Griatch: This still seems to be the case.

evennia/contrib/game_systems/clothing/tests.py
Griatch: It would be a good idea to make sure all unit tests work correctly so we don't merge a version that is actually broken for some use cases (even if the version is short-lived). So I think it makes sense to fix this contrib/unit test as part of this PR as well.

Griatch (changes requested):
:thinking: I can't reproduce the separate box-1,2,3 behavior I saw now either. It does make me a little worried :sweat_smile:

Overall, partial matching seems to be working fine with this change. As you mention, this does highlight a problem with multimatch display for searching for `""` though - the resulting list of multimatches is nonsense and not actionable (this is not a regression from this PR though, as you say, I see the same behavior in main). That could be a separate investigation.

I do think this should not merge until it actually comes with a fully updated unit test suite though (rather than commented out tests); those adjustments can/should be done in the same PR IMO. I also give some comments on the unit tests for this change.

InspectorCaracal:
> I do think this should not merge until it actually comes with a fully updated unit test suite though (rather than commented out tests); those adjustments can/should be done in the same PR IMO. I also give some comments on the unit tests for this change.

Thanks! I was waiting for an answer to this before tackling the needed changes so I could do everything at once.

evennia/objects/tests.py
InspectorCaracal: If we want explicit, then I'll need to check if it's equal to an empty list. I can do that. 👍

Griatch:
Ping, not much left to do on this one :)

InspectorCaracal:
Haven't forgotten about this one either, hopefully will be wrapping it up soon.

evennia/objects/manager.py
InspectorCaracal: The outer `if match_number is not None` is functionally relevant. In terms of code logic, it means that the inner checks regarding whether the match-number index is valid will only be checked if there is, in fact, a match-number provided in the term to begin with. I don't believe this to be superfluous or redundant.

InspectorCaracal: Ah, I've been misinterpreting what you meant by "the first if"; since it's below two ifs I thought you meant the first of the two you commended below.

Griatch (approved):
Thanks for your work addressing the comments. Looks good now, and with the fixed unit test too.

PR #3543, opened 2024-05-19
bugfix: change pathing to relative url

iLPdev:
changes url pathing for slot validation test

if following tutorial, runtime errors for conflicting models without fix

### DIFF
--- a/evennia/contrib/tutorials/evadventure/tests/test_equipment.py
+++ b/evennia/contrib/tutorials/evadventure/tests/test_equipment.py
@@ -89,7 +89,7 @@ class TestEquipment(EvAdventureMixin, BaseEvenniaTest):
         obj = MagicMock()
         obj.size = size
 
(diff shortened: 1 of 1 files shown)


Griatch (changes requested):
This breaks Evennia's test suite. Not sure if you can use a relative path in a mock patch.

InspectorCaracal:
I think this maybe was supposed to be a documentation fix? It looks as though the path here is the one the function would have while going through the tutorial in your own game.

Griatch:
If it could be patched relatively that would be a benefit, because I suppose this becomes an issue if you copy&paste verbatim.

iLPdev:
yeah, this is supposed to be about the documentation in the tutorial example.

Griatch:
This PR can't merge though; because while the tutorial assumes that you are working in your own game dir, the Evennia-side code must operate from inside the Evennia repository, and this PR breaks that. Closing.

PR #3550, opened 2024-05-28
Fix rpsystem contrib searching issues

InspectorCaracal:
This PR modifies the `get_search_result` method to avoid erroneous global searches and multimatches, and `parse_sdescs_and_recogs` to correctly remove the multimatch disambiguator from the actual string to be matched

I also added a new test to verify the correct behaviors.

Bug fixing

### DIFF
--- a/evennia/contrib/rpg/rpsystem/rpsystem.py
+++ b/evennia/contrib/rpg/rpsystem/rpsystem.py
@@ -442,7 +442,7 @@ def parse_sdescs_and_recogs(
                 (re.search(rquery, text, _RE_FLAGS), obj, text) for obj, text in candidate_map
             )
             # filter out any non-matching candidates
-            bestmatches = [(obj, m.group()) for m, obj, text in matches if m]
+            bestmatches = [(obj, mtch.group()) for mtch, obj, text in matches if mtch]
 
         else:
             # to find the longest match, we start from the marker and lengthen the
--- a/evennia/contrib/rpg/rpsystem/rpsystem.py
+++ b/evennia/contrib/rpg/rpsystem/rpsystem.py
@@ -392,7 +392,7 @@ def parse_sdescs_and_recogs(
         # if no sdesc, include key plus aliases instead
         else:
             candidate_map.append((obj, obj.key))
-        candidate_map.extend([(obj, alias) for alias in obj.aliases.all()])
+            candidate_map.extend([(obj, alias) for alias in obj.aliases.all()])
 
     # escape mapping syntax on the form {#id} if it exists already in emote,
     # if so it is replaced with just "id".
@@ -439,7 +439,7 @@ def parse_sdescs_and_recogs(
                 (re.search(rquery, text, _RE_FLAGS), obj, text) for obj, text in candidate_map
             )
             # filter out any non-matching candidates
-            bestmatches = [(obj, match.group()) for match, obj, text in matches if match]
+            bestmatches = [(obj, m.group()) for m, obj, text in matches if m]
 
         else:
             # to find the longest match, we start from the marker and lengthen the
@@ -1333,30 +1333,28 @@ class ContribRPObject(DefaultObject):
         """
         # we also want to use the default search method
         search_obj = super().get_search_result
-        is_builder = self.locks.check_lockstring(self, "perm(Builder)")
+        is_builder = self.permissions.check("Builder")
+        results = []
 
-        if candidates:
-            candidates = parse_sdescs_and_recogs(
+        if candidates is not None:
+            searched_results = parse_sdescs_and_recogs(
                 self, candidates, _PREFIX + searchdata, search_mode=True
             )
-            results = []
-            for candidate in candidates:
-                # we search by candidate keys here; this allows full error
-                # management and use of all kwargs - we will use searchdata
-                # in eventual error reporting later (not their keys). Doing
-                # it like this e.g. allows for use of the typeclass kwarg
-                # limiter.
-                results.extend(
-                    [obj for obj in search_obj(candidate.key, **kwargs) if obj not in results]
-                )
-
-            if not results and is_builder:
-                # builders get to do a global search by key+alias
-                results = search_obj(searchdata, **kwargs)
+            if not searched_results and is_builder:
+                # builders get to do a search by key
+                results = search_obj(searchdata, candidates=candidates, **kwargs)
+            else:
+                # we do a default search on each result by key, here, to apply extra filtering kwargs
+                for searched_obj in searched_results:
+                    results.extend(
+                        [
+                            obj
+                            for obj in search_obj(searched_obj.key, candidates=candidates, **kwargs)
+                            if obj not in results
+                        ]
+                    )
         else:
-            # global searches with #drefs end up here. Global searches are
-            # only done in code, so is controlled, #dbrefs are turned off
-            # for non-Builders.
+            # no candidates means it's a global search, so we pass it back to the default
             results = search_obj(searchdata, **kwargs)
         return results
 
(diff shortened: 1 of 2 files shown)


InspectorCaracal:
As a bit of a follow-up note: once/if my other PR for the search methods is merged, I will revisit `get_search_results` here as well as I'll be able to greatly simplify and streamline the results-filtering step.

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -439,7 +442,7 @@ def parse_sdescs_and_recogs(
                 (re.search(rquery, text, _RE_FLAGS), obj, text) for obj, text in candidate_map
             )
             # filter out any non-matching candidates
-            bestmatches = [(obj, match.group()) for match, obj, text in matches if match]
+            bestmatches = [(obj, m.group()) for m, obj, text in matches if m]
```
Griatch, line 445: Why turn `match` into a one-letter variable `m`? PEP8 clearly recommends staying >1 character length for all Python variables.

Griatch (commented):
It's a bit hard to determine what is being fixed here, because the issue doesn't reference an issue or gives any examples of what the issues with "global searches and multimatches" are. 

(Also, since many of the changes here appear to be removal of comments and minor restructuring, the actual functional changes become harder to spot when reviewing. It would be helpful with some PR line comment pointing out the particular intent and change applied (like passing `candidates=[searched_obj]` into `search_obj`  where it wasn't in the past) :pray: )

evennia/contrib/rpg/rpsystem/rpsystem.py
InspectorCaracal: Ah, my bad, I was mimicking the format I've seen used for list comprehensions etc. elsewhere in the code.

As for why, it's because `match` is a builtin keyword now and I get paranoid about overwriting those. >.> So I can just take this change out.

InspectorCaracal:
I'll see about condensing the support conversations on discord around the issues here to add to the PR and look into the commenting thing when I'm next at a computer.

evennia/contrib/rpg/rpsystem/rpsystem.py
Griatch: I shouldn't be using one-letter vars in list comprehensions, but it's possible it has snuck in somewhere (there's a lot of code after all). But it's actually a fair point concerning the use of `match`. We shouldn't use an existing python keyword. But one can still use something other than a one-letter one to appeal to PEP8 :)

InspectorCaracal: Sounds good! I'll do something like mat or mtch instead.

sintaxterror:
It doesn't appear that the underlying issue with the current version of this contribution was made clear, so let me see if I can fix that.

At the moment, if you use RpCharacter and you attempt to pick something up (note that you do not have to be super user to have this happen) the game will (if multiples of the same object exist) display every single one of those objects in your current location even if those objects should be out of reach. In particular, and most alarmingly, in someone else's inventory. Thankfully, it will not let you get an object from someone else, however if multiples of an object exist, get thing will trigger multimatch as you might expect, but get thing-1 will say you can't pick it up. Get 1 thing or get 3 things will let you pick up multiple of the same thing, but otherwise, you are unable to get anything. Side note, you can drop the same object in your inventory multiple times, even if your not holding it. As a builder or higher, the situation becomes worse, as trying to get an object will (quite spamfully) return every single object of that type in the game. even if you are in the same room with the object you intend to pick up.

InspectorCaracal:
Thank you sintax! I haven't had an opportunity to dig back up our conversation and relay it here, so I appreciate you taking the time to do so.

Griatch, sintaxterror is the one who ran across this issue and reported it on discord, where I helped work out what was going on and confirmed it.

Essentially, it is caused by what I described in the first post: the search is falling back to a global search when it shouldn't.

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -422,7 +422,10 @@ def parse_sdescs_and_recogs(
 
         # first see if there is a number given (e.g. 1-tall)
         num_identifier, _ = marker_match.groups("")  # return "" if no match, rather than None
+        # get the beginning of the actual text, minus the numeric identifier
         match_index = marker_match.start()
+        if num_identifier:
+            match_index += len(num_identifier) + 1
```
InspectorCaracal, line 428: strip the numeric identifier from the string to search for - without this, it's still trying to match a name of e.g. "2-box"

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -439,7 +442,7 @@ def parse_sdescs_and_recogs(
                 (re.search(rquery, text, _RE_FLAGS), obj, text) for obj, text in candidate_map
             )
             # filter out any non-matching candidates
-            bestmatches = [(obj, match.group()) for match, obj, text in matches if match]
+            bestmatches = [(obj, mtch.group()) for mtch, obj, text in matches if mtch]
```
InspectorCaracal, line 445: avoid overlapping the keyword `match` (which wasn't present when this was originally written)

InspectorCaracal: this wasn't supposed to be a review, just a comment, but i fatfingered it 😩

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -1333,30 +1336,30 @@ def get_search_result(
         """
         # we also want to use the default search method
         search_obj = super().get_search_result
-        is_builder = self.locks.check_lockstring(self, "perm(Builder)")
+        is_builder = self.permissions.check("Builder")
+        results = []
 
-        if candidates:
-            candidates = parse_sdescs_and_recogs(
+        if candidates is not None:
```
InspectorCaracal, line 1342: explicitly checks candidates being unset (`None`) versus just being an empty list - e.g. when searching for an object in an empty inventory inventory - avoiding *one* case where it was falling back to a global search when it shouldn't

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -1333,30 +1336,30 @@ def get_search_result(
...
+            searched_results = parse_sdescs_and_recogs(
```
InspectorCaracal, line 1343: this step should not modify the candidates, since it's doing an actual search. replacing candidates here broke the ability for builders to search the original candidates by key and instead did the key-search as a global search every time regardless of the candidates.

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -1333,30 +1336,30 @@ def get_search_result(
...
-                results.extend(
-                    [obj for obj in search_obj(candidate.key, **kwargs) if obj not in results]
-                )
-
-            if not results and is_builder:
-                # builders get to do a global search by key+alias
-                results = search_obj(searchdata, **kwargs)
+            if not searched_results and is_builder:
+                # builders get to do a search by key
+                results = search_obj(searchdata, candidates=candidates, **kwargs)
```
InspectorCaracal, line 1348: this search is intended as a fallback when no sdesc/recog matches are found, so it should be done only when there are no sdesc/recog matches

evennia/contrib/rpg/rpsystem/rpsystem.py
```diff
@@ -1333,30 +1336,30 @@ def get_search_result(
...
+                                searched_obj.key, candidates=[searched_obj], **kwargs
+                            )
+                            if obj not in results
+                        ]
+                    )
         else:
-            # global searches with #drefs end up here. Global searches are
-            # only done in code, so is controlled, #dbrefs are turned off
-            # for non-Builders.
+            # no candidates means it's a global search, so we pass it back to the default
```
InspectorCaracal, line 1362: there is no actual dbref check anywhere, this if/else is *only* checking if candidates were set and going to a global search if so. the dbref search itself is handled by the builder-permission fallback and the base search method.

Issue #2166, opened 2020-07-15
[BUG]When using Hangul, the columns in the table do not match.


aceamro:
#### Describe the bug
You suffer a lot.
If EvTable uses Hangul, etc., the columns do not match.
So I am temporarily using https://github.com/Robpol86/terminaltables, can it be improved?

who
Accounts:
+--------------+--------+------+-----------+-------+------+----------+--------------+
| Account Name | On for | Idle | Puppeting | Room  | Cmds | Protocol | Host         |
+~~~~~~~~~~~~~~+~~~~~~~~+~~~~~~+~~~~~~~~~~~+~~~~~~~+~~~~~~+~~~~~~~~~~+~~~~~~~~~~~~~~+
| amro(#1)     | 01:18  | 4s   | 암로(#1)    | 광장    | 9    | telnet   | 172.30.1.30  |
| angel        | 16:58  | 11s  | 천사        | 19_57 | 5980 | telnet   | 172.30.1.254 |
| devil        | 16:57  | 11s  | 악마        | 19_57 | 6664 | telnet   | 172.30.1.254 |
| lucifer      | 16:57  | 11s  | 루시퍼       | 19_57 | 6657 | telnet   | 172.30.1.254 |
| michael      | 16:58  | 11s  | 미카엘       | 19_57 | 5612 | telnet   | 172.30.1.254 |
| neo          | 01:39  | 0s   | 네오        | 19_57 | 6791 | telnet   | 172.30.1.254 |
| oracle       | 16:58  | 11s  | 오라클       | 19_57 | 5898 | telnet   | 172.30.1.254 |
| saby         | 01:41  | 2s   | 사비        | 7_46  | 6214 | telnet   | 172.30.1.254 |
| tintin       | 16:58  | 3s   | 틴틴        | 7_46  | 8094 | telnet   | 172.30.1.254 |
| tree         | 16:58  | 3s   | 트리        | 7_46  | 7202 | telnet   | 172.30.1.254 |
| wizard       | 16:58  | 3s   | 마법사       | 7_46  | 7122 | telnet   | 172.30.1.254 |
| zeze         | 16:58  | 3s   | 제제        | 7_46  | 8088 | telnet   | 172.30.1.254 |
+--------------+--------+------+-----------+-------+------+----------+--------------+
12 unique accounts logged in.
tickers
Active tickers:
+--------------+---------------+-----------------+----------+----+
| interval (s) | object        | path/methodname | idstring | db |
[...]

PR #3569, opened 2024-06-28
Update utils.py

hhsiao:
Added use_display_len to m_len, default to False so it won't change behavior for those using it. Justify now properly handle wide character, by using the new use_display_len for m_len.

### DIFF
--- a/evennia/utils/utils.py
+++ b/evennia/utils/utils.py
@@ -290,8 +290,8 @@ def justify(text, width=None, align="l", indent=0, fillchar=" "):
         # absolute mode - just crop or fill to width
         abs_lines = []
         for line in text.split("\n"):
-            nlen = m_len(line)
-            if m_len(line) < width:
+            nlen = m_len(line, True)
+            if m_len(line, True) < width:
                 line += sp * (width - nlen)
             else:
                 line = crop(line, width=width, suffix="")
@@ -306,7 +306,7 @@ def justify(text, width=None, align="l", indent=0, fillchar=" "):
     for ip, paragraph in enumerate(paragraphs):
         if ip > 0:
             words.append(("\n", 0))
-        words.extend((word, m_len(word)) for word in paragraph.split())
+        words.extend((word, m_len(word, True)) for word in paragraph.split())
 
     if not words:
         # Just whitespace!
@@ -2299,7 +2299,7 @@ def calledby(callerdepth=1):
     return "\n".join(out[::-1])
 
 
-def m_len(target):
+def m_len(target, use_display_len=False):
     """
     Provides length checking for strings with MXP patterns, and falls
     back to normal len for other objects.
@@ -2316,8 +2316,8 @@ def m_len(target):
     from evennia.utils.ansi import ANSI_PARSER
 
     if inherits_from(target, str) and "|lt" in target:
-        return len(ANSI_PARSER.strip_mxp(target))
-    return len(target)
+        return display_len(ANSI_PARSER.strip_mxp(target)) if use_display_len else len(ANSI_PARSER.strip_mxp(target))
+    return display_len(target) if use_display_len else len(target)
 
 
 def display_len(target):


Sulryn:
I can confirm that this commit fixes a problem EvTable has with East Asian characters.
Although #2166 solves most problem with Asian characters, it was not enough.
Since EvCell uses the justify() function  to add white spaces needed to correctly to fill the EvTable, which doesn't use display_len().
As a side note, EvMenu also uses the m_len() function  in its formatters, and it should work correctly if the formatters use m_len(str ,True) introduced in this commit. The current formatter doesn't get the correct display width and causes incorrect line breaks when I put Asian characters in the menu options.

hhsiao:
#3573 should be part of this, it handles wrapping of east asian words

evennia/utils/utils.py
```diff
@@ -2299,7 +2299,7 @@ def _stack_display(frame):
     return "\n".join(out[::-1])
 
 
-def m_len(target):
+def m_len(target, use_display_len=False):
```
Griatch, line 2302: If you update the signature with a new kwarg, you should also update the docstring appropropriately: see the `Args:` block below.

Griatch (changes requested):
This looks like a good fix, just need to update the docstring appropriately.

evennia/utils/utils.py
hhsiao: Updated the docstring :)

Griatch (approved):
Thanks, this looks good. Since https://github.com/evennia/evennia/pull/3573 is closely associated with this; I'll wait to merge this PR until that other PR is also ready to merge. :)

aceamro:
Hi everyone,

I've analyzed the issue and believe I've found a solution to correctly calculate the width of strings containing CJK characters (Chinese, Japanese, Korean) in tables. This issue causes uneven column widths when displaying strings with East Asian characters.

The problem originates from the `m_len` function in `evennia/utils/utils.py`, which uses the standard Python `len()` function that doesn't account for the double-width nature of CJK characters.

To fix this, I've made the following changes:

1.  **`get_real_width(text)` Function:**
    *   A new function `get_real_width(text)` was added, which calculates the real display width of a string.
    *  It iterates through the string, assigning a width of `2` to characters within the CJK Unicode range ( `0x2E80` to `0x110000` ) and `1` to all other characters. Newlines and tabs are ignored in the calculations.
    *  It uses Evennia's `raw_ansi` from `evennia.utils.ansi` to strip out ANSI escape sequences before calculating the width, allowing for correct calculations even when colored strings are being used.

2.  **Modified `m_len(target)` Function:**
    *  The `m_len(target)` function was updated to use the `get_real_width` function, ensuring that strings with CJK characters are measured using the correct width calculation.
     *   The modified version is as follows:
        ```python
        def m_len(target):
            """
            Provides length checking for strings with MXP patterns, and falls
            back to normal len for other objects.

            Args:
                target (str): A string with potential MXP components
                    to search.

            Returns:
                length (int): The length of `target`, ignoring MXP components.

            """
            # Would create circular import if in module root.
            from evennia.utils.ansi import ANSI_PARSER

            if inherits_from(target, str) and "|lt" in target:
                return get_real_width(ANSI_PARSER.strip_mxp(target))
            return get_real_width(target)
        ```
       

**How to Apply the Fix:**

1.  Replace the existing `m_len` function and add the `get_real_width` function in your `evennia/utils/utils.py` file with the code provided above.
2.  Restart the Evennia server to apply the changes.

**Explanation:**

*   This approach ensures that the width calculation used by `EvTable` and other related functions in Evennia correctly handles CJK characters, preventing layout issues and improving the display of text.
*   The `get_real_width` function is designed to accurately calculate the visual length of strings that might contain CJK characters.

By implementing this fix, Evennia can be better adapted for use in multi-lingual environments and have table layout display more predictably.

I have tested this solution and it seems to resolve the issue effectively. Please let me know if you have any feedback or encounter any issues.

Thank you

Before
스크린샷 2025-01-13 07 32 57

After
스크린샷 2025-01-13 07 32 45

Griatch:
@aceamro This original PR has been stale for a while; I would suggest you making a separate PR with your suggested changes that can be merged separately.

Griatch:
@hhsiao, @aceamro Ping.

Griatch:
Since this PR went stale, I have implemented this fix in https://github.com/evennia/evennia/commit/7b693d6577a420016ba791e164a12ff6124ca37d instead. I didn't go with neither this PR's solution, nor @aceamro's in the end, going for a simpler solution.

PR #3571, opened 2024-06-29
Group search multimatches by displayed name

### LINKED ISSUE #3570: [BUG] Multimatching message adds unneeded numbers to non-identical partial-match results
#### Describe the bug
When a search matches multiple objects as partial matches, rather than exact matches, the multimatch messaging system adds a disambiguator index to each item based on its position in the entire list of results. This results in it giving disambiguated searches that won't actually return anything.

#### To Reproduce
1. Create one object named "red box" and two objects named "red block"
2. `py self.search("red")` will return the following:
```
More than one match for 'red' (please narrow target):
 red box-1
 red block-2
 red block-3
 ```
3. `py self.search("red block-3")` will fail to find anything, because there is no third "red block" object

#### Expected behavior
Ideally, the output for partial-match multimatches would look like this:
```
More than one match for 'red' (please narrow target):
 red box
 red block-1
 red block-2
```

Evennia main aeb0cf6854
Debian linux

### PR DESCRIPTION
**Pull request by InspectorCaracal**
Modifies the multimatch output to iterate disambiguators only for results with matching names.

It's significantly more complex to make it remove the disambiguator entirely if only one match has a given display name, like the example in the related issue mentioned below, so the output instead looks like this:

```
>>> self.search('red')
More than one match for 'red' (please narrow target):
 red box-1
 red block-1
 red block-2
>>> self.search('red box-1')
red box
```

Bug fixing

Resolves #3570

### DIFF
--- a/evennia/utils/tests/test_utils.py
+++ b/evennia/utils/tests/test_utils.py
@@ -812,6 +812,65 @@ class TestJustify(TestCase):
         self.assertIn(ANSI_RED, str(result))
 
 
+class TestAtSearchResult(TestCase):
+    """
+    Test the utils.at_search_result function.
+
+    """
+
+    class MockObject:
+        def __init__(self, key):
+            self.key = key
+            self.aliases = ''
+
+        def get_display_name(self, looker, **kwargs):
+            return self.key
+        
+        def get_extra_info(self, looker, **kwargs):
+            return ''
+
+        def __repr__(self):
+            return f"MockObject({self.key})"
+
+    def test_single_match(self):
+        """if there is only one match, it should return the matched object"""
+        obj1 = self.MockObject("obj1")
+        caller = mock.MagicMock()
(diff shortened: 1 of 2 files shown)


Griatch (approved):
Hm, this is an interesting one. Looks like a good fix, but it would be good to also add some unit tests here to actually showcase this behavior and make sure these grouped results actually work to disambiguate the results correctly.

Griatch (changes requested):
Hm, this is an interesting one. Looks like a good fix, but it would be good to also add some unit tests here to actually showcase this behavior and make sure these grouped results actually work to disambiguate the results correctly.

InspectorCaracal:
Good call, I'll add that!

Griatch (approved):
I wonder if depending on the `get_display_name` return to group things can have confusing side effects for games with heavily modified or dynamic `get_display_name` methods. :thinking: But the current behavior is already confusing, so this is still an improvement, thanks!

PR #3580, opened 2024-07-06
fix(find/loc): show dbref of target location instead of target

erratic-pattern:
There is a minor bug in the `find/loc` command that displays the result's dbref instead of the result's location's dbref. This PR should fix it to display the correct dbref


Griatch (approved):
Nice catch :+1:

PR #3582, opened 2024-07-07
add true color parsing to ANSIString

0xDEADFED5:
test = evennia.utils.ansi.ANSIString('|#ff0000TEST')
caller.msg(test)

before this PR:

client receives: "\u001b[38;5;196mTEST\r\n"

after this PR:

client receives: "\u001b[38;2;255;0;0mTEST\r\n"

### DIFF
--- a/evennia/utils/ansi.py
+++ b/evennia/utils/ansi.py
@@ -822,7 +822,7 @@ class ANSIString(str, metaclass=ANSIMeta):
         if not decoded:
             # Completely new ANSI String
             clean_string = parser.parse_ansi(string, strip_ansi=True, mxp=MXP_ENABLED)
-            string = parser.parse_ansi(string, xterm256=True, mxp=MXP_ENABLED)
+            string = parser.parse_ansi(string, xterm256=True, mxp=MXP_ENABLED, truecolor=True)
         elif clean_string is not None:
             # We have an explicit clean string.
             pass
--- a/evennia/utils/hex_colors.py
+++ b/evennia/utils/hex_colors.py
@@ -121,27 +121,22 @@ class HexColors:
 
         r, g, b = self._hex_to_rgb_24_bit(tag)
 
-        # Is it greyscale?
-        if r == g and g == b:
-            return f"{indicator}=" + self._GREYS[self._grey_int(r)]
+        if not truecolor:
+            # Fallback to xterm256 syntax
+            r, g, b = self._rgb_24_bit_to_256(r, g, b)
+            return f"{indicator}{r}{g}{b}"
 
         else:
-            if not truecolor:
-                # Fallback to xterm256 syntax
-                r, g, b = self._rgb_24_bit_to_256(r, g, b)
-                return f"{indicator}{r}{g}{b}"
+            xtag = f"\033["
+            if "[" in indicator:
+                # Background Color
+                xtag += "4"
 
             else:
-                xtag = f"\033["
-                if "[" in indicator:
-                    # Background Color
-                    xtag += "4"
+                xtag += "3"
 
-                else:
-                    xtag += "3"
-
-                xtag += f"8;2;{r};{g};{b}m"
-                return xtag
+            xtag += f"8;2;{r};{g};{b}m"
+            return xtag
 
     def xterm_truecolor_to_html_style(self, fg="", bg="") -> str:
         """
--- a/evennia/utils/tests/test_truecolor.py
+++ b/evennia/utils/tests/test_truecolor.py
@@ -15,57 +15,78 @@ class TestANSIStringHex(TestCase):
     def setUp(self):
         self.str = "test "
         self.output1 = "\x1b[38;5;16mtest \x1b[0m"
+        self.output1_truecolor = "\x1b[38;2;0;0;0mtest \x1b[0m"
         self.output2 = "\x1b[48;5;16mtest \x1b[0m"
+        self.output2_truecolor = "\x1b[48;2;0;0;0mtest \x1b[0m"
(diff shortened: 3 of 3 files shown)


Griatch (changes requested):
This seems good, but it really needs to have an updated unit test to go with it.

0xDEADFED5:
oh wow, I can't believe I didn't run the tests first, sorry!

0xDEADFED5:
When I first started this PR (without thinking it through for more than 5 minutes) I assumed that I
was correcting a minor oversight.  But it turns out that I was actually disagreeing with the design
of the true color parser.  I don't want to step on any toes, but I think it could be changed a little.

Right now ANSIString silently downgrades hex color tags to 256 color by default.

I think the expected behavior is that a hex color ANSIString is parsed into true color ANSI by default.
But that's just my perspective as someone who tried to use long color hex codes and was surprised
to be seeing 256 color ANSI codes in my client.

I can see by the original tests that truecolor -> 256 by default is intended, and I think the
truecolor -> 256 is a really cool feature, I just don't think it should be the default. I think the
(truecolor or xterm256) logic should be on the protocol side.

This brings me to the greyscale handling for hex colors.  I don't think it's necessary (or expected).
Right now long hex tags and short hex tags that are greyscale are are automatically changed into the
correct xterm256 greyscale ANSI color (colors 232-255).  It's cool, and the ANSI codes are shorter, but there are
already xterm256 greyscale colors.  I think the special true color greyscale handling should just be removed.

I think the one downside is that people couldn't expect to work with true color ANSIStrings and have them work
later on xterm256, because they've already been parsed.  I think that's fair though.  Right now the way to support
true color and xterm256 at the same time is regular strings that are sent to ANSIParser with the correct 'xterm256'
and 'truecolor' options.  That would remain true after my proposed changes.

@michaelfaith84 What do you think?

Should I finish up my idea, or am I totally wrong?

michaelfaith84:
> I think the expected behavior is that a hex color ANSIString is parsed into true color ANSI by default.
> But that's just my perspective as someone who tried to use long color hex codes and was surprised
> to be seeing 256 color ANSI codes in my client.
...
> (truecolor or xterm256) logic should be on the protocol side.

I agree on this point. I was being overly cautious because truecolor support is hit or miss and client reporting of its support is bad. 

> This brings me to the greyscale handling for hex colors.  I don't think it's necessary (or expected).
> Right now long hex tags and short hex tags that are greyscale are are automatically changed into the
> correct xterm256 greyscale ANSI color (colors 232-255).  It's cool, and the ANSI codes are shorter, but there are
> already xterm256 greyscale colors.  I think the special true color greyscale handling should just be removed.

I'm confused here. But I think you want to remove greyscale from the truecolor code? I felt like the nice thing about greyscale was that it handled the foreground and background at the same time. It isn't strictly necessary, I suppose, because you could just use the xterm256 syntax to accomplish that... But it kinda seems off to be mixing up the color notations from a world building standpoint--at least to me. 

Thanks for reaching out! I trust you all to decide the best course of action for Evennia 😊

0xDEADFED5:
I misspoke a bit with my first version, I edited it a bit to clarify.  The simple version:  greyscale should only matter for true color to 256 conversion

michaelfaith84:
If you feel that's best. I like the behavior, personally, because it handles both foreground and background colors ensuring it's always readable (at least from what I recall). An alternative would be to duplicate the functionality using hex codes. 

If I'm remembering correctly, my inspiration was from the way Evennia handles xterm256 colors that overlap with xterm--it steps them back. I want to say I did the same thing with the hex codes but I could be mistaken.

michaelfaith84:
All of my comments come with the caveat of: I don't run a MUD or even play them anymore. I'm just nostalgic for them and enjoy contributing to Evennia 😊

0xDEADFED5:
Okay, thanks for taking the time to respond. I'm still forming my opinion, that's why I wanted to hear from you =)

0xDEADFED5:
Okay, thanks for taking the time to respond, I appreciate it. I'm still forming my opinion, that's why I wanted to hear from you =)

0xDEADFED5:
> The basic principle for downgrading is that most legacy clients don't support truecolor and won't tell you about it. As mentioned, the downgrading from xterm -> ansi is for this same reason. So if a developer uses hex codes they want to make sure that what comes out is at least colors (even if downgraded) and not some (to the user) random ansi code that messes up their display.
> 
> So if the client is not sending TRUECOLOR=True, I think it's wise to assume downgrading is necessary.
> 
> For greyscale conversion, using xterm is fine, as long as that in turn that downgrade (if xterm256 is also not supported)

Right, I understand that.  But should the downgrading happen by default in ANSIString or should it happen in protocol handling?  I think protocol.  Right now ANSIString isn't usable for true color tags and I think by default it should be usable and the coder should be trusted to know that they're requesting 24-bit ANSI codes.  The 16 color and 256 color tags are already there for the sane people =)

michaelfaith84:
@0xDEADFED5 I feel you are spot on that it should happen in protocol negotiation. Sorry I didn't square that away in the first place. Making the one change in your PR should accomplish the goal. If you check the negotiation in evennia/server/portal/ttype.py, it defaults to false (in the same way xterm256 does).  Rewriting the tests as well as testing manually (both in the console and doing an install) should confirm : )

Griatch:
The Protocol negotiation is indeed the place to handle this. I don't think the suggested code change is a problem - as seen we assume xterm256 is true there too despite not all clients supporting those either. I didn't block the merge of this because of the actual change, just the fact that it breaks unit tests. :grin:

Griatch (approved):
Thanks for the fixes and extended unit tests!

PR #3588, opened 2024-07-18
Add post-creation and spawn hooks to DefaultObject

InspectorCaracal:
This adds two additional hooks to `DefaultObject` and integrates them into the appropriate locations.

1. `at_object_post_creation` - An additional hook which is added to `at_first_save` *after* the attributes etc. from `create_object` or prototypes have been applied. Named to match the pattern of `at_object_post_copy`
2. `at_object_spawn` - A specific hook intended for additional post-spawning processing of updates applied by a prototype. It is called both when the object is initially spawned, and when the object is updated to match any changes to the prototype.

This should fulfill the feature requests and use-case requirements presented in #3095

#### Additional notes
I played it safe when integrating the new `at_object_spawn` hook into the spawner code and used `getattr`, just in case someone decides to use the prototype spawner in unexpected ways like spawning accounts or something.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -2052,8 +2052,8 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
 
     def at_object_spawn(self):
         """
-        Called once when this object is first spawned or updated from a prototype, after all the
-        creation hooks have been run and the object has been saved.
+        Called when this object is spawned or updated from a prototype, after all other
+        hooks have been run.
         """
         pass
 
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -310,7 +310,7 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
                             False, deletion is aborted. Note that all objects
                             inside a deleted object are automatically moved
                             to their <home>, they don't need to be removed here.
-     at_object_spawn()    - called when object is spawned from a prototype or updated
+     at_object_post_spawn() - called when object is spawned from a prototype or updated
                             by the spawner to apply prototype changes.
      at_init()            - called whenever typeclass is cached from memory,
                             at least once every server restart/reload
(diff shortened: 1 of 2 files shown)


Griatch (changes requested):
This is great, thanks for adding this, with the recent discussion in chat, this will be useful! 

- I think the hook should rather be `at_object_post_spawn` to be consistent; this is something that is explicitly happening the last thing in the spawn sequence
- I think it would be useful to have the hook include the prototype as an optional argument (`prototype=prototype`), or even better as `**prototype`. There could be some situations where having info about the prototype could be useful and not all prototype keys (like meta prototype keys) are reflected on the created object.  

> Going beyond the scope of this PR, I was thinking about if it would be useful to have a separate `at_object_spawn(**prototype)` method happening before creation where very industrious people could massage the prototype (maybe apply their own custom FuncParser?) just before it goes into spawning. But not sure if this is overkill, I've not heard anyone actually needing this. :sweat_smile:

InspectorCaracal:
> happening before creation where very industrious people could massage the prototype

Oh, good point! I actually do both custom pre- and post-processing on my prototype-spawned objects, so I can think of use-cases.

My main thought on those suggestions is that I feel that if there's both an `at_object_spawn` and an `at_object_post_spawn`, and one of them is passed a dictionary of the prototype, they should both be receiving it the same way to be consistent. That makes me think it would be better to do `prototype=prototype` rather than `**prototype`

Griatch:
@InspectorCaracal Ping :) Is there any more feedback you need for me?

InspectorCaracal:
Sorry this took so long. :sweat: I decided to opt for the prototype keyword to avoid potential future-update or subclassing conflicts between prototype keys and custom keyword arguments.

Griatch:
@InspectorCaracal Is this ready for re-review (I see you've made changes but you haven't pressed the little spinner up by my name, so I'm not sure)? :sweat_smile: )

InspectorCaracal:
I could have sworn I did, whoops.

Griatch (approved):
Looks good. Reminder to self to update the documentation with these hooks.

PR #3589, opened 2024-07-19
Update utils.py

hhsiao:
Fixed two regex that wasn't properly escaped, so it didn't actually do anything ;)  it's very minor, as it just remove empty spaces for line that only contains space

### DIFF
--- a/evennia/utils/utils.py
+++ b/evennia/utils/utils.py
@@ -473,7 +473,7 @@ def iter_to_str(iterable, sep=",", endsep=", and", addquote=False):
 list_to_string = iter_to_str
 iter_to_string = iter_to_str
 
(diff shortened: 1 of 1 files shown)


InspectorCaracal:
If the current code that this purports to fix was in fact doing nothing, then this assertion would be failing in the unit tests, and it isn't.

https://github.com/evennia/evennia/blob/45a1cb80e181551a293f7ef86ede21c50060d78c/evennia/utils/tests/test_utils.py#L72-L74

Griatch:
I think the original assessment of the regex doing nothing is not correct, but these regexes should also be correctly escaped since this will be a syntax error in Python3.13+. As far as I can see, the unit tests pass with these changes so it should be fine.

PR #3594, opened 2024-08-16
update dependencies a bit

0xDEADFED5:
@owllex recommended lunr and parameterized updates in Discord, and I had just been looking at dependencies because something I used complained about Jinja2 and pydantic.

Seems like pydantic isn't used anymore so I removed it, and seems like jinja2 isn't used, so I removed it.

tests are passing after i uninstalled/updated the stuff, but i'm not sure if anything more is needed


0xDEADFED5:
inflect doesn't seem to require pydantic? i'm definitely conjugatin' away with no pydantic installed and all the funcparser tests are passing. i've tested a few different setups now, nary a pydantic in sight.

InspectorCaracal:
> inflect doesn't seem to require pydantic? i'm definitely conjugatin' away with no pydantic installed and all the funcparser tests are passing. i've tested a few different setups now, nary a pydantic in sight.

I know that it did so I went to check this: pydantic was removed as a dependency in [version 7.2.0](https://github.com/jaraco/inflect/blob/726887aaf72c02dd3ef5e65b7a2822c2f7297930/NEWS.rst#v720)

Not sure what minimum version evennia is requiring right now, but that would be important for this decision.

Griatch (approved):
Hm, I checked again, and updating my existing virtualenv kept inflect at a 7.0 version for some reason, this does require pydantic. But I see that when testing with a completely fresh virtualenv, you get 7.3.1, so fair enough! 

That said, if you install the full evennia (`evennia`, `evennia[extras]` and the doc-building requirements in evennia/docs) you _do_ end up with pydantic in your requirements. But it's a fair point it doesn't need to be in the pyproject.toml file.

Issue #3596, opened 2024-08-16
test_new_task_waiting_input failing intermittently on Windows


0xDEADFED5:
evennia.commands.default.tests.TestCmdTasks.test_new_task_waiting_input is failing about half the time for me on Windows:

```
======================================================================
FAIL: test_new_task_waiting_input (evennia.commands.default.tests.TestCmdTasks.test_new_task_waiting_input)
Test task completing than a new task with the same ID being made while waitinf for input.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\ANON\repos\AI_Grotto\evennia\evennia\commands\default\tests.py", line 553, in test_new_task_waiting_input
    self.assertEqual(text, "Task completed while waiting for input.")
AssertionError: 'cancel request completed.The task function cancel returned: True' != 'Task completed while waiting for input.'
- cancel request completed.The task function cancel returned: True
+ Task completed while waiting for input.
----------------------------------------------------------------------
Ran 1 test in 0.463s
```
This is with a fresh venv on a freshly initialized game.  In Linux it's working fine for me.

0xDEADFED5:
I've looked into this a bit more, and I can't get the test to fail in the debugger.

I can catch this exception with an impressively long trace, but the test still passes:

```
Exception has occurred: _DefGen_Return       (note: full exception trace is shown but execution is paused at: _get_cmdsets)
(<commands.default_cmdsets.AccountCmdSet object at 0x0000023D80F853A0>, [<commands.default_cmdsets.AccountCmdSet object at 0x0000023D80F853A0>])
  File "C:\Users\ANON\repos\AI_Grotto\evenv\Lib\site-packages\twisted\internet\defer.py", line 1904, in returnValue
    raise _DefGen_Return(val)
  File "C:\Users\ANON\repos\AI_Grotto\evennia\evennia\commands\cmdhandler.py", line 411, in _get_cmdsets (Current frame)
    returnValue(obj.get_cmdsets(caller=caller, current=current))
  File "C:\Users\ANON\repos\AI_Grotto\evenv\Lib\site-packages\twisted\internet\defer.py", line 2000, in _inlineCallbacks
    result = context.run(gen.send, result)
  File "C:\Users\ANON\repos\AI_Grotto\evenv\Lib\site-packages\twisted\internet\defer.py", line 2168, in _cancellableInlineCallbacks
    _inlineCallbacks(None, gen, status, _copy_context())
[...]

PR #3597, opened 2024-08-18
make test_new_task_waiting_input more reliable on Windows

0xDEADFED5:
This is a fix for https://github.com/evennia/evennia/issues/3596

I'm not entirely sure why it works, but the test reliably passes for me now.
Is the clock advance creating a possible race condition?

`self.task = self.task_handler.add(self.timedelay, func_test_cmd_tasks)`

If the timedelay arg for self.task_handler.add is anything other than `self.timedelay`, it passes reliably for me.
if it's `self.timedelay` then on Windows it will fail 50% of the time or so.

### DIFF
--- a/evennia/commands/default/tests.py
+++ b/evennia/commands/default/tests.py
@@ -543,7 +543,7 @@ class TestCmdTasks(BaseEvenniaCommandTest):
         self.call(system.CmdTasks(), f"/cancel {self.task.get_id()}")
         self.task_handler.clock.advance(self.timedelay + 1)
         self.assertFalse(self.task.exists())
-        self.task = self.task_handler.add(self.timedelay, func_test_cmd_tasks)
+        self.task = self.task_handler.add(self.timedelay + 1, func_test_cmd_tasks)
         self.assertTrue(self.task.get_id(), 1)
         self.char1.msg = Mock()
         self.char1.execute_cmd("y")


evennia/commands/default/tests.py
```diff
@@ -543,7 +543,7 @@ def test_new_task_waiting_input(self):
         self.call(system.CmdTasks(), f"/cancel {self.task.get_id()}")
         self.task_handler.clock.advance(self.timedelay + 1)
         self.assertFalse(self.task.exists())
-        self.task = self.task_handler.add(self.timedelay, func_test_cmd_tasks)
+        self.task = self.task_handler.add(self.timedelay + 1, func_test_cmd_tasks)
```
Griatch, line 546: This is so arcane that we need a comment here for sure :D 

I think we can also do an explicit check for the os (you can use `if utils.host_os_is('nt')` to set this, since this appears to only be a thing for Windows.

Griatch (changes requested):
A weird fix like this (even if it works) needs a comment. I also think we can do this explicitly only for Windows to highlight the OS differences rather than mess with a functioning Linux test). 

(I also note there is a mysql 3.10 error, not sure what's going on there (the postgres failures can be ignored, it's github actions timing out).

0xDEADFED5:
it looks weird, but if the race condition is happening because of the clock advance, it might be possible for it to fail elsewhere?  i'm not 100% on what's happening under the hood, but i think `self.timedelay + 1` could even be `self.timedelay + 2` just to make sure that the cancel task always completes before the next task.

i think i just happened to hit the hardware/configuration lottery to reliably notice this

Griatch:
Re-running the mysql test just to be sure. And this still needs a comment before it merges.

Griatch (approved):
I'll merge this and add a comment myself.

0xDEADFED5:
sorry, i got really busy and totally forgot about this =/

PR #3599, opened 2024-08-19
Check msg from at_pre_cmd

blongden:
It's possible to display a msg from at_pre_cmd but it's not possible to test it. Rather than return when at_pre_cmd returns True, the handler should still run the msg checks is they're present.

Tests wont currently pick up any msg displayed in at_pre_cmd.

### DIFF
--- a/evennia/utils/test_resources.py
+++ b/evennia/utils/test_resources.py
@@ -433,7 +433,6 @@ class EvenniaCommandTestMixin:
         cmdobj.raw_string = raw_string if raw_string is not None else cmdobj.key + " " + input_args
         cmdobj.obj = obj or (caller if caller else self.char1)
         inputs = inputs or []
-
         # set up receivers
         receiver_mapping = {}
         if isinstance(msg, dict):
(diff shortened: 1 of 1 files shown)


InspectorCaracal:
I think this as it currently stands would cause unexpectedly different results when using the "additional inputs" feature of the test, since you have that being processed still even on initial command failure.

Griatch (changes requested):
~Thanks for working on this, but this seems to be the wrong approach to support it. The unit test suite must simulate the proper operation best it can. And if `at_pre_cmd` returns anything but False/None, the command _should_ abort immediately, that's in the spec. Ignoring this and continuing to `parse`/`func`/`at_post_cmd` means tests will not be representive of the real behavior.~ 

~If wanting to catch msg() calls in `at_pre_cmd`, one should output the mock's stored data early. Maybe it's worth breaking out the mock reporting section into a sub-function to make it easy to call from different parts of the test call execution.~~

Edit: Never mind; there is a `not` there which I somehow overlooked. So it will correctly bypass the parse/func test if `at_pre_cmd` returns a non-falsy value.

blongden:
It's actually not doing that - parse / func / at_post_cmd is only run if at_pre_cmd is falsey. I do need to check the additional inputs thing though. I'm not familiar with that functionality so I'm not sure how to test it (yet).

Rather than return early if at_pre_cmd is falsey I'm letting it run through and check the msg mock anyway (without running the rest of the cmd).

Griatch:
> It's actually not doing that - parse / func / at_post_cmd is only run if at_pre_cmd is falsey. 

This is what I said, but I see I did misinterpret your code nevertheless; there is a `not` in front of your `at_pre_cmd()` check I somehow missed. So ignore me on that :) The additional inputs is worth to look into though yes.

Griatch:
Ping.

aMiss-aWry:
@blongden I think you can cover a wider segment in `if not cmdobj.at_pre_cmd():` - i.e. all of this https://github.com/evennia/evennia/blob/fd05fe4c027695a618e23bf2bfb31149fb944479/evennia/utils/test_resources.py#L461-L496

Griatch:
@blongden Ping :)

Griatch:
@blongden Ping.

blongden:
I'm not sure why it took so long to get back to this (sorry!) but I've moved the execute_cmd into the check, and added a test in to check it works. Test may be in the wrong place though so let me know if you'd like me to move it and I promise I'll do it before another year goes by... 😁

Griatch (approved):
This looks good now, I think. Thanks for the unit test add as well, to confirm. I think it's in a fine location.

PR #3605, opened 2024-08-30
Pass node kwargs through `@list_node` pages

### LINKED ISSUE #3604: [BUG] Navigating `@list_node` pages loses original kwargs
#### Describe the bug
When navigating an EvMenu node through pages of options generated by the `@list_node` decorator, the original kwargs passed into the node are lost.

Choosing an option on the first page - *without* having navigated to any other pages at all - successfully passes the original kwargs in through the select function to the next node, but navigating to another page of options even once causes the kwargs to be lost.

#### To Reproduce
Here is a minimal EvMenu module to reproduce the issue:
```py
from evennia.utils.evmenu import list_node

def menunode_start(caller, raw_string, **kwargs):
    text = "This node will pass kwargs to the list node."
    options = [{"desc": "Okay", "goto": ("menunode_decorated", {"extra_keyword": True})}]
    return text, options

OPTIONS = ['A', 'B', 'C', 'X', 'Y', 'Z']
def _select(caller, selected, **kwargs):
    # log or output the received kwargs
    print(kwargs)
    return ("menunode_end", kwargs)

@list_node(OPTIONS, _select, pagesize=3)
def menunode_decorated(caller, raw_string, **kwargs):
    text = "This node has paginated options."
    return text, []

def menunode_end(caller, raw_string, **kwargs):
    text = f"Final kwargs: {kwargs}"
    return text, None
```

#### Expected behavior
The menunode should consistently pass the original kwargs through the select function, regardless of whether you've navigated away from the first page or not.

Evennia 4.3.0

InspectorCaracal:
Adds the `kwargs` dict to the generated goto options for option-page navigation.

Bug fixing

Closes #3604

### DIFF
--- a/evennia/utils/evmenu.py
+++ b/evennia/utils/evmenu.py
@@ -1414,21 +1414,21 @@ def list_node(option_generator, select=None, pagesize=10):
                     {
                         "key": (_("|Wcurrent|n"), "c"),
                         "desc": "|W({}/{})|n".format(page_index + 1, npages),
-                        "goto": (lambda caller: None, {"optionpage_index": page_index}),
+                        "goto": (lambda caller: None, kwargs | {"optionpage_index": page_index}),
                     }
                 )
                 if page_index > 0:
                     options.append(
                         {
                             "key": (_("|wp|Wrevious page|n"), "p"),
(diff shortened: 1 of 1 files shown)


Griatch (approved):
Nice. Thanks for addressing!

Issue #3622, opened 2024-09-23
[BUG] "@examine" an object with an .ndb of None produces no output and logs an error


aMiss-aWry:
#### Describe the bug
If an object has an .ndb attribute with a value of None, '@examine' breaks.

#### To Reproduce
1. `py self.ndb.test=None`
2. `@examine self`
3. See error

#### Expected behavior
.ndb attributes should be able to be set to None and not have any issues.

#### Environment, Evennia version, OS etc
         Evennia version: 4.3.0 (rev f2a5c5a85a)
         OS: posix
         Python: 3.11.9
         Twisted: 23.10.0
         Django: 4.2.13

#### Additional context
Server log:

```
2024-09-23 14:56:47 [!!] Unhandled error in Deferred:
2024-09-23 14:56:47 [!!] 
	Traceback (most recent call last):
	  File "/Users/redacted/evennia/commands/cmdhandler.py", line 629, in _run_command
	    ret = cmd.func()
	  File "/Users/redacted/evennia/commands/default/building.py", line 3172, in func
	    ).addCallback(_get_cmdset_callback)
	  File "/Users/I561397/Desktop/muddev/evenv/lib/python3.11/site-packages/twisted/internet/defer.py", line 620, in addCallback
	    return self.addCallbacks(callback, callbackArgs=args, callbackKeywords=kwargs)
	  File "/Users/I561397/Desktop/muddev/evenv/lib/python3.11/site-packages/twisted/internet/defer.py", line 533, in addCallbacks
	    self._runCallbacks()
	--- <exception caught here> ---
	  File "/Users/I561397/Desktop/muddev/evenv/lib/python3.11/site-packages/twisted/internet/defer.py", line 1075, in _runCallbacks
	    current.result = callback(  # type: ignore[misc]
	  File "/Users/redacted/evennia/commands/default/building.py", line 3160, in _get_cmdset_callback
	    self.msg(self.format_output(obj, current_cmdset).strip())
	  File "/Users/redacted/evennia/commands/default/building.py", line 3005, in format_output
	    objdata = self.get_formatted_obj_data(obj, current_cmdset)
	  File "/Users/redacted/evennia/commands/default/building.py", line 2989, in get_formatted_obj_data
	    objdata["Non-Persistent Attributes"] = self.format_nattributes(obj)
	  File "/Users/redacted/evennia/commands/default/building.py", line 2882, in format_nattributes
	    sorted(self.format_single_attribute(attr) for attr in ndb_attr)
[...]

Issue #3602, opened 2024-08-29
[BUG] Using `@examine` on a character breaks the active session


InspectorCaracal:
#### Describe the bug
When examining a character that has been active in the current server lifetime, the command will traceback in the logs and freeze the entire session with no error feedback or ability to do anything else. Disconnecting and reconnecting to force a new session will resolve the issue. Reloading the server to provide a new server-side session will also resolve the issue.

#### To Reproduce
1. Create a fresh game.
2. Do literally any command.
3. `ex me`

#### Environment, Evennia version, OS etc
Python 3.11, Debian, evennia main 11f0e8b01a

#### Additional context
Full traceback:
```
        Traceback (most recent call last):
          File "./evennia/evennia/commands/cmdhandler.py", line 629, in _run_command
            ret = cmd.func()
          File "./evennia/evennia/commands/default/building.py", line 3172, in func
            ).addCallback(_get_cmdset_callback)
          File "./devenv/lib/python3.11/site-packages/twisted/internet/defer.py", line 620, in addCallback
            return self.addCallbacks(callback, callbackArgs=args, callbackKeywords=kwargs)
          File "./devenv/lib/python3.11/site-packages/twisted/internet/defer.py", line 533, in addCallbacks
            self._runCallbacks()
        --- <exception caught here> ---
          File "./devenv/lib/python3.11/site-packages/twisted/internet/defer.py", line 1075, in _runCallbacks
            current.result = callback(  # type: ignore[misc]
          File "./evennia/evennia/commands/default/building.py", line 3160, in _get_cmdset_callback
            self.msg(self.format_output(obj, current_cmdset).strip())
          File "./evennia/evennia/commands/default/building.py", line 3005, in format_output
            objdata = self.get_formatted_obj_data(obj, current_cmdset)
          File "./evennia/evennia/commands/default/building.py", line 2989, in get_formatted_obj_data
            objdata["Non-Persistent Attributes"] = self.format_nattributes(obj)
          File "./evennia/evennia/commands/default/building.py", line 2882, in format_nattributes
            sorted(self.format_single_attribute(attr) for attr in ndb_attr)
          File "./evennia/evennia/commands/default/building.py", line 2882, in <genexpr>
            sorted(self.format_single_attribute(attr) for attr in ndb_attr)
          File "./evennia/evennia/commands/default/building.py", line 2853, in format_single_attribute
            if value is None and attr.strvalue is not None:
[...]

PR #3623, opened 2024-09-23
Fix for #3622 - check if attr.strvalue exists

aMiss-aWry:
Check if attr.strvalue exists before attempting to use it.

When an .ndb.item is set to None, it has no attribute db_strvalue. Later when examining an object we attempt to access its attributes' db_strvalue. InMemoryAttributes only have one if the value is not None.

Alternatively, we may want to set any default db_strvalue (i.e. in InMemoryAttribute's init function), but updating the check accomplishes the same result.

Bug fix

Fix for https://github.com/evennia/evennia/issues/3622 and https://github.com/evennia/evennia/issues/3602

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -2831,7 +2831,7 @@ class CmdExamine(ObjManipCommand):
 
         key, category, value = attr.db_key, attr.db_category, attr.value
         valuetype = ""
-        if value is None and attr.strvalue is not None:
+        if value is None and getattr(attr, "strvalue", None) is not None:
             value = attr.strvalue
             valuetype = " |B[strvalue]|n"
         typ = self._get_attribute_value_type(value)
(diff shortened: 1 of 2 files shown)


count-infinity:
I just ran into this and came here to see if already reported.  
The original error is in line 65 in `attributes.py`

```
....
File "./evennia/evennia/typeclasses/attributes.py", line 65, in <lambda>
            strvalue = property(lambda self: self.db_strvalue)
```
Changing building.py I think fixes the issue for @examine command but I think there's potential for other entrypoints to do the same. 

Changing line 65 to 
`strvalue = property(lambda self: getattr(self,'db_strvalue',None))`

 In the typeclasses/attributes.py file fixed this issue for me, and I think prevents alternate paths from running into the same issue..thoughts?

aMiss-aWry:
> I just ran into this and came here to see if already reported. The original error originates in line 65 in `attributes.py`
> 
> ```
> ....
> File "./evennia/evennia/typeclasses/attributes.py", line 65, in <lambda>
>             strvalue = property(lambda self: self.db_strvalue)
> ```
> 
> Changing building.py I think fixes the issue for @examine command but I think there's potential for other entrypoints to do the same.
> 
> Changing line 65 in the typeclasses/attributes.py file to `strvalue = property(lambda self: getattr(self,'db_strvalue',None))`
> 
> fixed this issue for me, and I think prevents alternate paths from running into the same issue..thoughts?

Hey, good idea. Tried it myself & looks good. Updated accordingly.

Griatch (approved):
I could have sworn I made this fix along with other changes, but maybe it got lost on my local drive somehow :shrug:  Thanks for fixing!

Griatch:
Was just an issue with the gh action. Merged.

PR #3626, opened 2024-09-29
Fix 'defense_type' typos in beginner tutorial + EvAdventure utils.py

feyrkh:
The beginner tutorial had mixed use of 'defend_type' vs 'defense_type'. The canonical utils.py uses 'defense_type', so I've switched the other uses to that.

utils.py also had a copy/paste error causing the defense_type to be reported as the attack_type in the object description string, this has been fixed.

Working tutorial code.


Griatch (approved):
The unit tests fail, but that looks like an issue with the build system that can't be cause by these changes. Merging.

PR #3632, opened 2024-10-04
Fix permission fallback when creating characters

InspectorCaracal:
At the moment, the fallback permissions for characters reference the entire permissions handler rather than the permissions within the handler. This causes the undesirable behavior of permission strings like `"developer,player"` being added to new characters, rather than the expected separate permissions of `"developer"` and `"player"`

Bug fixing


Griatch (approved):
Ooh, this is an interesting issue.

PR #3633, opened 2024-10-05
Allow simple customization of default object descriptions

### LINKED ISSUE #3617: [Feature Request] Allow simple customization of default object descriptions
#### Is your feature request related to a problem? Please describe.

When creating a new object, the object's `create` method sets a default description (such as "You see nothing special." for Objects and "This is an exit." for exits) if no description is provided to the call. These descriptions are hard-coded into the `create` methods and are not easy to change without copy/pasting the entire method for each object type in order to override it. In addition, these strings are slightly opinionated and should be more easily customizable.

#### Describe the solution you'd like

Define the default description for an object as a class variable on the relevant typeclass, and change object `create` methods to refer to that variable instead, similar to how object appearance templates are now done. In addition, explicitly permit an empty/None description rather than always falling back to the default. This makes customization easy and no one needs to re-implement `create` methods in order to change the default desc.

For example:

```py
class DefaultExit:
    default_description = "This is an exit."

# In DefaultObject.create
  desc = description if description is not None else cls.default_description
```

#### Describe alternatives you've considered
Fallback strings used to be display-only, where if no desc was set, the string was returned when someone looked at it. This allowed one to change the default by overriding the desc method, which could also work here. But I think providing the customization option is a better approach overall and fits with the general direction we've been taking for these kinds of customizations.

count-infinity:
Implements Feature Request #3617 Allow simple customization of default object descriptions

- Default typeclass descriptions (Object, Rooms, Exits, Characters) no longer set a default `db.desc` if not provided in the `create` method.
- typeclass `get_display_desc` methods now refer to a class level `default_description` variable
- CreateCmd No longer sets a default `db.desc`
- "You see nothing special." default description now runs through Django translator like the other typeclass defaults

Closes #3617

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -397,6 +397,9 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
 
     objects = ObjectManager()
 
+    # Used by get_display_desc when self.db.desc is None
+    default_description = _("You see nothing special.")
+
     # populated by `return_appearance`
     appearance_template = """
 {header}
@@ -1464,10 +1467,9 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
             if account:
                 obj.db.creator_id = account.id
 
-            # Set description if there is none, or update it if provided
-            if description or not obj.db.desc:
-                desc = description if description else "You see nothing special."
-                obj.db.desc = desc
+            # Set description if provided
+            if description:
+                obj.db.desc = description
 
         except Exception as e:
             errors.append(f"An error occurred while creating this '{key}' object: {e}")
@@ -1746,7 +1748,7 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
             str: The desc display string.
 
         """
-        return self.db.desc or "You see nothing special."
+        return self.db.desc or self.default_description
 
     def get_display_exits(self, looker, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


evennia/objects/tests.py
```diff
@@ -266,7 +294,7 @@ def test_get_objs_with_key_or_alias(self):
         query = ObjectDB.objects.get_objs_with_key_or_alias("")
         self.assertFalse(query)
         query = ObjectDB.objects.get_objs_with_key_or_alias("", exact=False)
-        self.assertEqual(list(query), list(ObjectDB.objects.all().order_by('id')))
+        self.assertEqual(list(query), list(ObjectDB.objects.all().order_by("id")))
```
count-infinity, line 297: black linter artifact

evennia/objects/objects.py
```diff
@@ -2055,7 +2057,7 @@ def at_object_post_spawn(self, prototype=None):
         """
         Called when this object is spawned or updated from a prototype, after all other
         hooks have been run.
-        
+
```
count-infinity, line 2060: black linter artifact

evennia/objects/objects.py
```diff
@@ -1972,7 +1974,7 @@ def at_first_save(self):
                     self.nattributes.add(key, value)
 
             del self._createdict
-        
+
```
count-infinity, line 1977: black linter

Griatch (changes requested):
Thanks for the PR! This is overall good, but it has side effects that also needs to be addressed; the unit tests fail across the board at the moment; those needs to be fixed. :)

count-infinity:
Fixed by changing expected output on test.

In this case `get_display_desc` used to return "You see nothing special." if there were no description, regardless of object type. 
Since the EvAdventureRoom is a DefaultRoom, it seems reasonable that it should now display the default room description.

Griatch (approved):

PR #3634, opened 2024-10-06
Contrib - Item Storage

aMiss-aWry:

This contrib adds tag-based off-grid item storage.

Players can store, retrieve, and list items stored in a room. Rooms can be marked as storerooms with the `storage` command by builders. Storerooms can have individual per-character storage or shared storage between all characters able to access the room. When storerooms are removed via `storage/delete`, all stored items are moved to the room contents. Stored items are moved off grid and returned to the grid when retrieved (after an at_pre_move check for both actions).

This is a common feature for MUDs that Evennia doesn't already have as a contrib.

### DIFF
--- a/evennia/contrib/game_systems/storage/README.md
+++ b/evennia/contrib/game_systems/storage/README.md
@@ -1,0 +1,37 @@
+# Item Storage
+
+Contribution by helpme (2024)
+
+This module allows certain rooms to be marked as storage locations.
+
+In those rooms, players can `list`, `store`, and `retrieve` items. Storages can be shared or individual.
+
+## Installation
+
+This utility adds the storage-related commands. Import the module into your commands and add it to your command set to make it available.
+
+Specifically, in `mygame/commands/default_cmdsets.py`:
+
+```python
+...
+from evennia.contrib.game_systems.storage import StorageCmdSet   # <---
+
+class CharacterCmdset(default_cmds.Character_CmdSet):
+    ...
+    def at_cmdset_creation(self):
+        ...
+        self.add(StorageCmdSet)  # <---
+
+```
+
+Then `reload` to make the `list`, `retrieve`, `store`, and `storage` commands available.
+
+## Usage
+
+To mark a location as having item storage, use the `storage` command. By default this is a builder-level command. Storage can be shared, which means everyone using the storage can access all items stored there, or individual, which means only the person who stores an item can retrieve it. See `help storage` for further details.
+
+## Technical info
+
+This is a tag-based system. Rooms set as storage rooms are tagged with an identifier marking them as shared or not. Items stored in those rooms are tagged with the storage room identifier and, if the storage room is not shared, the character identifier, and then they are removed from the grid i.e. their location is set to `None`. Upon retrieval, items are untagged and moved back to character inventories.
+
+When a room is unmarked as storage with the `storage` command, all stored objects are untagged and dropped to the room. You should use the `storage` command to create and remove storages, as otherwise stored objects may become lost.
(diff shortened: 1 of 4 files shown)


evennia/contrib/game_systems/storage/README.md
```diff
@@ -0,0 +1,37 @@
+# Item Storage
```
Griatch, line 1: Nice READMe, has all needed parts :+1:

evennia/contrib/game_systems/storage/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""
+Item storage integration - helpme 2022
```
Griatch, line 2: Should be 2024 I think :)

evennia/contrib/game_systems/storage/storage.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    Shared functionality for storage-related commands
+    """
+
+    def at_pre_cmd(self):
+        """
+        Check if the current location is tagged as a storage location
+        Every stored object is tagged on storage, and untagged on retrieval
+        """
+        success = super().at_pre_cmd()
```
Griatch, line 17: A nitpick: If `at_pre_cmd` returns `False`, the command execution should abort. By default the parent doesn't do anything, but if a dev changes the command base, I think the parent's result should be respected and return/abort already here.

evennia/contrib/game_systems/storage/storage.py
```diff
@@ -0,0 +1,180 @@
...
+
+class StorageCommand(MuxCommand):
+    """
+    Shared functionality for storage-related commands
+    """
+
+    def at_pre_cmd(self):
+        """
+        Check if the current location is tagged as a storage location
+        Every stored object is tagged on storage, and untagged on retrieval
```
Griatch, line 15: Evennia code style: Add a `Returns:` block that explains the return value's meaning.

evennia/contrib/game_systems/storage/storage.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        success = super().at_pre_cmd()
+        self.storage_location_id = self.caller.location.tags.get(category="storage_location")
+        if not self.storage_location_id:
+            self.caller.msg(f"You cannot {self.cmdstring} anything here.")
+            return True
```
Griatch, line 21: Here the parent's result is overridden (see previous comment).

evennia/contrib/game_systems/storage/storage.py
```diff
@@ -0,0 +1,180 @@
...
+
+        self.object_tag = (
+            "shared" if self.storage_location_id.startswith("shared_") else self.caller.pk
```
Griatch, line 24: Optional style comment: Can be a little cleaner to add a module constant like `TAG_NAME` at the top and use that here rather than inputing the string explicitly where it's used.

evennia/contrib/game_systems/storage/tests.py
```diff
@@ -0,0 +1,122 @@
+from evennia.commands.default.tests import BaseEvenniaCommandTest
+from evennia.utils.create import create_object
+
+from . import storage
+
+
+class TestStorage(BaseEvenniaCommandTest):
+    def setUp(self):
+        super().setUp()
+        self.char1.location = self.room1
```
Griatch, line 10: Nitpick: I don't think this is needed; `char1` is already located in in `room1` from the onset afaik.

Griatch (changes requested):
Cool contrib! I have some rather minor inline comments, but not much. :)

evennia/contrib/game_systems/storage/storage.py
aMiss-aWry: I think this is actually the opposite (https://www.evennia.com/docs/latest/_modules/evennia/commands/command.html#Command.at_pre_cmd) but point stands anyway, changed to return True if the super() returns True before checking further.

Griatch (approved):

Issue #3225, opened 2023-07-14
[BUG] Portal memory leak


Griatch:
#### Describe the bug

User _jumpscare_ (unclear what their github handle is) reported their game's Portal leaks memory over time: 
![c90f5aa3878a42998961c73cb8883857](https://github.com/evennia/evennia/assets/294267/4ce2e024-1de8-4485-9d79-16f7166cd834)

The server has 4GB memory, so this represents ~500MB rise over 14 days. 

Here's after a Server reload (Portal remained up):

![cffb988165f0e76a3fa700ab3a9cc3f0](https://github.com/evennia/evennia/assets/294267/0d75689d-91a1-4c97-b817-191028a968a1)

This behavior is not seen with an idle server, so it seems to have somethig to do with the session store. The server is confirmed to run with `DEBUG=False`. 

#### To Reproduce

Unclear how to reproduce. This appears like a memory leak. 

#### Expected behavior

The portal should not leak memory. 

#### Environment, Evennia version, OS etc

Linux, remote machine.

TehomCD:
I've noticed this same behavior over the years. I believe that it was observed with every version of Evennia I've used. For reference, here's what we're currently running until I have time to upgrade everything:
```
    Evennia 0.9.5 (rev 4ba752e) (rev 4ba752e)
    OS: posix
    Python: 3.8.1
    Twisted: 19.10.0
    Django: 2.2.9
```
It definitely seems tied to usage, as I've run a stage server in parallel on the same platform with a prod server, and stage remains fixed in used memory while prod grows over time. As our activity has fallen over the last year or two, resource usage has also slowed accordingly: I was able to slow my cadence on which I had to perform a full shutdown/restart in order to free up memory.

InspectorCaracal:
There was some hypothesizing that it has something specifically to do with creating sessions, so I did a small test to see if that holds water.

I spun up my default test server (which has no scripts or anything running, just the default Limbo, the superuser account and character, and a few random Objects) and let it idle for about half an hour, where it sat at this memory usage the whole time:
![image](https://github.com/evennia/evennia/assets/51038201/02e9ed64-a1aa-466c-8fd7-27f7462c6c27)
(that's about 82MB for the portal and 80MB for the server, if I read it correctly)

Then, after that half an hour, I ran a script to connect/authenticate/quit via telnet 200 times, which took a minute or two. The memory usage then looked like this:
[...]

PR #3635, opened 2024-10-06
fixing memory leaks

a-rodian-jedi:
**Brief overview of PR changes/additions**

- Fixes a memory leak in Portal Telnet connections
- Forces weak references on Telnet negotiation objects
- Stops _nop_keepalive LoopingCall on disconnected sessions

**Motivation for adding to Evennia**
Community complaints about memory leaks

**Other info (issues closed, discussion etc)**
https://github.com/evennia/evennia/issues/3225

This is not everything. There is also a leak in AttributeHandler, I believe, but I will separate that. I think these fixes should come 1 at a time for stability.

### DIFF
--- a/evennia/server/portal/mccp.py
+++ b/evennia/server/portal/mccp.py
@@ -15,6 +15,7 @@ This protocol is implemented by the telnet protocol importing
 mccp_compress and calling it from its write methods.
 """
 
+import weakref
 import zlib
 
 # negotiations for v1 and v2 of the protocol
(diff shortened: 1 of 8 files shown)


Griatch (changes requested):
Thanks for picking this up again (sorry to see the memory analysis tools go, but good to see these leaks fixed). 
This currently fails unit tests though; looks like there is an issue with weakref being used in a place where a dict is expected.

a-rodian-jedi:
Aha, the unit tests need () on the protocol as well. Very easy to address, I can get to it shortly.

a-rodian-jedi:
From my local testing, the unit tests should now be operational. Even though I checked many times and used edit searching and everything... I still managed to miss a couple of `self.protocol` lines that needed `()`. I think I have them all now.

Griatch (approved):
Thanks for finishing this up. Will be interesting to see if there are any other side effects of this down the line. And looking forward to the one for the AttributeHandler :wink:

PR #3639, opened 2024-10-09
Use consistent locale settings for system utilities

### LINKED ISSUE #3638: [BUG] CmdServerLoad produces traceback if system locale uses commas for decimal separator
#### Describe the bug
I was running tests on a machine that renders numbers as (1.234.567,89) and noticed one of the tests for system load fails to convert a float because of this (test_server_load), checked why and it was testing the `server` command, which runs a `ps` to get %mem info.

#### To Reproduce
1. Make the computer spricht Deutsch.
2. `server`
3. See traceback:
```
  File "[]/evennia/evennia/commands/default/system.py", line 872, in func
    pmem = float(
           ^^^^^^
ValueError: could not convert string to float: ' 0,3\n'
```

#### Expected behavior
Temporarily using a default locale in `LC_NUMERIC` in the env would work, I think we need to use `subprocess` then.

```
         Evennia version: 4.4.1 (rev f15835ad14)
         OS: posix
         Python: 3.11.10
         Twisted: 23.10.0
         Django: 4.2.11
```
#### Additional context
Maybe it should also yell at the user for having funny numbers. Just kidding...

aMiss-aWry:
Set the locale to the POSIX default [("C")](https://docs.oracle.com/cd/E19253-01/817-2521/overview-1002/index.html) when running system utility commands like `ps` to ensure standard output.

Use the `server` command without traceback.

Fixes #3638

### DIFF
--- a/evennia/commands/default/system.py
+++ b/evennia/commands/default/system.py
@@ -13,6 +13,7 @@ import traceback
 
 import django
 import evennia
+import subprocess
 import twisted
 from django.conf import settings
 from evennia.accounts.models import AccountDB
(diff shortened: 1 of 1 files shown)


Griatch (approved):
This is a nifty solution to this issue (I can't help but think there may be some more standardized way, but ... it works!).

PR #3644, opened 2024-10-15
Database backup contrib to schedule database backups into /server/backups

aMiss-aWry:
New `backup` command, which starts/manages a `DatabaseBackupScript`, which runs daily by default.

The script saves a copy of the database to the game `server/backups` folder. It uses `shutil.copy` for sqlite3 and `subprocess.run` `pg_dump` for postgresql. This PR only backs the db up to the same filesystem; users should decide what they want to do next (upload to the cloud, download to a hard drive etc).

I needed to write it for myself and could imagine others finding it useful.

Only supports sqlite3 and postgresql at the moment.

#### Screengrabs

Sqlite3:
image

In folder:
```
(evenv)  (master *) backups > pwd
[..]/desktop/muddev/mygame/server/backups

(evenv)  (master *) backups > ls
2024-10-15T21:05:06.db3	2024-10-15T21:05:10.db3
```

PostgreSQL:
image

In folder:
```
(evenv)  (master *) backups > pwd
[...]/desktop/muddev/gardenmud/server/backups
(evenv)  (master *) backups > ls
2024-10-15T21:06:55.sql	2024-10-15T21:07:02.sql
```


evennia/contrib/utils/database_backup/README.md
```diff
@@ -0,0 +1,69 @@
...
+This utility uses the settings.DATABASES dictionary.
+
+## Restoration
+
+Remember to `evennia stop` before restoring your db.
+
+### Restoring sqlite3 (.db3)
+
+* Copy the database backup you want to restore from back into the `server/` directory
+* Rename the database backup to `evennia.db3` if you have not modified `settings.DATABASES`, otherwise whatever name is in your `settings.DATABASES` dictionary.
```
Griatch, line 46: If you don't know that the sqlite3 file is named `evennia.db3`, this instruction becomes a little confusing. I think it would be clearer if this was phrased something like
```
By default (unless you changed the name of the file in `settings DATABASES`), the sqlite3 file `mygame/server/evennia.db3`. Copy your backup file over this file to recover your backup.
```

evennia/contrib/utils/database_backup/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Restoring postgres (.sql)
+
+* Prepare the following variables
+```
+export DB_USER=db_user from your settings.DATABASES
+export DB_NAME=db_name from your settings.DATABASES
+export BACKUP_FILE=the path to the backup file you are restoring from
+export PGPASSWORD=your db password
```
Griatch, line 55: This is hard to understand since it's all in \`\`\`. Is `export DB_USER=db_user from your settings.DATABASES` one command? Is `export DB_USER=db_user` the command you should run? Does 'export' here only mean you should find it and remember it? If you use \`\`\`, place the actual command alone (with comments between each block), or use clear `#`-markers to show what is a comment.

evennia/contrib/utils/database_backup/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+class CharacterCmdset(default_cmds.Character_CmdSet):
+    ...
+    def at_cmdset_creation(self):
+        ...
+        self.add(DbCmdSet)  # <---
+
+```
+
+Then `reload` to make the `backup` command available.
```
Griatch, line 29: It installs the backup command, but talks about a script. It doesn't actually describe how this is to be used - do you use the backup command to do the backup? Does it start a script? What? You shouldn't have to go read the code to find out.

evennia/contrib/utils/database_backup/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+This module schedules backups in-game, which saves a copy of your database to your game's `server/backups` folder. Database backups are *not* automatically uploaded to any cloud service, it is left to you to decide what to do with them (i.e. pushed to git, uploaded to the ether, downloaded to a hard drive).
+
+Backups can take place at any time. Restoring the game world from backup takes place during downtime, as documented below.
+
+Currently, the sqlite3 (the evennia default) and postgresql databases are supported. Others are welcome to add more.
+
+## Installation
+
+This utility adds the `backup` command. Import the module into your commands and add it to your command set to make it available.
```
Griatch, line 13: Since this is based on a global Script, why not also describe how to just add the script as a global script in settings; that way it could be set up without having to connect to the game and run a command ...

evennia/contrib/utils/database_backup/database_backup.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def backup_sqlite3(self, db_name, output_file_path):
+        """
+        Copy the sqlite3 db.
+        """
+
+        output_file_path += ".db3"
+        os.makedirs(os.path.dirname(output_file_path), exist_ok=True)
+        shutil.copy(db_name, output_file_path)
+        self.log(f"|wsqlite3 db backed up in: {BACKUP_FOLDER}|n")
```
Griatch, line 65: How safe is this operation? The sqlite3 file is just a binary and `shutil.copy` doesn't know anything about what a database is. What happens if this is being written to (by an active game) while this copy operation is performed?

Griatch (changes requested):
This is a cool contrib I think people will find useful! Some comments: 

- I made some clarifications suggestions for the README
- It's nice to have a command, but I imagine a lot of people would be happy to just add the global script to settings with some sane defaults without needing to run a command; seems like a good thing to add to readme and to consider as a use case! 
- The contrib's unit tests all fail. Most likely you tested this in isolation, but is not cleaning up the twisted deferreds of the script (I would suggest mocking instead of actually running a timed operation as part of a unit test).

Griatch:
Ping.

Griatch:
Ping :)

evennia/contrib/utils/database_backup/README.md
aMiss-aWry: Fixed wording according to suggestion

evennia/contrib/utils/database_backup/database_backup.py
aMiss-aWry: I think probably the easiest thing to do here is add a `PRAGMA integrity_check` post-op and log a failure if it doesn't work, indicating the backup was corrupted and needs to be retaken. Thoughts?

Griatch:
@aMiss-aWry The unit tests are failing across the board here.

evennia/contrib/utils/database_backup/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""
+Database backups - helpme 2024
+"""
+
+from .database_backup import DbCmdSet  # noqa
```
Griatch, line 5: You are supposed to use `DatabaseBackupScript` directly from this module I think. So it should probably be added here too.

evennia/contrib/utils/database_backup/tests.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+
+    def setUp(self):
+        super().setUp()
+
+    @patch("shutil.copy")
+    @patch("evennia.utils.logger.log_sec")
+    def test_sqlite_success(self, mock_logger, mock_copy):
+        mock_copy.return_value.returncode = 0
+        self.call(
```
Griatch, line 31: This is actually creating a script in the database afaik, and failures we see would be due to leaving Twisted deferred's dangling outside of the test scope. Either mock `create_script` or be careful to clean up in `tearDown`.

evennia/contrib/utils/database_backup/database_backup.py
```diff
@@ -0,0 +1,195 @@
...
+from evennia.utils import logger, search
+
+import datetime
+import os
+import shutil
+import subprocess
+import sqlite3
+
+_MUDINFO_CHANNEL = None
+BACKUP_FOLDER = "server/backups"
```
Griatch, line 16: This works, but it's more os-agnostic to use `os.path.join(settings.GAME_DIR, "server", "backups")` here.

evennia/contrib/utils/database_backup/database_backup.py
```diff
@@ -0,0 +1,195 @@
...
+from evennia import CmdSet, DefaultScript
+from evennia.commands.default.muxcommand import MuxCommand
+from evennia.utils.create import create_script
+from evennia.utils import logger, search
+
+import datetime
+import os
+import shutil
+import subprocess
+import sqlite3
```
Griatch, line 13: Is this actually used anywhere?

Griatch (changes requested):
@aMiss-aWry 
Looking back at this one (doing a lot of PR reviewing today), there are a few things that need to happen for this PR to merge: 

- See inline comments 
- The `postgresql_psycopg2` is `postgresql` these days.

evennia/contrib/utils/database_backup/database_backup.py
aMiss-aWry: It looks like you commented this on the import for os, shutil, subprocess, sqlite3 - yes they're all used in the script. If this was meant to be on the `from evennia.comms.models import ChannelDB` - yes that was unused, removed.

evennia/contrib/utils/database_backup/tests.py
aMiss-aWry: Added to teardown.

evennia/contrib/utils/database_backup/__init__.py
aMiss-aWry: Done

evennia/contrib/utils/database_backup/database_backup.py
aMiss-aWry: Done

evennia/contrib/utils/database_backup/database_backup.py
aMiss-aWry: Did the PRAGMA integrity_check thing.

evennia/contrib/utils/database_backup/database_backup.py
```diff
@@ -0,0 +1,202 @@
+from django.conf import settings
```
Griatch, line 13: As per CODESTYLE, the module needs a docstring describing what it's doing.

evennia/contrib/utils/database_backup/README.md
```diff
@@ -0,0 +1,84 @@
+# Database Backup Scheduler
+
+Contribution by helpme (2024)
+
+This module schedules backups in-game, which saves a copy of your database to your game's `server/backups` folder. Database backups are *not* automatically uploaded to any cloud service, it is left to you to decide what to do with them (i.e. pushed to git, uploaded to the ether, downloaded to a hard drive).
+
```
Griatch, line 6: It should be mentioned that MySQL/MariaSQL databases are not supported here.

evennia/contrib/utils/database_backup/database_backup.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def func(self):
+        """
+        Database backup functionality
+        """
+        caller = self.caller
+        args = self.args.strip()
+        interval = int(args) if args.isnumeric() else DEFAULT_INTERVAL
+
+        script = self.get_script()
```
Griatch, line 187: I think there needs to be a check for which database is configured, and the command aborting if finding a non-supported one (MySQL/MariaDB). Otherwise a user may run this, get a success message and then `at_repeat` will silently skip doing the backup every repeat, which is not great ...

evennia/contrib/utils/database_backup/database_backup.py
```diff
@@ -0,0 +1,202 @@
...
+
+        output_file_path += ".db3"
+        os.makedirs(os.path.dirname(output_file_path), exist_ok=True)
+        shutil.copy(db_name, output_file_path)
+
+        # Check the integrity of the copied db
+        con = sqlite3.connect(output_file_path)
+        cur = con.cursor()
+        try:
+            cur.execute("PRAGMA integrity_check")
```
Griatch: If this fails, the corrupt db3 file is left on disk - and will be counted as "latest backup". This gives a false sense of security. It seems better to delete the corrupted or move it to a `*_corrupted` suffix so the backup is recreated later.

evennia/contrib/utils/database_backup/database_backup.py
```diff
@@ -0,0 +1,202 @@
...
+        self.log(f"|wpostgresql db backed up in: {BACKUP_FOLDER}|n")
+
+    def backup_sqlite3(self, db_name, output_file_path):
+        """
+        Copy the sqlite3 db.
+        """
+
+        output_file_path += ".db3"
+        os.makedirs(os.path.dirname(output_file_path), exist_ok=True)
+        shutil.copy(db_name, output_file_path)
```
Griatch: It may be worth pointing out the risks of this in the README. The Shutil solution is fast but could clash with writes (maybe sqlite library's inbuilt backup command may be worth exploring?)

evennia/contrib/utils/database_backup/database_backup.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def get_latest_backup(self):
+        """
+        Returns:
+            str: Name of the most recent backup
+        """
+        try:
+            files = os.listdir(BACKUP_FOLDER)
+            paths = [os.path.join(BACKUP_FOLDER, basename) for basename in files]
+            last_backup = max(paths, key=os.path.getctime)
```
Griatch: Any file (like a README, tar file etc) becomes a 'backup' to this. While maybe unlikely, it's unexpected behavior, and pretty fragile. It would make sense to filter on `.db3` or `.sql` or whatever is needed.

evennia/contrib/utils/database_backup/database_backup.py
```diff
@@ -0,0 +1,202 @@
...
+            self.create_script(interval)
+            return
+
+        # Change backup script's interval (if the provided interval is different)
+        original_interval = script.interval
+        if args and original_interval != interval:
+            self.create_script(interval)
+            return
+
+        if "force" in self.switches:
```
Griatch: If the script didn't previously exist, the func() will have exited on line 172 before this is evaluated. That feels unexpected to the user.

evennia/contrib/utils/database_backup/database_backup.py
```diff
@@ -0,0 +1,202 @@
...
+            _MUDINFO_CHANNEL.msg(message)
+        logger.log_sec(message)
+
+    def backup_postgres(self, db_name, db_user, output_file_path):
+        """
+        Run `pg_dump` on the postgreSQL database and save the output.
+        """
+
+        output_file_path += ".sql"
+        with open(output_file_path, "w") as output_file:
```
Griatch: This already creates the sql file on disk. What happens if pg_dump fails (bad creds, not installed etc)? You'll end up with an empty/corrupted backup file  (false sense of security).

Griatch (changes requested):
Thanks for the fixes. Looked things over on the functional side / played around, and found a few more things that is behaving strangely, or is not as safe as one would expect a backup to be.

evennia/contrib/utils/database_backup/database_backup.py
aMiss-aWry: Fixed - deleting corrupt backup

evennia/contrib/utils/database_backup/database_backup.py
aMiss-aWry: Fixed - filtering on `.db3`/`.sql`

evennia/contrib/utils/database_backup/database_backup.py
aMiss-aWry: Reordered

evennia/contrib/utils/database_backup/database_backup.py
aMiss-aWry: Added cleanup in catch

evennia/contrib/utils/database_backup/database_backup.py
aMiss-aWry: Added check in at_pre_cmd()

evennia/contrib/utils/database_backup/database_backup.py
aMiss-aWry: Removed shutil - using sqlite's backup

evennia/contrib/utils/database_backup/README.md
aMiss-aWry: The supported dbs are mentioned in the next couple lines

Griatch (approved):
Thanks for addressing the feedback! 
The only caveat I can think of is that setting this up via `global_scripts` (as suggested in README) will bypass the db-check in `at_pre_cmd` (since no cmd is called), and thus unsupported engines will not be detected. But if you set things up like that it's probably a minor concern, so approving.

PR #3647, opened 2024-10-19
Properly reinitialize contents cache

### LINKED ISSUE #3646: [BUG] Weird, inconsistent behavior from contents cache handler
#### Describe the bug
On my live Evennia game, since I updated to 4.4.1 recently, I've been getting periodic very strange behavior from the contents cache where it thinks non-character things are characters in the room.

Here is an example of what happens in an affected room.

Examining the room (with unrelated output cropped)
```
ex here
------------------------------------------------------------------------------
Name/key: Wilderness (#1761)
Typeclass: OverworldRoom (typeclasses.rooms.OverworldRoom)
Exits: north(#1762), northeast(#1763), east(#1764), southeast(#1765), south(#1766), southwest(#1767), west(#1768), northwest(#1769)
Characters: Cal(#103)
Content: patch of wheat(#14578)
------------------------------------------------------------------------------
```
In-game `py` commands and output inspecting the output of various commands and contents cache:
```py
>>> here.contents_get(content_type='character')
[Cal, north, northeast, east, southeast, south, southwest, west, northwest, patch of wheat]
>>> print(','.join([f"{obj.key}#{obj.id}" for obj in here.contents_get(content_type='character')]))
Cal#103,north#1762,northeast#1763,east#1764,southeast#1765,south#1766,southwest#1767,west#1768,northwest#1769,patch of wheat#14578
>>> print(here.contents_cache._typecache)
defaultdict(<class 'dict'>, {'exit': {1762: True, 1763: True, 1764: True, 1765: True, 1766: True, 1767: True, 1768: True, 1769: True}, 'object': {14578: True}, 'character': {14575: True, 103: True}, 'player': {103: True}})
>>> print(here.contents_cache.get(content_type='character'))
[Cal, north, northeast, east, southeast, south, southwest, west, northwest, patch of wheat]
>>> print(here.contents_cache._typecache['character'].keys())
dict_keys([14575, 103])
```

In another case which I don't have a record of to share, it was returning an ordinary object as one of the room's characters that wasn't even *in* the room.

#### To Reproduce
I have not yet been able to intentionally reproduce this behavior. :(

#### Expected behavior
The `contents_cache.get` function should be just returning the objects under that key.

Evennia 4.4.1, Debian 12, Python 3.11, using MariaDB database backend

#### Additional context

So far, this has only happened in Wilderness rooms, but since I don't have reliable reproduction steps I can't confirm that it *wouldn't*. Most of the rooms in this game are part of a wilderness.

InspectorCaracal:
The contents cache handler currently has an oversights when reinitalizing:

1. The typecache is not reset before being initialized in `init`, allowing invalid data to persist and continue erroring.
2. ~~The `load` function includes objects that have been deleted but are still in memory, leading to idcache errors.~~

This PR corrects this.

Bug fixing

Resolves #3646 (or at least the parts of it that weren't my own fault)

### DIFF
--- a/evennia/objects/models.py
+++ b/evennia/objects/models.py
@@ -60,7 +60,7 @@ class ContentsHandler:
         Returns:
             Objects (list of ObjectDB)
         """
-        return list(obj for obj in self.obj.locations_set.all() if obj.pk)
+        return list(self.obj.locations_set.all())
 
     def init(self):
         """
--- a/evennia/objects/models.py
+++ b/evennia/objects/models.py
@@ -60,7 +60,7 @@ class ContentsHandler:
         Returns:
             Objects (list of ObjectDB)
         """
-        return list(self.obj.locations_set.all())
+        return list(obj for obj in self.obj.locations_set.all() if obj.pk)
 
     def init(self):
         """
@@ -68,6 +68,7 @@ class ContentsHandler:
 
         """
         objects = self.load()
+        self._typecache = defaultdict(dict)
         self._pkcache = {obj.pk: True for obj in objects}
         for obj in objects:
             try:


InspectorCaracal:
While lying awake failing to sleep last night, I realized that the change to load is completely unnecessary; it only needs the reinit of the typecache.

Griatch (approved):
This is really one of those bugs one has to _really_ work the system to find! :grin: Nice to get it fixed though!

PR #3648, opened 2024-10-20
adding a rate limiter to Evennia's server library

a-rodian-jedi:
- Adds a token-based rate limiter to Evennia's server module

- Proper rate limiting for those who want a true X/sec rate limit

### DIFF
--- a/evennia/server/rate.py
+++ b/evennia/server/rate.py
@@ -1,0 +1,178 @@
+"""
+Rate limiting module for Evennia.
+
+Usage example:
+    from evennia.server.rate import Limiter, RateLimitException
+
+    # Create a limiter that allows 5 actions per second, with a burst of 10
+    limiter = Limiter(limit=5, burst=10)
+
+    try:
+        # Check if an action is allowed
+        if limiter.ok():
+            # Perform the action
+            perform_action()
+            
+            # Mark the action as completed
+            limiter.mark_last()
+    except RateLimitException:
+        # Handle the case when the rate limit is exceeded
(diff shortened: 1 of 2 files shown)


Griatch (changes requested):
Thanks for the PR! I think this looks more like a contrib than something to go into core though (it's not used at all at the moment, and it more looks like something intended to plug into one's own code). 

So this should be added to `evennia/contrib/utils/` instead, I think. Also note that contribs need to follow the [contrib guidelines](https://www.evennia.com/docs/latest/Contribs/Contribs-Guidelines.html).

Griatch:
Ping.

Griatch:
@a-rodian-jedi Ping!

Griatch:
@a-rodian-jedi Ping.

evennia/server/rate.py
```diff
@@ -0,0 +1,177 @@
...
+
+        Returns:
+            bool: True if allowed, False otherwise.
+        """
+
+        _, tokens = self.advance(t)
+        
+        return tokens >= count
+
+    def ok(self):
```
Griatch, line 131: This both checks and raises and exception. Would something like `acquire()` returning a `bool` may be simpler?

Griatch, line 131: This both checks and raises and exception. Would something like `acquire()` returning a `bool` may be simpler?

Griatch:
@a-rodian-jedi  (whitenoise): Ping. Make this into a contrib and this should be fine to merge (just minor fixes/suggestions mentioned above).

PR #3651, opened 2024-10-23
Fix default justify behaviour in EvEditor

willmofield:

Changed the default align of the `:j` command in EvEditor to "left", to match both the behaviour indicated in EvEditor help and the default behaviour of `utils.justify`.

Align functionality between Python and in-game, and avoid unintended behaviour!


Griatch (approved):
It's indeed a discrepancy against the doc string. Thanks for the fix!

PR #3655, opened 2024-10-26
fix: remove erroneous links from the documentation

marado:
Sphinx assumes that WORD.py should be a link to https://WORD.py .

This is not a problem in most of the documentation, since Sphinx won't turn those python filenames into links when they are wrapped in backticks, like `WORD.py`. Unfortunately, not always that was being done, and so there were several wrong, broken links in the documentation.

This patch wraps all the occurrences I've found of this case with backticks, not only making the documentation more readible and homogeneous, but more importantly getting rid of those unwanted links in the generated HTML version of the documentation.

This just wraps a few python filenames, so they are presented as `file.py` instead of file.py 

Removes broken links on the HTML documentation.

The broken links can currently be seen in places like https://www.evennia.com/docs/latest/Howtos/Web-Help-System-Tutorial.html (search for `views.py`).


Griatch (approved):
This is a cool fix, thanks for your attention to detail!

PR #3657, opened 2024-10-27
Make FileHelpEntry hashable so it does not fail equality checks

a-rodian-jedi:

This hashes FileHelpEntry so it can pass equality checks. A lack of this causes `do_search` in help.py to fail at this line: https://github.com/evennia/evennia/blob/88afac2874160bada78189b82eff3782db1705f7/evennia/commands/default/help.py#L505

This is because removal has to check for equality.

Production bug on Secrets of Aelandris breaking several help files

### DIFF
--- a/evennia/help/filehelp.py
+++ b/evennia/help/filehelp.py
@@ -121,7 +121,7 @@ class FileHelpEntry:
         return f"<FileHelpEntry {self.key}>"
 
     def __hash__(self):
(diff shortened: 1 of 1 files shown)


Griatch (commented):
Looking at it, just the `key` should be enough; all the entries are made unique by the `key` alone.

Griatch (approved):

PR #3662, opened 2024-10-31
add_user_channel_alias was adding nicks into inputline category

hhsiao:
add_user_channel_alias was adding nicks into inputline category, it really should be adding it into "channel" category.


Griatch (changes requested):
Sorry, but this is incorrect and also breaks unit tests. There is nothing to fix here.

The `channel` nick category is for searching aliased channel names. The `inputline` nick category (as correctly used here) allows for users to be able to write `foo Hello!` that expands into `@channel mychannel = Hello!`.

PR #3663, opened 2024-10-31
Input processing should do nickreplace against inputline AND channel

hhsiao:
I believe this is the reason why add_user_channel_alias in evennia/comms/comms.py was adding nicks into "inputline" instead of "channel".


Griatch (changes requested):
No, inputfuncs knows nothing of entity types. It deals with the inputline. There are many other nick replacement types and these are handled in the search methods of each respective entity (such as channels). Closing.

PR #3665, opened 2024-11-03
Fix 'offer', 'hinder', 'alter' verb conjugation

aMiss-aWry:
For some reason these three have lines that do *no* conjugation, which means `$conj(offer)` results in every perspective seeing 'offer'.

Conjugations are only correct when the 'blanks' are removed

'offer' is correctly conjugated in https://github.com/evennia/evennia/blob/main/evennia/utils/verb_conjugation/verbs.txt#L510

'alter' in https://github.com/evennia/evennia/blob/main/evennia/utils/verb_conjugation/verbs.txt#L7677

'hinder' in https://github.com/evennia/evennia/blob/main/evennia/utils/verb_conjugation/verbs.txt#L8066

I assume this was partially a manually curated list and the original creator just missed removing the blanks


Griatch (approved):
Nice find. The list is taken from the XTAG English grammar project, so it's probably a mix of automatic and manually set up lines. :shrug:

PR #3669, opened 2024-11-07
Fix CmdPage tracebacks

InspectorCaracal:
In certain cases, CmdPage can fail with a traceback.

- When attempting to `tell` someone without providing a message
- When attempting to `tell` someone a single-word reply

This fixes the code to prevent those errors.

Bug fixing


Griatch (approved):
Looking this over a few times I think it does produce the same result :grin: Nice refactor as well as fix!

PR #3671, opened 2024-11-11
#3660 bugfix old name saved in alias

### LINKED ISSUE #3660: [BUG] name saved in alias
#### Describe the bug
I would like to change the name of some of my objects. I use the `get_display_name` method for this. My implementation does some checks with `self.tags.has` to choose the right template to render the name of my object. 
But I observed that the old names that are no longer valid are saved as aliases. That is disturbing to me because the object respond to user keywords that no longer describe the object. 

#### To Reproduce
1. `create closed box`
2. `inventory`
3. `name closed box=opened box`
4. See error : `alias opened box`
5. See an other error : `search closed box`

#### Expected behavior
I would like aliases are created only when I manually register them.

```
> about
Evennia MU* development system

Evennia version: 4.4.1
OS: posix
Python: 3.12.7
Twisted: 23.10.0
Django: 4.2.16
```

#### Additional context
Perhaps I should not dynamically change the name of an object. That would be disturbing to me because I need my objects to change state and to make it easily visible.

count-infinity:
Fixes issue where renaming an object keeps the old plural aliases.

- Moves plural_category to object level variable
- Modifies DefaultObject `at_rename` hook to clear plural aliases
 
Fixes issue #3660.

Root cause:
Plural aliases are added in the `DefaultObject.get_numbered_name` method which is called when a character uses `CmdInventory` or `CmdLook`.  If a builder renames an object, but doesn't use a command that calls get_numbered_name, the object can still be referenced by the old aliases.
 
Closes #3660

### DIFF
--- a/evennia/commands/default/tests.py
+++ b/evennia/commands/default/tests.py
@@ -1267,6 +1267,15 @@ class TestBuilding(BaseEvenniaCommandTest):
         )
         self.call(building.CmdName(), "Obj4=", "No names or aliases defined!")
 
+    def test_name_clears_plural(self):
+        box, _ = DefaultObject.create("Opened Box", location=self.char1)
+
+        # Force update of plural aliases (set in get_numbered_name)
+        self.char1.execute_cmd("inventory")
+        self.assertIn("one opened box", box.aliases.get(category=box.plural_category))
+        self.char1.execute_cmd("@name box=closed box")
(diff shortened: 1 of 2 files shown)


Griatch (approved):
This looks like a good fix, making use of an existing hook. Also great with a unit test as well. :+1:

PR #3677, opened 2024-11-15
Avoid certain tracebacks when creating accounts

InspectorCaracal:
Despite using code that assumes username and password are strings, `DefaultAccount.create` currently uses the default fallback value of `None` if they aren't provided.

This updates that to use an empty string instead, and additionally updates `normalize_username` to enforce that the provided username is normalized to a string.

Fixing edge cases


Griatch (approved):
This is a nice catch.

PR #3682, opened 2024-11-22
3564 - Fix leading wildcard help search

### LINKED ISSUE #3564: [BUG] Command help for commands starting with * does not display
#### Describe the bug

Certain commands whose names begin with * are unsearchable via in-game help. It's not obvious what the pattern is for the names that work and the names that don't.

Known broken command names:
*examine
*link
*shortcode

For example, this basic command, intended to override the built-in command of the same name, triggers the bug:

```py
class CmdExamine(building.CmdExamine):
    """
    Override of the default Examine command to change the command key.
    """

    __doc__ = building.CmdExamine.__doc__

    key = "*examine"
    aliases = ["*exam", "*ex", "@examine"]
```

When running `help *examine`, the following output is observed:
```
> help *examine
No help found

There is no help topic matching '*examine'.
... But matches were found within the help texts of the suggestions below.

Other topic suggestions:
*examine
```

After some debugging with @InspectorCaracal , it appears to have been broken sometime around evennia 4.1.0. It worked properly in previous versions, so it looks like a recent regression. Reproducible in an vanilla install.

#### To Reproduce
1. Sync to main:HEAD
2. Add the command above to the default CharacterCmdSet (it should override the built-in command).
3. help *examine

#### Expected behavior
Help works properly for this command.

         Evennia version: 4.1.1
         OS: posix
         Python: 3.10.5
         Twisted: 23.10.0
         Django: 4.2.8

Git commit hash: 3b84ec1b42 (Should be HEAD or close to it)

#### Additional Context
The bug is not reproducible if the command is ONLY named `*examine` and doesn't have the `@examine` alias. I don't know why.
`*` is not in my command ignore prefixes setting.

count-infinity:
- Factor out lunr related code from help_search_with_index function
- customize default lunr stemmer to return stemmed and wildcard matches
- unit tests

Fixes issue #3564  where commands with keys beginning with a '*' didn't return help results.

Closes #3564 

##### Root Cause:
This was related to lunr treating '\*' as a special wildcard search character.  Generally you can search for things like 'ex*ine' and it will return 'examine'.

When lunr indexes a document for searching it will stem words so it can match better, from docs `For example, the stem of “searching”, “searched” and “searchable” should be “search”`.  

`*examine` is stemmed to `*examin`.

Then when you search, lunr usually stems that term too, however it **won't stem wildcard searches**.  So the search is for exactly '*examine' and will be a miss for the indexed `*examin` without the ending 'e'.  (Although since it's wildcard it would match something like `hexamine`)

##### Fix:
There wasn't an easy way I saw to escape a wildcard, so best I could come up with was to create a custom stemmer that indexes exactly as the current one, but also keeps exactly words with leading wildcard, i.e. `*examine`.

### DIFF
--- a/evennia/help/tests.py
+++ b/evennia/help/tests.py
@@ -5,6 +5,7 @@ command test-suite).
 """
 
 from unittest import mock
+from parameterized import parameterized
 
 from evennia.help import filehelp
 from evennia.help import utils as help_utils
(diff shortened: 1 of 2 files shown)


Griatch (approved):
This is an interesting workaround. Thanks for the work. Also great with unit tests to verify. 

(We may want to move some of this stuff out of utils.py in the future ... it's getting pretty crammed in that module :sweat_smile: )

PR #3690, opened 2024-12-03
Avoid erroneous "direct match" targets when searching

InspectorCaracal:
This changes `get_search_direct_match` to cross-check the hardcoded input strings against the actual search candidates rather than completely ignoring them.

Usability and consistency.

While we do want the strings "me" and "here" to always reference the caller and the caller's location, it *doesn't* follow that the caller and the caller's location should always be valid search targets. For example, a command that restricts its search candidates to the caller's contents should not have to account for accidentally matching the caller's *location*, as that room isn't in the given candidates.


Griatch (approved):
This is an interesing observation, I agree!

Issue #3692, opened 2024-12-05
[BUG] Funcparser causes unexpected mutations to command help files


InspectorCaracal:
#### Describe the bug
When `FUNCPARSER_PARSE_OUTGOING_MESSAGES_ENABLED = True`, certain strings appear to be "eaten" by the funcparser parsing, despite not being valid func strings.

#### To Reproduce
1. Set `FUNCPARSER_PARSE_OUTGOING_MESSAGES_ENABLED = True` in your settings
2. Type `help nick` in game to view the CmdNick help text
3. The Examples section will incorrectly display as the following:
```
Examples:
  nick hi = say Hello, I'm Sarah!
  nick/object tom = the tall man
  nick build $1 $2 = create/drop $1;$2
  nick tell $1 $2=page $$2
  nick tm?$$1
  nick tm=$$1
  ```

The final three examples are the ones affected. These are the original lines:

```
      nick tell $1 $2=page $1=$2
      nick tm?$1=page tallman=$1
      nick tm\=$1=page tallman=$1
```

As you can see, these all have a consistent feature: the initial 1 after a $ is followed immediately by an equals sign without spaces. Strangely enough, the $2 being followed immediately by an equals sign does not seem to trigger this effect, only the 1

#### Expected behavior
It should display the help text as written, since there are no valid func strings involved.

#### Environment, Evennia version, OS etc
Evennia 4.5.0

count-infinity:
There were some bugs in the funcparser that I addressed in a first shot #3694 .

The issue mentioned here was caused by `curr_func.rawstr = curr_func.rawstr[: -len(infuncstr)]` which wiped the whole string when infuncstr has 0 length.  That only occurred when a $function was called inside another $function (or the parser thinks that's what's happening anyway).

As currently designed, the funcparser will eat escape character '\'  and add the next literal character no matter if it's parsing a function or not, so `nick tm\=$1=page tallman=$1` will become `nick tm=$1=page tallman=$1` without the '\'.

Ideally for this use case (default help responses) I think we'd be looking for a way to bypass the outgoing parser completely, overriding FUNCPARSER_PARSE_OUTGOING_MESSAGES_ENABLED value.

There is an `escape` kwarg that looks like it would escape so the function doesn't execute, but parser still eats backslashes.

Griatch:
Partially resolved in #3364.

PR #3694, opened 2024-12-16
Fixing funcparser issues

count-infinity:
Addresses some funcparser issues identified in #3692 where funcparser would eat characters.

Fix bugs seen in #3692 

This is not a full fix for #3692 as that specifically requests `help nick` (as the example) to return as it is written, and the parser seems to eat escapes '\' by design.  So in example case the literal
`nick tm\=$1=page tallman=$1` 
 becomes
`nick tm=$1=page tallman=$1`
Without the escape character.

### DIFF
--- a/evennia/utils/tests/test_funcparser.py
+++ b/evennia/utils/tests/test_funcparser.py
@@ -231,6 +231,8 @@ class TestFuncParser(TestCase):
             ("Test literal3 $typ($lit(1)aaa)", "Test literal3 <class 'str'>"),
             ("Test literal4 $typ(aaa$lit(1))", "Test literal4 <class 'str'>"),
             ("Test spider's thread", "Test spider's thread"),
+            ("Test invalid syntax $a=$b", "Test invalid syntax $a=$b"),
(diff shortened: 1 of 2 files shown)


evennia/utils/tests/test_funcparser.py
```diff
@@ -231,6 +231,8 @@ def test_constructor_raises(self, patched_variable_from_module):
             ("Test literal3 $typ($lit(1)aaa)", "Test literal3 <class 'str'>"),
             ("Test literal4 $typ(aaa$lit(1))", "Test literal4 <class 'str'>"),
             ("Test spider's thread", "Test spider's thread"),
+            ("Test invalid syntax $a=$b", "Test invalid syntax $a=$b"),
+            (r"Test invalid syntax $a\= b", "Test invalid syntax $a= b"),
```
Griatch, line 235: This is an incorrect unit test, while it's ok

Griatch (approved):
The \= is still an issue, but this seems to fix most of the issue, so approved. We cannot have incorrect unit tests in the suite though, so I will remove that until we can have a correct behavior tested.

Issue #3699, opened 2024-12-19
[BUG] dbserialize gets stuck at process_item while handling IntFlag enum type


0xDEADFED5:
#### Describe the bug
infinite recursion while iterating the enum

#### To Reproduce
1. add IntFlag to .db attribute
2. See error

```
File "C:\Users\ANON\repos\AI_Grotto\evennia\evennia\utils\dbserialize.py", line 720, in process_item
	    return item.__class__([process_item(val) for val in item])
	  File "C:\Users\ANON\repos\AI_Grotto\evennia\evennia\utils\dbserialize.py", line 720, in process_item
	    return item.__class__([process_item(val) for val in item])
	  File "C:\Users\ANON\repos\AI_Grotto\evennia\evennia\utils\dbserialize.py", line 720, in process_item
	    return item.__class__([process_item(val) for val in item])
	  File "C:\Python312\Lib\enum.py", line 1515, in __iter__
	    yield from self._iter_member_(self._value_)
	  File "C:\Python312\Lib\enum.py", line 1393, in _iter_member_by_value_
	    for val in _iter_bits_lsb(value & cls._flag_mask_):
	builtins.RecursionError: maximum recursion depth exceeded
```
i'll take a look at it in a moment

PR #3700, opened 2024-12-19
fix infinite recursion for serializing enum.IntFlag types

0xDEADFED5:
fix for https://github.com/evennia/evennia/issues/3699

### DIFF
--- a/evennia/utils/dbserialize.py
+++ b/evennia/utils/dbserialize.py
@@ -35,6 +35,7 @@ from django.utils.safestring import SafeString
 import evennia
 from evennia.utils import logger
 from evennia.utils.utils import is_iter, to_bytes, uses_database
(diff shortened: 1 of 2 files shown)


0xDEADFED5:
i think this isn't fully baked yet, looks like i broke stuff

0xDEADFED5:
looks good now, but it's a hacky way to check the type. couldn't figure out a better way

0xDEADFED5:
i just built python 3.10.16 from source to have a look.  3.10 fails because IntFlag is a different type there, working on a fix

0xDEADFED5:
sorry for spamming the shit outta this, it looks like serializing IntFlags isn't going to be a simple fix for Python 3.10 unless someone else has a better idea, I think this is as good as it gets for now

PR #3706, opened 2025-01-06
Update Korean

aceamro:


Griatch (changes requested):
Thanks for the PR, but it cannot merge like this: 
- This PR introduces more than just the Korean language, it also updates python code and there is no information in the PR, nor links to any issues explaining what this is doing. Please add this.
- This PR has failing unit tests. They need to be fixed.

PR #3710, opened 2025-01-10
Remove redundant and incorrect session-clearing from `at_server_cold_start`

InspectorCaracal:
The server-start code which calls this hook already correctly handles whether or not the session IDs should be cleared or not. The removed code is thus only serving to break session tracking when using `reset` mode.

Bug fixing


Griatch (approved):
Might be a left-over from an older start-process.

PR #3718, opened 2025-01-25
remove twistd.bat creation and usage

0xDEADFED5:

remove twistd.bat creation and usage

while testing Python 3.13 things were failing when they shouldn't.  i finally realized twistd.bat was calling an older version of twistd.exe from a completely different venv.

tested on Python 3.10 & Python 3.13, with and without venv


0xDEADFED5:
i removed the pywin32 error msg, Evennia runs fine so far without it. is there anything in Evennia which requires pywin32? if there is, none of the tests are hitting it.

Griatch (approved):
This is interesting. Maybe twistd has started being added on the Windows path since I first created this bootup sequence (I don't run Windows regularly). If we can skip the batch file that would make things a lot easier. Let's merge it and see if we get any complaints.

PR #3721, opened 2025-01-28
Avoid loading cmdsets that don't need to be checked

InspectorCaracal:
At the moment, the cmdhandler will load all of the command sets on all of the objects in a room when checking for merged commands, *regardless* of whether those objects will even allow you to call the commands on them. In most cases, this has no noticeable effect - however, in cases where there are a large number of commands attached to these objects, it can cause significant overhead by loading those cmdsets, despite them never being used.

This PR moves the lock check on the objects to the portion of the code which builds the list of objects to check, avoiding that cmdset-building for objects that are not callable.

Avoiding certain edge-case lag issues.


Griatch (approved):
I read the discussion in chat. Interesting that it would actually have an effect in some situations!

PR #3723, opened 2025-02-02
Fix bug that prevented exits with an ordinal alias from displaying

aMiss-aWry:
```
left ass cheek: Hi - Using the in-game map display contrib, I'm confused how checking for compass directions in exit aliases is supposed to work in this code from exit_name_as_ordinal:
        if exit_name not in _COMPASS_DIRECTIONS:
            compass_aliases = [
                direction in ex.aliases.all() for direction in _COMPASS_DIRECTIONS.keys()
            ]
            if compass_aliases[0]:
                exit_name = compass_aliases[0]
            if exit_name not in _COMPASS_DIRECTIONS:
                return ""

compass_aliases is a list of booleans, which makes sense checking a value with an if statement (although - it only checks north for some reason anyway?) but in the next line, exit_name is set to that boolean value, and then compared to _COMPASS_DIRECTIONS keys, which are strings representing north, south, etc. So since True and False aren't compass directions, the entire alias-checking block always returns False, even if it were checking all the directions (and not just [0] which corresponds to north)
I assume this isn't intended behavior? If so, if I fix it would it be valuable to share?
```

Fix above bug noticed by discord user

### DIFF
--- a/evennia/contrib/grid/ingame_map_display/ingame_map_display.py
+++ b/evennia/contrib/grid/ingame_map_display/ingame_map_display.py
@@ -125,16 +125,11 @@ class Map(object):
         Returns:
             string: The exit name as a compass direction or an empty string.
         """
-        exit_name = ex.name
-        if exit_name not in _COMPASS_DIRECTIONS:
-            compass_aliases = [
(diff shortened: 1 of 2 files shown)


Griatch (approved):
Thanks for the fix!

Issue #3724, opened 2025-02-03
[BUG] Required version of Twisted (>=24.11) breaking Evennia `yield`?


aMiss-aWry:
#### Describe the bug
Stated Twisted requirement (24.11) does not work well with Evennia's usage of `yield`.

#### To Reproduce
1. `pip install --upgrade twisted`
2. `reload`
3. Try to `destroy` any object in game

#### Environment, Evennia version, OS etc

Pulled the latest changes and on `evennia restart`, saw:
```
    ERROR: Twisted 23.10.0 found. Evennia requires
    version 24.11 or higher.
```
I upgraded twisted with `pip install --upgrade twisted` and now:

```
evennia -v

    Evennia 4.5.0 (rev 8e6474c8d1)
    OS: posix
    Python: 3.11.10
    Twisted: 24.11.0
    Django: 4.2.13
```

```
@create box
You create a new Object: box.

destroy box
Are you sure you want to destroy box [yes]/no?
Traceback (most recent call last):
  File "/[redacted]/muddev/evennia/evennia/commands/cmdhandler.py", line 633, in
_run_command
    ret = yield ret
          ^^^^^^^^^
  File "/[redacted]/muddev/evenv/lib/python3.11/site-packages/twisted/internet/defer.py",
line 2017, in _inlineCallbacks
    result = context.run(gen.send, result)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/[redacted]/muddev/evennia/evennia/commands/default/building.py", line 948, in
func
    elif answer.strip().lower() in ("n", "no"):
         ^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'strip'

An untrapped error occurred.
(Traceback was logged 25-02-03 14:54:58).
```

Just to make sure that was the problem, I went back to twisted 23.10.0

```
evennia -v

    ERROR: Twisted 23.10.0 found. Evennia requires
    version 24.11 or higher.


    Evennia 4.5.0 (rev 8e6474c8d1)
    OS: posix
    Python: 3.11.10
    Twisted: 23.10.0
    Django: 4.2.13
```
```
destroy box
Are you sure you want to destroy box [yes]/no?
y
box was destroyed.
```

0xDEADFED5:
if you're seeing that message it means you haven't upgraded Evennia

0xDEADFED5:
i see the same thing =/

0xDEADFED5:
[...]

PR #3726, opened 2025-02-08
fix create/destroy bug that surfaced in Twisted 24, get rid of some uses of returnValue()

0xDEADFED5:

fixes https://github.com/evennia/evennia/issues/3724
twisted.internet.defer.returnValue is deprecated and slated for removal.

bugfix

in the lines i removed(`ret = yield ret`), `ret` is always None or a generator in my testing.... never a Deferred (this is inside func decorated with \@inlineCallbacks).  Since it's not a Deferred I don't think it can be yielded.  Why is it failing now?  Not sure, but Twisted did make a change that allows you to yield coroutines as well, so my best guess is that change is causing it to fail now.

3.13 tests were failing because scipy 1.12.0 wheels weren't available for 3.13...bumped version so tests can run

### DIFF
--- a/evennia/commands/cmdhandler.py
+++ b/evennia/commands/cmdhandler.py
@@ -630,14 +630,12 @@ def cmdhandler(
             if isinstance(ret, types.GeneratorType):
                 # cmd.func() is a generator, execute progressively
                 _progressive_cmd_run(cmd, ret)
-                ret = yield ret
                 # note that the _progressive_cmd_run will itself run
                 # the at_post_cmd etc as it finishes; this is a bit of
                 # code duplication but there seems to be no way to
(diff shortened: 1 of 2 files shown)


0xDEADFED5:
for the last commit, i updated the code based on my assumptions.  if i'm right the code is functionally the same but more technically correct. i don't think the result of _run_command is ever useful or used anywhere

Griatch (approved):
Tested and veritied; this does fix the issue. There are other uses of `returnValue` in the code than this; and we may also want to start replacing the `inlineCallbacks` call for `async/await` as per the latest Twisted docs, but that's a separate project. 

Thanks for investigating this!

PR #3728, opened 2025-02-11
Update godot contrib README.md

ChrisLR:
Using localhost causes a significant delay when connecting, using 127.0.0.1 avoids that.

Bugfix

Thanks to CatDadKai for finding out the solution!


Griatch (approved):
Interesting find. I would expect this to be OS dependent though.

PR #3739, opened 2025-03-10
Don't use f-string for i18n

JohniFi:
Python f-strings don't work with i18n (gettext). Refactor to use `.format` instead.

Bux Fixing

Similar issue already discussed in 2021 on [Discord](https://discordapp.com/channels/246323978879107073/246323978879107073/885436654825902090)

In that file (accounts.py) are also some Strings that are not marked for translation (`"..."` instead of `_("...")`). Is that intentionally? For example `"You are already puppeting this object."` on [Line 474](https://github.com/evennia/evennia/blob/main/evennia/accounts/accounts.py#L474). That one was at some point translated (see e.g. current german (de) PO file) but does not get pulled into the PO file on `evennia makemessages` at the moment.

### DIFF
--- a/evennia/accounts/accounts.py
+++ b/evennia/accounts/accounts.py
@@ -516,7 +516,7 @@ class DefaultAccount(AccountDB, metaclass=TypeclassBase):
                 and len(self.get_all_puppets()) >= _MAX_NR_SIMULTANEOUS_PUPPETS
             ):
                 self.msg(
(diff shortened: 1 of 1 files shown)


Griatch (approved):
Thanks! Not all parts of Evennia's core should be under i18n. The account.py is a bit of a greyzone (an account can easily be overridden in code, but it's much less common to do than e.g. the character/object typeclasses). And having some strings translated and some not is just confusing, so I'll homogenize that.

PR #3743, opened 2025-03-13
Log stack trace when object creation fails

aMiss-aWry:
Log the full trace when object creation fails.

`["An error occurred while creating object 'orange cat object: '>' not supported between instances of 'int' and 'NoneType'"]`

That was the only error available to me when the create command failed in-game. I figured out the problem by adding the trace line in evennia core and feel that it will be useful to others. It would have taken more effort to debug.

Found it a prior issue on discord: https://discord.com/channels/246323978879107073/980878618685231194/1311335819168186412
image

The same search string ("An error occurred while creating") in discord comes up with many results of people trying to debug where they went wrong with insufficient information to get help.

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -1481,8 +1481,8 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
                 obj.db.desc = description
 
         except Exception as e:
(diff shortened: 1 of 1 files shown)


Griatch (approved):
This is fair enough, doesn't heart to be more informative here. :+1:

PR #3744, opened 2025-03-15
fix format strings for i18n

JohniFi:
The formating of the strings via `.format()` has to be called "after" translation via `gettext()`.
`_("...").format()` instead of `_("...".format())`
Otherwise the string will not be translated because at runtime the format method returns a String with already filled out content which then does not match any translation message via gettext().

Bux fixing.

similar discussion: [Stackoverflow: Translating formatted strings in Django not working](https://stackoverflow.com/questions/44957261/translating-formatted-strings-in-django-not-working)


Griatch (approved):
This is a good catch. Thanks!

Issue #3613, opened 2024-09-12
[Documentation] All keyword arguments of type `list` in the docs are being linked to a specific contrib's list() function


aMiss-aWry:
#### Existing page / new
many, example: https://www.evennia.com/docs/latest/api/evennia.utils.create.html

#### Documentation issue
keyword arguments of type (list) link to https://www.evennia.com/docs/latest/api/evennia.contrib.game_systems.barter.barter.html#evennia.contrib.game_systems.barter.barter.TradeHandler.list

because the documentation thinks all references to `list` must mean this specific function?

#### Suggested change
Don't link all things of type `list` to a specific function named `list()`

https://github.com/user-attachments/assets/f4396074-c2ed-41ff-badf-201aea5fcb2a

Griatch:
Hm, that `list` is a method on a class, so it shouldn't clash with the regular Python `list` ... :thinking: Building this is happening automatically by the sphinx build system. Not sure how to fix this, really, needs more research :disappointed:

aMiss-aWry:
Appears to be somewhat widely known/long-running issue: https://github.com/sphinx-doc/sphinx/issues/2549

It seems they had a fix three years ago on Sphinx 4.0.0.beta1 released on 2021-04-12.

We appear to be using a much older version: Created using [Sphinx](https://www.sphinx-doc.org/) 3.2.1.

This stackoverflow answer suggests a patch as of 2023 for those on older versions (if breaking changes make updating unrealistic): https://stackoverflow.com/questions/31784830/sphinx-ivar-tag-goes-looking-for-cross-references

>For those running older Sphinx versions than 4.0, here is a much simpler fix than the monkey-patch from @mzjn that was tested back to Sphinx 1.6.7 from Ubuntu 18.04. The fix basically just does the same thing that was done in PR [#8638](https://github.com/sphinx-doc/sphinx/pull/8638) that fixed the issue.

>Put the following code into your conf.py:

```
from sphinx.domains.python import PyObject
PyObject.doc_field_types[map(lambda f: f.name == 'variable', PyObject.doc_field_types).index(True)].rolename = None
```

So there's some options to try, maybe?

Griatch:
Thanks for looking! Yeah, there are other improvements in 4+ too. 

[...]

PR #3745, opened 2025-03-16
upgrade docs to Sphinx 8

0xDEADFED5:
fixes https://github.com/evennia/evennia/issues/3613

i've only had a quick look at this, and i haven't tested multi-version.

i did notice that the API listing is expanded by default now.

for instance /evennia/docs/build/latest/api/evennia.utils.html#evennia-utils

looks like this now:

docs_new

instead of this:

docs_old

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 Updated dependencies: Django >5.1 (<5,2), Twisted >24 (<25).
 Python versions: 3.11, 3.12, 3.13.
 
+This upgrade requires running `evennia migrate` on your existing database
+(ignore any prompts to run `evennia makemigrations`).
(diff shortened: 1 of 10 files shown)


Griatch:
Thanks for updating this! It builds, but I see there are _a lot_ of build problem since the update. I will work off this branch to see if I can resolve some of it, as well as clean up some of the local changes I've made as well :smiling_face_with_tear:

0xDEADFED5:
erm, yeah. i had a hard time grokking that nightmare.

0xDEADFED5:
ugh, i suck with git

0xDEADFED5:
i don't know what i did, i'll redo it, sorry

PR #3751, opened 2025-03-23
Handle missing account in Command

### LINKED ISSUE #3750: [BUG] Cannot run some default commands without a connected account
#### Describe the bug
Executing some commands on a `Character` object without an attached account causes errors. For instance, running `inventory` on a character with an item in their inventory (it can't be empty), throws the following error:

```
  File "/Users/elias/PycharmProjects/evennia/.venv/lib/python3.11/site-packages/evennia/commands/cmdhandler.py", line 629, in _run_command
    ret = cmd.func()
          ^^^^^^^^^^
  File "/Users/elias/PycharmProjects/evennia/.venv/lib/python3.11/site-packages/evennia/commands/default/general.py", line 373, in func
    table = self.styled_table(border="header")
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/elias/PycharmProjects/evennia/.venv/lib/python3.11/site-packages/evennia/commands/command.py", line 641, in styled_table
    border_color = self.account.options.get("border_color")
                   ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'options'
```

#### To Reproduce
1. Set up some way to see the error messages thrown by a character that you aren't controlling. I'm not aware of any built-in way to do this, so I made a custom class for it:
```python
from typeclasses.characters import Character

class FakeCharacter(Character):
    def msg(self, text=None, from_obj=None, **kwargs) -> None:
        if text is not None:
            try:
                is_say = text[1]["type"] == "say"
            except Exception:
                is_say = False

            if from_obj is None and not is_say:
                self.execute_cmd(f"say {text}")

        super().msg(text=text, from_obj=from_obj, **kwargs)
```
2. Spawn the character: `create/drop npc:typeclasses.fake_character.FakeCharacter`
3. Give it an item: `create/drop sword`, `py self.search("npc").execute_cmd("get sword")`
4. Run the `inventory` command as that character: `py self.search("npc").execute_cmd("inventory")`
5. The character will `say` the error:
```
>>> self.search("npc").execute_cmd("inventory")
npc says, "Traceback (most recent call last):
  File "/Users/elias/PycharmProjects/evennia/.venv/lib/python3.11/site-packages/evennia/commands/cmdhandler.py", line 629, in _run_command
    ret = cmd.func()
          ^^^^^^^^^^
  File "/Users/elias/PycharmProjects/evennia/.venv/lib/python3.11/site-packages/evennia/commands/default/general.py", line 373, in func
    table = self.styled_table(border="header")
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/elias/PycharmProjects/evennia/.venv/lib/python3.11/site-packages/evennia/commands/command.py", line 641, in styled_table
    border_color = self.account.options.get("border_color")
                   ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'options'

An untrapped error occurred.
(Traceback was logged 25-03-22 20:24:41)."
<Deferred at 0x1083b48d0 current result: None>
```

#### Expected behavior
All default commands that aren't account-specific should execute correctly without needing an attached account.

I'm on MacOS 15.4 with an M2 processor.

```
    Evennia 4.5.0
    OS: posix
    Python: 3.11.11
    Twisted: 23.10.0
    Django: 4.2.19
```

#### Additional context
The reason I'm trying to do this is because I want to create an NPC that is fully controlled by an LLM. My NPC class extends `Character` because I want it to be able to do everything a player can do.

I've only run into this bug with the `inventory` command, but I imagine it exists on other commands too. I'll comment on this issue if I find any more affected commands.

EliasWatson:
Fixes the bug mentioned in https://github.com/evennia/evennia/issues/3750. I tested that the `inventory` and `access` commands both work now.

Closes #3750

### DIFF
--- a/evennia/commands/command.py
+++ b/evennia/commands/command.py
@@ -623,6 +623,22 @@ Command \"{cmdname}\" has no defined `func()` method. Available properties on th
             )[0]
         return settings.CLIENT_DEFAULT_WIDTH
 
+    def _get_account_option(self, option):
+        """
+        Retrieve the value of a specified account option.
+
+        Args:
+            option (str): The name of the option to retrieve.
+
+        Returns:
+            The value of the specified account option if the account exists,
+            otherwise the default value from settings.OPTIONS_ACCOUNT_DEFAULT.
+
+        """
+        if self.account:
+            return self.account.options.get(option)
+        return settings.OPTIONS_ACCOUNT_DEFAULT.get(option)
+
     def styled_table(self, *args, **kwargs):
         """
         Create an EvTable styled by on user preferences.
@@ -638,8 +654,8 @@ Command \"{cmdname}\" has no defined `func()` method. Available properties on th
                 or incomplete and ready for use with `.add_row` or `.add_collumn`.
 
         """
-        border_color = self.account.options.get("border_color")
-        column_color = self.account.options.get("column_names_color")
+        border_color = self._get_account_option("border_color")
+        column_color = self._get_account_option("column_names_color")
 
         colornames = ["|%s%s|n" % (column_color, col) for col in args]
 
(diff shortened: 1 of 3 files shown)


Griatch (approved):
These are some nice finds. Running a character 'headless' is not so common (like you want to do with an LLM), so these were not encountered before. It would generally be a good idea to establish a set of unit tests for these fixes too, something to think about for future PRs - as it is, I may look into adding them later.

PR #3756, opened 2025-03-28
update german translation

JohniFi:
Updated the german .po file to reflect the latest main branch status and added some missing translations.

Internationalization

Disclaimer: I skipped `utils/validatorfuncs.py` and `utils.eveditor.py` because they contain many strings that are hard/unreasonable to translate due to combinations of sentences with placeholders like `{option_key}` that are not translated and overall containing many technical words etc.


JohniFi:
The test-suite failed because of:
`docker: Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit.`
Not related to my code at all i guess?

PR #3757, opened 2025-03-28
Add internationalization (i18n via gettext)

JohniFi:
Marked some more strings for translation so you don't need to overwrite whole methods of e.g. `DefaultObject` in your own game just to translate the simple messages defined there (without changing the behavior of the methods).

**Disclaimer:** [get_numbered_name()](https://github.com/evennia/evennia/blob/main/evennia/objects/objects.py#L1674) is impossible to translate via gettext because it uses [inflect](https://pypi.org/project/inflect/) for inflection (only works in english). So you still need to overwrite this method in your foreign language.

Internationalization

### DIFF
--- a/evennia/objects/objects.py
+++ b/evennia/objects/objects.py
@@ -1382,13 +1382,14 @@ class DefaultObject(ObjectDB, metaclass=TypeclassBase):
 
             if obj.has_account:
                 if home:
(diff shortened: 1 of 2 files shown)


Griatch (approved):
These are some nice convenience changes for i18n! :+1:

PR #3758, opened 2025-03-29
GaugeTrait check for last_update

RianeDev:
This is just copying the fix for CounterTrait over to GaugeTrait, since they had the same issue.

As per: https://github.com/evennia/evennia/pull/3495

There may be a better way to just reference it from CounterTrait as this feels somewhat pedantic, but python isn't my strong suit and this does the job. Tested both on player going ooc and back to ic, as well as on reload.


Griatch (changes requested):
Are we seeing actual issues with `GaugeTrait`?  Since `GaugeTrait` inherits from `CounterTrait`, this static method is already available on `GaugeTrait` without having to duplicate any code ...?

Issue #3759, opened 2025-03-29
[BUG] saving store_key from TickerHandler in db attribute breaks it


0xDEADFED5:
![Image](https://github.com/user-attachments/assets/bf0dd6df-1a7d-4c14-8ac8-446257b9321c)

PR #3760, opened 2025-03-29
change store_key format for TickerHandler

0xDEADFED5:
fixes https://github.com/evennia/evennia/issues/3759

not sure if this is the best way to fix it, but it seems reasonable?

i tested to see if existing tickers would be affected by: adding a ticker with original code, restarting, applying this patch, and restarting the server again.  the ticker remained.  as far as i can tell this shouldn't affect existing games.

anyone relying on the old store_key format will have a breaking change, but since it didn't seem to work (after being serialized) that should hopefully be nobody

### DIFF
--- a/evennia/scripts/tickerhandler.py
+++ b/evennia/scripts/tickerhandler.py
@@ -364,8 +364,7 @@ class TickerHandler(object):
         Tries to create a store_key for the object.
 
         Args:
-            obj (Object, tuple or None): Subscribing object if any. If a tuple, this is
-                a packed_obj tuple from dbserialize.
+            obj (Object, tuple or None): Subscribing object if any.
             path (str or None): Python-path to callable, if any.
             interval (int): Ticker interval. Floats will be converted to
                 nearest lower integer value.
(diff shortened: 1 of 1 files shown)


Exsavio:
Hello, I have tested this change; and whilst this does create a consistent store_key, it introduces a new issue. If you create a persistent ticker using this code; and reload the server you now receive this error:

```
2025-03-30 15:00:11 [!!]   File "C:\Projects\FreshEvennia\evenv\Lib\site-packages\evennia\scripts\tickerhandler.py", line 132, in _callback
2025-03-30 15:00:11 [!!]     if not obj or not obj.pk:
2025-03-30 15:00:11 [!!]                       ^^^^^^
2025-03-30 15:00:11 [!!] AttributeError: 'int' object has no attribute 'pk'
```

I believe this is because the original serialized information contained in the key is being used after a reload to deserialize the target object and re-establish the ticker in restore().

0xDEADFED5:
crap, thanks for the report

0xDEADFED5:
obviously i didn't look at this closely enough. i don't think this is going in the right direction

PR #3764, opened 2025-03-30
upgrade docs to sphinx 8

electroglyph:
upgrade docs to Sphinx 8

fixes https://github.com/evennia/evennia/issues/3613

i've only had a quick look at this, and i haven't tested multi-version.

i did notice that the API listing is expanded by default now.

for instance /evennia/docs/build/latest/api/evennia.utils.html#evennia-utils

looks like this now:

docs_new

instead of this:

docs_old

sorry for the repost, i suck with git.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,18 +1,17 @@
 # requirements for building the docs
 
-sphinx==3.2.1
-myst-parser==0.15.2
-myst-parser[linkify]==0.15.2
-Jinja2 < 3.1
+sphinx==8.2.3
+myst-parser==4.0.1
+myst-parser[linkify]==4.0.1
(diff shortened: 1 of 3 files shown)


Griatch:
@0xDEADFED5 As an update, I am working off the [now closed PR branch](https://github.com/evennia/evennia/pull/3745) (locally), to fix the numerous issues caused by the move to Sphinx 8. It seems later versions of Sphinx are more pedantic than they used to be, so this takes some time. 

I'll see if I'll rebase to this one, but my local version might have diverted too much for this to be practical. If so, I might end up closing this PR. Either way, I'll keep it open as a tracker for now.

Griatch:
Closing this now, thanks for your work on it!

PR #3765, opened 2025-03-30
rebuild TickerHandler store_key if it's been serialized

0xDEADFED5:

fixes https://github.com/evennia/evennia/issues/3759


Exsavio:
This does appear to fix the issue, and not mess up reload/restore(). Thanks!

Griatch (approved):
I tried a few different approaches, but this turns out to be easiest way around this issue. So LGTM. Normally, I'd also request a new unit test to catch this error, but since I have the code open anyway, I can add that. :)

PR #3768, opened 2025-03-31
Copied objects retain plural alias category.

### LINKED ISSUE #3746: [BUG] Preserve alias (tag) category on CmdCopy
#### Describe the bug
When copying an object ingame via the [copy <obj>](https://github.com/evennia/evennia/blob/main/evennia/commands/default/building.py#L365) command, the aliases get copied over with added "_copy" but the tag category of the aliases (e.g. ["plural_key"](https://github.com/evennia/evennia/blob/main/evennia/objects/objects.py#L416) used in [DefaultObject.get_numbered_name()](https://github.com/evennia/evennia/blob/main/evennia/objects/objects.py#L1717) is lost. So it for example can't be cleared by [get_numbered_name()](https://github.com/evennia/evennia/blob/main/evennia/objects/objects.py#L1719) because it just clears aliases of that category.

#### To Reproduce
1. `create/drop apple`
2. `look`
3. `copy apple`
4. `examine apple_copy`
5. `name apple_copy = apple`
6. `examine apple-2`
7. copied apple still has plural aliases for "apple_copy" (now without tag category "plural_key")

#### Expected behavior
The tag category should be copied too, not just the tag name and type (alias). 

    Evennia 4.5.0 (rev bf0c7bc5c)
    OS: posix
    Python: 3.12.9
    Twisted: 24.11.0
    Django: 5.1.7

#### Additional context
Probably just need to change the [CmdCopy](https://github.com/evennia/evennia/blob/main/evennia/commands/default/building.py#L365) somehow.

jaborsh:
`CmdCopy` now receives (alias,category) pairs from ObjectDB to ensure plural categories are not lost on the new object.

Of note is #3432 which describes a longstanding issue regarding plural aliases in the first place.

Resolves #3746

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -397,7 +397,10 @@ class CmdCopy(ObjManipCommand):
             if not from_obj:
                 return
             to_obj_name = "%s_copy" % from_obj_name
-            to_obj_aliases = ["%s_copy" % alias for alias in from_obj.aliases.all()]
+            to_obj_aliases = [
+                (f"{alias}_copy", category)
+                for alias, category in from_obj.aliases.all(return_key_and_category=True)
+            ]
             copiedobj = ObjectDB.objects.copy_object(
                 from_obj, new_key=to_obj_name, new_aliases=to_obj_aliases
             )


Griatch (approved):
Nice catch!

PR #3785, opened 2025-04-10
Refactor is_ooc lockfunc and remove objtag redefinition

### LINKED ISSUE #3583: [BUG] MULTISESSION_MODE 3 / Command Locks Bug
#### Describe the bug
Commands that make use of the is_ooc() lock (and maybe others?) encounter an interesting bug when combined with `MULTISESSION_MODE = 3`. Should `CmdIC` be called with an `is_ooc()` lockstring, the session will have a puppet, but the account itself will be treated as IC rather than just the session.

Should a second session initialize and disconnect, it no longer has access to any of the commands blocked by the `is_ooc()` lockstring.

#### To Reproduce
1. Set `MULTISESSION_MODE = 3` in `settings.py`
2. Add `locks = "cmd:is_ooc()"` to a command.
3. Start two sessions: one IC with a puppet and one OOC with none.
4. Attempt to puppet an object with the latter session and fail.

#### Expected behavior
`is_ooc()` lock should apply to session, not to account.

jaborsh:

Codebase simplification:

* Removed the `objtag` function, which was redundant as it simply called the `tag` function with different parameters.
* Simplified the `is_ooc` function. In doing so, high multisession mode complications resolved. `Account` objects do not have `sessid` but all other objects do, when puppeted.

Resolves #3583

### DIFF
--- a/evennia/locks/lockfuncs.py
+++ b/evennia/locks/lockfuncs.py
@@ -477,18 +477,6 @@ def tag(accessing_obj, accessed_obj, *args, **kwargs):
     return bool(accessing_obj.tags.get(tagkey, category=category))
 
 
-def objtag(accessing_obj, accessed_obj, *args, **kwargs):
-    """
-    Usage:
-        objtag(tagkey)
-        objtag(tagkey, category):
-
-    Only true if `accessed_obj` has the given tag and optional category.
-
-    """
-    return tag(accessed_obj, None, *args, **kwargs)
-
-
 def objloctag(accessing_obj, accessed_obj, *args, **kwargs):
     """
     Usage:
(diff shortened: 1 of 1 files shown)


Griatch:
As a heads-up, I won't review or look at this while it's marked `Draft` :)

jaborsh:
> As a heads-up, I won't review or look at this while it's marked `Draft` :)

No worries. I marked as draft as a result of our conversation discussing making this change with a SessionHandler you said you've made some progress on locally. Just keeping this open until you have a chance to push it to a branch or some such.

Griatch:
I have indeed a WIP SessionHandler locally, but it depends on other, bigger changes (probably SessionDB), so this will need to be parked a while.

PR #3787, opened 2025-04-11
Add type hint to lazy_property to return the underlying type

Problematic:
Adds proper type hinting for the type that `lazy_property` wraps. Hat tip to @owllex on discord for pointing me in the right direction.

I like type hints :)

### DIFF
--- a/evennia/utils/utils.py
+++ b/evennia/utils/utils.py
@@ -27,6 +27,8 @@ from inspect import getmembers, getmodule, getmro, ismodule, trace
 from os.path import join as osjoin
 from string import punctuation
 from unicodedata import east_asian_width
+from collections.abc import Callable
+from typing import Generic, TypeVar, overload
 
 from django.apps import apps
 from django.conf import settings
@@ -2180,8 +2182,9 @@ def deepsize(obj, max_depth=4):
 # lazy load handler
 _missing = object()
 
+TProp = TypeVar("TProp")
 
-class lazy_property:
+class lazy_property(Generic[TProp]):
     """
     Delays loading of property until first access. Credit goes to the
     Implementation in the werkzeug suite:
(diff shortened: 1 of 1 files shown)


evennia/utils/utils.py
```diff
@@ -2202,18 +2204,24 @@ def attributes(self):
...
+
+    def __get__(self, obj, type=None) -> TProp | 'lazy_property':
         """Triggers initialization"""
         if obj is None:
             return self
-        value = obj.__dict__.get(self.__name__, _missing)
+        value: TProp = obj.__dict__.get(self.__name__, _missing)
         if value is _missing:
             value = self.func(obj)
         obj.__dict__[self.__name__] = value
```
Problematic, line 2227: I didn't touch it in case it's this way on purpose, but can this caching be moved into the if block? Seems like a redundant step if it's not missing.

Problematic, line 2227: I didn't touch it in case it's this way on purpose, but can this caching be moved into the if block? Seems like a redundant step if it's not missing.

Griatch: It's not redundant, we need the `value` variable to be calculated also if the `if` statement doesn't pass. We could maybe use the newish `:=` operator, but it would probably just make this harder to read.

Griatch (approved):

Griatch (changes requested):
The idea is not bad, but this doesn't pass unit tests.

evennia/utils/utils.py
```diff
@@ -2181,7 +2183,7 @@ def _recurse(o, dct, depth):
 _missing = object()
 
 
-class lazy_property:
+class lazy_property[TProp]:
```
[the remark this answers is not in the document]
Problematic: Ah, yep, looks like this syntax is only valid for 3.12+. I'll make the change; is 3.10 still the earliest supported version?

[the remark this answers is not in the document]
Griatch: Since this is for `main` branch, Python 3.11 is the earliest supported version (we dropped 3.10 support.

Issue #3791, opened 2025-05-22
[BUG] Hooks in DefaultObject do not call super()


Iiiiiiigor:
#### Describe the bug

Most hooks (e.g. at_object_creation) in DefaultObject are missing the call to super() method -- this makes multiple inheritance fragile and is often tricky to spot.

#### Expected behavior
All DefaultObject hooks that are meant to be overridden in derived classes should call super() -- otherwise multiple inheritance breaks easily: depending on order of inheritance some overloaded implementations might not be called at all.

At least this peculiar behaviour should be documented if it is hard to fix for some reason.

Griatch:
Thanks for the issue, but this makes no sense. 🤔 There is no peculiar behavior here -  `DefaultObject` _defines_ hooks like `at_object_creation`, there is no `super()` parent to call. If you _inherit_ from `DefaultObject` you can add `super()` but for most hooks, the parent is empty so `super()` doesn't matter.

Iiiiiiigor:
Hmm. sorry i have to nag. Let me start by saying, i really like evennia so far, *but* you seem to misunderstand python inheritance. Read for example https://rhettinger.wordpress.com/2011/05/26/super-considered-super/

There are a lot of edge-cases that can pop up, which can be avoided, if **every** implementation of a method calls super() and that includes the BaseClass with no other parents that have the method. Why? Because with multi-inheritance you easily (and almost accidentally) can inherit from the same BaseClass twice. Under long-standing python rules, this might mean that in the *method-resolution order* your BaseClass implementation is called somewhere in the middle, and not in the end of the inheritance chain: No call to super() -> chain broken, code not executed. Read up on it, really. Python has a pretty good approach to this problem.

I think you managed to avoid problems so far by using MixinClasses the way you do and avoiding to mix one mixin with another further up the chain (thus inheriting twice from BaseClass with some not Baseclass classes in between).

If you don't want to change this, fine by me. With proper precaution and can work just fine.
Just be aware that this slightly fishy by python standards.

Thanks again for your hard work. I find evennia a joy to work with so far.

Griatch:
@Iiiiiiigor Thanks for the link! I think there is a miscommunication here though.

You write

> Most hooks (e.g. at_object_creation) in DefaultObject are missing the call to super() method -- this makes multiple inheritance fragile and is often tricky to spot.

[...]

PR #3792, opened 2025-05-23
add warning about mro and DefaultObject

Iiiiiiigor:

add a warning about MRO and DefaultObject to the Tutorial.

It might seem obvious to some, but it took me a bit to debug a problem like this, so why not add the warning. 

https://github.com/evennia/evennia/issues/3791


Griatch:
Ah, this is something else than what you talk about in your other issue. You are referring to `MyObject` here, which I presume is a _child_ of `Object` or `DefaultObject`. If you have a mix-in overriding some relevant method of `Object/DefaultObject`, you must indeed make sure that your mixin is added earlier in the chain. Maybe you are confusing what `DefaultObject` is and what `MyObject` is - the two are not interchangeable in your example.

This has nothing to do with `DefaultObject` having  `super()` calls though (as discussed in #3791). It _could_ be relevant to add `super()` calls to your `MyObject` though, if it makes use of any `DefaultObject` parent resources.

Griatch (changes requested):
Thanks for the suggestion. I can see it being useful to clarify - elaborate a bit on the inheritance here, but I don't think confusing this with `DefaultObject` is helping.

Griatch:
Ping.

Griatch:
Thanks for bringing this up. I clarified the pitfalls with not providing the multiple inheritance in the right order in the tutorial now. Closing this.

PR #3793, opened 2025-05-28
Added Chinese Translation of the Evennia Documentation

bluedeep:
**Added Chinese Translation of the Evennia Documentation**

- A new `source_zh_CN` folder was added under the `evennia/docs/` path to translate the documentation in source into Simplified Chinese.
- Under `evennia/docs/source_zh_CN/pylib/`, the build scripts were modified to automatically handle contrib_readmes and dynamic_pages for different languages.
- The `evennia/docs/Makefile` was also updated to support the `SPHINX_LANGUAGE` environment variable for building multilingual documentation.

You can translate the source files located in `evennia/docs/source/` into different languages. 
Documentation source files for each language should be placed in the corresponding 
`evennia/docs/source_${SPHINX_LANGUAGE}/` directory. Here, ${SPHINX_LANGUAGE} should adhere to the
[ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes) standard.

For example, the source files for Simplified Chinese documentation should be located under the path 
`evennia/docs/source_zh_CN/`.

During compilation, you need to define the `SPHINX_LANGUAGE` environment variable to build the 
documentation for a specific language. For instance:
   ```  
   SPHINX_LANGUAGE=zh_CN make local
   ```


bluedeep:
> Thanks for this work! I am not certain we really want to support multiple languages like this in the main repo (we can't really maintain it), but also, we are in the process of updating the doc build process, so holding off on this one.

Got it, thanks for the clarification!

Totally understand the concerns about maintenance and the ongoing changes to the doc build process. I’ll go ahead and close this PR for now.

PR #3799, opened 2025-07-10
Update tb_basic.py to resolve type in scripts caller

Tharic99:
Original scripts.add was (self.command_handler_class), which does not exist. It should be (self.combat_handler_class)

This fix updates that only.


Tharic99:
I was tempted to add some generic code so the NPC actually fights back that I have working, but I didn't want to go outside of the original boundaries/intent of the contrib, so this just fixes the error that prevents it from successfully running.

Let me know if you'd like me to add NPC auto-combat though, I've got it sitting in a fork.

Griatch (approved):
Thanks for the change! The fighting back bit is something I wanted to delay until later in the tutorial, but I can see it also being useful for debugging :thinking: 

For now I'm just merging this fix though.

PR #3800, opened 2025-07-24
Update HAProxy Config Docs

holl0wstar:
Add config example for HAProxy on how to proxy telnet through it, so secure telnet with SSL can be used instead of just regular telnet.

Just changed the HAProxy examples in the documentation to include SSL Telnet as well.

This was not something that was covered for HAProxy at all- Only NGINX. I had to do some research on it and I figured it would be a nice to have since it's a question I've seen asked before a few times (and people were just directed to nginx, which is fine), but I feel like having a HAProxy example is also helpful.


Griatch (approved):
This is an interesting addition. Thanks for researching it. Merging.

Issue #3775, opened 2025-04-02
[Feature Request] Merge Caracal's cache buster into Evennia


0xDEADFED5:
#### Is your feature request related to a problem? Please describe.

actively development on the webclient is a pain for the dev, and possibly a pain for end users when stale files are in the browser cache

#### Describe the solution you'd like

@InspectorCaracal has a nice solution at https://github.com/InspectorCaracal/evelite-client with a cache buster that several people have included in their own webclients

PR #3812, opened 2025-08-25
Implement "cache buster"

InspectorCaracal:

Adds a new templatetag filter that adds a short query string to the end of static file paths. This is a common "cache-busting" technique to force browsers to redownload particular files if they've been updated, rather than waiting for the files to be cleared from the browser's cache on their own.

Feature request: #3775

My original implementation linked in the feature request *works*, but uses the file's mtime, which is generally less desirable than a contents hash as it forces the file to be re-downloaded even if it's the same just by virtue of it being saved again. It also requires hard-coding the specific file paths into the context, which isn't great! This method is more widely and easily reusable.

### DIFF
--- a/evennia/web/templates/webclient/base.html
+++ b/evennia/web/templates/webclient/base.html
@@ -6,7 +6,7 @@ with evennia set up automatically and get the Evennia JS lib and
 JQuery available.
 -->
 
-{% load static %}
+{% load static cachebuster %}
 <html dir="ltr" lang="en">
   <head>
     <title> {{game_name}} </title>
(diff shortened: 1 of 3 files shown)


InspectorCaracal:
Okay hold up. I found a Django thing while looking into something completely different which I think will be infinitely better than this and I'm real mad it's never come up before, but I'll need to try it out first to see

Griatch:
As a note, since this is marked `Draft`, I'm not reviewing it. :)

Griatch:
Ping.

PR #3818, opened 2025-09-01
Update spawn hook to use `new_prototype`

InspectorCaracal:
The initial implementation of the `at_object_post_spawn` hook erroneously used the `prototype` variable in the spawn-update logic, which could be a prototype key rather than the prototype itself, resulting in an unexpected input of a string. This corrects that to `new_prototype` which should be guaranteed to contain the actual homogenized prototype dict.

Resolving edge-case issues.


Griatch (approved):
Good catch!

PR #3825, opened 2025-10-09
Updated Portuguese translation

marado:

Back when there was a `develop` branch, I submitted and update to the Portuguese translation with #3285, which was accepted and merged into the `develop` branch, but never reached `main`.

This pull request presents that same patch to be pulled into `main`, as well as a new commit that fixes two Portuguese strings that where wrongly translated.

Enhance the Portuguese translation of Evennia.


Griatch:
Hm, that is curious - I see a change from you, from a few days earlier, was merged. But the change you merge here was later. I wonder if that was accidentally lost in a rebase since they looked similar by name. Thanks for catching that!

PR #3850, opened 2025-12-09
Fix potential error from clone-num logic in `DefaultObject.copy`

InspectorCaracal:
This adds handling for the situation where the object being copied has no location, and also makes it so that the produced key follows the rules described in the docstring (appends the next available integer suffix from the largest integer suffix present in the location).

Bug fixing

Unit tests included :tm:

### DIFF
--- a/evennia/objects/tests.py
+++ b/evennia/objects/tests.py
@@ -419,6 +419,10 @@ class TestObjectManager(BaseEvenniaTest):
         copied2.key = "CopyMe01"
         copied3 = self.obj1.copy()
         self.assertEqual(copied3.key, "CopyMe02")
+        # and that nothing breaks if something in the room doesn't share the prefix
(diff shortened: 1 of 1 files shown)


evennia/objects/objects.py
```diff
@@ -1506,15 +1506,20 @@ def find_clone_key():
...
+                obj.key.removeprefix(key)
                 for obj in self.location.contents
-                if obj.key.startswith(key) and obj.key.lstrip(key).isdigit()
-            )
-            return "%s%03i" % (key, num)
+            ]
+            num = 1
+            if nums := [int(suffix) for suffix in suffixes if suffix.isdigit()]:
+                num = max(nums) + 1
+            return f"{key}{num:02d}"
```
Griatch, line 1522: This changes behavior from `obj002` to `obj02` which could affect anyone already using it.

PR #3853, opened 2025-12-16
Fix multimatch when '-' in name.

### LINKED ISSUE #3691: [BUG] search multimatching stops working when object key contains a hyphen
#### Describe the bug
When searching for an object that has duplicates (same key) and contains a hyphen, e.g. "t-shirt", the returned list of duplicates with the appended numbers don't seem to work.

#### To Reproduce
1. Create two default objects called "t-shirt"
2. Create a search for said "t-shirt" objects. Should return a list something like...
```
More than one match for 't-shirt' (please narrow target):
 t-shirt-1 (carried)
 t-shirt-2 (carried)
```
4. Attempt to `get t-shirt-1`
5. See error
```
get t-shirt
More than one match for 't-shirt' (please narrow target):
 t-shirt-1
 t-shirt-2
get t-shirt-1
Could not find 't-shirt-1'.
```
#### Expected behavior
You should be able to pick up one of the t-shirts using `t-shirt-1`

    Evennia 4.4.1
    OS: nt
    Python: 3.12.7
    Twisted: 23.10.0
    Django: 4.2.16

count-infinity:
Fixes #3691 - search multimatching stops working when object key contains a hyphen (like "t-shirt")

This changes the default SEARCH_MULTIMATCH_REGEX to support hyphenated object names (e.g., t-shirt-1, pull-lever-2). The new regex matches up to the first -<number> that ends a word (followed by space or end of string), whereas the old regex only matched names without hyphens.

Fix bug.

Closes #3691

### DIFF
--- a/evennia/commands/tests.py
+++ b/evennia/commands/tests.py
@@ -1223,6 +1223,38 @@ class TestCmdParser(TestCase):
         self.assertEqual(cmdparser.try_num_differentiators("look me-3"), (3, "look me"))
         self.assertEqual(cmdparser.try_num_differentiators("look me-567"), (567, "look me"))
 
+    def test_num_differentiators_hyphenated_names(self):
+        """Test that hyphenated object names like 't-shirt-1' are parsed correctly.
+
+        This tests the default SEARCH_MULTIMATCH_REGEX which uses the format 'name-number'.
+        Objects with hyphens in their names (e.g., 't-shirt') should be correctly parsed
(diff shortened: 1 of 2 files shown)


Griatch (approved):
That's an interesting edge case. 
I note that the settings docstring is not (neither now, nor before) mentioning the `<args>` named capturing group which is also used by the cmdparser ...

PR #3854, opened 2025-12-16
Normalize CRLF to LF

### LINKED ISSUE #3847: [BUG] CRLF breaks batch code #INSERT
#### Describe the bug
When a batch code file is saved using CRLF line endings (as Windows does by default) instead of LF newlines, it cannot be correctly parsed by the batch code processor, causing it to error out.

#### To Reproduce
1. Save batchcode that uses #INSERT with LF line endings and execute within Evennia.
2. Batchcode executes without issue.
3. Save the same batchcode with CRLF line endings and execute within Evennia.
4. See error.

#### Expected behavior
Ideally, files should work with #INSERT regardless of whether they use LF or CRLF.

Evennia version: 5.0.1, running under Python venv
OS: observed on Windows

#### Additional context
Honestly not the biggest problem, VSCode and other editors make switching between CRLF and LF very easy so the bug can be avoided without much hassle.

count-infinity:

Fixes #3847 - CRLF breaks batch code #INSERT

Normalizes the batchcode input to change windows CRLF to just LF, or singular \r to \n. 

Close #3847

### DIFF
--- a/evennia/utils/batchprocessors.py
+++ b/evennia/utils/batchprocessors.py
@@ -235,6 +235,9 @@ def read_batchfile(pythonpath, file_ending=".py"):
     if not text and decoderr:
         raise UnicodeDecodeError("\n".join(decoderr), bytearray(), 0, 0, "")
 
+    if text:
+        text = text.replace("\r\n", "\n").replace("\r", "\n")
+
     return text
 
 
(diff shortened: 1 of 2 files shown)


Griatch (approved):
Clean fix, with unit tests, nice!

PR #3856, opened 2025-12-30
Add alias "look at" to look command.

pierremonteux:
Added alias to look command of "look at"

Motivation for doing this is twofold.  I have years of playing a mud where the command "look at" was required to look at something.  I am constantly correcting my own usage of "look at" to "look" on two different MUDs now, one of which I am not a developer on.

### DIFF
--- a/evennia/commands/default/general.py
+++ b/evennia/commands/default/general.py
@@ -69,7 +69,7 @@ class CmdLook(COMMAND_DEFAULT_CLASS):
     """
 
     key = "look"
-    aliases = ["l", "ls"]
+    aliases = ["l", "ls", "look at"]
     locks = "cmd:all()"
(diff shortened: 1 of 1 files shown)


InspectorCaracal:
I can't say whether or not Griatch might merge this anyway, but this kind of thing is usually left up to the game developer to add to their game if they want it. The command system in Evennia is very flexible that way and designed with the expectation that the devs will customize things like this.

If one of those games is an evennia game you're a developer on, you would do it something like this:

```py
# in a file of your choice
from evennia.commands.default.general import CmdLook

# light custom overide to add look at alias
class CmdLookAt(CmdLook):
    aliases = ['l', 'ls', 'look at']
```

And then you just import and add that to the CharacterCmdSet in default_cmdsets.py

PR #3865, opened 2026-02-15
Add WebSocket subprotocol negotiation per MUD Standards proposal

daiimus:

Implements RFC 6455 `Sec-WebSocket-Protocol` negotiation for Evennia's WebSocket server, adding support for three standard subprotocols from the [MUD Standards WebSocket specification](https://mudstandards.org/websocket/) alongside Evennia's existing format:

- **`json.mudstandards.org`** — JSON envelope format (`{"proto", "id", "data"}`) with ANSI text in BINARY frames and structured OOB in TEXT frames
- **`gmcp.mudstandards.org`** — GMCP strings in TEXT frames, raw ANSI in BINARY frames
- **`terminal.mudstandards.org`** — Raw ANSI/UTF-8 in BINARY frames, no OOB channel
- **`v1.evennia.com`** — Evennia's existing JSON array format (unchanged behavior)

Architecture uses a strategy pattern: `WebSocketClient` delegates to a `WireFormat` instance selected during `onConnect()`. Wire formats live in a `wire_formats/` package — adding a new format means adding one file with zero changes to `webclient.py`.

Additional changes:
- Extracts shared GMCP encode/decode logic into `gmcp_utils.py` (used by both telnet and websocket GMCP paths, reducing ~90 lines of duplication in `telnet_oob.py`)
- Adds `WEBSOCKET_SUBPROTOCOLS` setting for server operators to control which formats third-party clients can negotiate
- Built-in JS webclient now sends `Sec-WebSocket-Protocol: v1.evennia.com` to properly identify itself (previously sent no subprotocol header)
- Comprehensive test suite (~870 lines) covering all wire formats, round-trips, edge cases, and negotiation logic

**Backward compatibility:** Clients that send no `Sec-WebSocket-Protocol` header (including the current Evennia webclient) automatically get the `v1.evennia.com` format. Existing behavior is fully preserved — this is a zero-breaking-change addition.

MUD clients that speak standard WebSocket subprotocols (Mudlet's upcoming WebSocket mode, third-party web clients, bespoke MUD clients) currently cannot connect to Evennia's WebSocket endpoint in a standards-compliant way. The [MUD Standards WebSocket proposal](https://mudstandards.org/websocket/) defines a common set of subprotocols that multiple MUD servers and clients are converging on. Adding this support makes Evennia interoperable with the broader MUD ecosystem out of the box, without requiring game developers to write custom protocol code.

The `telnet.mudstandards.org` subprotocol was deliberately omitted — it requires tunneling a full telnet byte stream over WebSocket, which adds substantial complexity for marginal benefit given that Evennia already has a native telnet server.

- MUD Standards WebSocket specification: https://mudstandards.org/websocket/
- The GMCP refactor into `gmcp_utils.py` also fixes a latent bug where incoming `Core.Supports.Get` GMCP messages were routed to `supports_get` (not a registered inputfunc) instead of `client_options`. The shared `decode_gmcp` now uses the reverse mapping correctly.
- `telnet_oob.py` is modified only to delegate to the shared `gmcp_utils` — all GMCP encoding/decoding behavior is preserved.

#### Disclosure

This PR was developed with significant AI assistance (Claude via OpenCode). All code, tests, and documentation were reviewed by a human, but we want to be transparent about the process. If anything looks off or doesn't fit Evennia's conventions, please don't hesitate to point it out — we're happy to iterate.

Inside every WebSocket server there are five wolves: `json.mudstandards.org`, `gmcp.mudstandards.org`, `terminal.mudstandards.org`, `v1.evennia.com`, and `telnet.mudstandards.org` — though that last one is chained up in the yard because it wouldn't stop trying to tunnel a full telnet session through the fence.

### DIFF
--- a/evennia/server/portal/gmcp_utils.py
+++ b/evennia/server/portal/gmcp_utils.py
@@ -1,0 +1,143 @@
+"""
+Shared GMCP (Generic MUD Communication Protocol) utilities.
+
+This module provides encoding and decoding functions for GMCP messages,
+shared between telnet OOB and WebSocket wire format implementations.
+
+GMCP messages follow the format: "Package.Subpackage json_payload"
+
+The mapping dictionaries translate between Evennia's internal command
+names and standard GMCP package names.
+"""
+
+import json
+
+from evennia.utils.utils import is_iter
+
+# Mapping from Evennia internal names to standard GMCP package names.
+EVENNIA_TO_GMCP = {
+    "client_options": "Core.Supports.Get",
+    "get_inputfuncs": "Core.Commands.Get",
+    "get_value": "Char.Value.Get",
+    "repeat": "Char.Repeat.Update",
+    "monitor": "Char.Monitor.Update",
+}
+
+# Reverse mapping from GMCP package names to Evennia internal names.
+GMCP_TO_EVENNIA = {v: k for k, v in EVENNIA_TO_GMCP.items()}
+
+
+def encode_gmcp(cmdname, *args, **kwargs):
+    """
+    Encode an Evennia command into a GMCP message string.
+
+    Args:
+        cmdname (str): Evennia OOB command name.
+        *args: Command arguments.
+        **kwargs: Command keyword arguments.
+
+    Returns:
+        str: A GMCP-formatted string like "Package.Name json_data"
+
+    Notes:
+        GMCP messages are formatted as:
+            [cmdname, [], {}]          -> Cmd.Name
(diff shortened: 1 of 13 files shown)


daiimus:
@Griatch For sure. Let me know if you have a preference on one of the potential websocket standard approaches. The backwards compatibility for current web client and updated JSON seem to be the most straight forward, but I kinda like the idea of throwing them all out there and letting the dev decide.

Issue #3874, opened 2026-03-01
[BUG] Clientside (Mudlet) command stacking causing disconnect


aMiss-aWry:
#### Describe the bug
On Mudlet 4.20.0, if you send stacked commands the connection breaks with an error logged in `portal.log`:
```
|Portal| 2026-03-01 23:27:36 [!!] Unhandled Error
	Traceback (most recent call last):
	  File "/Users//.pyenv/versions/3.11.0/envs/venv/lib/python3.11/site-packages/twisted/internet/selectreactor.py", line 165, in _doReadOrWrite
	    why = getattr(selectable, method)()
	  File "/Users//.pyenv/versions/3.11.0/envs/venv/lib/python3.11/site-packages/twisted/internet/tcp.py", line 250, in doRead
	    return self._dataReceived(data)
	  File "/Users//.pyenv/versions/3.11.0/envs/venv/lib/python3.11/site-packages/twisted/internet/tcp.py", line 255, in _dataReceived
	    rval = self.protocol.dataReceived(data)
	  File "/Users//Desktop/muddev/evennia/evennia/server/portal/telnet.py", line 105, in dataReceived
	    super().dataReceived(data)
	  File "/Users//.pyenv/versions/3.11.0/envs/venv/lib/python3.11/site-packages/twisted/conch/telnet.py", line 599, in dataReceived
	    self.applicationDataReceived(b"".join(appDataBuffer))
	  File "/Users//Desktop/muddev/evennia/evennia/server/portal/telnet.py", line 336, in applicationDataReceived
	    if len(data) > 2 and _HTTP_REGEX.match(d):
	builtins.TypeError: cannot use a string pattern on a bytes-like object
```

#### To Reproduce
1. Download Mudlet
2. Connect to fresh Evennia instance
3. Enter `test;;test`
4. See connection fail

#### Expected behavior
Stacked commands should execute as expected

#### Environment, Evennia version, OS etc
```
    Evennia 6.0.0 (rev 3761a7cb2)
    OS: posix
    Python: 3.12.10
    Twisted: 24.11.0
    Django: 6.0.2
```

#### Additional context
Looks like this is because Mudlet sends stacked commands as bytes instead of strings. Should be fixable in telnet.py

PR #3875, opened 2026-03-02
Fix TypeError in telnet HTTP check when command stacking sends bytes

aMiss-aWry:
Normalize bytes for _HTTP_REGEX match where needed. Added test.

Prevent clientside (Mudlet) command stacking from causing disconnect

Closes https://github.com/evennia/evennia/issues/3874

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,7 +75,7 @@ dependencies = [
   "inflect >= 5.2.0",
   "autobahn >= 20.7.1, < 21.0.0",
   "lunr == 0.7.0.post1",
-  "simpleeval == 1.0.3",
+  "simpleeval == 1.0.5",
   "uritemplate == 4.1.1",
   "tzdata >= 2022.6",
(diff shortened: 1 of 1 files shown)


evennia/server/portal/telnet.py
```diff
@@ -330,8 +330,10 @@ def applicationDataReceived(self, data):
             data = [_IDLE_COMMAND]
         else:
             data = _RE_LINEBREAK.split(data)
-
-            if len(data) > 2 and _HTTP_REGEX.match(data[0]):
+            # Normalize to bytes for regex match if needed.
+            if len(data) > 2 and _HTTP_REGEX.match(
+                data[0].encode("utf-8", errors="replace") if isinstance(data[0], str) else data[0]
+            ):
```
Griatch, line 336: This is unnecessary; the data is always bytes at this point (after the regex fix). I'll merge this, but clean out this bit I think.

PR #3881, opened 2026-03-03
Refactor connection commands to use deferToThread for password hashing

### LINKED ISSUE #3879: [Feature Request] Speed Up Account.create / Account.authenticate
I'm just going to copy/paste from the Discord:

[5:54 PM]Uncle Buff
: Hey everyone! I've been spending some time identifying a number of bottlenecks in Evennia that I'm wondering if anyone else has experienced and attempted to rectify. These bottlenecks manifest in the form of significant lag to my entire game, so I consider them pretty problematic. Figured I'd share my findings thus far in hopes that others might both find them useful and have solutions.

There are two primary points of lag that stick out to me regularly. The first is when a user is attempting to authenticate, create an account, or when an account password is changed. I did some benchmarking and have isolated the biggest contributor of this to be the password hasher. At the default value of 1200000 iterations, it takes anywhere from 0.47S to 0.93S to complete the process (tested on both Windows and Linux systems). This doesn't include the roundtrip of actually logging in and hitting the database, so the end result tends to be at least a full second of lag for everyone on the game when someone performs a password-related operation. My initial test was to lower the number of iterations to 200000, which significantly reduced the operation time to 0.15S. However, after deciding I would rather not decrease password security if possible, I switched Evennia's password hasher to use argon2, which at its default reasonable settings benchmarked operations between 0.05S and 0.13S. This is more than reasonable to me, so if anyone else has noticed this sort of lag when engaging in password-related shenanigans, this will likely solve the problem (ensure you keep the default hasher at the bottom of your PASSWORD_HASHERS setting so that older accounts can still log in and will be automatically upgraded to argon2 upon doing so).

jaborsh:
Addresses lag spikes caused by sync password hashing. `PBKDF2PasswordHasher` is intentionally expensive for security reasons; a single call to `Account.authenticate` or `Account.create` blocks the main thread for 100ms-300ms. During this time, the entire server freezes for all players.

Using `yield` and Twisted Deferreds, we can offload hashing to a background worker thread. This resolves login lag spikes.

```
Sync Auth Time:  0.14147s
Async Auth Time: 0.00001s
```

Closes #3879

### DIFF
--- a/evennia/commands/cmdhandler.py
+++ b/evennia/commands/cmdhandler.py
@@ -36,7 +36,7 @@ from traceback import format_exc
 from django.conf import settings
 from django.utils.translation import gettext as _
 from twisted.internet import reactor
-from twisted.internet.defer import inlineCallbacks
+from twisted.internet.defer import Deferred, inlineCallbacks
 from twisted.internet.task import deferLater
 
 from evennia.commands.cmdset import CmdSet
@@ -251,6 +251,9 @@ def _progressive_cmd_run(cmd, generator, response=None):
             utils.delay(value, _progressive_cmd_run, cmd, generator)
         elif isinstance(value, str):
             _GET_INPUT(cmd.caller, value, _process_input, cmd=cmd, generator=generator)
+        elif isinstance(value, Deferred):
+            value.addCallback(lambda result: _progressive_cmd_run(cmd, generator, response=result))
+            value.addErrback(lambda fail: generator.throw(fail.type, fail.value, fail.tb))
         else:
             raise ValueError("unknown type for a yielded value in command: {}".format(type(value)))
 


count-infinity:
@jaborsh 
That did resolve the ValueError issue.

But new account creation is actually much slower now when using evennia out of the box.  ~4s account creation with deferToThread vs less than a second without. 

Are you testing against sqlite?

```bash
Main branch (no deferToThread):
total: 0.741s
create_account: 0.599s   (mostly password hashing - 0.578s in pbkdf2_hmac)
create_character: 0.073s
join_channels: 0.021s
...
PR branch (with deferToThread):

total: 4.081s
create_character: 1.555s
join_channels: 1.055s
create_account: 0.755s
sqlite3.Connection.commit: 2.657s  <-- !!!
```

I think sqlite has issues with multiple threads connecting? This could just be my machine though:
    Evennia 6.0.0 (rev 3761a7cb2)
    OS: posix
    Python: 3.13.0
    Twisted: 24.11.0
    Django: 6.0.3

Griatch:
This is very interesting, and also a bit surprising to me - because the normal wisdom is that thread operations that are CPU limited (like this one is) *shouldn't* improve performance due to the Python GIL. The normal way to handle it is normally by using a subprocess. 

I think the only reason this works is because the hasher is a C (or Fortran) library to goes beyond the GIL. You cannot offload generic CPU-bound Python code this way - the GIL would just block both threads.

I agree that the ability to inject a custom deferred in the command execution is useful though - at least it's what makes it work for this very specific use case. I just worry that people may think it's a valid way to parallelize any heavy code - but that's not a blocker for adding it.

Anyway, SQlite is fast for some things and slow for others. In this particular case, I do think it makes sense to investigate if one can offload only the hash operation itself (it's the only thing actually helped by it after all), without it needing to do any ORM operations. 

That said, if moving to argon2 is so considerably faster (as mentioned in #3879), it may be worth to also default to it - that might resolve this for most new games at least.

PR #3888, opened 2026-03-07
fix: close stale DB connections before first query

### LINKED ISSUE #3887: [BUG] Python 3.14 MemoryError with SQLite forks
#### Describe the bug
When running Evennia with Py 3.14, the server crashes on startup with a MemoryError. The crash occurs because twisted daemonizes by forking the process and the child inherits SQLite connections opened during `django.setup()`. Py 3.14's `sqlite3` is not fork-safe, so reusing these connections raises `MemoryError`.

#### To Reproduce
1. Create new Evennia game directory
2. `cd newgame`
3. `evennia migrate`

#### Expected behavior
No errors.

Evennia 6.0.0
Python 3.14.0

jaborsh:
When twisted daemonizes via fork, the child process inherits SQLite connections from the parent. Python 3.14's sqlite3 module is not fork-safe and raises MemoryError when these inherited connections are reused.

Closes #3887

### DIFF
--- a/evennia/server/service.py
+++ b/evennia/server/service.py
@@ -9,6 +9,7 @@ import traceback
 
 import django
 from django.conf import settings
+import django.db
 from django.db import connection
 from django.db.utils import OperationalError
 from django.utils.translation import gettext as _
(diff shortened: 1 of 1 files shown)


aMiss-aWry:
By the way, if anyone still has their postgres setup from the old docs - and has `pip show psycopg2-binary` - and is finding that `evennia` hangs when you try to use it now - and bisected their way to finding this commit as the apparent problem - all you have to do is:
```
pip uninstall psycopg2-binary
pip install "psycopg[binary]"
```
psycopg2 is an old driver. Looks like 2 and 3 handle safety differently, and 2 was just waiting indefinitely for a response that would never come.

And remember to update your `requirements.txt`.

PR #3899, opened 2026-03-29
Fix @teleport showing double error message on failed search

blongden:

CmdTeleport.parse() called caller.search() without quiet=True, so a failed search would show Evennia's default "Could not find X" message. Then func() would show a second "Destination not found." message.

Now uses quiet=True and extracts [0] from the results list, giving a single clean error message from func().

No double error message


Griatch (approved):
I guess this was not strictly a bug, but maybe it gave unnecesary information. Approved.

PR #3900, opened 2026-03-29
Fix TypeError in MSDP encode when args contain non-string values

blongden:

Wraps args[0] in str() before string concatenation in encode_msdp. Without this, MSDP clients sending numeric values cause a TypeError on the += operation.

Avoids a crash when MDSP clients send numeric values


Griatch (approved):
Nice fix. Thanks!

PR #3901, opened 2026-03-29
Fix O(N²) cmdset merge by hashing commands on key instead of constant

blongden:

Command.__hash__ returned hash("command") for all instances, putting every command in the same hash bucket. This made set() operations in CmdSet._union O(N) per lookup instead of O(1). With large command sets (700+ commands), a single merge took ~1000ms.

Now hashes on self.key. This isn't perfectly consistent with __eq__ (which uses _matchset intersection - two commands sharing only an alias would be equal but have different hashes), but that edge case is extremely rare in practice and the performance difference is massive for games with large command sets.

Benchmarked: cmdset merge drops from ~1000ms to <50ms with 700 commands.

Games with any serious number of command suffer crazy performance lag, constantly spinning on cmdset merge. This will speed up all evennia games, especially those with more than a handful of commands.

### DIFF
--- a/evennia/commands/command.py
+++ b/evennia/commands/command.py
@@ -266,12 +266,14 @@ class Command(metaclass=CommandMeta):
         implement __hash__ and that the corresponding hashes for equivalent
         instances are themselves equivalent.
 
-        Technically, the following implementation is only valid for comparison
-        against other Commands, as our __eq__ supports comparison against
-        str, too.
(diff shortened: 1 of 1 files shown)


Griatch (approved):
The performance improvement makes this worth it, but it will be interesting to see what effect this actually has on existing games - if anyone depends on the alias matching for equalization determination (such as when overriding commands), this does break that.

PR #3902, opened 2026-04-01
Fix telnet disconnect when multiple commands arrive in one packet

blongden:

The HTTP detection regex in applicationDataReceived uses a string pattern (r"...") but the data it matches against is bytes. When 3+ commands arrive in a single TCP packet (e.g. rapid input or semicolon-separated commands from a MUD client), the data splits into 3+ parts, triggering the len(data) > 2 guard. The subsequent _HTTP_REGEX.match(data[0]) then crashes with:

    TypeError: cannot use a string pattern on a bytes-like object

This unhandled exception kills the telnet connection.

Fix: change the regex from a string pattern (r"...") to a bytes pattern (rb"...") to match the bytes data it receives.

Fix a crash bug

### DIFF
--- a/evennia/server/portal/telnet.py
+++ b/evennia/server/portal/telnet.py
@@ -44,7 +44,7 @@ _IDLE_COMMAND = str.encode(settings.IDLE_COMMAND + "\n")
 
 # identify HTTP indata
 _HTTP_REGEX = re.compile(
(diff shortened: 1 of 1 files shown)


count-infinity:
There are a few existing PR fixes out there already it seems. 
It looks like this duplicates #3875

Griatch:
Thanks for the PR, closed since it's a duplicate of fix in #3875 (and a few others) :grin:

PR #3904, opened 2026-04-01
Fix TypeError in telnet HTTP detection regex

tezafofeza:
## Summary

- `_HTTP_REGEX` uses a string pattern (`r"..."`) but `applicationDataReceived` passes bytes data from `_RE_LINEBREAK.split()`, causing `TypeError: cannot use a string pattern on a bytes-like object` when a scanner/bot sends an HTTP request to the telnet port
- Fix restores the bytes prefix (`rb"..."`) that was present in earlier versions

## Steps to reproduce

1. Start Evennia with telnet port exposed
2. Send an HTTP request to the telnet port (e.g. `curl http://localhost:4000`)
3. Portal crashes with `TypeError` at `telnet.py:334`

## Test plan

- [ ] Verify telnet connections still work normally
- [ ] Verify HTTP requests to telnet port are rejected gracefully instead of crashing the portal

### DIFF
--- a/evennia/server/portal/telnet.py
+++ b/evennia/server/portal/telnet.py
@@ -44,7 +44,7 @@ _IDLE_COMMAND = str.encode(settings.IDLE_COMMAND + "\n")
 
 # identify HTTP indata
 _HTTP_REGEX = re.compile(
-    r"(GET|HEAD|POST|PUT|DELETE|TRACE|OPTIONS|CONNECT|PATCH) (.*? HTTP/[0-9]\.[0-9])", re.I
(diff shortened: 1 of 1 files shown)


count-infinity:
Dang, this PR looks like duplicate of PR (https://github.com/evennia/evennia/pull/3902).  That one is only a few hours sooner than this one.

jaborsh:
> Dang, this PR looks like duplicate of PR (https://github.com/evennia/evennia/pull/3902).  That one is only a few hours sooner than this one. 

https://github.com/evennia/evennia/pull/3875 - a month ago.

Griatch:
Thanks for the PR, closed since it's a duplicate of fix in #3875 (and a few others) :grin:

PR #3905, opened 2026-04-05
Fix system commands duplicated in merged CmdSets

kvmet:
I was encountering an issue with EvMore where the pager would consistently wind up having duplicates of all of its commands. After a bunch of poking around I determined this is because CmdSets merge was picking up system commands from multiple locations and duplicating them. 

First commit worked but I didn't like that it was using a string compare so I refactored it to a more elegant fix. 

---

## Problem

When a CmdSet using `mergetype = "Replace"` or `"Union"` is merged onto another set that contains system commands (keys starting with `__`), those system commands appear twice in the result. The most visible symptom is with EvMore: after the pager is displayed, any input produces:

```
More than one match for 'q' (please narrow target):
 q-1 [n;end;a;next;p;abort;previous;quit;t;top;q;e]
 q-2 [n;end;a;next;p;abort;previous;quit;t;top;q;e]
```

## Root cause

`_replace` and `_union` both copy `cmdset_a.commands[:]` verbatim into `cmdset_c.commands`. Because `CmdSet.add()` appends system commands to both `self.commands` and `self.system_commands`, that raw slice includes them. `CmdSet.__add__` then unconditionally calls `cmdset_c.add(sys_commands, allow_duplicates=True)` to re-add the correctly-merged system commands so they end up in `cmdset_c.commands` twice. `cmdparser.build_matches` then finds two identical matches and reports the ambiguity error.

`_intersect` and `_remove` build their results differently and are not affected.

## Fix

Rather than filtering in each merge method, the deduplication is applied in `__add__` at the exact point of re-adding. Strip any system commands already present in `cmdset_c.commands` before calling `add(sys_commands, allow_duplicates=True)`. This uses `Command.__eq__`/`__hash__` rather than string prefix matching, keeps the merge methods untouched, and automatically covers any future merge type.

## Tests

Added two regression tests to `TestCmdSetMergers` covering system command deduplication after both Replace and Union merges, with system commands present on either side.

### DIFF
--- a/evennia/commands/cmdset.py
+++ b/evennia/commands/cmdset.py
@@ -243,13 +243,17 @@ class CmdSet(object, metaclass=_CmdSetMeta):
 
         """
         cmdset_c = cmdset_a._duplicate()
-        # we make copies, not refs by use of [:]
-        cmdset_c.commands = cmdset_a.commands[:]
+        # Exclude system commands; __add__ re-adds them after merging.
+        cmdset_c.commands = [c for c in cmdset_a.commands if not c.key.startswith("__")]
         if cmdset_a.duplicates and cmdset_a.priority == cmdset_b.priority:
-            cmdset_c.commands.extend(cmdset_b.commands)
+            cmdset_c.commands.extend(
+                [c for c in cmdset_b.commands if not c.key.startswith("__")]
+            )
         else:
             existing_commands = set(cmdset_a.commands)
-            cmdset_c.commands.extend([cmd for cmd in cmdset_b if cmd not in existing_commands])
+            cmdset_c.commands.extend(
+                [cmd for cmd in cmdset_b if cmd not in existing_commands and not cmd.key.startswith("__")]
+            )
         return cmdset_c
 
     def _intersect(self, cmdset_a, cmdset_b):
@@ -295,7 +299,8 @@ class CmdSet(object, metaclass=_CmdSetMeta):
 
         """
         cmdset_c = cmdset_a._duplicate()
-        cmdset_c.commands = cmdset_a.commands[:]
+        # Exclude system commands; __add__ re-adds them after merging.
+        cmdset_c.commands = [c for c in cmdset_a.commands if not c.key.startswith("__")]
         return cmdset_c
 
     def _remove(self, cmdset_a, cmdset_b):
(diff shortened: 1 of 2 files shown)


kvmet:
P.S. I have a monkey patch for this if anyone else is in a spot where they want to fix it before a fix makes it into main.

in `/server/conf/at_server_startup.py`:

```python
def at_server_init():
    """
    This is called first as the server is starting up, regardless of how.
    """
    _patch_cmdset_sys_cmd_dedup()

def _patch_cmdset_sys_cmd_dedup():
    """
    Fix Evennia bug: CmdSet._replace and _union both copy cmdset_a.commands[:] into
    cmdset_c, which includes system commands. CmdSet.__add__ then calls
    cmdset_c.add(sys_commands, allow_duplicates=True), appending them again — no dedup.
    Result: typing "q" to exit EvMore yields "More than one match for 'q'".

    Fix: mirrors the upstream fix exactly — strip system commands from cmdset_c.commands
    in __add__ before re-adding, using Command.__eq__/__hash__ rather than string prefix
    matching.

    Remove this patch once the upstream fix is pulled in.
    """
    import functools
    from evennia.commands.cmdset import CmdSet

    original_add = CmdSet.__add__

    @functools.wraps(original_add)
    def _patched__add__(self, cmdset_a):
        if not cmdset_a:
            return original_add(self, cmdset_a)
        # Mirror the priority logic from __add__ to reconstruct sys_commands before
        # calling the original, since it's a local variable we can't access after.
        sys_a = cmdset_a.get_system_cmds()
        sys_b = self.get_system_cmds()
        if self.priority <= cmdset_a.priority:
            sys_commands = sys_a + [c for c in sys_b if c not in sys_a]
        else:
            sys_commands = sys_b + [c for c in sys_a if c not in sys_b]
        # original_add will add sys_commands once via add(sys_commands, allow_duplicates=True).
        # The raw commands[:] copy in the merge methods may have already included them,
        # so the result could have duplicates. Strip and re-add exactly once.
        cmdset_c = original_add(self, cmdset_a)
        sys_set = set(sys_commands)
        cmdset_c.commands = [c for c in cmdset_c.commands if c not in sys_set]
        cmdset_c.add(sys_commands, allow_duplicates=True)
        return cmdset_c

    CmdSet.__add__ = _patched__add__

```

Griatch (approved):
Looks like a solid fix, and also has tests. Nice!

PR #3906, opened 2026-04-06
Omkar first issue

omkarbasnet1:

### DIFF
--- a/evennia/utils/tests/test_utils.py
+++ b/evennia/utils/tests/test_utils.py
@@ -25,6 +25,31 @@ class TestIsIter(TestCase):
         self.assertEqual(True, utils.is_iter([1, 2, 3, 4]))
         self.assertEqual(False, utils.is_iter("This is not an iterable"))
 
+class TestMakeIter(TestCase):
+    def test_non_iterable_to_list(self):
+        """ Single non-iterable val should be wrapped in list """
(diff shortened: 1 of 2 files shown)


evennia/utils/tests/test_utils.py
```diff
@@ -54,6 +82,95 @@ def test_dedent(self):
         self.assertEqual(expected_string, utils.dedent(input_string))
 
 
+from django.test import TestCase
+from evennia.utils.utils import class_from_module
```
Griatch, line 86: This should not be re-imported again in the middle of the module.

Griatch (changes requested):
Thanks for contributing! Based on emails, was this some sort of class contribution? :) You must write a better title and description when making a PR, or it easily looks like a mistake posting.

Alas, you are testing functionality that already have extensive tests, and there are  code quality issues here too, like re-importing in the middle of tests. This can't merge.

The `TestClassFromModule` test does add new testing, so that could be useful, but the rest appear to be duplicates, alas. So please update the PR to only include the code adding to the code coverage, fixing the python code.

Griatch (changes requested):
Sorry, you cannot keep adding new fixes to the same branch. Make separate branches/PRs for your features. Since there is no sign of responding to feedback, I'm closing this PR - please open new, separate PRs for these fixes.

PR #3909, opened 2026-04-08
Fix funcparser treating $" as malformed function call

### LINKED ISSUE #3907: [BUG] Funcparser treats $" as malformed func
#### Describe the bug
Image

#### To Reproduce
1. Have inline funcs enabled
2. Message yourself '$"x"'

#### Expected behavior
As a non-func it should be parsed as plain text.

```
         Evennia MU* development system

         Evennia version: 6.0.0 (rev 45ce2e2cc9)
         OS: posix
         Python: 3.12.10
         Twisted: 24.11.0
         Django: 6.0.2
```

aMiss-aWry:
Added test that $"x" gets parsed as a non-func
Added look-ahead in FuncParser.parse() to check viability of func

Bug fix

Closes https://github.com/evennia/evennia/issues/3907

I don't think this is the best way, but it was the quick fix that worked for me. I do mildly wonder if regex would be better.

### DIFF
--- a/evennia/utils/funcparser.py
+++ b/evennia/utils/funcparser.py
@@ -352,8 +352,16 @@ class FuncParser:
                 continue
 
             if char == start_char:
+                # peek ahead - if next char can't start a funcname, treat $ as literal
+                next_char = string[ichar + 1 : ichar + 2]
+                if next_char and not (next_char.isalpha() or next_char == "_"):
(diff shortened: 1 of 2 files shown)


Griatch (approved):
This looks good, and I think this look-ahead approach is better than a regex.  And the test gives additional confidence. Approved!

PR #3910, opened 2026-04-08
Fix MXP parse to not escape surrounding text as HTML entities

### LINKED ISSUE #3860: [BUG] Mudlet 4.20.0 clickable link tags parsed as raw text in EvMenu
#### Describe the bug
Clickable link tags (|lc) are appearing as raw text with Mudlet 4.20.0. Something to the effect of

```py
 <SEND HREF="1">1: Give the darkness a Tongue. (Enable Screenreader)  
 <SEND HREF="2">2: Give the darkness a Shape. (Disable Screenreader)
```

#### To Reproduce
1. Create EvMenu with keys between nodes
2. Enter menu
3. See error

#### Expected behavior
Should not see <SEND HREF=

- Mudlet 4.20.0
- Evennia 5.0.1

#### Additional context
Is not due to #3859.

aMiss-aWry:
THIS WAS A FUN ONE.

mxp_parse previously HTML-escaped all &, <, and > characters in non-MXP text portions of a message.

This caused literal angle brackets surrounding MXP tags (e.g. `<|lchelp eat|lthelp eat|le>`) to be sent to the client as `&lt;` and `&gt;` ~~(funny story, originally this PR said < and > instead of < and >, I'm sure you can guess how that happened)~~ instead of `<` and `>`, since MXP-enabled clients do not decode HTML entities outside of MXP tag contexts.

The fix removes the HTML escaping of non-MXP content entirely. MXP clients only process tags within \x1b[4z secure mode markers, so surrounding plain text is treated as literal text regardless of angle brackets.

A test is added to TestTelnet covering correct link substitution, passthrough of surrounding angle brackets, and the early-return path for messages containing no MXP markup.

image
->
image

Fixes https://github.com/evennia/evennia/issues/3860#issue-3898585173

### DIFF
--- a/evennia/server/portal/mxp.py
+++ b/evennia/server/portal/mxp.py
@@ -32,19 +32,48 @@ MXP_URL = MXP_TEMPSECURE + '<A HREF="\\1">' + "\\2" + MXP_TEMPSECURE + "</A>"
 
 def mxp_parse(text):
     """
-    Replaces links to the correct format for MXP.
+    Parse Evennia's MXP link markup into MXP escape sequences suitable for
+    sending to MXP-enabled clients.
+
+    Converts ``|lc<cmd>|lt<label>|le`` to a clickable SEND tag and
+    ``|lu<url>|lt<label>|le`` to a clickable URL tag. Non-MXP content
+    has ``&``, ``<``, and ``>`` HTML-escaped to prevent them from being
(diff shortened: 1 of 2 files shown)


Griatch (approved):
I think this identifies the right place to update, not sure we need all the extra complexity of replacing though. I'll merge this and ponder some more.

Issue #3807, opened 2025-08-19
[Feature Request] Enter a brief description here


blongden:
#### Is your feature request related to a problem? Please describe.

Unsure that this stage if this is a bug or a feature request. I am working on upgrading an old evennia 0.9.5 project to 5.0.1. It's gone relatively smoothly so far however I noted 5-6s of lag on every command by users with a high number of cmdsets. After some debugging it emerges that the issue is two fold. One, the WeakValueDictionary used to cache the merged command set always seems to be empty, resulting in the full merged command set being recaculated every time, and two, the merged command set is stored as a single blob of data, which includes the exit objects in each room which contain custom cmdsets. This means the whole merged command set is stored with every combination of exits for the whole evennia world.

#### Describe the solution you'd like

I have prepared a patch which fixes this issue by splitting the cache into exits, which will cache per room, and everything else. This seems to work nicely.

I have not yet investigated what is going on with the WeakValueDictionary, but replacing is with a regular python dictionary works fine in the meantime (though appreciate it wont clear things out the cache if things are deleted.

PR #3911, opened 2026-04-09
Fix/cmdset merge cache performance

kvmet:
<h2>Problem</h2>
<p>The merge cache in <code>cmdhandler.get_and_merge_cmdsets</code> used Python object identity (<code>id()</code>) as the cache key. Since cmdset objects are recreated on each call, the cache never produced a hit. For players in rooms with many exits or objects carrying custom cmdsets this resulted in a <em>ton</em> of merge operations. O(N).</p>
<h2>Changes</h2>
<p><strong><code>evennia/commands/cmdset.py</code></strong></p>
<p>Added a <code>fingerprint</code> property to <code>CmdSet</code>. It lazily computes a hashable tuple from the cmdset's semantic content (key, priority, mergetype, options, and the sorted matchsets of all commands), caches it on the instance, and invalidates it when commands are added or removed via <code>add()</code> or <code>remove()</code>. This means the fingerprint is computed once per cmdset per mutation rather than on every command input.</p>
<p><strong><code>evennia/commands/cmdhandler.py</code></strong></p>
<ul>
<li>Replaced <code>id()</code>-based cache keys with <code>tuple(cmdset.fingerprint for cmdset in cmdsets)</code>. Two calls with semantically identical cmdsets now share a cache entry regardless of object identity.</li>
<li>Replaced the plain <code>dict</code> with a bounded <code>OrderedDict</code> (max 1000 entries) with LRU eviction, preventing unbounded memory growth.</li>
</ul>
<h2>Performance</h2>
<p>Benchmarked against a simulated scenario of 33 cmdsets (515 total commands):</p>

  | Time | vs full merge
-- | -- | --
Full merge (no cache) | 36.3 ms | baseline
Fingerprint, cold (first compute) | 0.15 ms | 242x faster
Fingerprint, hot (cached on CmdSet) | 0.002 ms | 18,150x faster

[ EDIT: I think it is _only_ 500x best case due to erroneous cache hits pointed out by Griatch :'(. ]

<p>On the hot path the cache lookup is essentially free.</p>
<h2>Tests</h2>
<p>Added <code>TestGetAndMergeCmdSets.test_merge_cache_hit</code> which verifies that two calls with the same logical cmdsets produce a cache hit (no growth in cache size, same returned object identity).</p></body></html>

# Motivation for adding to Evennia

Cache was implemented but basically non-functional. This should be closer to the original intent.

# Other info (issues closed, discussion etc)

Addresses issue #3807. (Plus should just be a very nice performance increase)

### DIFF
--- a/evennia/commands/cmdhandler.py
+++ b/evennia/commands/cmdhandler.py
@@ -28,7 +28,7 @@ command line. The processing of a command works as follows:
 """
 
 import types
-from collections import defaultdict
+from collections import OrderedDict, defaultdict
 from copy import copy
 from itertools import chain
 from traceback import format_exc
@@ -48,7 +48,37 @@ _IN_GAME_ERRORS = settings.IN_GAME_ERRORS
 
 __all__ = ("cmdhandler", "InterruptCommand")
 _GA = object.__getattribute__
-_CMDSET_MERGE_CACHE = {}
+_CMDSET_MERGE_CACHE = OrderedDict()
+_CMDSET_MERGE_CACHE_MAXSIZE = 1000
+
+
+def _cmdset_fingerprint(cmdset):
+    """
+    Build a hashable fingerprint from a cmdset's semantic content. Two cmdsets
+    with identical commands and merge properties will produce the same fingerprint,
+    regardless of Python object identity.
+
+    Args:
+        cmdset (CmdSet): The cmdset to fingerprint.
+
+    Returns:
+        tuple: A hashable fingerprint.
+
+    """
+    cmd_ids = tuple(sorted(tuple(sorted(cmd._matchset)) for cmd in cmdset.commands))
+    sys_cmd_ids = tuple(sorted(tuple(sorted(cmd._matchset)) for cmd in cmdset.system_commands))
+    return (
+        cmdset.key,
+        cmdset.priority,
+        cmdset.mergetype,
+        cmdset.duplicates,
+        cmdset.no_exits,
+        cmdset.no_objs,
+        cmdset.no_channels,
+        tuple(sorted(cmdset.key_mergetypes.items())) if cmdset.key_mergetypes else (),
+        cmd_ids,
+        sys_cmd_ids,
+    )
 
 # tracks recursive calls by each caller
 # to avoid infinite loops (commands calling themselves)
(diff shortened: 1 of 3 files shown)


kvmet:
After some more thought, the test I added was too dependent on the cache implementation so I removed it. The existing tests should be good enough for verifying no regression.

Griatch (changes requested):
This is an impressive speedup and good work! 

I think there is a blocker here though: Take this example: 

room1: has ExitCmdSet with exits north/south
room2: has ExitCmdSet with exits north/south 

Both rooms produce exitcmdsets with the same signature, both contain same matchset commands (north/south). 

- When arriving in room1, the fingerprint is built - OK. 
- Moving to room2, the cmdset would match the same fingerprint and return the exit objects of room1, not room2. 

I think what needs to happen here is that the fingerprint must also contain `self.obj` to identify just where the cmdset sits (`self.obj.id` is not enough since cmdsets could sit on Players or Objects, and could thus have colliding dbrefs).

kvmet:
@Griatch  Good catch! I added the object references to the fingerprint. 

Direct object references should keep the objects from getting GCed and crash loudly if unhashable (Should never happen as Evennia `TypedObject` should always be hashable unless something is very wrong).

A couple other small tweaks:
- Switched from sorted tuples to frozensets for command/key_mergetypes collections (order-independent matching). Requires the obj to be hashable but removes requirement to sort.
- Added fingerprint cache invalidation in `make_unique()` Unlikely that this would have mattered but it was another case that I missed before.

P.S. I am getting less speedup now: ~500x best case. It's totally possible that my test conditions were all winding up as erroneous cache hits because of the issue you pointed out, so this is probably a lot more accurate. (I was enjoying the idea of such an extreme speedup! But 500x is still pretty great, lol.)

### New tests:

- **Merge preserves `cmd.obj` bindings**: Build two `cmdset`s with different `cmdsetobj` and different `cmd.obj`. Merge them. Verify each command in the result has the correct `obj` reference.
- **Same-key commands from different objects both survive merge**: Two objects each contribute a command with the same key (like how exits or object commands might collide). Merge the `cmdset`s. Verify the result contains the command from the higher-priority (or last-merged) set, and that its `obj` is correct.
- **Merge after `add()` reflects new state**: Build a `cmdset`, merge it with another. Then `add()` a new command to the first set. Merge again. Verify the new command appears in the result with correct `obj`.
- **Merge after `remove()` reflects new state**: Same as above but `remove()`. Verify it's gone from the merge result.

### Caveats

- The cache returns shared cmdset/command instances. Safe by default: `cmdhandler` copies commands before execution (`copy(cmd)` at line 758). Commands with `retain_instance = True` opt into shared mutable state and already carry a cross-talk warning. The cache doesn't meaningfully change that risk.
- Storing `cmd.obj` and `cmdsetobj` in a frozenset requires them to be hashable. Evennia's `TypedObject` inherits from Django models (integer PK), so this should always hold. An unhashable object here would indicate something else is very wrong.
- On cache hit, `merged_from` still references input cmdsets from the original miss. Considered updating it, but the cached result is a shared object: overwriting `merged_from` on each hit would cause concurrent lookups to clobber each other. Content is equivalent anyway. Left as-is.

blongden:
I'll give this a test too (I raise the original issue that this addresses). I 'fixed' it by splitting the cache into two but I always felt it was the wrong way to do it and I never went back to fixing it. Will let you know how it goes!

Griatch (approved):
This looks good now. Thanks for the contribution!

Issue #3912, opened 2026-04-13
[BUG] Searching tags by category returns no results


llzzies:
#### Describe the bug
When trying to use **tag/search** to find objects with a certain tag category (ex: tag/search :cityhall) it appears to append a space before the category and thus returns no matches.

#### To Reproduce
1. Create a new object.
2. tag object = boxes:furniture
3. tag/search :furniture

#### Expected behavior
I'm sure this worked in previous versions of Evennia and appropriately listed out all objects that matched the tag category rather than returning no matches.

#### Environment, Evennia version, OS etc
Evennia version: 6.0.0 (rev 747e834bf)
         OS: posix
         Python: 3.12.3
         Twisted: 24.11.0
         Django: 6.0.3

#### Additional context
<img width="721" height="85" alt="Image" src="https://github.com/user-attachments/assets/e1063cb8-96f2-406e-b035-efe6c8ab68df" />

Griatch:
Fixed in #3913

PR #3913, opened 2026-04-13
Fix tag/search failing when searching by category only

blongden:

The tag/search command failed to find objects when searching by category alone (e.g. `tag/search :cityhall`) due to two issues:

1. Leading whitespace from command parsing was not stripped, causing the tag key to contain a space
2. An empty tag key was passed as "" instead of None, which searched for tags with an empty string key rather than matching all keys in the given category

Strip self.args and convert empty tag to None so category-only searches work correctly. Added tests for both cases.

### DIFF
--- a/evennia/commands/default/building.py
+++ b/evennia/commands/default/building.py
@@ -3991,10 +3991,11 @@ class CmdTag(COMMAND_DEFAULT_CLASS):
             return
         if "search" in self.switches:
             # search by tag
-            tag = self.args
+            tag = self.args.strip()
             category = None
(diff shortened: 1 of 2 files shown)


blongden:
Fixes #3912

Griatch (approved):
Looking through the history, this bug was actually here from day one, but it was masked by an imprecise handling of 'falsy' input tje Typeclass manager. That imprecise handling was fixed earlier this year, which made this bug become visible as a side effect. So this fix is correct. Thanks!

Issue #3701, opened 2024-12-20
[BUG] (Enter a brief description here)


aMiss-aWry:
#### To Reproduce
1.
2.
3.

#### Additional context

'''

0xDEADFED5:
i kinda like it how it is. devs can utilize screenwidth when necessary (i.e. help command), but otherwise wrapping is left up to the client...it feels like something that should be the client's job? is this something that servers traditionally handle?

i think it would be a cool option, but personally don't think it feels right to have it be enabled by default

MorquinDevlar:
One of the issues with the current implementation is that not all commands adhere to changes made by the player. Commands like examine does not adhere to what the player defines in options.

Before we decide on how/if this should be changed, I'd prefer we spend time making sure all commands behave in a similar way.

But, that being said, I agree with the "problem" of setting something in options, and then the game just ignores it. That seems counter intuitive.

I had massive issues with the web-client, because it did not adhere to the screen-width I defined. In the end I had to modify the actual web client, and hard code a window width limit for the text output, which then would wordwrap the room contents.

This is the same room, with and without wordwrap. I am sure you can see which one would be easier to read while the screen is scrolling fast.
<img width="1903" alt="Image" src="https://github.com/user-attachments/assets/d4e1c8cb-e404-4634-af8e-b60ff950d19f" />

InspectorCaracal:
While I'm generally on the side of "the client should handle it" for things like this, I believe respecting client line length/screen width is a thing servers have historically done for telnet clients, due to the limitations of the clients?

As such, I think that such a value, IF provided by the client (or the user on behalf of the client), should be respected by the server across the board - because it's a client compatability value. It's conceptually the same as stripping MXP from the output of the client doesn't report supporting it.

That said, I also think that the default behavior when such a value was not reported or set should be to not hard-wrap the lines at all, rather than to use a default line length.

MorquinDevlar:
Old school terminal screens were only 80 characters wide, which afaik is where it comes from. Most clients default 100 chars, since it's easier to follow along when you don't have to read a very side text, while at the same time watching shorter messages flowing on the screen.

I use my client to break tings up as well but there are some commands, like examine, where Evennia chooses how much text to include, in a description for instance, based on the setting of the screen width. If I the also break using my client, it looks mess.

Client breaking at 100, options in Evennia set to 80. Good to go.
[...]

PR #3914, opened 2026-04-16
NAWS updates were being checked before super was called. 

michaelfaith84:
Checked for a NAWS updated after we actually had data. Also fixed a bug in the validation input funcs and added tests I should have originally.

Maintaining functionality from #3498 

#3701

### DIFF
--- a/evennia/server/portal/tests.py
+++ b/evennia/server/portal/tests.py
@@ -341,6 +341,50 @@ class TestTelnet(TwistedTestCase):
         self.assertIn("fish & chips", result)
         self.assertNotIn("&amp;", result)
 
+    @mock.patch("evennia.server.portal.portalsessionhandler.reactor", new=MagicMock())
+    def test_naws_resize_syncs_updated_width(self):
+        """
+        Verify that a NAWS resize packet causes sessionhandler.sync to be called
+        AFTER negotiate_sizes has updated SCREENWIDTH, not before.
+        Regression test for the ordering bug introduced in #3498.
+        """
+
+        self.transport.client = ["localhost"]
+        self.transport.setTcpKeepAlive = Mock()
+        d = self.proto.makeConnection(self.transport)
+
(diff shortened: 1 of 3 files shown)


evennia/server/portal/tests.py
```diff
@@ -394,3 +394,42 @@ def test_data_out(self):
         args, kwargs = call_args
         is_binary = kwargs.get("is_binary", args[1] if len(args) > 1 else False)
         self.assertFalse(is_binary)
+
+
+@mock.patch("evennia.server.portal.portalsessionhandler.reactor", new=MagicMock())
```
Griatch: This is outside of any testing class, so this will never run, or fail since it has `self` outside of a class.

evennia/server/portal/tests.py
```diff
@@ -394,3 +394,42 @@ def test_data_out(self):
...
+        return original_sync(session)
+
+    self.proto.sessionhandler.sync = capturing_sync
+    self.proto.dataReceived(b"".join([IAC, SB, NAWS, b"\x00\xa0", b"\x00\x32", IAC, SE]))
+
+    # SCREENWIDTH must be updated to 160 BEFORE sync fires
+    self.assertEqual(self.proto.protocol_flags["SCREENWIDTH"][0], 160)
+    self.assertEqual(synced_widths, [160])  # sync saw the NEW width, not 120
+
+    self.proto.nop_keep_alive.stop()
```
Griatch: This will never be reached if an assertion fails above, leaving the reactor in an unclean state. The addCleanup helper or breaking this into a separate class with tearDown() method would be better.

Griatch (changes requested):
Good fix, but the tests won't work this way, so they need to be updated.

michaelfaith84:
🤦

Griatch (approved):
Thanks!

PR #3917, opened 2026-04-18
Add configurable server lifecycle broadcast messages

blongden:

Introduces SERVER_RELOAD_INITIATE_MSG, SERVER_RESET_MSG, and SERVER_RESTART_MSG settings so games can override the hardcoded restart/reload strings without patching core. Defaults preserve the existing messages exactly.

The reload message has always been easy to change, but the '... Server restarted' one has always been hard coded.

### DIFF
--- a/evennia/server/sessionhandler.py
+++ b/evennia/server/sessionhandler.py
@@ -39,6 +39,7 @@ from evennia.utils.utils import (
 
 _FUNCPARSER_PARSE_OUTGOING_MESSAGES_ENABLED = settings.FUNCPARSER_PARSE_OUTGOING_MESSAGES_ENABLED
 _BROADCAST_SERVER_RESTART_MESSAGES = settings.BROADCAST_SERVER_RESTART_MESSAGES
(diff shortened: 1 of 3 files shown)


evennia/server/sessionhandler.py
```diff
@@ -415,7 +416,7 @@ def portal_sessions_sync(self, portalsessionsdata):
         evennia.EVENNIA_SERVER_SERVICE.at_post_portal_sync(mode)
         # announce the reconnection
         if _BROADCAST_SERVER_RESTART_MESSAGES:
-            self.announce_all(_(" ... Server restarted."))
+            self.announce_all(_(_SERVER_RESTART_MSG))
```
Griatch, line 419: This will break the i18n of this message, so may be better to drop the i18n if you can now change it from settings.

Griatch (approved):
I think this is a useful fix; potentially it will break the i18n of these strings, but maybe i18n can be turne off those strings if they can be modified in settings anyway.

Issue #2790, opened 2022-07-02
[BUG - Develop] (Resurrecting the < > & MXP issues)


aMiss-aWry:
#### Describe the bug
< > & breaking the game in Mudlet

- When I say 'breaking the game', I mean 'the Mudlet profile freezes, awaiting a closed pointy bracket >, and until receiving one will show no output to the profile'.

- Sub-issue 1: Messaging "<break" in-game with in-game messages and commands like 'say' while MXP was on was not breaking the game, but using the same functions with the in-game py interpreter like 'caller.msg("<break")' was.
- Solution: Py code messaging was being sent with 'options={"raw": True}', meaning it was indeed not running through the mxp parser. This could be a reasonably quick fix or not considered something that requires fixing at all, really - but the discrepancy was a head-scratcher until we figured that out.

- Primary Issue: Turning off your MXP protocol flag in the game (with 'option mxp=off') means that mxp parsing, which replaces < with &lt; and so on, will no longer be done. However, if the 'MXP' check is still toggled on in Mudlet settings, this means that now ANY instance of '<break' would break the profile because Mudlet still expects the incoming text to be MXP - because Evennia's do_mxp() function (part of the handshake protocol) always responds the GLOBAL setting for MXP_ENABLED, not the player's protocol flag. 
- Solution: I'm personally shunting this off onto players. They have to turn off MXP in their client settings if they don't want MXP. In my opinion, this could be 'fixed' by removing MXP as a player protocol option... otherwise do_mxp() has to check for the session's MXP protocol flag as well as the global setting and send the corresponding value.

Locations of relevant code: 
server/portal/mxp.py - do_mxp() - Evennia's response to client checks for MXP
commands/default/system.py - _run_code_snippet() - Py code transmitting as raw text (not really sure what the solution here is if we don't want to parse it, which we may not for a variety of reasons)

While on a fundamental level games shouldn't be able to break player client profiles in this fashion and Mudlet should probably have more graceful handling of some kind, we definitely need to fix something in here.

#### To Reproduce
1. Using mudlet, make sure settings -> special options -> force MXP negotiation off is UNCHECKED
2. In mudlet in evennia make sure option mxp=on
3. In mudlet in evennia with developer perms, try 'py self.msg("<break")'
4. See error.
5. To fix, try 'py self.msg("fix>")'
6. Try 'say <hello' and some more random input and notice this does not break anything.
8. Try option mxp=off
9. Try 'say <hello' and more random input and notice this breaks everything.

[...]

PR #3922, opened 2026-04-30
Fix MXP OPEN mode eating unknown tags like <username> in plain text

blongden:

After MXP negotiation, clients default to OPEN mode and attempt to parse all <word> text as MXP elements. This silently removes unknown tags such as <username> and <password> from server output.

Send \x1b[7z (LOCK_LOCKED) immediately after the MXP handshake to put clients into permanently-locked mode. Explicit clickable links created by mxp_parse() already use \x1b[4z (TEMP_SECURE) wrappers, which work correctly from a locked baseline and return to locked mode afterwards.

I think this was the intended way for it to work since it correctly send the TEMP_SECURE escape string when links are used with the pipe commands. It also fixes an issue for us that cropped up which was one of those 'how did this ever work' sort of scenarios.

May help with #2790 but don't quote me on that!

### DIFF
--- a/evennia/server/portal/mxp.py
+++ b/evennia/server/portal/mxp.py
@@ -103,6 +103,10 @@ class Mxp:
         if settings.MXP_ENABLED:
             self.protocol().protocol_flags["MXP"] = True
             self.protocol().requestNegotiation(MXP, b"")
+            # Switch client to permanently-locked mode so plain text containing
+            # '<' is never parsed as MXP. Actual clickable links use \x1b[4z
+            # (TEMP_SECURE) wrappers from mxp_parse(), which work fine from here.
+            self.protocol().sendLine(b"\x1b[7z")
         else:
             self.protocol().wont(MXP)
         self.protocol().handshake_done()
--- a/evennia/server/portal/mxp.py
+++ b/evennia/server/portal/mxp.py
@@ -26,6 +26,7 @@ URL_SUB = re.compile(r"\|lu(.*?)\|lt(.*?)\|le", re.DOTALL)
 MXP = bytes([91])  # b"\x5b"
 
 MXP_TEMPSECURE = "\x1b[4z"
+MXP_LOCK_LOCKED = b"\x1b[7z"
 MXP_SEND = MXP_TEMPSECURE + '<SEND HREF="\\1">' + "\\2" + MXP_TEMPSECURE + "</SEND>"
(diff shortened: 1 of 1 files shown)


jaborsh:
Is this even with https://github.com/evennia/evennia/pull/3910?

blongden:
Summoning @aMiss-aWry 😁 

But yes I think this is even with that fix (in fact @aMiss-aWry's fix may have been why we started noticing `<tags>` being eaten by MXP clients) MXP defaults to OPEN mode which would mean any and all lines from the server would be parsed for MXP tags (see https://www.zuggsoft.com/zmud/mxp.htm - mode 0 is default). This means that anything with a real angle bracket would be parsed by a MXP client and discarded as a probable invalid flag (`<username>` for example).

This change sets default line mode to 7 which is NOT MXP by default. Evennia does the right thing already and explicitly sets mode 4 - temp secure mode - for actual MXP tags.

evennia/server/portal/mxp.py
```diff
@@ -103,6 +103,10 @@ def do_mxp(self, option):
         if settings.MXP_ENABLED:
             self.protocol().protocol_flags["MXP"] = True
             self.protocol().requestNegotiation(MXP, b"")
+            # Switch client to permanently-locked mode so plain text containing
+            # '<' is never parsed as MXP. Actual clickable links use \x1b[4z
+            # (TEMP_SECURE) wrappers from mxp_parse(), which work fine from here.
+            self.protocol().sendLine(b"\x1b[7z")
```
Griatch: This constant should really be a named constant, like `MXP_LOCK_LOCKED` (up where MXP_TEMPSECURE` is defined.

Griatch (changes requested):
MXP is really a whack-a-mole of issues. I have a comment below, to fix.

winkie5970:
I opened a new bug report related to this issue: https://github.com/evennia/evennia/issues/3959
Not sure if this is also solved by your fix above.

evennia/server/portal/mxp.py
blongden: Done - added `MXP_LOCK_LOCKED = b"\x1b[7z"` alongside `MXP_TEMPSECURE` and replaced the inline bytes literal with it.

blongden:
> I opened a new bug report related to this issue: #3959 Not sure if this is also solved by your fix above.

Looks like it yes. Every <> 'tag' is parsed as MXP in the current version - this stops that from happening.

Griatch (approved):
Thanks for the fix and addressing comments!

PR #3936, opened 2026-05-20
Add -> bool return annotations to override-intended hook methods

Problematic:

Fourteen hook methods on DefaultObject, DefaultScript, ScriptBase, DefaultChannel, DefaultAccount, and ServerSession returned a bare `True` literal, causing type checkers to infer Literal[True] rather than bool. Userland subclasses that return False to abort behavior (the documented contract) would be flagged as incompatible overrides. Adds explicit `-> bool` annotations; no runtime behavior changes.

Types are good, we like types :)


Griatch (approved):
Nice fix!

PR #3938, opened 2026-05-22
Fix ReferenceError in notifications plugin when window unfocused

Problematic:

The desktop notifications plugin's onText handler takes (args, kwargs) but the requestPermission callback references a bare `text` identifier that isn't defined anywhere in scope, throwing:

```
    Uncaught ReferenceError: text is not defined
        at notifications.js:43
```

This fires when a message arrives while the webclient window is unfocused, but only for users who have notification_popup enabled and have granted notification permission.

Read the message body from args[0] to match the convention every other text-handling plugin uses (default_out.js, text2html.js, html.js, goldenlayout.js).

This seems to be a rather old issue with a straightforward fix.


Griatch (approved):
Good catch.

PR #3939, opened 2026-05-25
Make AttributeProperty and NAttributeProperty generic for type inference

Problematic:
AttributeProperty is now Generic[T], with overloaded __init__, __get__, and __set__ signatures so type checkers can infer the value type from the default argument. NAttributeProperty inherits the same generic behavior.

This is a purely additive typing change — zero runtime behavior change.

Usage:

    class Character(DefaultCharacter):
        name = AttributeProperty(default="")       # inferred str
        count = AttributeProperty(default=0)        # inferred int
        tags = AttributeProperty(default=list)      # callable default -> inferred list

    obj.name.upper()      # type checker knows this is str
    obj.count + 1         # type checker knows this is int
    greet(obj.name)       # passes to fn(name: str) without complaint

The existing manual annotation workaround (my_attr: str = AttributeProperty(...)) was already a type error before this change and is now unnecessary. Users should drop the annotation and let the descriptor infer the type, or annotate with the descriptor type if they prefer being explicit:

    name = AttributeProperty(default="")                          # valid
    name: AttributeProperty[str] = AttributeProperty(default="")  # also valid
    name: str = AttributeProperty(default="")                     # type error (was already broken)

When no default is provided, AttributeProperty() types as AttributeProperty[Any] for backwards compatibility.

The at_get/at_set hook signatures are unchanged.

Types :)

### DIFF
--- a/evennia/typeclasses/attributes.py
+++ b/evennia/typeclasses/attributes.py
@@ -12,7 +12,9 @@ which is a non-db version of Attributes.
 import fnmatch
 import re
 from collections import defaultdict
+from collections.abc import Callable
 from copy import copy
+from typing import Any, Generic, Self, TypeVar, overload
 
 from django.conf import settings
 from django.db import models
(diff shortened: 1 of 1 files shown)


Griatch (approved):
Applied generics, looks ok as far as I can tell!

PR #3940, opened 2026-05-25
Add @overload signatures to TagHandler.get() for type narrowing

Problematic:
Type checkers can now infer the return type based on return_list and return_tagobj arguments, eliminating the need for manual type annotations at call sites.


Griatch (approved):
Looks good too!

PR #3942, opened 2026-06-01
Add AMP_CONNECT_TIMEOUT for portal probe

### LINKED ISSUE #3941: [BUG] AMP blocks ~30s when Portal is down/unreachable
#### Describe the bug
`evennia_launcher.send_instruction()` connects to the AMP port with `endpoints.TCP4ClientEndpoint(reactor, AMP_HOST, AMP_PORT)`. The default timeout for this is 30s. Every launcher command probes this port to check if the portal is running. Normally, a local port returns an instant refusal so the probe is fast, but if the OS does not fast-refuse (like with WSL2 with mirrored networking) the probe blocks for the full 30s before deciding the portal is not, in fact, running.

#### To Reproduce
1. WSL2 in `networkingMode=mirrored`
2. With Evennia stopped, run `evennia start`
3. Observe ~30s added latency.

#### Expected behavior
The launcher should decide the portal is unreachable quickly. A connect to a loopback AMP port can do so in <1 ms. Waiting 30s here is unreasonable.

- Evennia 6.0.0
- Python 3.14.3
- WSL2 `networkingMode=mirrored`

jaborsh:

The launcher probes the AMP port to decide if the portal is running. `TCP4ClientEndpoint` defaults to 30s timeout. Usually, local port returns an instant response, but when OS doesn't fast-refuse (as is the case with WSL2 + Tailscale), the probe blocks for the full 30s before concluding the portal is down. Every launcher command probes at least once, so `start` and `stop` will each stall for 30s.

A short timeout is safe because it is only ever needed to bound the failure case. A live portal accepts connection in under a millisecond, so 2s is well above the expected threshold.

Closes #3941

### DIFF
--- a/evennia/server/evennia_launcher.py
+++ b/evennia/server/evennia_launcher.py
@@ -81,6 +81,7 @@ NO_REACTOR_STOP = False
 AMP_PORT = None
 AMP_HOST = None
 AMP_INTERFACE = None
+AMP_CONNECT_TIMEOUT = None
 AMP_CONNECTION = None
 
 SRELOAD = chr(14)  # server reloading (have portal start a new server)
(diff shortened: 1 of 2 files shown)


Griatch (approved):
A nice adjustment, I think.

PR #3943, opened 2026-06-11
Replace `unique_together` and `indexes` Tag options with UniqueConstraint

### LINKED ISSUE #3897: [BUG] Migration 0017 get_operations() not called during test DB creation
#### Describe the bug
  Migration typeclasses/migrations/0017_use_index_instead_of_index_together_in_tags.py uses a custom get_operations() method to return database-specific operations. However, Django's migration framework reads the operations class attribute directly — it does not call get_operations(). Since the migration sets operations = [], no operations run when applied to a fresh database (e.g., test DB creation).

  This causes migration 0018 to fail with:

  ValueError: Found wrong number (0) of indexes for typeclasses_tag(db_key, db_category, db_tagtype, db_model)

#### To Reproduce

  1. Install Evennia 6.0.0
  2. Create a game directory with evennia --init
  3. Run evennia test --settings settings <any_test_module>

  The test runner creates a fresh test database, replays all migrations, and hits the error at migration 0018.

  Why It Works in Production

#### Expected behavior

  On existing databases, the index_together constraint from older migrations already exists. Migration 0017 was applied when first released. But on a fresh DB built from scratch (like test runner creates), 0017's empty operations = [] means the index is never created, and 0018's RenameIndex fails.

  Root Cause

  In 0017_use_index_instead_of_index_together_in_tags.py:

  class Migration(migrations.Migration):
      dependencies = [...]

      def get_operations(self, app_labels, schema_editor):
          if schema_editor.connection.vendor == "sqlite":
              return [
                  migrations.AddIndex(...),
                  migrations.RenameIndex(...),
              ]
          else:
              return [migrations.AddIndex(...)]

      operations = []  # Will be populated at runtime by get_operations()

  Django does not call get_operations() — it reads operations directly.

  Environment

  - Evennia 6.0.0
  - Django 6.0.3
  - Python 3.12
  - SQLite backend

maxsond:

Replaces the `unique_together` and `indexes` properties of the Tag meta with `UniqueConstraint`

Seems to resolve https://github.com/evennia/evennia/issues/3897 in a Django-idiomatic way.

Another thing that seemed to work for me was to just delete the `indexes` property alone, but that left `unique_together` which is [considered obsolete](https://docs.djangoproject.com/en/6.0/ref/models/options/#django.db.models.Options.unique_together), so I went ahead and upgraded that.

My understanding is that this change is functionally sound (the old "indexes" property, as I understand it, is redundant with the "unique_together" property given that it only contained the full set of "unique_together" fields, so the index of that unique set is effectively just the tag's PK, and this change retains that set of unique fields), but has one wrinkle which is that any existing databases that migrate using this change will retain an orphaned `typeclasses_db_key_be0c81_idx` index created in migration 0017.

### DIFF
--- a/evennia/typeclasses/tags.py
+++ b/evennia/typeclasses/tags.py
@@ -81,8 +81,12 @@ class Tag(models.Model):
         "Define Django meta options"
 
         verbose_name = "Tag"
-        unique_together = (("db_key", "db_category", "db_tagtype", "db_model"),)
-        indexes = [models.Index(fields=["db_key", "db_category", "db_tagtype", "db_model"])]
+        constraints = [
+            models.UniqueConstraint(
+                fields=["db_key", "db_category", "db_tagtype", "db_model"],
+                name="tag_unique_key_category_tagtype_model",
+            )
+        ]
 
     def __lt__(self, other):
         return str(self) < str(other)


Griatch (changes requested):
I don't disagree with the change itself, but meta changes like this still requires a migration -  you are renaming the restraints and you need to include the migration (generated `makemigrations typeclasses` in the PR.

PR #3952, opened 2026-06-27
Update intro_menu to match the history.js webclient plugin

persimmonb:
The text in `contrib.tutorials.tutorial_world.intro_menu` says the key combination to scroll up/down through the history is Ctrl+up/down, but `web/static/webclient/js/plugins/history.js` uses Shift instead of Ctrl. This just updates the text in intro_menu to match what history does. Likewise with adding a new line to input.

Thanks all for putting so much effort into this library, it's been a joy to poke around in!


Griatch (approved):
Good catch!

PR #3962, opened 2026-07-19
fix: prevent aborting channel message distribution when at_pre_channel_msg returns False/None

### LINKED ISSUE #3961: [BUG] Channels drop messages if any subscriber returns False from `at_pre_channel_msg`
#### Describe the bug

In `evennia/comms/comms.py`, within the `DefaultChannel.msg` method, the loop that distributes a channel message to individual subscribers has an early return statement instead of continue. As a result, if *any* subscriber's `at_pre_channel_msg` returns None or False, the distribution loop terminates entirely. This prevents all remaining subscribers in the receivers list from receiving the message.

Evennia's default implementation of `at_pre_channel_msg` never returns False, but is designed to be overridden to support custom behavior, and its docstrings suggests that it is meant to abort for only the receiver(s) for which it returns False, as opposed to everyone.

I discovered this in my own game when I built a custom channel that isn't heard while the player is in certain states, and applies this block in `at_pre_channel_msg`. If any connected player lacked the requisite state, no messages were ever delivered. Even the person sending the message would not see it as it doesn't echo back to them.

#### To Reproduce
1. Create a new mygame.
2. Implement Account.at_pre_channel_msg to return False for one account (e.g. `if self.key == 'Test': return False`), and connected that account.
3. Have another account send a message to the channel
4. Observe no message is received by any subscribers

#### Expected behavior
`at_pre_channel_msg` should only filter messages for the subscribers for which it returns False, not everyone.

I believe this is a very simple fix. In DefaultChannel.msg:
```
         for receiver in receivers:
             try:
                 recv_message = receiver.at_pre_channel_msg(message, self, **send_kwargs)
                 if recv_message in (None, False):
-                    return
+                    continue
                 receiver.channel_msg(recv_message, self, **send_kwargs)
```

         Evennia version: 5.0.1
         OS: posix
         Python: 3.13.7
         Twisted: 24.11.0
         Django: 5.2.6

Also observed in evennia 6.1.0

owllex:
This PR fixes a bug where channel message distribution would abort early if any subscriber's `at_pre_channel_msg` returned `False` or `None`. The bug was caused by a `return` statement in the receiver iteration loop, which has been corrected to `continue` to allow the message to proceed to subsequent subscribers.

A new unit test class `ChannelMessageDistributionTests` has been added to `evennia/comms/tests.py` to verify this logic and prevent regression.

Fixes evennia/evennia#3961

N/A

### DIFF
--- a/evennia/comms/comms.py
+++ b/evennia/comms/comms.py
@@ -656,7 +656,7 @@ class DefaultChannel(ChannelDB, metaclass=TypeclassBase):
             try:
                 recv_message = receiver.at_pre_channel_msg(message, self, **send_kwargs)
                 if recv_message in (None, False):
-                    return
+                    continue
 
                 receiver.channel_msg(recv_message, self, **send_kwargs)
 
--- a/evennia/comms/tests.py
+++ b/evennia/comms/tests.py
@@ -1,3 +1,5 @@
+from unittest.mock import MagicMock
+
 from django.test import SimpleTestCase
 
 from evennia.commands.default.comms import CmdChannel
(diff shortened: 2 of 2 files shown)


Griatch (approved):
Clean fix, thanks!

PR #3968, opened 2026-07-25
fix(evmenu): format dict helptext per entry

### LINKED ISSUE #3755: [BUG] EvMenu Multi-Help raises Attribute Error
#### Describe the bug
When providing an EvMenu node with a tuple for its text with a dict as its second item, Evennia returns with Attribute Error: 'dict' object has no attribute 'strip'

Bug appears to occur during formatting in EvMenu util goto function.

#### To Reproduce
1. Define EvMenu 
2. Create node with return (main_text, {str: st)), options
3. Run EvMenu to node

#### Expected behavior
When providing EvMenu node with tuple for text, dict should provide help commands based off of their key/alias-value pairs.

 Evennia version: 4.5.0
         OS: posix
         Python: 3.11.10
         Twisted: 23.10.0
         Django: 4.2.16

#### Additional context
When looking into EvMenu's functions, handling of helptext may be the cause. When processing helptext dict, it defines _help_text as a dict, then parses help_text dict. Finally, it assigns help_text = _help_text.

Following this parsing, it checks if help_text has a value, before passing it to helptext_formatter, which expects a string rather than a dict. It attempts to strip the help_text, leading to the error.

Image

Image

Image

Sanjays2402:

`EvMenu.helptext_formatter()` called `.strip()` on the help text unconditionally, but a node may return its help text as a dict to provide per-command tooltips — so any such node raised `AttributeError: 'dict' object has no attribute 'strip'`. Dict helptext is now formatted entry by entry, which keeps the tooltip lookup in `parse_input()` working.

The dict form is already parsed and supported in `_execute_node`/`parse_input` (including `(key, aliases)` tuples), so it was only the formatter that made it unusable.

Closes #3755. Added `TestEvMenuDictHelptext`, which fails with the reported AttributeError without the fix and passes with it; `evennia.utils.tests.test_evmenu` is green (8 tests).

### DIFF
--- a/evennia/utils/evmenu.py
+++ b/evennia/utils/evmenu.py
@@ -1165,12 +1165,17 @@ class EvMenu:
         Format the node's help text
 
         Args:
-            helptext (str): The unformatted help text for the node.
+            helptext (str or dict): The unformatted help text for the node. A dict
+                maps a tooltip command to its own help text; each entry is
+                formatted separately.
 
         Returns:
-            helptext (str): The formatted help text.
+            helptext (str or dict): The formatted help text, of the same type as
+                the input.
 
         """
+        if isinstance(helptext, dict):
+            return {key: self.helptext_formatter(entry) for key, entry in helptext.items()}
         return dedent(helptext.strip("\n"), baseline_index=0).rstrip()
 
     def options_formatter(self, optionlist):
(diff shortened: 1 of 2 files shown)


Griatch (approved):
This is a nice fix, also appreciate the unit test to back it up! Thanks!

